PR #132, opened 2026-07-17
feat(download): --dry-run, --model weights-skip, clearer auth errors + transient read retry

ZacxDev:
## What & why

Four self-contained CLI papercut fixes surfaced by a blind dogfood of **v0.1.63**. Each is focused; no scope creep.

### 1. Auto-retry transient failures on idempotent read GETs (the 503-flakiness)
The dogfood hit **5–8 consecutive `503`s** on `models search --query` with no retry. Read GETs now route through a bounded retry (`internal/api/retry.go`):
- Retries **429 / 502 / 503 / 504** and transient network errors (timeout / connection-refused / reset). Terminal 4xx (**400/401/403/404**) return on the first try — **401 keeps its existing refresh path**, no double-handling.
- Up to **4 total tries**, exponential backoff; honors a `429` **`Retry-After`** (capped at 5s).
- One brief stderr note per retry (`transient 503 from Civitai, retrying (2/4)…`).
- Wired at the shared getter seam (`getRaw` → `getWithRetry`), so **all** read subcommands benefit (models/versions/images/tags/creators/users/articles/collections). The **download stream** and **mutating calls** (submit/dev-token/withdraw/dev-tunnel) are deliberately not wrapped.
- **Design:** backoff base is injectable (`Client.RetryBackoffBase`) so tests run with zero sleeps.

### 2. `civitai download --dry-run`
Resolve the version + selected file(s) and print the plan — name, size, SHA256, resolved target path, whether auth is required — then exit `0`. Creates **no file and no `.part`**, transfers nothing. Works with `--file` / `--all` / `--model` / `--out` / `--out-dir`. A non-weights file shows as *would be refused* in the plan instead of aborting it.

### 3. `download --model` skips non-weights default versions
`resolveVersionID` blindly took `modelVersions[0]`, which can be a training-data / on-site-generation registration → errors on a ZIP. It now picks the **first version whose primary file is downloadable weights** (`IsModelWeights()`), noting the skip on stderr, and errors clearly if none has weights. If the models endpoint embeds no per-version file info, it falls back to the prior `[0]` behavior (no regression). **The explicit positional `<version-id>` path is unchanged** — naming a non-weights version directly still hits the `--allow-nonmodel` guard.

### 4. Clearer "login required" download error
An anonymous download of even a public file 401s (Civitai requires a token for **any** model-file download). The 401 message now says so and points at `civitai login` / `CIVITAI_API_KEY`; the **403** message is distinct — the file is *gated* (early-access / subscriber-only), and re-logging-in won't help.

## Test coverage (complete, table-driven, deterministic — no wall-clock sleeps)

**`internal/api/retry_test.go`** — retries-then-succeeds on 502/503/504/429; gives-up-after-max naming status+attempts (`4` requests asserted via hit counter); no-retry on 400/401/403/404 (exactly 1 request); honors `Retry-After`; transient network error retried (stub RoundTripper); pure `retryAfterDelay` parse/cap table; `isTransientNetErr` unit; backoff base 0 → instant.
**`internal/cmd/download_dryrun_test.go`** — plan prints name/size/sha/target/auth + writes nothing (out-dir asserted empty, `dlHits==0`); `--all` + `--model` + `--out`; single non-weights shows refusal without erroring; exit 0.
**`internal/cmd/download_model_weights_test.go`** — non-weights default → resolves to the weights version (asserts chosen id + note); all-non-weights → clear error; default-already-weights → unchanged, no note; explicit positional non-weights still refuses.
**`internal/cmd/download_cmd_test.go`** — 401 → contains `civitai login`; 403 (with token) → indicates *gated*, not "run login".
`internal/api/read_test.go` updated: the retriable rows now assert the exhaustion message with instant backoff.

Full `go build ./... && go vet ./... && go test ./... && go test -race ./internal/cmd/... ./internal/api/...` all green; `gofmt -s -l .` clean.

## Follow-up (not in this PR)
developer.civitai.com's CLI guide should document `--dry-run` and the `--model` weights-skip behavior — separate repo, intentionally untouched here.


ZacxDev:
## Audit follow-up: deep-paging 429 + backoff jitter + retry cosmetics

Pushed `dcf71ec` addressing one 🟡 finding and two cheap nits from the retry-loop audit.

### 🟡 Fix 1 — don't retry a deterministic deep-paging 429; preserve the `--cursor` hint
Civitai returns **429 deterministically on deep `--page` offsets** (a paging-depth limit, not a transient outage — the whole reason `--cursor` exists). The new retry loop was retrying *every* 429 four times and then surfacing the generic `retryExhaustedError` ("…temporarily unavailable, try again shortly"), which is wrong for that case and dropped the actionable hint.

Now a **429 is retriable only when it carries a `Retry-After` header** (a genuine server-signalled throttle, honored up to the cap). A **429 without `Retry-After`** is terminal for reads — exactly **1 request** — and is returned cleanly (`status, raw, nil`) so it routes back through `readError`'s 429 branch, which revives the bespoke `for deep paging use --cursor instead of --page` guidance. `502/503/504` and Retry-After-bearing 429s are unchanged.

### Fix 2 (nit) — backoff jitter to de-correlate the herd
`backoffFor` now adds **±20% jitter** (`math/rand` top-level, concurrency-safe on Go 1.25) to the exponential delay so a fleet-wide 503 storm doesn't re-fire in synchronized waves on the fixed 250ms/500ms/1s schedule. Jitter is **gated on base>0** (the injected zero base still yields exactly `0`, keeping tests instant + deterministic) and the cap is re-applied **after** jitter. `Retry-After` overrides are honored verbatim (not jittered). The ctx-cancel path is untouched.

### Fix 3 (cosmetics)
- Exhaustion message no longer emits a dangling `": "` when the body snippet is empty.
- Renamed builtin-shadowing params: `cap` → `capDelay` (`clampDelay`, and `retryAfterDelay` which had the same shadow), `max` → `maxN` (`retryNote`). Pure rename.

### Tests
- `TestRead429WithoutRetryAfterIsTerminalWithCursorHint` — 429 w/o Retry-After ⇒ **1 request**, error contains `--cursor`, does **not** contain "temporarily unavailable", no retry notice.
- `TestRead429WithRetryAfterExhaustsGeneric` — persistent 429+`Retry-After` ⇒ retried to the cap, generic exhaustion message.
- `TestReadHonorsRetryAfterOn429` (existing) — 429+Retry-After recovers on a later attempt.
- `TestReadRetriesTransientThenSucceeds` — 429 dropped from the loop (502/503/504 unchanged).
- `TestReadErrorSurfacesBody` — 429 row updated to assert the `--cursor` message.
- `TestBackoffForJitterBounds` — base=0 ⇒ 0; base>0 ⇒ delay within `[0, cap]` across attempts.
- `TestRetryExhaustedNoTrailingColonOnEmptyBody` — always-503 empty body ⇒ no trailing `": "`.

### Gates (all green)
`go build ./...`, `go vet ./...`, `go test -count=1 ./...`, `go test -race ./internal/api/... ./internal/cmd/...`, `gofmt -s -l .` (clean).

### Also updated
`retry.go` header/doc comments and the README retry note to match the new 429 behavior (out-of-scope 🟢 nits — Retry-After asymmetry, dead 503 branch, `--model --dry-run` ordering, doc "~4 tries" — intentionally left).

PR #137, opened 2026-07-18
feat(cli): type-aware download folder routing, base-model compat warning, article content rendering + auth-required download docs

ZacxDev:
## Summary

Adds four REST endpoints for **articles** and **collections**, closing the Phase-2 CLI gap (models/images/tags/creators/users already have public REST; articles + collections did not). Each endpoint **wraps the existing service layer** (the same functions the website's tRPC routers use) — no business logic is reimplemented.

> **⚠️ Gated preview — NO LONGER public/anonymous.** These four endpoints are now gated behind the **`appBlocksAuthor`** cohort (moderators + the `app-dev-testers` Flipt cohort), the same authoring cohort the App Blocks v1 endpoints use. Every non-cohort caller — **including anonymous** — gets a **`403 { error: 'This API is in preview — access is restricted to Civitai moderators and app developers.' }`**. This is a deliberate DX choice for an *invited* cohort: rather than a dark 404 that hides the endpoint's existence, the 403 reveals that the API exists but explains WHY the caller is restricted — so someone who's been invited but isn't yet provisioned into the cohort understands the situation instead of chasing a phantom "not found". The gate is evaluated FIRST in each handler, before the rate limit + service, via `isAppBlocksAuthorEnabled({ user })` (async; mod static-floor OR the `app-blocks-author` Flipt cohort). Ownership/user always comes from the authenticated session (`MixedAuthEndpoint`'s optional user), never a client param.
>
> **Consumer note:** the `civitai` CLI must be **logged in as a cohort member** (a moderator / app-author OAuth session) to reach these — an anonymous or ordinary-user call gets the 403 preview message. Widen later by widening the `app-blocks-author` Flipt flag, not by editing these handlers.

| Endpoint | Wraps | Shape |
|---|---|---|
| `GET /api/v1/articles` | `getArticles` | `{ items, metadata: { nextCursor, nextPage } }` (composite keyset cursor) |
| `GET /api/v1/articles/{id}` | `getArticleById` | article object (moderator-only fields stripped) |
| `GET /api/v1/collections` | `getAllCollections` (public) / `getUserCollectionsWithPermissions` (`?mine=true`) | `{ items, metadata: { nextCursor, nextPage } }` (keyset cursor on id) |
| `GET /api/v1/collections/{id}` | `getUserCollectionPermissionsById` + `getCollectionById` | collection projection |

All four use `MixedAuthEndpoint` and match the `{ items, metadata }` envelope + `limit`/`cursor` param style of the existing `models` endpoint.

## Cohort gate + visibility (defense in depth)

The `appBlocksAuthor` gate is the OUTER boundary — but the per-resource visibility gates below are retained UNCHANGED behind it, so even a cohort member only ever sees what the service layer would show their session. Read-only over existing data — **no DB migration**.

- **Articles list** — passes the session user through to `getArticles` unchanged and never a client-supplied userId. For a non-owner/non-moderator the service returns only `status = Published` + `ingestion = Scanned` and drops `availability = Private` (`forceHidePrivate: true` pins this even with `?username=`). `favorites`/`hidden` (own-engagement) are authenticated-only → 401.
- **Article detail** — `getArticleById` gates non-moderators to `(published + scanned) OR owner`; a draft/unpublished article 404s for a non-owner (indistinguishable from non-existent). `moderatorNsfwLevel` is stripped from the response.
- **Collections list** — `getAllCollections` pins privacy to `[Public]` for any non-moderator (the service re-clamps too). `?mine=true` is authenticated-only (401 otherwise) and keys `getUserCollectionsWithPermissions` on the **session** user id, never a client param.
- **Collection detail** — reuses `getUserCollectionPermissionsById` for the read decision (the same authority `getCollectionByIdHandler` uses); no read permission → the same bare 404 as a non-existent collection (no existence oracle). `getCollectionById` does no gating on its own, so this check is mandatory before it's called.
- **Maturity/region** — all four derive the browsing-level ceiling the same way the existing public endpoints do (region-restricted → SFW) and clamp covers / drop over-ceiling collections accordingly.

## Conservative rate limiting

Helper `src/server/utils/public-api-rate-limit.ts`, modeled on the existing block-catalog / shared-storage fixed-window limiters (same Redis `INCR`+`EXPIRE` shape, same **fail-open** posture — a Redis blip must never hard-fail a GET). Runs AFTER the cohort gate, so a non-cohort caller never even reaches the limiter.

- **Keying:** now that every caller is authenticated (the cohort gate rejects anon), the limiter keys on **user id**; the client-**IP** path (CF-Connecting-IP-derived) remains for defense but is effectively unreachable.
- **Limits (named, tunable constants at the top of the helper):**
  - Unauthenticated: **60 req/min per IP** (path now unreachable behind the gate).
  - Authenticated: **120 req/min per user** — 2× headroom for legitimate batch reads (e.g. the CLI).
- On exceed: **429** with a `Retry-After` header (+ `Cache-Control: no-store` so the edge can't cache the throttle response). Applied before the expensive service call.

## Tests

`src/tests/api/v1/{articles,collections}/index.test.ts` (28 tests, under `src/tests/**` so they don't break `next build`):
- the **author-cohort gate**: a non-cohort authed user AND an anonymous caller each get a **403 carrying the preview message**, with neither the rate limiter nor the service invoked (the flag helper `isAppBlocksAuthorEnabled` is mocked — no real Flipt); the happy-path suites now mock the flag ON (in-cohort) so they exercise the handler body,
- envelope + cursor serialization, the visibility gate (a private/unpublished item 404s / is excluded for a non-owner; collection detail never calls `getCollectionById` without read permission), the authenticated-only params (401), and the rate-limit path (429 + `Retry-After`, service skipped).

`tsc --noEmit`: the changed files are clean (the only errors in the run are pre-existing `Cannot find module` noise from an un-checked-out submodule in the isolated worktree, plus downstream implicit-`any` in an untouched file — none in this PR's files).


ZacxDev:
## Fix pass — audit findings addressed (`0052c07`)

Following the adversarial audit (passed, no blockers) that flagged the hand-rolled HTML renderer as fragile, this pass replaces it with a real tokenizer and closes the nits.

### Fix 1 (primary) — `articles get --content` now uses `golang.org/x/net/html`
`internal/cmd/htmlrender.go` is rebuilt on `html.NewTokenizer` instead of regex-scanning HTML by hand. Output shape is unchanged (headings `#..######`, `- `/`N.` lists, `[text](url)` links, `` ``` `` code fences, bold/italic, blockquote, `<br>`→newline). Correctly handles what the regex version got wrong:

- **`<script>` / `<style>`** — the entire subtree is skipped; inner text never renders.
- **HTML comments / doctype** — dropped, never printed.
- **Unclosed `<pre>` / list / paragraph** — flushed at EOF so buffered content is never silently dropped.
- **Malformed / unclosed / deeply-nested markup** — cannot panic or loop (tokenizer handles it by construction; covered by a test).
- **Entities** — decoded once (the tokenizer returns decoded `TextToken` text; no double-decode).

`go.mod`: adds direct `require golang.org/x/net v0.57.0` (a sibling of the existing `x/crypto`, `x/term`, `x/text`); `go mod tidy` run, `go.sum` updated.

### Fix 2 (security) — strip terminal control characters
Even via the tokenizer, decoded text can carry raw C0 control bytes (e.g. `ESC`) a malicious article author embeds → ANSI-escape injection into the reader's terminal. `sanitizeControl()` strips all C0 controls (`0x00`–`0x1F`) + `0x7F` except `\n`/`\t` (`\r` dropped) over the final rendered string. Tested with `&#27;`, a literal `\x1b[2J`, and a raw ESC byte → no control bytes survive, surrounding visible text does.

### Fix 3 — `mixedTypeWarning` names the correct file
`internal/cmd/layout.go`: the reference category is now the **primary/weights** file (or the most common known category), not `files[0]`. A bundled VAE listed before the weights file no longer perversely names the checkpoint as the mis-filed one.

### Fix 4 — `--layout` path-traversal test
Added an end-to-end + unit assertion that a hostile server filename (`../../evil.txt`) under `--layout … --root <tmp>` stays basename-sanitized inside the routed type folder, never escaping `--root`.

### Left as-is (audit judged acceptable)
- `--for-base` with a model *name* → silent no-op (documented, conservative).
- `base model:` line always printed on download (intended per the lever).

### Gates
`go build ./...`, `go vet ./...`, `go test -count=1 ./...`, `go test -race ./internal/cmd/... ./internal/api/...` all green; `gofmt -s -l .` clean.

New/updated tests: renderer — script/style skipped, comments dropped, unclosed pre flushed, unclosed list/paragraph flushed, malformed/deep-nested no-panic, control-char stripping; layout — off-type-when-ancillary-first + most-common-fallback naming; download — `--layout` hostile-name sanitization (e2e + unit).

PR #193, opened 2026-08-03
docs(handoff): land both app-analytics handoff docs on main

ZacxDev:
Gets the app-analytics handoff doc onto `main` instead of leaving it on a branch, and updates it so it stops reading as a to-do list for work that has since shipped.

## Why now

The doc is the source for everything in this stream — cli#190/#191/#192 and civitai#3557/#3561/#3572/#3574/#3581. It has been sitting on an unmerged branch describing "3 PRs open, nothing merged", with a prominent **"Do not merge until that re-gate reports"** directive that is now three merges out of date. A stranded doc that is also stale is worse than no doc: the next reader acts on the imperatives.

## What changed

- **RESOLVED banner** with the merged-PR table, replacing the "nothing merged" status.
- **Neutralised the stale directive** — and recorded that the rebase it anticipated *was* needed, for a different reason than it guessed: `#3566` landed later still and edited the same `detail: {}` object, turning #3561 `CONFLICTING` after its gate had already passed.
- **Struck follow-ups #1, #2 and #4** with outcomes, including the two places **this doc was wrong**:
  - #1's entry missed a second proc with the identical live defect — `getMyForgejoCloneInfo`, which `civitai app pull` drives. An audit found it, not the list. Worth knowing generally: a ranked follow-up list is not a survey.
  - #2's suggested fix (reuse `humaniseScopeEndpoint`) would have shipped a bug. Measured against the real function it returns `'(no workflow id)'` and `''` for the bounded tokens, because it's the per-*row* labeller and an aggregate bucket has no `detail`.
- **Recorded #1's scope decision with the prod evidence** that confirmed it after the fact: 331 live tokens unblocked, **30 of which lack bit 26** — so copying the nearest precedent would have left those still 403ing. Plus the measurement trap: `(mask & Full) = Full` is also true when `mask == Full` and reports 145 false hits.
- **New "Still open — start here"** section, so the remaining work is findable without reading the history: the CI `component`-tier gap, the stale-`node_modules` trap that silently removes ~1,126 tests, the unverified `addCollaborator` downgrade lead, and a note that `installs: 0` should be assumed broken until a positive control exists.
- **"What actually caught the bugs"** — across 12 adversarial audit rounds, every fix round found a defect in the previous fix, and the mechanical gate caught none of them.

## The two things in here I'd most want a reviewer to see

**CI does not run the `component` (browser) project at all.** Three PRs merged today ship browser tests — #3574's 7, #3581's 3, plus #3557's existing panel tests — that have **never executed on a canonical browser**. They ran only against a nixpkgs `playwright-chromium-headless-shell` shimmed in locally, because Playwright's vendored binary can't exec on NixOS. As things stand those guards would catch nothing in CI.

**A stale `node_modules` silently removes ~1,126 tests.** A worktree installed before main gained `@civitai/db-queries` fails at `kyselyDb.ts` module load, so 89 files' tests are never collected and the runner still prints a plausible total (10,131 passed vs 11,257 after `pnpm install`). Filed on [civitai#3579](https://github.com/civitai/civitai/issues/3579).

Doc-only; no code touched.

### DIFF
--- a/claudedocs/2026-08-03-app-analytics-handoff.md
+++ b/claudedocs/2026-08-03-app-analytics-handoff.md
@@ -1,8 +1,57 @@
-# Handoff — App analytics: CLI command + two platform fixes
+# Handoff — App analytics: CLI command + platform fixes
 
-**Session date:** 2026-08-02 → 2026-08-03
+**Session dates:** 2026-08-02 → 2026-08-03 (two sessions)
 **Repos touched:** `civitai/cli`, `civitai/civitai`, `devrc`
-**Status at handoff:** 3 PRs open and audited, 1 re-gate in flight, nothing merged.
+
+> ## ✅ RESOLVED — everything below shipped, plus follow-ups #1, #2 and #4
+>
+> This doc is kept as the record of how the work was found and reasoned about, not as a
+> to-do list. **Read the "Still open" section at the bottom first** — the rest is history.
+>
+> | PR | what | state |
+> |---|---|---|
+> | [civitai#3561](https://github.com/civitai/civitai/pull/3561) | endpoint cardinality fix | merged |
+> | [civitai#3557](https://github.com/civitai/civitai/pull/3557) | dark-flag fabricated-zero fix | merged |
+> | [civitai#3572](https://github.com/civitai/civitai/pull/3572) | follow-up #1 — scope-gate `getMyAppAnalytics` **and** `getMyForgejoCloneInfo` | merged |
+> | [civitai#3574](https://github.com/civitai/civitai/pull/3574) | follow-up #2 — humanise both analytics top-N cards | merged |
+> | [civitai#3581](https://github.com/civitai/civitai/pull/3581) | follow-up #4 — `getMyRevenue` undiscriminated zero | merged |
+> | [cli#190](https://github.com/civitai/cli/pull/190) | `civitai app metrics <slug>` | merged |
+> | [cli#191](https://github.com/civitai/cli/pull/191) | AGENTS.md items 5–7 | merged |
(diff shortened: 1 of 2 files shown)


ZacxDev:
## Scope grew by one file: a second handoff doc

Now lands **two** docs, not one:

- `claudedocs/2026-08-03-app-analytics-handoff.md` — the predecessor, updated with shipped outcomes and two retractions.
- `claudedocs/handoff-app-analytics-followups.md` — **new.** Where the stream stands now, plus one fully-diagnosed open bug.

That open bug is the reason this is worth reviewing rather than rubber-stamping. **`preview / component-tests` went success → failure exactly at #3574's audit-fix round, and the root cause is mine:**

- `AppAnalyticsPanel.tsx:263` renders a stat whose tooltip is `"Generations run through your app within the selected range…"`
- my test asserts `page.getByText('Generations')`, and `getByText` is **substring + case-insensitive**
- so the locator resolves to **2 elements** whenever that hover-only tooltip is mounted

[civitai#3593](https://github.com/civitai/civitai/pull/3593) established that vitest browser mode shares **one page across every `.browser.test.tsx` file**, so a leftover pointer position from an earlier file can have the tooltip open at mount. Fails in a full-suite run, never in a single-file run — exactly the observed pattern. Confirmed by computation that the previous label (`'Generation submits'`) did **not** substring-match that tooltip, so the vocabulary rename is what introduced it.

The doc carries the verbatim fix, and the eliminations so nobody re-runs them: the cold-`optimizeDeps` flake is genuinely fixed; PR 3591 passes `component-tests` on a base containing the merge (so the suite is not universally broken); and a local repro cannot settle it, because the full-project run fails on this NixOS host **with and without** the change.

PR #200, opened 2026-08-04
feat(scaffold): drop the redundant Buzz balance readout; raise buzzBudgetPerGen to a real ceiling; rebrand Comfy Cloud -> Comfy on Civitai

ZacxDev:
Four scaffold/branding cleanups from a recorded design review.

Docs counterpart: **civitai/civitai-developer-docs#31**

---

## A — Remove the redundant Buzz balance readout from the default scaffold

Every app an AI agent generates rendered blue/green/yellow Buzz chips + a total, because `page-money` (the **default** template — `internal/cmd/app_create.go`) did. Apps run inside the Civitai chrome, which **already shows the viewer's balance**, so the in-app copy is pure redundancy — and it competes with the real one whenever the two are briefly out of sync. Agents copy the scaffold, so it propagated into every generated app.

### Consumers of `useBuzzBalance()` — removed vs kept

| Consumer | Verdict | Why |
|---|---|---|
| `BalancePanel` (3× `PoolChip` + total) | **Removed** | The readout itself. Pure duplication of host chrome. |
| `PoolChip`, `POOL_COLORS`, `balancePanelStyle`, `balanceTotalStyle`, `poolChipStyle`, `poolDotStyle`, `poolValueStyle` | **Removed** | Dead once the panel is gone — no other referents (verified by grep + a clean `tsc --noEmit` on a generated app). |
| `loading` / `error` from the hook | **Removed** | Only ever read by the panel's degraded states. |
| **`balance` → `AccountPicker`** | **KEPT** | Annotates *which account can actually fund this generation* (`· 0` + a `title` on an empty pool). This is app-specific context the host chrome cannot provide — it's reasoning about spend, not re-displaying a number. |
| **`refetch` → post-success effect** | **KEPT** | A generation debits the pools; without the refetch the picker's 0-annotations go stale immediately after the spend that caused them. |

`useBuzzBalance` itself is **not** blanket-removed — only the readout is.

### Coverage

Not deleted to make things pass — **net +1 test** (baseline 137 → 138):

- `renders the 3-pool Buzz balance` → **`does NOT render a Buzz balance readout`** (asserts all five `pm-balance*` testids absent, after waiting for a post-`BLOCK_INIT` frame so it can't pass vacuously).
- **NEW** `the balance IS consumed: the account picker annotates a 0-Buzz pool` — the positive control. A zero-assertion alone can't distinguish "readout removed" from "balance no longer reaching the app"; this proves the hook still delivers (yellow 80 / blue 20 / **green 0** → only green annotated).
- `gracefully shows "unavailable"` → **`a failed balance read degrades silently`** — the graceful-degradation path is retained, retargeted at the picker.
- The anon test's `pm-balance` assertion was **dropped, deliberately**: with no readout for any viewer it would pass regardless and prove nothing about the anon path. Its absence is guarded by its own test.
- Go side: `mustContain(app, "pm-balance")` → `mustNotContain` for `pm-balance`, `BalancePanel`, `PoolChip`, `POOL_COLORS` — so re-adding a readout under a *different* testid still trips it.

The generated README now teaches the rule instead of the pattern.

---

## B — "Comfy Cloud" → "Comfy on Civitai"

The Comfy project holds a trademark on "Comfy Cloud". **39 occurrences** in this repo (more than a line-grep suggests — 3 were line-wrapped across a newline, and 4 more were case/hyphen variants a case-sensitive pass missed: `/comfy cloud recipe/i` in an e2e selector and 3× `Comfy-Cloud`). The regex variant was caught by *running the generated app's suite*, not by review.

### User-facing (39 renamed)

| Surface | Count |
|---|---|
| CLI help — `app_create.go` | 3 |
| CLI stdout — `app_init.go` | 4 (1 string + 3 comment) |
| Issue template `request-recipe.yml` (name, description, body) | 3 |
| Generated README `page-money/README.md.tmpl` | 2 |
| Generated app UI + comments `App.tsx.tmpl` (incl. the tab label + 2 Alert titles) | 6 |
| Generated app `comfy.ts` / `generation.ts` / tests | 12 |
| Go test assertions + comments | 9 |

### Internal — deliberately NOT renamed

- `customComfy` — the `WorkflowBody` discriminant (**35** in-repo / 25 in a generated app)
- `starter-comfy-txt2img` — the server-registered recipe id (**13** / 12)
- `civitai_app_block_customcomfy_*` — **not present in this repo** (metric names live in `civitai/civitai`); nothing to protect here.

---

## C — `page-vite`: React `^18.3.1` → `^19.0.0`

Bumped. The template uses only `createRoot` (`react-dom/client`), `React.StrictMode`, `useState`, `useEffect` — none of React 19's removals (`ReactDOM.render`, `findDOMNode`, string refs, legacy context, `defaultProps` on function components) apply. Left `@vitejs/plugin-react@^4` / `vite@^6` alone — verified working rather than assumed.

---

## D — `buzzBudgetPerGen` default `40` → `300` (safety ceiling, not an estimate)

`buzzBudgetPerGen` bounds what a **malicious or exploited app can spend per generation**. It is not a forecast. The scaffold shipped **40** — the starter Comfy recipe's per-job ceiling (30) **plus 10**. That is worst-case-plus-margin, i.e. estimate-shaped, and since `page-money` is the *default* template, every generated app inherited it and taught authors' agents to size the field like a forecast.

It also blocked a fix: a CLI "this value looks too small" warning can't exist while any threshold catching an author's ~50 would also fire on our own 40.

### Why 300

The scaffold's worst case per generation is **30** — the hard, server-enforced per-job ceiling of `starter-comfy-txt2img`, the priciest thing this template can invoke. (The txt2img sample is cheaper; that's exactly why 40 "also covered" it.) **300 = 10× that.**

The multiple comes from the asymmetry, not from taste:

| | Consequence |
|---|---|
| **Too low** | Refused at a **pre-submit gate** — nothing runs, nothing is charged — but the app is **broken for every user** until a new manifest version ships **and is re-approved**. |
| **Too high** | Nothing. The headroom is **never spent**: the server re-prices and hard-caps every generation at the real per-job ceiling, and clamps the manifest value to the platform per-gen cap at mint. The schema sets only `exclusiveMinimum: 0` — no upper bound. |

So the value must absorb a pricier checkpoint, a full 5-LoRA stack (`MAX_LORAS`), a future recipe with a higher ceiling, and Buzz repricing — and must be unmistakably *not* a forecast. 300 against a 30 worst case cannot be misread as one; 40 can.

### Sites changed

- `templates/page-money/block.manifest.json.tmpl` — authoritative (the server reads this at mint)
- `templates/page-money/src/generation.ts.tmpl` — the `PAGE_BUZZ_BUDGET` mirror
- `scaffold_test.go` — assertion + a **new seam guard**

### New seam guard

The budget lives in **two** files and nothing at runtime reconciles them, so a drifted mirror silently misinforms the author. The guard now parses the number the **rendered** manifest actually emitted and requires `PAGE_BUZZ_BUDGET` to equal it — pinning the *relationship*, not just each value. Mutation-verified: drifting only the mirror fails on `PAGE_BUZZ_BUDGET`; reverting both to 40 fails the literal assertion **and** the parsed-value check, with distinct messages.

Also corrected a comment that contradicted the code: `PAGE_BUZZ_BUDGET` was documented as driving a client-side "Top up · N" CTA, but it has **no consumer** in the scaffold.

### 🔴 Stale prose left in place — needs routing to cli#199

The generated README still names the old number. I did not edit it (budget guidance is owned by cli#199):

- `templates/page-money/README.md.tmpl:113` — *"That's why `page.buzzBudgetPerGen` is **40**: it must reserve at least the recipe's ceiling (30) and also covers a cheap txt2img gen."*
- `templates/page-money/README.md.tmpl:119` — *"so **40** is the per-gen ceiling for the txt2img sample too"*

Both now contradict the manifest, **and both articulate the very "ceiling + margin" reasoning this change removes** — so they need rewriting, not just a number swap.

Two more in the docs repo, same owner (docs#30), also untouched by my docs PR:
- `apps/guide/comfy-cloud.md:158` — a manifest sample with `"buzzBudgetPerGen": 40`
- `apps/guide/comfy-cloud.md:167` — *"pairs `buzzBudgetPerGen: 40` with the `starter-comfy-txt2img` recipe"*

### Not changed: `examples/buzz-generator.block.manifest.json` (sets `10`)

Same anti-pattern, and it would trip any "too small" warning — but the repo README states these examples are **"copied from the shipping `civitai-block-*` apps"**, so editing the value would falsify that provenance claim. It is a record of a real app, not a scaffold default. Flagging rather than silently rewriting: either the shipped app's budget needs raising, or the "a good reference for a correct manifest" framing needs dropping.

---

## Verification

- `go build ./...` clean; `gofmt -l` empty.
- `go test ./...` — **1332 RUN / 1330 PASS / 2 SKIP / 0 FAIL**, 16 packages ok (counted and reconciled exactly, not read off an exit code).
- **Generated a real app from the edited templates** and ran its own gates: `npm test` **138/138 pass** (11 files), `npm run typecheck` **0 errors**. Baseline generated from `origin/main`: **137 pass** → net +1, nothing dropped.
- Typechecker validated against a known-bad input first (injected type error → 2 errors; 0 under test), so the clean result is not a fact about the command line.
- **Mutation-checked every new guard**, each failing for its *own* reason: re-adding a `pm-balance` testid → `TestRenderPageMoney`: `expected NOT to contain "pm-balance"`; reverting the help string → `TestAppCreateHelpMentionsComfy`. Both green again after restore.
- `page-vite`: generated, `npm install` (react **19.2.8**), `npm run build` succeeded (29 modules, dist emitted).
- Budget: generated an app and confirmed the **emitted** manifest carries `"buzzBudgetPerGen": 300`; `civitai app validate` passes; generated suite still 138/138 with 0 TS errors.

https://claude.ai/code/session_01R7ZCxbgcsv3WfsSJrYdSCi

### DIFF
--- a/.github/ISSUE_TEMPLATE/request-recipe.yml
+++ b/.github/ISSUE_TEMPLATE/request-recipe.yml
@@ -1,12 +1,12 @@
-name: Request a Comfy Cloud recipe
-description: Request a new Civitai Comfy Cloud (customComfy) recipe your App can invoke.
+name: Request a Comfy on Civitai recipe
+description: Request a new Comfy on Civitai (customComfy) recipe your App can invoke.
 title: "[Recipe request] "
 labels: ["recipe-request"]
 body:
   - type: markdown
     attributes:
       value: |
-        **Comfy Cloud recipes are server-registered + code-reviewed** — an App can't ship its own graph. The iframe only picks a **recipe id** plus a small, per-recipe-validated `params` object; the Civitai server owns the ComfyUI graph, the resource allowlist, the checkpoint policy, and a hard per-job Buzz ceiling.
+        **Comfy on Civitai recipes are server-registered + code-reviewed** — an App can't ship its own graph. The iframe only picks a **recipe id** plus a small, per-recipe-validated `params` object; the Civitai server owns the ComfyUI graph, the resource allowlist, the checkpoint policy, and a hard per-job Buzz ceiling.
 
         Use this form to request a NEW recipe. Describe the graph/workflow you want, the params your App needs to expose, and the use case. We'll review it and, if approved, register it as a recipe id you can invoke via `buildComfyBody({ recipe, params })` (see `src/comfy.ts` in the page-money scaffold).
   - type: input
--- a/internal/cmd/app_create.go
+++ b/internal/cmd/app_create.go
@@ -29,17 +29,17 @@ immediately runnable (npm install && npm run dev:harness), test-green, and
 validates clean.
 
 The default scaffold ships TWO samples in the one app: a runnable txt2img money
-path AND a Comfy Cloud (customComfy) sample that runs a server-registered recipe
+path AND a Comfy on Civitai (customComfy) sample that runs a server-registered recipe
 (invite-only beta) — both share the estimate -> consent -> submit -> poll driver,
 switched by an on-screen mode toggle. Both work end-to-end in "npm run
-dev:harness" (mock host); see the generated README's "Comfy Cloud sample" section.
+dev:harness" (mock host); see the generated README's "Comfy on Civitai sample" section.
 
 Templates (override with --template):
   static      a no-build page app (index.html + a tiny JS, no build step)
   page-vite   a vite + React page app (config-as-code build: buildCommand + outputDir)
   page-money  a vite + React + TS full-page (W10) money-path app wired to the
               published App SDK (estimate -> consent -> submit -> poll -> Buzz
-              spend); includes a txt2img + a Comfy Cloud (customComfy) sample
+              spend); includes a txt2img + a Comfy on Civitai (customComfy) sample
               [default for create]
 
 The display name can be free-form ("My Cool Block"); it is slugified for the
--- a/internal/cmd/app_create_cmd_test.go
+++ b/internal/cmd/app_create_cmd_test.go
@@ -108,10 +108,10 @@ func TestAppCreateDefaultsToPageMoney(t *testing.T) {
 	if !strings.Contains(stdout, "beta access") {
 		t.Errorf("create should gate submit/dev-tunnel under a beta-access heading:\n%s", stdout)
 	}
-	// The Comfy Cloud (customComfy) sample is surfaced (finding #2: it was
+	// The Comfy on Civitai (customComfy) sample is surfaced (finding #2: it was
 	// undiscoverable) — honestly flagged as invite-only beta.
-	if !strings.Contains(stdout, "Comfy Cloud") {
-		t.Errorf("create should surface the Comfy Cloud sample in the output:\n%s", stdout)
+	if !strings.Contains(stdout, "Comfy on Civitai") {
+		t.Errorf("create should surface the Comfy on Civitai sample in the output:\n%s", stdout)
 	}
 	// The trimmed message drops the old multi-line Buzz/OAuth/personal-key
 	// paragraph (dev:live now lives in dev-token/README, not the scaffold banner).
@@ -137,7 +137,7 @@ func TestAppCreateDefaultsToPageMoney(t *testing.T) {
 	}
 }
 
-// TestAppCreateHelpMentionsComfy guards finding #2: the Comfy Cloud sample must
+// TestAppCreateHelpMentionsComfy guards finding #2: the Comfy on Civitai sample must
 // be discoverable from `civitai app create --help` (it was previously described
 // as a txt2img-only app).
 func TestAppCreateHelpMentionsComfy(t *testing.T) {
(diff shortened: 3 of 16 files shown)


ZacxDev:
### README budget guard: measured, but it cannot live in this PR

Feedback was that my seam guard here pins manifest ↔ `generation.ts` — the pair that did **not** drift — while the README (the writer with the actual drift history) is unguarded. Correct, and I wrote + mutation-verified the README guard.

It can't ship in #200. The guard asserts the rendered README names the budget the rendered manifest emitted, and each PR moves exactly one of those two values:

| tree | manifest | README | guard |
|---|---|---|---|
| **#200 alone** | 300 | 40 | 🔴 RED |
| **#199 alone** (`d6c7889`) | 40 | 300 | 🔴 RED |
| **merged** | 300 | 300 | ✅ GREEN |

Since **#200 must merge before #199**, adding it here makes this PR red and deadlocks the chain. I did not push it and did not weaken it. This PR stays green: `1332 RUN / 1330 PASS / 2 SKIP / 0 FAIL`, 16 pkgs.

Full evidence + a ready-to-apply patch is posted on **#199**, where it goes green once that branch rebases onto a `main` containing this PR.

Note this PR does still ship manifest `300` against a README saying `40` — that inconsistency is real, is what #199 fixes, and the merge of the two is clean (`git merge-tree --write-tree` exit 0, verified).

https://claude.ai/code/session_01R7ZCxbgcsv3WfsSJrYdSCi

PR #208, opened 2026-08-05
fix(scaffold): the static and page-vite templates never posted BLOCK_READY (#206)

### LINKED ISSUE #206: Scaffolded 'static' (default) and 'page-vite' apps never send BLOCK_READY — the host times out and collapses them
Two of the three `civitai app init` templates — including the **default** — appear to scaffold an app that can never signal readiness to the host, so the host times out and collapses it. This is the same *symptom* as #196 but a different cause, and the dev-tunnel preflight added in #198 deliberately does not catch it (it is gated to SDK projects).

## The contract

The host marks a block ready only on a `BLOCK_READY` message. In `civitai/civitai`, `src/components/AppBlocks/PageBlockHost.tsx` has exactly one transition into `ready`, inside the `BLOCK_READY` handler:

```js
// PageBlockHost.tsx:864
const off = onMessage<unknown>('BLOCK_READY', () => {
  ...
  setStatus((current) => (current === 'loading' ? 'ready' : current));  // :871
```

Enumerating every `setStatus(` in that file, `:871` is the only path to `'ready'`. The failure path is `:733`, `loading → 'timeout'`, after `BLOCK_READY_TIMEOUT_MS = 10_000` (`pageBlockHostLogic.ts:571`).

A timeout is not a visible error card. `hostRenderDecision.ts` documents:

> Every TERMINAL-FAILURE state collapses → `'collapse'` (the host renders null; the slot takes no space — no visible broken card, no reserved gap). This covers: … `'timeout'` (no BLOCK_READY in 10s) …

## Who actually sends it

Not the SDK. `@civitai/app-sdk@0.31.0`'s runtime JS contains no `BLOCK_READY` at all — only a type in `dist/blocks/messages.d.ts:467` and a README line. The sender is `@civitai/blocks-react`:

```js
// @civitai/blocks-react@0.39.0 — dist/internal/iframeTransport.js:311
this.dispatch('BLOCK_READY', { height: 0 });
```

## What each template does

| template | depends on `blocks-react` | messages it sends | reaches `data-block-ready="true"` |
|---|---|---|---|
| `page-money` | yes | via `IframeTransport` | yes |
| `page-vite` | **no** | `RESIZE_IFRAME` only | **no** |
| `static` (**default**) | **no** | `RESIZE_IFRAME` only | **no** |

Grepping the whole of `internal/scaffold/templates/page-vite/` and `.../static/` for `BLOCK_` returns **zero** matches. Each contains exactly one `postMessage` call, sending `RESIZE_IFRAME` — which `PageBlockHost.tsx` does not reference at any line, so it is not treated as a liveness signal either.

All three manifests declare `page` + `iframe`, so all three route through the same host path. `static` is the default: `civitai app init "My Block"` with no `--template` produces it (`internal/cmd/app_init.go:59`).

## Ruled out

- **A server-injected ready shim.** No bootstrap/shim injection found in the block-serving path; every server-side `BLOCK_READY` reference is the analytics beacon (`app-views.service.ts`, `track/block-render.ts`), not a sender.
- **A render mode that skips the handshake.** `hostRenderDecision` collapses on `timeout` regardless, and `IframeHost.tsx:1956` gates `data-block-ready` on the same `isReady`.

## Not verified

I have not loaded a scaffolded `page-vite` or `static` app in a real host browser and watched it time out — the tunnel endpoint is not exposed and the feature is kill-switched, the same limit recorded on #198. The evidence above is read directly from both sides of the contract (host source and template source), not observed end-to-end. Worth one browser confirmation before acting, since a fix touches the default template.

## Possible directions

Not prescribing one — the trade-off is a product call:

1. **Emit `BLOCK_READY` from the raw templates.** A few lines of `window.parent.postMessage({ type: 'BLOCK_READY', height: 0 }, '*')` keeps them dependency-free. Needs care: `IframeTransport` freezes `parentOrigin` to the first sender and dedupes re-inits (`iframeTransport.js:291-295`), so a hand-rolled version should not re-emit on every render.
2. **Give them `@civitai/blocks-react`.** Correct by construction, but turns the "no-build" `static` template into a build.
3. **Document them as non-hosted previews** and change the default to one that works.

Whichever way, a scaffold seam guard like `internal/scaffold/dev_embed_contract_test.go` — asserting each template that declares `page` emits the ready ack — would keep this from regressing silently.

ZacxDev:
Fixes #206.

## What was broken

A `page` app is not shown by the Civitai host until it posts `BLOCK_READY` —
that handler (`PageBlockHost.tsx:864-871`) is the **only** transition into
`ready`. All three templates declare `page`, but only `page-money` ever sent
it, and only because `@civitai/blocks-react`'s `IframeTransport` acks
internally (`dist/internal/iframeTransport.js:311`). `static` — the
`civitai app init` **default** — and `page-vite` are deliberately SDK-free, so
they had to say hello themselves and did not. `BLOCK_READY` appeared nowhere in
this repo.

Original omission, not a regression: `git log --follow` shows both template
files declared `page` from their first commit.

Two knock-on problems, fixed in the same commit:

* **Both templates modelled the wrong envelope.** The host dispatches
  `event.data.payload` to its subscribers (`usePostMessage.ts:189`), so every
  block→host message is `{ type, payload }`. Both templates put fields at the
  top level. The host ignores the `BLOCK_READY` payload, so this would not have
  broken the ack itself — it teaches the wrong shape for every message an
  author adds next.
* **`RESIZE_IFRAME` is N/A for a page surface** (`hostHandlerParity.ts:100-110`:
  the page renders the iframe full-viewport and does not size to content). Both
  templates demoed it as "the minimal useful thing an app can do". Removed.

## Two corrections to the issue text

The issue is right about the bug and wrong about the symptom, both times in the
direction of making it look milder than it is:

1. **It does not collapse the slot silently.** `pageBlockHostLogic.ts:476-479`
   is explicit: *"a FULL-PAGE surface that collapses is just a blank viewport.
   So a failed page renders a message INSIDE the trust frame instead."* The user
   sees *"This app didn't load in time … We already retried 2 times
   automatically."*
2. **It is ~37s, not 10s.** `MAX_AUTO_RETRIES = 2` with
   `AUTO_RETRY_BACKOFF_MS = [2000, 5000]` → 10+2+10+5+10.

Also: the snippet the issue suggests as the fix reproduces the envelope bug
(top-level fields, `payload: undefined` at the host).

## Measured control matrix

Real `PageBlockHost` from `civitai@851f759c20`, real Chromium, apps served over
HTTP into an iframe with the manifest's own `allow-scripts allow-forms` sandbox
(no `allow-same-origin` ⇒ opaque `null` origin). "after" apps re-scaffolded from
this branch's binary and rebuilt, not hand-edited.

| case | before | after (2 runs) |
|---|---|---|
| `static` (default template) | **never ready** → failure card ~37s | **ready 453 / 459 ms** |
| `page-vite` | **never ready** → failure card ~37s | **ready 477 / 434 ms** |
| `page-money` (regression check) | ready 431–472 ms | ready **489 / 435 ms** |
| inert page, no script (negative control) | timeout ~37s | **still times out ~38.7s** |
| naive `BLOCK_READY` poster (positive control) | ready | ready 269 / 218 ms |

Both templates also verified at `trustTier: 'internal'` (pinned origin rather
than opaque `null`): ready in 470–531 ms. After the fix each block frame posts
**exactly one** `BLOCK_READY` and **zero** `RESIZE_IFRAME`; before, it posted
`RESIZE_IFRAME` ≥1 and no `BLOCK_READY` — which is what separates "the JS never
ran" from "the JS ran and said the wrong thing".

The negative control still failing is what makes the four greens readable.

## Why ack on `BLOCK_INIT`, not on load

**On-load is not broken.** An on-load poster was *measured working* in the same
harness (ready in 178–265 ms). Ack-on-INIT is a robustness choice, and the
comments in the emitter say so rather than claiming otherwise:

* **No race.** The host registers its `BLOCK_READY` subscriber inside an effect
  and silently **drops** a message whose type has no subscriber yet, with no
  retry (`usePostMessage.ts:152-153`). It re-posts `BLOCK_INIT` every 400 ms
  until acked (`iframeInitController.ts:51`), so answering INIT cannot be early.
* **No `'*'`.** The init event carries `event.origin`, so the ack is addressed
  at the host instead of broadcast.
* **Correct timing.** The host fades its launch overlay and enables
  `pointerEvents` on ready, so acking too early reveals an empty interactive
  frame.
* **It is what the SDK does**, so both paths behave alike.

Repeats are a no-op (mirroring `iframeTransport.js:293-296`), which matters
because the host's inbound channel is rate-limited to 30 msg/sec **across all
types**, with the budget consumed after subscriber lookup — re-acking every
retry could starve `BLOCK_ERROR`.

## Structure: one authority, copied verbatim

`internal/blockproto` holds the emitter as embedded bytes;
`scaffold.Render` copies them **with no `text/template` pass** into the path
`Template.ReadyAckPath()` names (`civitai-host.js` / `src/civitai-host.js`).
No per-template `.tmpl` copy, because two hand-maintained copies drift and drift
here is invisible locally: the app renders fine and dies only inside the real
host. `static` stays no-build (a bare `<script src>`, no package.json);
`page-vite` stays SDK-free (a side-effect `import`).

This is the **fourth vendored mirror** in the CLI, alongside `schema/`, the slot
registry, and the dev-embed dotenv mirror. AGENTS.md item 11 records it and
item 10's "third vendored mirror" is amended.

## What the two guards prove — and what they don't

**Guard A** (`ready_ack_contract_test.go`, in `make ci`) enumerates
`AllTemplates()` and decides subject-hood from **evidence in the rendered
output**, never a list in the test: subject if the rendered manifest declares
`page`; expected to ship the emitter unless the rendered `package.json` depends
on `@civitai/*`. It asserts byte-equality with `blockproto.ReadyAckSource()` and
that the entry point actually **reaches** the file (index.html `<script src>`,
or an import in the module index.html loads) — as independent sub-tests, after
stripping comments. Plus a count positive control (≥2 SDK-free page templates
examined) and a shape predicate with a negative corpus (comment / README prose /
wrong envelope / untriggered post / `'*'` target), each required to be rejected
**by the expected rule**.

**Guard A does not prove the ack fires.** No Go static check can prove
JavaScript executes.

**Guard B** (`ready_ack_runtime_test.go`) loads each shipped emitter in node
under a `window` stub, plays the host's side, and reads the outbound message:
exactly one, only after a real `BLOCK_INIT`,
`{type:'BLOCK_READY',payload:{height:0}}`, addressed at the host origin and not
`'*'`, with a second `BLOCK_INIT` producing no second ack. Env-gated
(`CIVITAI_CHECK_SCAFFOLD_RUNTIME=1`, mirroring `pins_guard_test.go`) so
`make ci` stays Go-only and offline, and wired into the **daily**
`bump-scaffold-pins` workflow as its own job — a gated test with no scheduled
runner is a meaningless green. It validates its own harness first: an inert
fixture and a naive-ack fixture go through the same assertions and must be
faulted, and the naive one doubles as the count positive control (it posts on
every message, so the driver's counter is watched moving off zero).

**Guard B does not prove the real host accepts the ack** — the stub is this
repo's reading of the contract. The browser matrix above is that evidence.

### Mutation results

Each failure confirmed to come from *that* assertion's own error message, with
`go test ./...` otherwise clean:

| mutation | outcome |
|---|---|
| delete the `<script>` tag from `static/index.html` | A **wiring** fails; A byte-equality **passes** |
| render an emitter one byte off canonical | A **byte-equality** fails; A wiring **passes** |
| drop `"page"` from `static`'s manifest | static skipped **and** the count control fails |
| emitter posts `{type,height}` with no `payload` | A shape **and** B fail |
| emitter acks on load instead of `BLOCK_INIT` | A shape **and** B fail |
| invert `event.source !== window.parent` | **A passes entirely, whole Go module green, B fails** |

The last row is the reason Guard B exists: the emitter keeps every structural
feature Guard A can see and is completely inert. (The brief predicted
ack-on-load would play that role; measured, Guard A's shape rules catch that one
too — the inverted source check is the real blind spot. Reported as measured,
not as predicted.)

## Verification

* `make ci` green: 16 packages ok, **1545 `=== RUN` / 1542 `--- PASS` / 0
  `--- FAIL` / 3 `--- SKIP`**, no `panic: test timed out`, `gofmt -s -l .`
  silent, `go mod tidy` a no-op.
* `golangci-lint v2.12.2`: **0 issues**.
* `actionlint` clean on the changed workflow, with a positive control (it
  rejects `runs-on: ubuntu-lates`, so its acceptance is a real green).
* `CIVITAI_CHECK_SCAFFOLD_RUNTIME=1 go test -run TestScaffoldedReadyAckActuallyFires`
  green, controls reported.

## Deliberately deferred

* **The `internal/antipattern` rule** for a page manifest with no ready-ack, and
* **a `civitai app validate` warning** for already-scaffolded apps.

Both are the right follow-ups — an author who scaffolded last week still has a
broken app and nothing tells them. They are not in this PR because landing them
now would turn the scaffolds' own self-validation red before the templates that
satisfy them have shipped.

## Note for the schema owners (not touched here)

`schema/app-block.manifest.schema.json` describes `iframe.resizable` in
size-to-content terms, which is arguably wrong for a `page` app — the host
renders those full-viewport and `RESIZE_IFRAME` is marked N/A for
`PageBlockHost`. All three templates still set `"resizable": true`. That file is
a vendored mirror and ask-first per AGENTS.md, so it is left alone; flagging it
rather than fixing it.

### DIFF
--- a/.github/workflows/bump-scaffold-pins.yml
+++ b/.github/workflows/bump-scaffold-pins.yml
@@ -49,11 +49,14 @@ jobs:
   # in node against a fake host and asserts a single correctly-addressed
   # `{type:'BLOCK_READY',payload:{height:0}}` goes out.
   #
-  # It rides this workflow because this is the repo's DAILY scheduled runner and
-  # the test is env-gated (`CIVITAI_CHECK_SCAFFOLD_RUNTIME=1`) so `make ci` stays
-  # Go-only and offline. A gated test with no runner is a green that means
-  # nothing. It is deliberately independent of the `bump` job below: it must run
-  # on every sweep, including the (usual) no-op ones where the pins are current.
+  # 🔴 This is the DRIFT runner, NOT the gate. The gate is the identically-named
+  # job in ci.yml, which runs on every pull request. A daily job cannot block a
+  # merge — a regression would land green and surface up to 24h later, on a
+  # workflow with no dependent job — so this copy alone would be a guard that
+  # gates nothing. It exists to catch drift on `main` between PRs (a merged-tree
+  # interaction, a node bump, an npm-side change), and is deliberately
+  # independent of the `bump` job below so it runs on every sweep, including the
+  # usual no-op ones where the pins are already current.
   ready-ack-runtime:
     runs-on: ubuntu-latest
     permissions:
--- a/.github/workflows/bump-scaffold-pins.yml
+++ b/.github/workflows/bump-scaffold-pins.yml
@@ -41,6 +41,43 @@ permissions:
   pull-requests: write
 
 jobs:
+  # Runtime half of the ready-ack contract (issue #206). The offline guard in
+  # `internal/scaffold/ready_ack_contract_test.go` proves each SDK-free page
+  # template ships a byte-identical copy of `internal/blockproto/ready-ack.js`
+  # and that the entry point loads it — it CANNOT prove the ack fires, because
+  # no Go static check can prove JavaScript executes. This job runs the emitter
+  # in node against a fake host and asserts a single correctly-addressed
+  # `{type:'BLOCK_READY',payload:{height:0}}` goes out.
+  #
+  # It rides this workflow because this is the repo's DAILY scheduled runner and
+  # the test is env-gated (`CIVITAI_CHECK_SCAFFOLD_RUNTIME=1`) so `make ci` stays
+  # Go-only and offline. A gated test with no runner is a green that means
+  # nothing. It is deliberately independent of the `bump` job below: it must run
+  # on every sweep, including the (usual) no-op ones where the pins are current.
+  ready-ack-runtime:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-go@v5
+        with:
+          go-version: "1.25"
+          cache: true
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "22"
+
+      # -v so the run REPORTS its controls (an inert emitter and a naive ack are
+      # both driven through the same assertions and must be rejected) rather than
+      # just an exit code.
+      - name: scaffolded ready-ack actually fires
+        env:
+          CIVITAI_CHECK_SCAFFOLD_RUNTIME: "1"
+        run: go test ./internal/scaffold -run TestScaffoldedReadyAckActuallyFires -count=1 -v
+
   bump:
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,7 +88,7 @@ jobs:
           CIVITAI_CHECK_PUBLISHED_PINS: "1"
         run: go test ./internal/scaffold -run TestScaffoldPinsSatisfyPublished -v
 
-  # Guard B for the block->host ready-ack (issue #206), and the half that GATES.
+  # Guard B for the block->host ready-ack (issue #206) — the RUNTIME half.
   #
   # `internal/scaffold/ready_ack_contract_test.go` runs in the build-test job
   # above and proves each SDK-free page template ships a byte-identical copy of
@@ -99,9 +99,15 @@ jobs:
   #
   # This job executes the emitter in node against a fake host. It is env-gated
   # so `make ci` stays Go-only and offline. The daily `bump-scaffold-pins` sweep
-  # runs it too, but a DAILY runner gates nothing: a regression would merge green
-  # and surface up to 24h later, on a workflow with no dependent job. A PR check
-  # is what makes it a guard.
+  # runs it too, but a daily runner reports up to 24h AFTER a merge, on a
+  # workflow nobody watches — running it per-PR puts the result in front of a
+  # reviewer while the change is still open.
+  #
+  # 🔴 NOTE: this REPORTS, it does not GATE. `main`'s required contexts are
+  # exactly `pins-vs-published` and `scaffold-currency` (no rulesets), so a red
+  # run here does not block a merge — `build-test` doesn't either. Adding this
+  # to the required contexts is the outstanding maintainer decision recorded in
+  # AGENTS.md item 11.
   #
   # -v so the run REPORTS its controls (an inert emitter and a naive ack are
   # driven through the same assertions and must be rejected) rather than just an
@@ -170,29 +176,47 @@ jobs:
         working-directory: ./_ci-page-vite
         run: |
           set -euo pipefail
-          bundles=$(find dist -name '*.js' -type f | wc -l)
-          echo "JS files emitted into dist/: $bundles"
-          if [ "$bundles" -lt 1 ]; then
+
+          # Every grep below is `|| true`. Without it this whole step is a trap:
+          # grep exits 1 on NO MATCH, pipefail propagates that through the pipe,
+          # and `set -e` kills the step AT THE ASSIGNMENT — so the count line and
+          # the ::error:: annotation never print. Measured: the step exited 1
+          # having printed only the positive-control line, which is precisely the
+          # ambiguity the positive control exists to remove ("is the harness not
+          # reading the bundle, or did the ack not survive the build?").
+          # It failed CLOSED, so this was a diagnosis bug, not a safety one.
+
+          # The exact JS files under test, captured once and reused, so the
+          # assertions cannot be satisfied by dist/index.html or a public/
+          # passthrough that never goes through the bundler.
+          mapfile -t js < <(find dist -name '*.js' -type f | sort)
+          echo "JS files emitted into dist/: ${#js[@]}"
+          printf '  %s\n' "${js[@]}"
+          if [ "${#js[@]}" -lt 1 ]; then
             echo "::error::the build emitted no JS into dist/ — nothing to check, so a pass here would be meaningless"
             exit 1
           fi
+
           # POSITIVE CONTROL for the grep itself, FIRST: a string that must be in
           # the bundle whatever the handshake does. If this misses, the path or
           # the pattern is wrong and the two assertions below prove nothing.
-          control=$(grep -rlF 'CI Vite Block' dist/ | wc -l)
+          control=$(grep -lF 'CI Vite Block' "${js[@]}" | wc -l || true)
           echo "positive control (app display name): $control file(s)"
           if [ "$control" -lt 1 ]; then
-            echo "::error::positive control failed — this step is not reading the built bundle"
+            echo "::error::positive control failed — this step is not reading the built bundle, so its verdict on the ready-ack means nothing"
             exit 1
           fi
+
+          rc=0
           for needle in BLOCK_INIT BLOCK_READY; do
-            n=$(grep -rlF "$needle" dist/ | wc -l)
+            n=$(grep -lF "$needle" "${js[@]}" | wc -l || true)
             echo "$needle: $n file(s)"
             if [ "$n" -lt 1 ]; then
               echo "::error::$needle is absent from the built bundle — the ready-ack did not survive the build, so the shipped app never says hello (issue #206)"
-              exit 1
+              rc=1
             fi
           done
+          exit "$rc"
 
       - name: validate
         working-directory: ./_ci-page-vite
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,6 +88,116 @@ jobs:
           CIVITAI_CHECK_PUBLISHED_PINS: "1"
         run: go test ./internal/scaffold -run TestScaffoldPinsSatisfyPublished -v
 
+  # Guard B for the block->host ready-ack (issue #206), and the half that GATES.
+  #
+  # `internal/scaffold/ready_ack_contract_test.go` runs in the build-test job
+  # above and proves each SDK-free page template ships a byte-identical copy of
+  # `internal/blockproto/ready-ack.js` that the entry point RESOLVES to. It
+  # cannot prove the ack FIRES — no Go static check can prove JavaScript
+  # executes, and an emitter with an inverted `event.source` check passes every
+  # static assertion while being completely inert.
+  #
+  # This job executes the emitter in node against a fake host. It is env-gated
+  # so `make ci` stays Go-only and offline. The daily `bump-scaffold-pins` sweep
+  # runs it too, but a DAILY runner gates nothing: a regression would merge green
+  # and surface up to 24h later, on a workflow with no dependent job. A PR check
+  # is what makes it a guard.
+  #
+  # -v so the run REPORTS its controls (an inert emitter and a naive ack are
+  # driven through the same assertions and must be rejected) rather than just an
+  # exit code.
+  ready-ack-runtime:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-go@v5
+        with:
+          go-version: "1.25"
+          cache: true
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "22"
+
+      - name: scaffolded ready-ack actually fires
+        env:
+          CIVITAI_CHECK_SCAFFOLD_RUNTIME: "1"
+        run: go test ./internal/scaffold -run TestScaffoldedReadyAckActuallyFires -count=1 -v
+
+  # Rot-guard: build the page-vite template. It is the only SDK-free template
+  # with a build step, and nothing else in CI builds it — `template-page-money`
+  # and `scaffold-currency` both scaffold page-money.
+  #
+  # Since #206 its entry module has a HARD RESOLUTION DEPENDENCY on
+  # `./civitai-host.js`, so a wrong import path is a build failure rather than a
+  # subtle one. Measured: `import '../nonexistent/civitai-host.js'` survived the
+  # entire Go suite, and would have shipped a template that cannot build at all.
+  #
+  # The bundle assertion is separate from Guard A on purpose. Guard A pins the
+  # SOURCE tree; Vite output is what the platform actually serves, and a
+  # tree-shaken or mis-externalised side-effect import would leave a bundle with
+  # no handshake in it. Control pair, both measured locally: with the import the
+  # bundle carries BLOCK_INIT/BLOCK_READY, with it commented out it carries
+  # neither — while the positive control (the app's display name) matches in
+  # both cases, proving the grep reads the bundle either way.
+  template-page-vite:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-go@v5
+        with:
+          go-version: "1.25"
+          cache: true
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "22"
+
+      - name: scaffold page-vite
+        run: go run ./cmd/civitai app init "CI Vite Block" --dir ./_ci-page-vite --template page-vite
+
+      - name: install
+        working-directory: ./_ci-page-vite
+        run: npm install --no-audit --no-fund
+
+      - name: build
+        working-directory: ./_ci-page-vite
+        run: npm run build
+
+      - name: the ready-ack survives bundling
+        working-directory: ./_ci-page-vite
+        run: |
+          set -euo pipefail
+          bundles=$(find dist -name '*.js' -type f | wc -l)
+          echo "JS files emitted into dist/: $bundles"
+          if [ "$bundles" -lt 1 ]; then
+            echo "::error::the build emitted no JS into dist/ — nothing to check, so a pass here would be meaningless"
+            exit 1
+          fi
+          # POSITIVE CONTROL for the grep itself, FIRST: a string that must be in
+          # the bundle whatever the handshake does. If this misses, the path or
+          # the pattern is wrong and the two assertions below prove nothing.
+          control=$(grep -rlF 'CI Vite Block' dist/ | wc -l)
+          echo "positive control (app display name): $control file(s)"
+          if [ "$control" -lt 1 ]; then
+            echo "::error::positive control failed — this step is not reading the built bundle"
+            exit 1
+          fi
+          for needle in BLOCK_INIT BLOCK_READY; do
+            n=$(grep -rlF "$needle" dist/ | wc -l)
+            echo "$needle: $n file(s)"
+            if [ "$n" -lt 1 ]; then
+              echo "::error::$needle is absent from the built bundle — the ready-ack did not survive the build, so the shipped app never says hello (issue #206)"
+              exit 1
+            fi
+          done
+
+      - name: validate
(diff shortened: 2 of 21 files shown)


ZacxDev:
## Audit round — corrections to this PR's body, and what changed

An independent blind audit found **no 🔴 findings** (the shipped emitter is correct and the browser matrix held up), but it falsified two claims I made in the description above. Correcting them here rather than quietly editing the body, plus the fixes in `eddaa44` and `738ecb6`.

### ❌ Correction 1 — the wiring guard did NOT prove what I said it proved

The body says Guard A asserts "that the entry point actually **reaches** the file". That was **not true as written**. It matched a **basename**, so two mutants shipped a broken app past the *entire* suite — both measured green, both reproducing #206:

| mutant | what it ships |
|---|---|
| `static/index.html` → `src="./vendor/civitai-host.js"` | a 404; the app never says hello |
| `page-vite/main.jsx` → `import '../nonexistent/civitai-host.js'` | `npm run build` fails outright |

This is the "spelled rather than structural" class: the check could pass while the hazard existed in a different shape.

**Fixed.** Every reference is now **resolved** — a `<script src>` against the document root, an import against the importing module's directory — and compared with where the emitter actually is; a bare specifier or a URL is not a project path at all. On failure it prints each reference with what it resolved to and whether that file exists. Backed by a 10-case corpus on synthetic trees (wrong subdir, unresolvable relative, bare specifier, HTML comment, commented-out import, orphan module, no script tags, CDN URL ending in the name). The claim now holds; AGENTS.md item 11 records that a basename or `strings.Contains` match must never come back.

### ❌ Correction 2 — Guard B ran on nothing that could block a merge

The body presents Guard B as the thing that catches an inverted `event.source` check. It was **env-gated with no PR runner**: `ci.yml` never set `CIVITAI_CHECK_SCAFFOLD_RUNTIME`, so it skipped on every pull request and ran only on a daily schedule, on the default branch, with no dependent job. The exact failure I advertised it catching would have **merged green** and surfaced up to 24h later.

**Fixed.** `ready-ack-runtime` is now a job in `ci.yml` (per-PR, gating); the daily copy stays, relabelled as the drift runner it actually is.

This matters more than it sounds: **four of sixteen mutations are invisible to the entire offline suite** and are caught only by Guard B — inverted source check, inverted type check, latch-before-post, and a listener attached inside a `setTimeout`.

**Gate proven, not assumed** — #211 was opened as a temporary draft carrying the inverted source check (a mutation `go test ./...` passes cleanly), purely to watch `ready-ack-runtime` go red on a pull request. It did. That PR is closed and its branch deleted; it was never mergeable and exists only as evidence.

### Also fixed in this round

- **Nothing in CI built `page-vite`** — only page-money was ever scaffolded and built. Since this PR its entry module has a hard resolution dependency on `./civitai-host.js`, so mutant #2 above would have shipped a template that cannot build. New `template-page-vite` job builds it and asserts **the ack survives bundling** (Guard A pins the source tree; Vite output is what the platform serves). Control pair measured: with the import the bundle carries `BLOCK_INIT`/`BLOCK_READY`, with it commented out it carries neither, while the positive control (the app's display name) matches either way.
- **`acked` latched before a post that can throw.** Now measured rather than spec-derived: in Chromium 1228, `postMessage(msg, 'null')` throws `SyntaxError: Invalid target origin 'null'`, and `event.origin` is the string `"null"` when the sender is at an opaque origin. Latching first would silence the emitter forever after one throw. The latch now sets after the post; pinned by a new `--throw-first-post` driver mode (which also models a browser's per-listener exception isolation — previously a throwing emitter would have killed the driver rather than being observed). Not reachable through today's `PageBlockHost`, and the comment says so.
- **The emitter pins the sender WINDOW, not the sender ORIGIN.** The comment said "only the framing host" and the README said "the host's real origin" — both overstated it. It establishes *our embedder*, not *Civitai*. Sound for this message (the ack carries no data); insufficient for anything you add next, which both scaffold READMEs explicitly invite. All four surfaces corrected, with an explicit warning that an origin allowlist is required before handling any message carrying data. **Decision recorded, not left implicit:** no allowlist is vendored into the emitter — the real set (prod, preview subdomains, `dev-*.civit.ai`) is platform state that moves without notice, would become a *fifth* mirror needing lockstep, and breaks the dev tunnel when wrong, while `@civitai/blocks-react` already maintains it from `VITE_BLOCK_ALLOWED_PARENT_ORIGINS`.
- **The documented upgrade path could wedge the host into a fake-ready state.** "If you later adopt `@civitai/blocks-react`, delete it" was a parenthetical, and the only line reversing "don't delete it". Keeping both means whichever handshake answers the first `BLOCK_INIT` clears the host's retry loop *and* readiness timeout; the SDK's `waitForInit` then rejects at 10s and the host sits "ready" showing an app that never started — no retry, no card. Worse than #206 and silent. Now a callout with the mechanism in four places.
- **`useBlockResize` doc was self-inconsistent** — described as "for the non-page block surfaces" while page-money (a page template) calls it. Corrected: `RESIZE_IFRAME` is ignored on a page surface and the hook is surface-agnostic, so calling it there is a harmless no-op.
- **Provenance path was wrong** in two comments and in the body above: `usePostMessage.ts` is at `src/components/AppBlocks/`, **not** `src/hooks/` — it never lived there.
- **`app init`'s handshake lines had no test**, unlike every sibling caveat. Added, with a control asserting page-money does *not* get them, and proven to fail when the block is deleted.

### Mirror currency — the reference moved

The platform checkout moved `851f759c20` → **`35a9598dc9`**. **`BLOCK_HELLO` now exists host-side** (`iframeInitController.notifyHello`, `hostHandlerParity.ts`, host browser tests); the body's "zero host references" was true when measured and is now stale.

Re-verified independently rather than taken on report: `notifyHello()` leaves the retry interval and readiness timeout armed by `start()` untouched, and its own comment forbids it becoming a precondition for sending init — so a block that never announces is served by the unchanged retry loop, which is exactly what this emitter relies on. **This PR is unaffected.** The emitter deliberately does not send `BLOCK_HELLO` (cost: one retry tick of latency; benefit: no second vendored message), and that is now recorded in AGENTS.md along with the fact that the init-fragment fast path ships gated off and refuses `surface === 'dev-tunnel'`. Emitting it stays out of scope.

### Re-verification (the fix round resets the gate)

**Full mutant battery re-run — 16 mutations, ZERO survivors**, including the classes the first battery missed:

| mutation | offline suite | Guard B |
|---|---|---|
| wrong subdir in `<script src>` | **FAIL** | pass |
| unresolvable relative import | **FAIL** | pass |
| `<script>` tag deleted | **FAIL** | pass |
| import commented out (still greppable) | **FAIL** | pass |
| bare specifier | **FAIL** | pass |
| rendered emitter drifts one byte | **FAIL** | pass |
| static manifest drops `"page"` | **FAIL** | pass |
| wrong envelope (no `payload`) | **FAIL** | **FAIL** |
| acks on load instead of `BLOCK_INIT` | **FAIL** | **FAIL** |
| wrong bind: listener on `document` | **FAIL** | **FAIL** |
| swapped operands `(origin, message)` | **FAIL** | **FAIL** |
| **inverted source check** | pass | **FAIL** |
| **inverted type check** | pass | **FAIL** |
| **latch before post** | pass | **FAIL** |
| **listener inside a `setTimeout`** | pass | **FAIL** |
| `app init` drops the handshake caveat | **FAIL** | pass |

Each offline failure was attributed to the specific test that killed it (four other tests render these same templates, so "something went red" would not be attribution).

⚠️ **One process note worth recording:** the first battery run reported `static manifest drops "page"` as a **survivor**. It wasn't — my mutation regex stopped at the `}` inside `{{ .Name }}`, so the mutation silently never applied and a no-op read exactly like a guard gap. The battery now has a positive control that each mutation actually changed the tree, and the real mutation is killed by the count control.

**Browser matrix re-run** (real `PageBlockHost`, real Chromium, apps re-scaffolded and rebuilt from this branch), 2 runs:

| case | result |
|---|---|
| `static` | ready **459 / 483 ms** |
| `page-vite` | ready **453 / 491 ms** |
| `page-money` (regression) | ready **152 / 458 ms** |
| inert control | **still times out ~38.6 s** ✅ |

Both templates also verified at `trustTier: internal` (413–456 ms), and each posts **exactly one** `BLOCK_READY` and zero `RESIZE_IFRAME`. The inert control still failing is what makes the rest readable. (Incidentally visible in the tap: page-money now posts `BLOCK_HELLO`, independently confirming the mirror-currency finding above.)

**`make ci`**: 16 packages ok, **1560 `=== RUN` / 1557 `--- PASS` / 0 `--- FAIL` / 3 `--- SKIP`**, no timeout panic, `gofmt -s -l .` silent. `golangci-lint v2.12.2`: 0 issues. `actionlint` (with shellcheck) clean on both changed workflows, with a positive control proving it can go red on them.

### Still deferred, unchanged

The `internal/antipattern` rule and the `civitai app validate` warning for already-scaffolded apps; and `"resizable": true` in the manifests / the `schema/` `resizable` wording (vendored, ask-first).

ZacxDev:
## Round 3 — a false claim about gating, an unreachable diagnostic, and a vacuous control

A delta re-audit of `4bc88ad..738ecb6` came back **safe to merge, no 🔴** — the ten prior findings verified fixed by mutation. It then found three more 🟡 in the delta plus leftovers. All verified against the code before implementing; **all were real**, and one of them falsifies a claim in my previous comment on this PR.

### ❌ Correction — "a PR check is what makes it a guard" was false

My last comment and `ready_ack_runtime_test.go` both said the per-PR `ready-ack-runtime` job **blocks the merge**. It does not.

```
$ gh api repos/civitai/cli/branches/main/protection --jq .required_status_checks.contexts
["pins-vs-published","scaffold-currency"]
$ gh api repos/civitai/cli/rulesets
[]
```

`ready-ack-runtime` is not a required context. Neither is `template-page-vite`. **Neither is `build-test`.** A red run reports on the PR and stops nothing — so #211's red job was evidence the check *detects* the regression, not evidence it would have *blocked* it. That distinction was mine to make and I got it wrong.

**Fixed by correcting the wording, not by changing branch protection.** Making these contexts required is the honest alternative, but it is a repo-policy change affecting every open PR including #210, and that is the maintainer's decision — it has not been given, so I did not take it. `ready_ack_runtime_test.go`, `ci.yml` and AGENTS.md item 11 now say what is true (runs and reports on every PR; not a required context) and record making it required as the outstanding step, with the measurement inline so the next reader can re-check it.

### 🟡 Every `::error::` in the bundle assertion was unreachable

`set -euo pipefail` + `n=$(grep -rlF "$needle" dist/ | wc -l)`: grep exits 1 on no match, pipefail propagates, `set -e` kills the step **at the assignment**. Reproduced — the step exited 1 having printed only the positive-control line, never the count and never the annotation.

It failed *closed*, so this was a diagnosis bug, not a safety one — but the log could not distinguish "the harness isn't reading the bundle" from "the ack didn't survive the build", which is exactly the ambiguity the positive control exists to remove. Each grep now carries `|| true`, the loop accumulates `rc` instead of dying, and the needles are scoped to the JS files actually counted (previously a match in `dist/index.html` or a `public/` passthrough would have satisfied them).

Watched working, both directions, with a decoy planted in `dist/index.html`:

```
=== ack survives (expect rc=0) ===          === ack MISSING + decoy (expect rc=1) ===
JS files emitted into dist/: 1              JS files emitted into dist/: 1
positive control: 1 file(s)                 positive control: 1 file(s)
BLOCK_INIT: 1 file(s)                       BLOCK_INIT: 0 file(s)
BLOCK_READY: 1 file(s)                      ::error::BLOCK_INIT is absent …
--- step rc=0 ---                           BLOCK_READY: 0 file(s)
                                            ::error::BLOCK_READY is absent …
                                            --- step rc=1 ---
```

The decoy is the scoping proof: the strings are present in `dist/index.html` and the counts are still 0.

### 🟡 The new `app_init` control test was vacuous

Replacing the guard at `app_init.go:238` with `if true` **survived the entire suite**. `page-money`'s `ReadyAckPath()` is `""`, so the line printed with an empty filename — `   performs the host handshake — keep it. …` — and my assertion, which checked that `civitai-host.js` was *absent*, was satisfied **by the very bug it should have caught**. Spelling an assertion on a value the failure mode erases can only ever pass.

Now asserted on the phrase `"performs the host handshake"`, which an empty path cannot erase, and shared as a constant between the positive and negative tests. **Watched going red** on the exact mutation that survived.

On the conjunct question: I checked, and **neither half should go** — they're coextensive across today's three templates but each guards a different failure. `ReadyAckPath() != ""` is what stops the empty-filename line (demonstrated above); `!NeedsHarness()` is what keeps *"there's no host to send BLOCK_INIT"* from being printed at a template whose `dev:harness` mounts one. A future template shipping both needs the wording reconsidered, not a conjunct deleted. Recorded as a comment at the site.

### 🟡 Dead branch removed, three live branches pinned

- **`://` half of the URL guard: genuinely dead, removed.** Verified by exhaustively resolving candidate specifiers with each half toggled: dropping `://` changes the accept/reject outcome for *no* input, including `https://x/../civitai-host.js`. The CDN corpus case is rejected by the fallthrough and now says so instead of appearing to exercise a guard it never reached.
- **`//` half: NOT dead — load-bearing, and now pinned.** Without it, `//evil.example.com/../civitai-host.js` has its leading slash stripped, `filepath.Join` cleans the `..`, and it resolves to **exactly the emitter's path** and is accepted. New corpus case; deleting the guard now fails.
- **`errNoScriptTags` early return** and **`?`/`#` suffix stripping** (N6) were both correct but deletable without any test failing. Pinned: the corpus gained an optional `wantErr` so a rejection has to come from the *expected* branch (previously the no-script-tags case was green via the fallthrough either way), plus an accept case for `import './civitai-host.js?url'`.

### Leftover sweep

Two sites still carried the "**the real host origin**" overclaim I corrected everywhere else in round 2 — `ready_ack_contract_test.go:92` and `ready_ack_runtime_test.go:151`, both introduced in `a456c95` and missed by that sweep. Both now say "BLOCK_INIT carries the sender's origin". Re-grepped: zero remaining occurrences across `internal/`, `README.md` and `AGENTS.md`.

### Not taken

**N5** (resolver over-strict on layouts no shipped template uses — fail-closed and documented), **N7** (residual accept-shapes: string-literal-only import, `type="text/plain"`, `<script>` inside `<template>`), branch protection, and the still-deferred antipattern rule / `validate` warning — all out of scope by agreement.

### Addendum — AGENTS.md item numbering (prose only)

#210 (`feat/generate`) adds items **11–16** to the same list while this PR adds item **11**; `git merge-tree --write-tree` exits **1** on `AGENTS.md`. Both PRs also independently corrected the intro's stale *"items 5–8 cover `civitai app metrics`"* to *"5–9"* while adding a different clause, so the correct merged sentence is neither side's and a take-one-side resolution silently drops the other's range.

Per agreement this PR does **not** resolve that conflict, renumber, or touch #210 — item 11 stays item 11, and #210 renumbers to 12–17 on its next rebase. What's added is three sentences recording the convention so the next collision doesn't re-derive it: append-only, numbered by arrival, the PR merging **second** renumbers **its own** items, never an existing one (items cross-reference each other by number — item 10 → item 11, item 9 → item 6 — and so do code comments), and re-grep every `item N` after any renumber.

Applied the rule to the file as it stands: all four references resolve correctly. The Layout section carries no item references in this branch.

PR #210, opened 2026-08-05
feat(generate): add `civitai generate` and `civitai workflows`

ZacxDev:
Adds `civitai generate` — submit a generation, watch it, download the outputs — plus
`civitai workflows list|get|cancel`. Built in five gated phases; every gate was
re-verified independently rather than accepted from the phase that produced it.

## Surface

```bash
civitai generate "a cat wearing sunglasses"                   # everything server-defaulted
civitai generate "a cat" --checkpoint 1759168 --lora 250712:0.8 --quantity 4
civitai generate "a cat" --dry-run                            # price it, spend nothing
civitai generate --input graph.json                           # raw graph passthrough
civitai generate "a cat" --print-input > g.json               # graduate flags -> file
civitai workflows list | get <id> | cancel <id>
```

Five content flags, not twelve. Everything omitted stays reachable via `--input`, which is
also why the CLI needs no per-engine code.

## 🔴 The spend story, stated plainly

There is **no server-side spend ceiling reachable from a personal API key.** The quote is
not binding, realized cost can exceed it with **no refund**, and the per-API-key
`buzzLimit` does **not** bind this path (the orchestrator meters a separate System key, so
a user who sets it is falsely reassured). `--max-cost` is an **estimate check, not a cap**;
`--timeout` stops *waiting*, not *paying*. All of that is in `--help` and the README in
those words, because the failure mode is someone running an unattended loop believing
otherwise.

Defaults are conservative: `--dry-run` costs nothing, a TTY run confirms, and a **non-TTY
run refuses without `--yes`** so a pipeline can never silently spend.

## Why it looks the way it does

The server is **permissive, not validating** — every one of these was measured live, and
each returns HTTP 200:

| Input | Server behaviour |
|---|---|
| `steps: 0` | accepted; degenerate, half-price, wrong job |
| unknown key | silently dropped; your parameter vanishes |
| invalid sampler | silently ignored |
| **nonexistent model id** | **200, and you are billed for a substituted model** |
| `quantity: 10000` | clamped to 10, no error |

So the CLI resolves `--checkpoint`/`--lora` against `/api/v1/model-versions/{id}` **before**
submitting, turning that fourth row into a hard local 404. It is a live lookup, not a
vendored table, so it carries no drift cost.

Three things that look like bugs and are load-bearing (all in `AGENTS.md` items 11–16):

- **Two different tRPC envelopes.** `whatIf` takes the graph flat; `generate` takes it
  nested under `.input`. A mismatch returns 200 with a constant ~8 cost while every
  "did it 200?" test stays green.
- **`externalId` is minted unconditionally.** The server retries submits 3× with no
  idempotency key — its own comment says the caller must supply one — so omitting it risks
  a triple charge.
- **`DownloadPresigned` carries no credential.** `isTrustedDownloadHost` trusts any
  `*.civitai.com` host and blobs come from `orchestration.civitai.com`, so folding this
  back into `DownloadFile` would send a full-scope personal API key to the orchestrator.

## Verification

- **1786 tests, 1783 pass, 0 fail, 3 skip** (all three skips pre-existing/env-gated), gofmt
  and vet clean — run on the **merged tree** against current `origin/main`, not just the
  branch.
- Guards were **mutation-tested**: each was broken deliberately and shown to fail *with its
  own message*, then reverted. Two guards initially **survived** and the tests were
  strengthened until they didn't.
- **One real generation** at the phase-2 gate: estimated 8 Buzz, **charged exactly 8**,
  1024×1024 JPEG downloaded with no credential attached, re-attach via `workflows get`
  confirmed. That single data point does not disprove the overage risk above.

`claudedocs/civitai-generate-design.md` carries the full trace and an appendix retracting
every design claim that did not survive implementation.

## Requires

A **personal API key** with the AI Services scope. An OAuth `civitai login` token cannot
generate — the `civitai-cli` OAuth client lacks the AI scope bits, which needs a
server-side migration to change. `civitai whoami` already reports the capability.

## Known gaps

- `--input` is **txt2img-only** on purpose. The server's prompt audit is gated on
  `'prompt' in data`, and the validated graph is rebuilt from declared nodes only, so a
  graph carrying its prompt elsewhere could slip the audit. The CLI should not be that
  vector; other workflows are refused with that reason.
- Silent model **substitution within an ecosystem** remains undetectable — the tRPC reply
  carries no substitution record (only the App Blocks path does). Worth a server-side ask.
- No img2img yet.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -94,6 +94,12 @@ These go beyond the global defaults because this repo's release pipeline
   `root.go`.
 - `internal/{scaffold,validate,pkgzip,manifest,api,config,auth}` — the building
   blocks behind those commands.
+- `internal/genapi` — the orchestrator **generation** client behind
+  `civitai generate` and `civitai workflows …`: tRPC transport, the two envelope
+  shapes, the graph payload, model-version resolution. Deliberately not in
+  `pkg/civitai` (that is the public read/download SDK) because generation is a
+  money-spending surface whose wire shape is not a public contract. Read
+  items 11–16 before touching it.
 - **Module root** (`package cli`, `main.go` + `schema.go`) exists *only* to
   `go:embed` the vendored `schema/` and `examples/`. It is not the executable.
 
@@ -153,10 +159,16 @@ func newWhoAmICmd() *cobra.Command {
 ## Intentional decisions that look wrong (read before "fixing")
 
 Items 1–3 are deliberate mirrors of the platform (items 4 and 8 are deliberate
-*non*-mirrors); items 5–8 cover `civitai app metrics`, the CLI's only analytics
-read path. The durable fix for the mirroring is a server-side
-`civitai app validate` endpoint that calls the real `BlockManifestValidator` —
-until that exists, vendoring is on purpose.
+*non*-mirrors); items 5–9 cover `civitai app metrics`, the CLI's only analytics
+read path; items 11–16 cover `civitai generate`, the CLI's only path that
+**spends the user's money irreversibly**. The durable fix for the mirroring is a
+server-side `civitai app validate` endpoint that calls the real
+`BlockManifestValidator` — until that exists, vendoring is on purpose.
+
+The generate items pull in the opposite direction from items 1–3, and that is
+deliberate: `validate` mirrors the platform because a local answer is *cheaper*
+than a round-trip, while `generate` mirrors nothing because a local answer that
+is *wrong* costs real Buzz. Read item 12 before adding any check to that path.
 
 1. **`civitai app validate` is a best-effort LOCAL mirror, not the authority.**
    The server-side `BlockManifestValidator`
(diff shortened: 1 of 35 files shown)


ZacxDev:
## Correction to this PR's description, after an adversarial audit

The body claims: *"Guards were mutation-tested: each was broken deliberately and shown to fail with its own message."* **That is overstated, and I'm correcting it here rather than editing the body, since anyone who already read it saw the stronger claim.**

Every **deletion-shaped** mutant does die. But an independent audit ran **50 semantic mutants** — operand swaps, branch inversions, off-by-ones, stale re-binds — and **8 survived**. Deletion is the weak half of mutation testing; the surviving ones delete nothing:

| Mutation | Where | Consequence |
|---|---|---|
| `case "y","yes"→ok; default→cancel` rewritten as `case "n","no"→cancel; default→ok` | `generate.go:979` | The spend prompt's **default-deny** is unpinned. Tests only feed `"y"` and `"n"`, so a refactor to a deny-list keeps the suite green while `[y/N]` silently becomes `[Y/n]`. The current code *is* correct (bare Enter, `maybe`, `yy`, `Ye` all cancel) — it just isn't asserted. |
| drop `filepath.Base` in `pendingPath` | `generate_state.go:82` | The documented traversal guard on user-supplied `--external-id` has **no test**. Verified working manually. |
| `cost > balance` → `cost >= balance` | `generate.go:709` | The `cost == balance` boundary on the money check is untested. Fails safe. |
| replay trigger `status == 401` → `status != 200` | `genapi/client.go:175` | Nothing pins that the mutation replay happens *only* on 401. Not a live double-charge — the unconditional `externalId` bounds it — but that's half of the argument in AGENTS.md item 15, and only the other half is asserted. |

Plus four lower-impact survivors (clamp-warning boundary, a redundant overwrite stat, a `<400` vs `<300` status check, and the unread crash-recovery record).

### Two findings fixed in 2cb3193, both on the money story

1. **The realized charge was only ever printed on `--no-wait`.** The default waiting path announced the workflow and never showed what was billed — on a command whose documented behaviour is that realized cost can exceed the estimate with no refund. Now printed, and printed as *nothing* (not `0`) when the server reported no cost.
2. **"You MAY have been charged" fired when nothing had left the machine.** `StatusOf(err)==0` conflated *"the request went out and the reply was lost"* with *"the request never left this process"*. A cancelled context produced the second, and the warning sent the user to re-attach with an `externalId` no workflow exists behind — which submits a **new** job and charges for real.

Both are mutation-verified by **branch inversion**, not deletion: inverting the ctx guard makes the false warning appear *and* the genuine one disappear, so both directions are pinned. Each assertion carries a positive control.

### Not fixed here

The four table rows above are real gaps in *test coverage*, not defects — the audit verified the code behaves correctly in each case. The unread `PayloadHash` (its comment claims a re-attach comparison that nothing implements) and the missing confirmation on `workflows cancel` are worth follow-ups.

### What the audit confirmed independently

The envelope asymmetry, reproduced live and blind: flat = **12**, nested = **8**, empty `{}` = **8** — i.e. a nested payload is byte-identically the default-job price. Buzz balance identical before and after; zero spend; no mutation procedure was ever called. Also confirmed: no double-charge path, no unapproved-charge path, no credential leak, the priced graph is byte-identically the submitted graph, and the 401 replay re-sends the same `externalId`.

Audit verdict: **safe to merge**, no deploy-blocking findings.

ZacxDev:
## Correcting my own previous fix — 2cb3193 was a regression, reverted in 75cb114

My last comment said fix #2 was "mutation-verified by branch inversion". It was — and it was still **wrong**. A delta re-audit of `79b76c5..2cb3193` caught it, and I've verified both of its findings independently.

### What I broke

2cb3193 added an early return on `ctx.Err() != nil` before the "you MAY have been charged" warning, on the theory that a cancelled context means the request never left the process.

**That is true only for the microseconds before the POST.** Ctrl-C *during* the round trip cancels the same context, and that window is up to the submit timeout (~120 s) — a slow orchestrator hop being exactly why a user reaches for Ctrl-C. So the guard traded a harmless false warning in a microsecond window for **silence in a two-minute window where the charge is real**.

Silence there is the worst outcome available: this warning is the **only** surface that hands the user their `externalId`, and nothing reads the crash-recovery record back (`readPendingGeneration` has no production consumer). The user is left with an error, no workflow id, no idempotency key, and no hint that money may have moved.

Red/green, cancel fired *inside* the submit seam with `submitCalls == 1` as the control:

| build | stderr |
|---|---|
| 79b76c5 (before my "fix") | `⚠ …MAY still have been accepted and charged… --external-id 91da0a02-…` |
| 2cb3193 (my "fix") | **empty** |

### Three claims in my own comment were also false

- *"`--timeout` produces the never-sent case"* — `--timeout` doesn't bound the submit at all; it's applied in `waitAndCollect`. The submit's bound is a client-level `http.Client{Timeout}`, which leaves the caller's ctx live.
- *"re-attach with an externalId no workflow exists behind… submits a NEW job and charges them for real"* — overstated. The orchestrator dedupes on `(userId, externalId)`, so re-attaching is strictly **safer** than the bare re-run the warning talks the user out of.
- *"Same guard the poll loop already applies at `generate_wait.go:190`"* — the line number is right and the analogy is **inverted**. There, a cancelled ctx routes to `printReattach` and says the job *"has already been charged"*. I cited a precedent that does the opposite of what I did, and per this repo's "a comment is a claim too" rule that comment would have led the next maintainer to defend the regression.

### The fix in 75cb114

The uncertainty belongs in the **message**, not the control flow. We cannot observe whether the request bytes were written, so the warning **always** fires; a cancelled context is worded *"the submit was interrupted before any answer arrived"* rather than claiming server silence. The false comments are gone.

### And the seam I'd missed

The re-audit also found that my charge assertions all called `printSubmitted` **directly** — so a mutant making the *call site* pass a nil cost **survived all 913 package tests**. My headline claim ("the default path shows the charge") was asserted about the renderer, never about the path. Now covered by waiting-path tests that drive `runGenerate`, with a nil-cost contrast.

Both regressions are now pinned: re-introducing the `ctx.Err()` return fails two tests, and the call-site mutant fails one.

Suite: **1795 tests, 1792 pass, 0 fail, 3 skip** (the three known env-gated), gofmt and vet clean.

### Still open (unchanged, follow-ups)

The four coverage gaps from the previous comment stand — the spend prompt's default-deny, the `--external-id` traversal guard, the `cost == balance` boundary, and the 401-only replay trigger. All are missing *tests*, not defects. Plus `PayloadHash`, whose comment still claims a re-attach comparison nothing implements.

ZacxDev:
## Audit loop converged — round 4 clean, no findings

Four rounds, stopping on a clean round rather than on my say-so.

| round | scope | outcome |
|---|---|---|
| 1 | full PR, **blind** | no 🔴; 2 🟡 fixed; 8 of 50 semantic mutants survived |
| 2 | `79b76c5..2cb3193` | 🔴 **my fix #2 was a money-safety regression** — reverted |
| 3 | `2cb3193..75cb114` | functionally clean; 🟡 guards were *spelled*, not structural |
| 4 | `75cb114..5517be6` | **no findings** |

Round 4 independently re-ran every kill: blanking the interpolated `externalId` now fails 2 tests (previously survived all 917), deferring `printSubmitted` past the wait fails the ordering assertion and *only* that one — proving the new assertion discriminates rather than duplicating. It also confirmed the round-3 kills were preserved through the rewrite, that the precondition guards can't pass vacuously (`lastExternalID` is assigned only inside the submit closure), and `-race` clean.

**The round-2 finding is the one worth remembering.** I added a `ctx.Err()` early return to stop a false "you MAY have been charged" warning. It was mutation-verified by branch inversion and it was still wrong: `ctx.Err()` cannot distinguish *cancelled before the POST* (microseconds, no charge possible) from *cancelled mid-flight* (up to ~120s, charge entirely possible). It traded a harmless false warning for **silence in the window where the charge is real** — and that warning is the only surface handing back the `externalId`. Mutation testing proved the guard did what I intended; it could not tell me the intent was wrong. Only the delta re-audit did.

Outstanding, all **test-coverage gaps rather than defects** (each verified to behave correctly): the spend prompt's default-deny, the `--external-id` traversal guard, the `cost == balance` boundary, and the 401-only replay trigger. Plus `PayloadHash`, whose comment claims a re-attach comparison nothing implements — either implement it or reword.

State: `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`, 0 commits behind `main`, 1795 tests / 1792 pass / 0 fail / 3 pre-existing skips, gofmt + vet clean, `-race` clean. Ready for human review.

PR #211, opened 2026-08-05
DO NOT MERGE — temporary: prove the ready-ack CI gate goes red

ZacxDev:
Temporary draft, opened only to watch `ci.yml`'s new `ready-ack-runtime` job FAIL on a pull request, then closed immediately.

The commit inverts `event.source !== window.parent` in the vendored emitter, making it completely inert. Measured locally: `go test ./...` is fully green with this mutation — it is invisible to every offline check. If `ready-ack-runtime` does not go red here, the gate added in #208 does not gate.

Evidence for #208. Closing as soon as the job reports.

### DIFF
--- a/.github/workflows/bump-scaffold-pins.yml
+++ b/.github/workflows/bump-scaffold-pins.yml
@@ -49,11 +49,14 @@ jobs:
   # in node against a fake host and asserts a single correctly-addressed
   # `{type:'BLOCK_READY',payload:{height:0}}` goes out.
   #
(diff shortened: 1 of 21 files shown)


ZacxDev:
Purpose served: `ready-ack-runtime` went **red** here — the only failing check, with Guard B's own assertions (the emitter answered a foreign window and addressed the ack at the wrong origin) — while `build-test` (the entire offline Go suite) passed, exactly as measured locally.

That is the evidence for #208: the per-PR gate added there really does block this class of regression. Closing; never mergeable.

PR #212, opened 2026-08-06
fix(validate): tell existing page apps they never post BLOCK_READY (#206)

ZacxDev:
Follow-up to #208 (merged as `4018e2c`), closing #206 for apps that already exist.

`4018e2c` fixed the `static` and `page-vite` **templates**, which declared a
`page` surface and never posted `BLOCK_READY`. **Every app scaffolded before that
commit is still broken, and nothing tells its author** — it renders perfectly
everywhere you can look locally and is replaced by a failure card in the real
host after ~37s. This was deliberately deferred out of #208 so it would land
*after* the templates were fixed; landing it first would have turned the shipped
scaffolds' own self-validation red.

Two pieces, with deliberately different evidentiary strength.

---

## 1. `internal/antipattern`: the `resize-iframe-page` rule (a GATE)

`hostHandlerParity.ts` marks `RESIZE_IFRAME` **N/A for `PageBlockHost`** — a page
block fills the surface and does not size to content — so a page app posting it
is dead code that also models the wrong envelope for every message the author
writes next. Both raw templates demoed it until #206 removed them.

This one **is** allowed to fail the build (it fails the `scaffold-currency` job),
because it fires on a **literal that is present in the file**. No inference.

It is the first non-REST entry in that denylist, which until now was strictly
"dead REST route behind a host bridge". Two precision decisions:

- It matches only a **matching `'` or `"` pair** — deliberately *not* the
  backtick form. This repo's README, both scaffold READMEs and AGENTS.md item 11
  all name `` `RESIZE_IFRAME` `` in markdown backticks *precisely to tell authors
  not to post it*. A gate that fails the documentation of its own rule is a false
  positive at a correct project.
- It **assumes a page surface rather than reading a manifest.** `Rule` is a
  per-line regex with no manifest context by construction, and every template
  this CLI ships declares `page`. A non-page template would need that assumption
  revisited, not the rule deleted — the code says so.

## 2. `civitai app validate`: a page-app-without-ack advisory (a WARNING)

🔴 **`Warnings`, never `Errors`, and that must stay true.** The lockfile check
next door earns hard-error status because the platform *provably* fails (`npm ci`
dies). This one infers **runtime** behaviour from **static text**, and there are
correct projects it cannot read: an ack from a bundled dependency, a framework
wrapper, a code-split chunk, an extension the scan does not open. Hard-failing on
a heuristic is the false-warning-at-a-correct-project failure AGENTS.md item 10
spent four measured corrections avoiding — and `--strict` already gives anyone
who wants a gate one.

**The dependency is evidence, and it is checked FIRST.** On
`@civitai/blocks-react` the ack comes from `IframeTransport` and the literal
`BLOCK_READY` **never appears in `src/`** — measured on a rendered `page-money`:
zero occurrences anywhere outside `node_modules`. A source-only scan therefore
false-warns at *every correct money-path app on the platform*, which is the worst
possible outcome for advisory output. `TestSDKTemplateIsQuietOnlyBecauseOfTheDependency`
pins that the dep gate is the **only** reason page-money is accepted, so that
branch cannot rot into a case that passes for an unrelated reason.

Other constraints, each from the #208 audit rounds:

- **Only fires where it can observe.** No `page` surface, no check. Unreadable
  `package.json`, no check. Unwalkable tree, no check. **And reading ZERO source
  files reports "no finding"** — a zero-hit scan over a zero-file tree is
  indistinguishable from a scanner wired to nothing. That is not defensive
  padding: a manifest-only fixture in `warnings_test.go` failed without it.
- **Source only** — never `outputDir` (not built yet, gitignored), never
  `node_modules`, never `.md`.
- **Comments are stripped, and that is load-bearing.** Both SDK-free templates
  carry a source comment reading "The ONE message a page app must send is
  `BLOCK_READY`" — and it *survives deleting `civitai-host.js`*. Without the
  strip the check is inert on the exact population it was written for. Measured
  both ways.
- **Placement.** `warningChecks` runs unconditionally *including under
  `ManifestOnly`*, which `app init` uses to self-check the template it just
  wrote. A check that reads `src/` lives in the `projectState` branch beside
  `lockfileChecks`, or init's self-validation starts reading files that are not
  its business.

---

## Measured before / after

Built binary, real scaffolds on disk. The unit tests are not a substitute for
this.

| Project | `civitai app validate` | antipattern gate |
|---|---|---|
| Scaffold reconstructed from the templates at `4018e2c^` | ⚠ **warns** (exit 0) | ✗ **`app.js:18 — resize-iframe-page`** |
| Fresh `static` (today) | ✓ valid, silent | ✓ clean |
| Fresh `page-vite` (today, + lockfile) | ✓ valid, silent | ✓ clean |
| Fresh `page-money` (today, + lockfile) | ✓ valid, silent | ✓ clean |
| Fresh `static`, `civitai-host.js` deleted | ⚠ **warns** (exit 0; exit 1 under `--strict`) | ✓ clean |
| Fresh `page-vite`, `src/civitai-host.js` deleted | ⚠ **warns** (exit 0) | ✓ clean |

The pre-fix reconstruction is the real population: it trips **both** checks, for
two independent reasons (no emitter, and a live `RESIZE_IFRAME` post).

## What these checks do NOT prove

- **Not that the ack fires.** No static check can — an emitter with an inverted
  `event.source` guard satisfies every assertion here. That is what Guard B
  (`ready_ack_runtime_test.go`) exists for, and there is no runtime guard at all
  for an app an author already has. The advisory says this in its own message.
- **Not that a silent project is correct.** A missing warning means "the literal
  appears in code somewhere", which is weak evidence deliberately treated as
  conclusive: being generous costs a missed warning, being strict costs a false
  one.
- **The server remains the authority.** `validate` is still a best-effort local
  mirror.

## Mutation results

Four rounds, 42 mutant runs. Every mutation was verified to have **actually
changed the tree** (`git diff --stat` non-empty for the expected file) before its
verdict was read, and each round carried a **null mutant** (comment-only edit)
that had to SURVIVE — it did, every time. Kills are **attributed**: several
neighbours in this repo render the same templates, so an unattributed "KILLED" is
not evidence about my guard.

Final state — all attributed to a test added in this PR:

| Mutant | Killed by |
|---|---|
| page gate inverted | `TestReadyAckWarning/a_non-page_app_never_warns` |
| page gate deleted | `.../a_non-page_app_never_warns` |
| `declaresPage` accepts a null page | `.../an_explicit_"page":_null_does_not_warn` |
| **SDK dep gate deleted** | `TestSDKTemplateIsQuietOnlyBecauseOfTheDependency`, `TestShippedTemplatesDoNotTripTheReadyAckWarning/page-money`, `.../page_app_depending_on_@civitai/blocks-react_does_not_warn` |
| SDK gate drops the `unknown` case | `TestReadyAckCannotObserve/an_unreadable_package.json_…` |
| dev/peerDependencies no longer count | `.../the_SDK_dep_counts_from_devDependencies_too` |
| any package.json counts as an SDK | `.../a_package.json_with_no_@civitai_dep_still_warns` |
| missing package.json → `unknown` | `TestDeletingTheEmitterMakesTheWarningFire/static` |
| comment stripping removed | `.../an_ack_named_only_in_a_comment_warns` |
| HTML comments not stripped | `.../an_ack_named_only_in_an_HTML_comment_warns` |
| string literals unprotected in the stripper | `.../a_URL_literal_does_not_swallow_the_ack_after_it` |
| markdown scanned as source | `.../an_ack_described_only_in_the_README_warns` |
| **outputDir scanned** | `.../an_ack_present_only_in_a_NON-STANDARD_outputDir_warns` |
| node_modules scanned | `.../an_ack_in_node_modules_does_not_count` |
| `scanned == 0` no longer silences | `TestReadyAckCannotObserve/a_directory_holding_only_a_manifest_…` |
| walk error warns instead of silencing | `TestReadyAckCannotObserve/an_unreadable_source_file_…` |
| advisory promoted to a hard Error | `TestReadyAckIsAdvisoryOnly` |
| check moved out of `projectState` | `TestReadyAckSkippedByManifestOnly` |
| check removed entirely | `TestReadyAckWarning/page_app_with_no_ack_at_all_warns` |
| resize rule removed from the registry | `TestScanDir/RESIZE_IFRAME_posted_from_a_page_app_fails` |
| resize rule widened to bare prose | `TestScanDir/docs_naming_RESIZE_IFRAME_in_backticks_are_not_flagged` |
| resize rule narrowed to `'` only | `TestResizeIframeBoundary` |
| resize rule's `Replacement` emptied | `TestScanDir/RESIZE_IFRAME_posted_from_a_page_app_fails` |

**The first sweep left four of these unproven, and the second commit here is the
fix.** Three were real gaps and one was a weak mutant of mine:

- **`"page": null`** — nothing covered it, so `declaresPage`'s `v != nil` half
  was indistinguishable from a presence-only test.
- **the outputDir skip** — the fixture used `dist`, which is *also* in the
  name-based skip list, so an **earlier check always won** and the
  manifest-driven branch never executed. The mutation disabling it survived the
  entire suite. A non-standard outputDir (`public`) is the only way in.
- **the string-literal case of the comment stripper** — the URL fixture put the
  URL and the ack on *different lines*, so deleting the quote handling only ate
  the rest of the URL's own line.
- **the walk-error branch** — initially written off as unreachable by any fixture
  (chmod is a no-op for root). It is not: a dangling symlink with a scanned
  extension fails `os.ReadFile` deterministically for any user. The "expected to
  survive" was a rationalisation for an untested branch.

Two further mutants were **weak, not surviving guards**, and were re-expressed:
one left a variable unused and only broke the build; one replaced a prefix of
`Replacement` while leaving the string non-empty.

## Verification

- `make ci` green. Full suite: **1613 `=== RUN`, 1609 `--- PASS`, 0 `--- FAIL`,
  4 `--- SKIP`**, 16 packages `ok`, 0 `FAIL`, 0 `panic: test timed out`. All four
  skips are pre-existing env-gated tests (`TestScanDirFromEnv`,
  `TestSubmissionsCapDriftAgainstLiveAPI`, `TestScaffoldPinsSatisfyPublished`,
  `TestScaffoldedReadyAckActuallyFires`).
- `gofmt -s -l .` prints nothing, against 213 `.go` files found in the tree (the
  count is the positive control — a silent `gofmt` over zero files looks
  identical to a clean one).
- The env-gated `ready-ack-runtime` guard run locally with
  `CIVITAI_CHECK_SCAFFOLD_RUNTIME=1`: PASS, including its two inert/naive
  controls, on node v26.5.0.
- The live scaffold-and-validate demonstration in the table above, with the
  binary built from this branch.

## Deliberately not done

- **No `app doctor` command.** Out of scope.
- **`schema/` untouched** (vendored, ask-first). Worth flagging though:
  `iframe.resizable`'s description still reads *"Whether the host honours
  RESIZE_IFRAME messages"*, which is arguably wrong for a page app — the host
  marks that message N/A on the page surface regardless of the flag. AGENTS.md
  item 11 already notes it as a schema-side wording issue; it needs a
  server-side decision, not a unilateral edit here.
- **No new dependency; templates unchanged.**
- **AGENTS.md numbering:** this takes item **12**, per the append-only convention
  added in `4018e2c`. #210 (`feat/generate`) is also open and adds items — per
  that convention, whichever of us merges **second** renumbers **its own** new
  items and re-greps the `items N–M` range clauses.

Closes the #206 gap for existing apps.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -103,11 +103,13 @@ jobs:
   # workflow nobody watches — running it per-PR puts the result in front of a
   # reviewer while the change is still open.
   #
-  # 🔴 NOTE: this REPORTS, it does not GATE. `main`'s required contexts are
-  # exactly `pins-vs-published` and `scaffold-currency` (no rulesets), so a red
-  # run here does not block a merge — `build-test` doesn't either. Adding this
-  # to the required contexts is the outstanding maintainer decision recorded in
-  # AGENTS.md item 11.
+  # 🔴 NOTE: this both reports AND gates. `main`'s required contexts are
+  # `pins-vs-published`, `scaffold-currency`, `build-test`, `ready-ack-runtime`
+  # and `template-page-vite` (no rulesets), so a red run here blocks the merge.
+  # That was a policy change made after this job landed — it did NOT gate when
+  # it shipped. Re-measure with
+  # `gh api repos/civitai/cli/branches/main/protection` before calling any job
+  # here a gate; see AGENTS.md item 11.
   #
   # -v so the run REPORTS its controls (an inert emitter and a naive ack are
   # driven through the same assertions and must be rejected) rather than just an
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -154,7 +154,9 @@ func newWhoAmICmd() *cobra.Command {
 
 Items 1–3, 10 and 11 are deliberate mirrors of the platform (items 4 and 8 are
 deliberate *non*-mirrors); items 5–9 cover `civitai app metrics`, the CLI's only
-analytics read path. The durable fix for the mirroring is a server-side
+analytics read path; item 12 covers the two checks that tell an author their
+EXISTING app is missing the item-11 handshake. The durable fix for the mirroring
+is a server-side
 `civitai app validate` endpoint that calls the real `BlockManifestValidator` —
 until that exists, vendoring is on purpose.
 
@@ -501,6 +503,72 @@ neither one's.
       of one retry tick, and sending it would add a second vendored message for
       no correctness gain. If you reconsider, note the init-fragment fast path
       ships gated off and refuses `surface === 'dev-tunnel'` by construction.
+
+12. **The ready-ack checks for EXISTING apps are two deliberately different
+    tiers, and neither is allowed to be a hard failure.** #206 fixed the
+    templates; every app scaffolded before `4018e2c` is still broken and nothing
+    told its author. These close that gap, and the tier split is the whole
+    design:
+    - **`internal/antipattern`'s `resize-iframe-page` rule is a GATE** (it fails
+      the scaffold-currency job), because it fires on a *literal that is present
+      in the file* — no inference. It is the first non-REST entry in that
+      denylist: the package's original doctrine was "dead REST route behind a
+      host bridge", and this is a dead host MESSAGE, marked N/A for
+      `PageBlockHost` by `hostHandlerParity.ts`. It matches only a `'`/`"`-quoted
+      literal, deliberately NOT the backtick form, because this repo's README,
+      both scaffold READMEs and item 11 all name `RESIZE_IFRAME` in markdown
+      backticks precisely to tell authors not to post it — a gate that fails the
+      documentation of its own rule is a false positive. It also assumes a page
+      surface rather than reading a manifest: `Rule` is a per-line regex with no
+      manifest context, and every template this CLI ships declares `page`. A
+      non-page template would need that assumption revisited, not the rule
+      deleted.
+    - 🔴 **`validate`'s page-without-ack check is a WARNING and must stay one**
+      (`internal/validate/readyack.go`). It is the mirror image of item 3's
+      reasoning: `lockfile.go` earns hard-error status because the platform
+      *provably* fails (`npm ci` dies), whereas this infers RUNTIME behaviour
+      from STATIC TEXT and there are correct projects it cannot read — an ack
+      from a bundled dependency, a framework wrapper, a code-split chunk, an
+      extension the scan does not open. Hard-failing on a heuristic is the
+      false-warning-at-a-correct-project failure item 10 spent four measured
+      corrections avoiding, and `--strict` already lets anyone who wants a gate
+      have one.
+    - **The dependency is evidence, and it is checked FIRST.** For a project on
+      `@civitai/blocks-react` the ack comes from `IframeTransport` and the
+      literal `BLOCK_READY` NEVER appears in `src/` — measured on a rendered
+      `page-money` (zero occurrences outside `node_modules`). A source-only scan
+      therefore warns at every correct money-path app, which is the worst
+      outcome for advisory output. `sdkDependency` mirrors `civitaiSDKDeps` in
+      `ready_ack_contract_test.go` on purpose: the two answer the same question,
+      and disagreeing would mean the CLI warns at a shape its own scaffold guard
+      calls correct. `TestSDKTemplateIsQuietOnlyBecauseOfTheDependency` pins that
+      the dep gate is the ONLY reason page-money is accepted.
+    - **Comments are stripped, and that is load-bearing rather than tidy.** Both
+      SDK-free templates carry a source comment reading "The ONE message a page
+      app must send is `BLOCK_READY`", and it SURVIVES deleting
+      `civitai-host.js` — so without the strip the check is inert on the exact
+      population it was written for. Measured both ways. `.md` is excluded from
+      the scan for the same reason: a README describing the handshake is not an
+      implementation of it.
+    - 🔴 **Reading NOTHING is not finding nothing.** `emitsReadyAck` reports "no
+      finding" when it opened ZERO source files — a zero-hit scan over a
+      zero-file tree is indistinguishable from a scanner wired to nothing (a
+      broken extension table, an over-eager skip list, or `validate` pointed at a
+      directory holding a manifest and no checkout, which is the shape of every
+      manifest-only fixture in that package). This is not defensive padding: it
+      is what a manifest-only fixture in `warnings_test.go` failed on, and
+      removing it re-breaks that test.
+    - **Placement is a real constraint, not style.** `warningChecks` runs
+      unconditionally *including under `ManifestOnly`*, which `app init` uses to
+      self-check the template it just wrote. A check that reads `src/` must live
+      in the `projectState` branch beside `lockfileChecks`, or init's
+      self-validation starts reading files that are not its business.
+      `TestReadyAckSkippedByManifestOnly` pins it, with a `Dir()` positive
+      control so a green there cannot mean "the check never fires at all".
+    - **What it does NOT prove:** that the ack fires. Only that the message is
+      mentioned in code. The runtime proof is Guard B (item 11), and there is
+      none at all for an app the author already has — which is why this is
+      advisory and says so in its own message.
 
 **When you change a validation rule, keep both vendored mirrors (`schema/` + the
 ported Go checks in `internal/validate/`, including the slot registry) in sync
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -504,18 +504,18 @@ neither one's.
       thing that BUILDS an SDK-free template, and it asserts the ack survives
       bundling (Guard A pins the source tree; Vite output is what the platform
       serves).
-      🔴 **REPORTING IS NOT GATING — no ready-ack check currently blocks a
-      merge.** Measured, not assumed, via
+      🔴 **REPORTING VS GATING — measure it, never infer it from the job
+      existing.** These jobs now gate. Measured via
       `gh api repos/civitai/cli/branches/main/protection`: the required contexts
-      are exactly `pins-vs-published` and `scaffold-currency`, with no rulesets.
-      So `ready-ack-runtime`, `template-page-vite` and even `build-test` all
-      report and stop nothing. **Outstanding step:** adding `ready-ack-runtime`
-      (and arguably `build-test`) to the required contexts is what converts
-      reporting into gating. That is a repo-policy change touching every open
-      PR, so it is the maintainer's call and was deliberately not taken by the
-      agent that wrote this. Until it is, do not describe any of these jobs as a
-      gate — an earlier revision of this item, and of
-      `ready_ack_runtime_test.go`, claimed one "BLOCKS the merge". It was false.
+      are `pins-vs-published`, `scaffold-currency`, `build-test`,
+      `ready-ack-runtime` and `template-page-vite`, with no rulesets. That was a
+      deliberate repo-policy change made AFTER this item first shipped; until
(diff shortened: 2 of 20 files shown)


ZacxDev:
## Correcting two claims this PR made, both falsified by review

Posting rather than silently editing the body: the original text asserted these,
and both were wrong. I re-verified each myself before changing anything.

### ❌ "The dependency is evidence" — false for four of the six published packages

The gate tested the **`@civitai/` prefix**. Measured from fresh npm tarballs:

| package | version | acks? | evidence |
|---|---|---|---|
| `@civitai/blocks-react` | 0.39.0 | **yes** | `dist/internal/iframeTransport.js:311` — `this.dispatch('BLOCK_READY', {height: 0})` |
| `@civitai/app-sdk` | 0.31.0 | **no** | 17 runtime `.js` files, **zero** contain `BLOCK_READY`; only hits are a `.d.ts` type and the README. Declares no dependencies, so it cannot ack transitively either. |
| `@civitai/theme` | 0.2.0 | no | 0 occurrences — framework-agnostic CSS tokens |
| `@civitai/components` | 0.3.0 | no | 0 occurrences — "Plain HTML or any framework" CSS |
| `@civitai/components-react` | 0.3.0 | no | 0 occurrences |
| `@civitai/cli` | 0.1.89 | no | 0 occurrences — installs this Go binary |

The two CSS packages are exactly what a hand-written no-build page app installs,
so the scope test went **silent on a genuinely broken app**. Reproduced live:
adding `@civitai/theme` to a `static` scaffold with `civitai-host.js` deleted —
the literal #206 shape — suppressed the warning.

**Fixed** by an exact set in `blockproto.PackageAcksReady`, carrying the
per-package evidence, with the two widening mutants (`HasPrefix`, `Contains`)
now pinned. `@civitai/app-sdk` is deliberately **excluded**: an app-sdk-only
project genuinely does not ack and *should* be warned.

The same predicate now backs `civitaiSDKDeps` in the scaffold contract guard,
which had the **identical hole**: a future template depending on
`@civitai/theme` would have been classified SDK-backed and excused from shipping
an emitter — a born-broken template passing its own guard. One rule, one place.

### ❌ "It never false-warns at a correct project" — three reachable shapes did

All three reproduced live before fixing.

| shape | cause | fix |
|---|---|---|
| `var OPEN = '<!--';` above an emitter **in the same file** | `stripHTMLComments` has no string awareness, discarded everything after an unterminated marker, and ran on `.js` | HTML stripping gated to markup extensions **and** made non-lossy |
| `src` is a symlink into a shared package | `filepath.WalkDir` does not follow symlinks and does not report them as directories, so the whole source tree was skipped | directory symlinks followed, visited set on `EvalSymlinks`-resolved paths |
| `"outputDir": "src"` (a manifest with **zero** validation errors) | the manifest-driven skip removed the directory holding the emitter; `"outputDir": "."` removed the entire tree | that skip is **gone** |

The reasoning behind dropping the `outputDir` skip, since it reverses an earlier
decision in this PR: this is a **presence** check, so scanning an extra directory
can only ever *add* evidence, while skipping one can only ever *create* a false
warning. Skipping is therefore a cost decision, never a correctness one. What
remains is a fixed list of never-source names applied to **entries only**, so no
rule can remove the root. The residual trade is stated rather than hidden — a
stale build under a non-conventional output directory can retain an ack the
source lost, which is a false negative, the cheap direction.

### Also in this round

- The scan is **three-valued** (found / absent / unobservable) and only `absent`
  warns. A read error, an over-cap file or the file budget is a *gap*, not an
  absence.
- **Caps**: 5000 files, 2 MiB each. `validate` used to be a manifest-only read.
  Measured on 20,008 files with one 88 MB `.js`: **0.50 s / 316 MB RSS →
  0.04 s / 21 MB**.
- **`app submit` now prints warnings.** It branched only on `res.OK()` and
  dropped them — the highest-traffic path, and the last point before review, told
  nobody. It prints and does **not** block; blocking stays the `--strict`
  contract.
- The antipattern rule is **scoped to code extensions**. A message *name* is
  quotable in prose and usable as a JSON key; unscoped, the rule flagged its own
  documentation. `What` now reads "reference to", not "postMessage of" — the
  regex matches the quoted token anywhere in code, and claiming a call was a
  claim it cannot make.
- Corrected the "first non-REST entry" claim (`deprecated-blocks-cli` predates
  it — it is the first dead *message*, the third family).
- ⚠️ **Upgrade note now in the README**: existing `validate --strict` users can go
  red on any of the above. Called out rather than left to be discovered.

### Mutation battery: 4 rounds, 52 runs

Null mutant in every round (all survived); every mutation's diff verified
non-empty before its verdict was read. Added the three classes the audit showed
my first sweep never had: **widening** (`==`→`HasPrefix`→`Contains`,
prefix-boundary), **message content**, and the **comment-stripper internals** —
all eight of the audit's survivors were in those classes, and my original sweep
only ever *deleted* or *inverted*, never *relaxed*.

**Two real survivors found, both now fixed:**

1. **Applying the skip list to the ROOT survived the entire suite** — every
   fixture's project root is a random `t.TempDir()` name, so nothing covered the
   "entries only" claim. A project directory called `build`/`out`/`dist` is
   ordinary, and under that mutant it scanned nothing and went silent. Now pinned
   by `TestReadyAckSkipListNeverRemovesTheRoot`, each case with its own positive
   control.
2. **Deleting the visited set survived, including my cycle test** — that test was
   proving the wrong thing. A self-referential symlink terminates anyway, because
   Linux gives up with `ELOOP` after ~40 resolutions and the resulting error makes
   the scan partial. The OS guarantees termination; the visited set's real job is
   avoiding duplicate work. Only *counting files* can see that, so
   `TestReadyAckScannerVisitsEachDirectoryOnce` walks the scanner directly and
   asserts three routes to one directory read its files once. The cycle test stays
   as a backstop, relabelled to say what it does not prove.

Five further "survivors" were **broken scripts, not surviving guards** (a missing
`strings` import in `blockproto`, unbalanced parens, a pattern `gofmt` had
reformatted). A `BUILD-BROKEN` or `NOT-APPLIED` mutant is a fact about the
script; each was re-expressed to compile and genuinely remove the property, and
all then died.

### Verification

- Merged tree, not just the branch: `origin/main` merged in (#210/#213/#214
  landed meanwhile) → **1896 `=== RUN` / 1892 `--- PASS` / 0 `--- FAIL` /
  4 `--- SKIP`**, 17 packages `ok`, 0 `FAIL`, 0 `panic: test timed out`.
  `gofmt -s -l .` silent over **242** `.go` files.
- Live matrix re-run on a binary built **from the merge**: fresh
  `static`/`page-vite`/`page-money` silent; emitter deleted → warns; `+
  @civitai/theme` → still warns; `+ @civitai/blocks-react` → silent; symlinked
  `src` → silent.
- #210 took AGENTS.md items 12–17, so per this repo's convention the PR merging
  **second** renumbers its own — the ready-ack item is now **18**. Both sides had
  edited the "which items are which" sentence, the exact collision AGENTS.md
  warns about; it was rewritten to name both ranges rather than taking either
  side, and every `item N` cross-reference re-grepped.

### Still not proven

Neither check proves the ack **fires** — only the real host can, and there is no
runtime guard at all for an app an author already has. A silent project means
"the literal appears in code somewhere", which stays weak evidence deliberately
treated as conclusive.

PR #219, opened 2026-08-06
fix(validate): the ready-ack advisory passed an emitter nothing loads

ZacxDev:
`civitai app validate`'s page-without-ack advisory (#206) told authors to do
**two** things — copy `civitai-host.js` in, and **load it** from `index.html` or
the entry module — and verified only the first. A blind dogfood run of the
shipped CLI found the false pass; I reproduced it independently.

Measured on a genuinely pre-fix scaffold (`app init` from the CLI at `0ce0025`),
emitter copied in and never referenced:

```
$ grep -c 'civitai-host' index.html
0
$ civitai app validate <dir> --strict
✓ <dir> is valid          # rc=0
```

The browser never loads the file, so the app is exactly as broken as before —
and the author is told they are fine **while obeying our own advice**. That is
worse than the silence it replaced. An orphan file containing the literal
anywhere in the tree passed identically.

## What changed

The check now resolves the **entry graph** — `index.html`, its `<script src>`
tags and inline modules, and the imports below them — and asks whether a file
the *browser loads* mentions the message.

That resolution already existed in this repo, as test-only code behind scaffold
Guard A (`internal/scaffold/ready_ack_contract_test.go`). It moves to
`internal/blockproto` (`entrygraph.go` + `wiring.go`), so one predicate answers
the question for both a template and an author's project — the precedent is
exact: `PackageAcksReady` was consolidated there for the same reason, and the
second copy was wrong. The three copies of the comment stripper collapse into
`blockproto/comments.go` alongside it. Guard A's control corpus moved with the
predicate; Guard A's strength is unchanged.

**Two tiers, and the message names which one ran:**

| tier | when | what it says |
|---|---|---|
| **reachability** (strong) | the entry graph resolves **completely** | `nothing index.html loads reaches it` (orphan emitter) or `nothing index.html loads posts it either` |
| **presence only** (weak) | the graph is **incomplete** | the old whole-tree scan — *and the message states* `it did NOT check that the file is loaded` |

Still a **Warning**, never an Error, for every reason already recorded in
AGENTS.md item 18.

## The decidability boundary, and why

The obvious boundary is "static shape ⇒ decidable, bundler ⇒ not". **I did not
use it**, and the first draft that did was wrong in both directions: an ordinary
Vite project resolves perfectly (`index.html` *is* Vite's entry, which is how
Guard A has checked `page-vite` all along), while a *no-build* project can still
carry a specifier we cannot follow. Drawing the line at project type would have
left the whole `page-vite` population — half the pre-fix apps — on the broken
tier.

So the boundary is **completeness of the resolution itself**. `ResolveEntryGraph`
reports `Complete`, and *any* reference it cannot account for clears it:

- a bundler alias, or any bare specifier that is not a declared npm dependency
  (`import '@/civitai-host.js'`);
- an off-project or protocol-relative URL, or a path escaping the project root;
- a reference resolving to a file that is not there;
- a file it could not read, or an exhausted budget;
- no `index.html` at the project root at all.

A bare specifier naming a **declared dependency** is *accounted for* — it is a
package, not a file in this project — which is the only reason a React app
resolves at all. `packageDeps` therefore returns the dependency names alongside
the ack verdict, from one read of `package.json`.

🔴 **A reference to a file that is not there is a GAP, not a decided absence.**
Tempting to read it as "the browser 404s that, so it cannot be the ack" — but a
project that presumably builds having an unresolvable reference means *our model
is wrong*, and a confident finding on a wrong model is the failure this PR is
about. It also keeps two documented trades alive unchanged: deleting the emitter
from a current scaffold (leaving a dangling `<script src>`) drops to the presence
tier, and item 18's `public/`-holds-a-stale-ack false negative stays a false
negative instead of silently becoming a warning.

**"Cannot observe" does not become "everything passes."** An incomplete graph
never produces a *wiring* finding (item 10's doctrine), but the presence finding
still fires when nothing in the tree mentions the message at all, and its text
discloses its own weakness. The residual is stated rather than hidden: an
unresolvable graph **plus** the literal somewhere is silence — a false negative,
the cheap direction. An *unobservable tree scan* (unreadable file, size cap, file
budget) still gates both tiers.

## Measured before/after

`before` = `e800129` (current `main`, the shipped defect). `after` = this branch.
Projects scaffolded with the CLI at `0ce0025`, so they are genuinely pre-#206.

```
FIXTURE                BEFORE (e800129)               AFTER (this branch)
control-noemitter      WARN         rc=0 strict=1     WARN/missing   rc=0 strict=1
static-unwired         silent       rc=0 strict=0     WARN/unwired   rc=0 strict=1
static-orphan          silent       rc=0 strict=0     WARN/unwired   rc=0 strict=1
vite-unwired           silent       rc=0 strict=0     WARN/unwired   rc=0 strict=1
static-wired           silent       rc=0 strict=0     silent         rc=0 strict=0
vite-wired             silent       rc=0 strict=0     silent         rc=0 strict=0
fresh-static           silent       rc=0 strict=0     silent         rc=0 strict=0
fresh-page-vite        silent       rc=0 strict=0     silent         rc=0 strict=0
fresh-page-money       silent       rc=0 strict=0     silent         rc=0 strict=0
```

`control-noemitter` is the **harness positive control**: a pre-fix project with
no emitter anywhere, which must warn on *both* binaries. Without it, an
all-silent BEFORE column is indistinguishable from a probe wired to nothing.
`--strict` moves the exit code to 1 on exactly the warning rows and nowhere else.

## What the check now proves, and what it still cannot

Proves, on the strong tier: **a file your `index.html` actually loads mentions
`BLOCK_READY` outside a comment** — resolved, not grepped, and not basename-
matched.

Still cannot prove:

- **that the ack fires.** No static check can; an inverted `event.source` guard
  passes everything here. That is Guard B's job for templates, and there is no
  equivalent for an app the author already has.
- **wiring behind a bundler alias, a computed `import(expr)`, a plugin-injected
  entry, or a framework-owned entry with no root `index.html`.** These land on
  the presence tier, which says so in its own text.
- **anything about a project whose tree it could not fully read.** Silence, as
  before.

## Mutation results

20 mutants, restore by file-copy snapshot with a sha256 check both directions
(the first attempt restored with `git checkout` and silently left every untracked
file mutated — noted so nobody rebuilds that harness). Every mutant was confirmed
to change the file before the run; each kill is attributed to the test *and* the
assertion message.

| mutant | verdict | killed by |
|---|---|---|
| **M0 null** (comment-only) | **SURVIVED** ✓ | — (must survive) |
| M1 restore the defect (presence only) | KILLED | `TestEmitterPresentButNotLoadedWarns` — `tier = "", want "unwired"` |
| M2 old precedence (a literal anywhere wins) | KILLED | same |
| M3 every bare specifier is a package | KILLED | `TestReadyAckDecidabilityBoundary/a_bundler_alias…` |
| M4 dependency match `==` → prefix | KILLED | same |
| M5 `//` guard → `///` | KILLED | `TestReadyAckWiringPredicate/reject: protocol-relative URL…` |
| M6 drop root containment | KILLED | `…/reject: an import that escapes the project root` |
| M7 missing file is not a gap | KILLED | `TestReadyAckWarning/a_non-conventional_outputDir…` + boundary test |
| M8 stop reading inline `<script>` imports | KILLED | `…/accept: an inline module in index.html imports it` |
| M9 drop extensionless resolution | KILLED | `…/accept: an extensionless import…` |
| M10 file budget is not a gap | KILLED | `TestEntryGraphBudgetsAreGaps/the_file_budget_is_a_gap` |
| M11 drop the graph's source-extension gate | KILLED | `…/a_stylesheet_in_the_graph_is_not_ack_evidence` |
| M12 delete the presence tier's disclosure | KILLED | `TestReadyAckAdvisoriesStateTheirOwnStrength` |
| M13 unterminated `<!--` becomes lossy again | KILLED | `TestReadyAckWarning/an_unterminated_<!--_in_.html…` |
| M14 stop respecting JS string literals | KILLED | `…/a_URL_literal_does_not_swallow_the_ack_after_it` |
| M15 HTML-strip every extension | KILLED | `…/a_matched_<!--_-->_pair_in_.js_strings…` |
| M16 drop the unobservable-tree gate | KILLED | `TestReadyAckCannotObserve/a_directory_holding_only_a_manifest…` |
| M17 missing `index.html` is not a gap | KILLED | `…/an_ack_in_a_.vue_single-file_component_does_not_warn` |
| M18 make `ErrNoScriptTags` unreachable | KILLED | `…/reject: no script tags at all` |
| M19 Guard A depth `2` → `99` | **survived, then fixed** | now `…/reject: the emitter is two import levels below the entry` |

M19 is the finding worth reading: nothing in the suite held Guard A's documented
"one level of imports on purpose" contract. The corpus case that pins it is in
this PR, and re-running M19 kills it.

## Tests

New fixtures are **rendered templates, mutated** — a hand-written string fixture
carries none of the template's comments, import graph or near-miss shapes. The
mutation helper *fails* if its replacement is a no-op, so a template rename
cannot silently turn a test into "the project was never broken".

- emitter present and wired (`<script src>`, and moved to `vendor/` with the tag
  followed) → silent
- emitter present, **not** wired, on `static` **and** `page-vite` → warns
  (`unwired`)
- emitter wired via an entry-module import → silent
- orphan file containing `BLOCK_READY` → warns
- `@civitai/blocks-react` project → silent, asserted to be silent *because of the
  dependency* (`scanForReadyAck` = `ackAbsent` on that tree)
- undecidable shapes asserted explicitly: bundler alias with the emitter present
  → **silent** (documented residual), same alias with it deleted →
  `presence-only`, no root `index.html` → `presence-only`, dangling reference →
  `presence-only`, genuinely pre-fix project → `missing` (reachability)
- `TestShippedSDKFreeTemplatesResolveCompletely` — a seam guard so the strong
  tier cannot go quietly dead: if a template edit made the shipped scaffolds
  unresolvable, every real project would silently fall to the presence tier while
  every "no warning" assertion stayed green.

## `make ci`

Green. Counts from `go test ./... -v -count=1`, not an exit code:
`=== RUN 2024`, `--- PASS 2020`, `--- FAIL 0`, `--- SKIP 4`, 0 timeout panics,
18/18 packages ok, `gofmt -s -l .` silent. The 4 skips are the pre-existing
env-gated ones (`TestScanDirFromEnv`, `TestSubmissionsCapDriftAgainstLiveAPI`,
`TestScaffoldPinsSatisfyPublished`, `TestScaffoldedReadyAckActuallyFires`);
Guard B was additionally run locally with `CIVITAI_CHECK_SCAFFOLD_RUNTIME=1` and
passes, including both of its own controls.

## Docs

AGENTS.md gains **item 20** (append-only numbering, next free number) for the
tier split, the boundary and its rejected alternative, and the measured matrix.
Items **11** and **18** are corrected in place: 11 records the resolver's move
into `blockproto` and the newly-pinned depth bound; 18's "mentioned in code"
claim becomes "code the browser loads" and points at 20. README documents both
tiers plus the thing an author actually needs — that copying the emitter in is
half the fix, with the exact reference each template uses.

Refs #206, #212. No changes to `schema/`, the emitter, or any template.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -479,6 +479,18 @@ neither one's.
       Don't reintroduce it as a "minimal example". Note `iframe.resizable` in
       the vendored `schema/` still describes itself in size-to-content terms;
       that is a schema-side wording issue, not a licence to re-add the message.
+    - **The entry-graph resolver moved OUT of Guard A and into
+      `internal/blockproto`** (`entrygraph.go` + `wiring.go`, with the comment
+      stripper in `comments.go`), because `internal/validate` needed the same
+      question answered for an AUTHOR's project and had answered it with a
+      whole-tree grep instead — see item 20 for the false pass that produced.
+      Guard A now calls `blockproto.ReadyAckWiring`; its control corpus moved
+      with the predicate, into `internal/blockproto/entrygraph_test.go`. Nothing
+      about Guard A's strength changed, and the depth bound it relies on
+      (`readyAckWiringDepth = 2`: index.html's `<script src>` entries plus their
+      DIRECT imports) is now pinned by a corpus case — widening it to 99 was a
+      SURVIVING mutant, i.e. the "one level deep on purpose" contract was
+      documented and unheld.
     - **Three guards, and none subsumes the others.**
       `ready_ack_contract_test.go` (Guard A, runs in `make ci`) enumerates
       `AllTemplates()`, decides subject-hood from the RENDERED manifest and
@@ -816,9 +828,16 @@ neither one's.
       `TestReadyAckSkippedByManifestOnly` pins it, with a `Dir()` positive
       control so a green there cannot mean "the check never fires at all".
     - **What it does NOT prove:** that the ack fires. Only that the message is
-      mentioned in code. The runtime proof is Guard B (item 11), and there is
-      none at all for an app the author already has — which is why this is
-      advisory and says so in its own message.
+      mentioned in code the browser loads. The runtime proof is Guard B (item
+      11), and there is none at all for an app the author already has — which is
+      why this is advisory and says so in its own message.
+    - 🔴 **"MENTIONED IN CODE" WAS NOT ENOUGH, AND ITEM 20 IS THE REPAIR.** As
+      first shipped this bullet read "mentioned in code" full stop, and the
+      check's own remedy asked for TWO edits while verifying one. Read item 20
+      before touching `readyack.go`: the whole-tree presence scan described
+      above is now the WEAK tier, reached only when the entry graph cannot be
+      resolved, and the advisory it emits is a different string that discloses
+      the difference.
 
 19. **img2img sends `workflow: "txt2img"` PLUS `images[]`, requires
     `--ecosystem`, and uploads with NO credential — three things that each read
@@ -928,6 +947,83 @@ neither one's.
     blob URLs rather than local paths. An upload spends no Buzz, but it is a
     network write — so `--print-input`'s "reaches no money seam" claim is still
     true while its "no request at all" claim is not, with `--image`.
+
+20. **The ready-ack advisory has TWO TIERS, the message names which one ran, and
+    that disclosure is the fix — not a nicety.** `validate`'s page-without-ack
+    check (item 18) used to ask only "does any file in this tree mention
+    `BLOCK_READY`". Its own remedy asked the author for TWO edits — copy
+    `civitai-host.js` in, and LOAD it from index.html or the entry module — and
+    it verified the first. Measured on a genuinely pre-fix scaffold (`app init`
+    from the CLI at `0ce0025`, emitter copied in, never referenced):
+    `civitai app validate --strict` printed `✓ … is valid` and exited **0** for
+    an app that was still exactly as broken as before #206, and an orphan file
+    containing the literal anywhere in the tree passed identically. 🔴 **A green
+    check earned by obeying our own advice is worse than the silence it
+    replaced** — that is the whole reason this item exists, and it is the same
+    "presence is not reachability" hole Guard A had before it was rewritten,
+    regenerated at a second call site.
+    - **The tiers.** REACHABILITY (strong) runs when
+      `blockproto.ResolveEntryGraph` resolves the project COMPLETELY — a root
+      index.html, and every `<script src>`, inline-module import and import
+      below it accounted for. Then "nothing the browser loads posts
+      `BLOCK_READY`" is a real finding, and an unreferenced emitter is reported
+      as the orphan it is (`readyAckAdviceUnwired`). PRESENCE ONLY (weak) runs
+      when the graph is INCOMPLETE, and falls back to the whole-tree scan
+      (`readyAckAdvicePresenceOnly`).
+    - 🔴 **THE WEAK TIER'S MESSAGE MUST KEEP SAYING WHAT IT DID NOT CHECK**
+      ("it did NOT check that the file is loaded"), and the strong tier's must
+      NOT carry that disclaimer. A check that changes strength SILENTLY between
+      project shapes is exactly how the false pass above shipped, so the
+      strength is part of the output, not an implementation detail.
+      `TestReadyAckAdvisoriesStateTheirOwnStrength` pins both directions.
+    - **THE DECIDABILITY BOUNDARY IS COMPLETENESS, NOT PROJECT TYPE.** The first
+      draft drew it at "does the manifest declare a `buildCommand`" — bundled
+      means undecidable — and that is wrong in both directions: an ordinary Vite
+      project resolves perfectly (index.html IS Vite's entry), while a
+      *no-build* project can still carry a specifier we cannot follow. So the
+      resolver reports `Complete`, and ANY reference it cannot account for
+      clears it: a bundler alias or other bare specifier that is not a declared
+      dependency, an off-project or protocol-relative URL, a path escaping the
+      project root, a reference resolving to a file that is not there, an
+      unreadable file, an exhausted budget. **A bare specifier naming a declared
+      npm dependency is ACCOUNTED FOR** — it is a package, not a file in this
+      project — which is the only reason a React app resolves at all; that is
+      why `packageDeps` returns the dependency NAMES alongside the ack verdict
+      from one read of package.json.
+    - 🔴 **A reference to a file that is NOT THERE is a GAP, not a decided
+      absence.** It is tempting to read it as "the browser 404s that, so it
+      cannot be the ack" — but a project that presumably builds having an
+      unresolvable reference means OUR MODEL is wrong, and a confident finding
+      built on a wrong model is the failure this item is about. It also keeps
+      two documented trades alive: deleting the emitter from a current scaffold
+      (dangling `<script src>`) drops to the presence tier, and the
+      `public/`-holds-a-stale-ack false negative in item 18 stays a false
+      negative instead of silently becoming a warning.
+    - **"Cannot observe" must not become "everything passes" either.** An
+      incomplete graph never produces a WIRING finding (item 10's doctrine), but
+      the presence finding still fires when nothing in the tree mentions the
+      message at all. The residual is stated rather than hidden: an unresolvable
+      graph PLUS the literal somewhere is silence — a false negative, the cheap
+      direction. And an UNOBSERVABLE tree scan (unreadable file, size cap, file
+      budget) gates BOTH tiers: a tree we could not read is not one we can draw
+      a wiring conclusion about.
+    - **Both tiers share the `readyAckSourceExts` gate.** A `.css` an entry
+      module imports really is loaded by the browser and really cannot implement
+      a handshake, so it is in the graph but is not ack evidence. Dropping that
+      gate on the graph scan was a mutant killed by a rendered page-vite fixture
+      with the literal in `src/index.css`.
+    - **Still a Warning, never an Error** — every reason in item 18 stands, and
+      the stronger tier does not change that: reachability is still inference
+      from static text, and `--strict` already gives anyone who wants a gate
+      one.
+    - **Measured, before (`e800129`) → after**, on pre-fix `static` and
+      `page-vite` projects: emitter copied but not wired
+      `silent rc=0 strict 0` → `WARN rc=0 strict 1`; orphan `BLOCK_READY` file
+      the same; properly wired → silent both; fresh `static` / `page-vite` /
+      `page-money` scaffolds → silent both. The harness carries a POSITIVE
+      CONTROL — a pre-fix project with no emitter warns on BOTH binaries — so
+      the silent cells are real false passes rather than a probe wired to
+      nothing.
 
 **When you change a validation rule, keep all four vendored mirrors in sync with
 the server — `schema/`, the ported Go checks in `internal/validate/` (including
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1002,9 +1002,21 @@ neither one's.
       scheme-qualified or protocol-relative is DOCUMENT-RELATIVE, there are no
       bare specifiers, and there is no extension or directory-index guessing.
       Under module syntax a bare specifier is a package and a bundler's
-      extension resolution applies. The scheme regex is ANCHORED (`^`) because
-      the scheme check runs BEFORE the `?`/`#` strip, so an unanchored pattern
-      throws away `./x.js?from=https://cdn`. Do not re-merge these paths.
+      extension resolution applies. The scheme regex is ANCHORED (`^`), or an
+      unanchored pattern matches a `https:` in a QUERY STRING and throws away
+      `./x.js?from=https://cdn`. (An earlier revision of this item credited the
+      ORDER of the scheme check and the `?`/`#` strip. It does not matter — two
+      independent sweeps swapped them with nothing failing. Only the anchor is
+      load-bearing.) Do not re-merge these paths.
+      🔴 **And the boundaries are HTML boundaries, not `\b`.** `\b` matches
+      between `-` and `s`, so `\bsrc\s*=` read `data-src=` as `src=` and
+      `<script\b` read `<script-loader>` as a script. Measured on a CORRECT
+      `static` scaffold with `data-src` — a real consent-manager / lazy-load
+      pattern — the wrong reference resolved, `Complete` stayed true, and the
+      STRONG tier warned with `--strict` rc=1: the same false-warning class the
+      unquoted-`src` fix had just closed, reintroduced by the fix itself. RE2
+      has no lookahead, so the tag boundary is spelled as "the character after
+      `<script` is whitespace, `/` or `>`".
     - **THE DECIDABILITY BOUNDARY IS COMPLETENESS, NOT PROJECT TYPE.** The first
       draft drew it at "does the manifest declare a `buildCommand`" — bundled
       means undecidable — and that is wrong in both directions: an ordinary Vite
@@ -1026,21 +1038,43 @@ neither one's.
       three separate comments — this item included — claimed an exhausted budget
       made the graph incomplete. Measured: a CORRECT project whose ack sits ten
       imports below index.html got the STRONG tier's warning and `--strict`
-      rc=1, on a graph we had stopped walking. Only imports that are NOT
-      declared packages count as truncated: a leaf importing `react` and nothing
-      else has nothing left to see, and counting it would make every project as
-      deep as the bound unobservable.
-    - 🔴 **THE GRAPH RETAINS NO FILE CONTENTS, and `EntryFile` must not grow a
-      `Code` field again.** It had one, and a 200-module graph entirely INSIDE
-      the file-count and size budgets peaked **410 MB RSS**, against ~40 MB for
-      the same tree before the graph existed — a check whose caps exist
-      precisely so `validate` cannot become a memory event, blowing past the
-      316 MB figure those caps were sized against, with `app submit` paying it
-      too. Contents now go to `EntryGraphOptions.Inspect` while they are in
-      hand, so the walk holds ONE file at a time exactly like the tree scan.
-      After the fix: 39.5 MB, matching the pre-graph baseline on that fixture.
-      `TestEntryGraphRetainsNoContents` pins it structurally rather than by
-      benchmark.
+      rc=1, on a graph we had stopped walking.
+      🔴 **But the gap must fire ONLY when something was actually truncated, and
+      getting that wrong silenced the defect this whole PR exists to catch.** A
+      declared package was never going to be followed, and a target ALREADY IN
+      THE GRAPH was read by another route. The first version checked only the
+      first, so a DIAMOND — the normal shape of any real module graph — whose
+      deepest module re-imports an already-walked file produced a gap whose
+      message was literally false about a file sitting in `g.Files`, dropped the
+      project to the presence tier, and let an orphaned emitter pass with rc=0.
+      Measured; a self-cycle at the bound did the same. A gap that over-fires is
+      not "conservative", it is the false pass wearing a different hat. See
+      `truncatedBelow`.
+    - 🔴 **THE GRAPH MUST STAY O(ONE FILE), AND THAT TOOK TWO FIXES — THE FIRST
+      ONE ALONE WAS CLAIMED AS COMPLETE AND WAS NOT.** `EntryFile` must not grow
+      a `Code` field again, AND every import specifier must be `strings.Clone`d:
+      a Go substring shares its parent's backing array, so an un-cloned
+      specifier pins its whole file, and it is held in both the BFS queue and
+      `EntryFile.Spec`. Measured on a 200-module graph entirely INSIDE the
+      file-count and size budgets, where every module is ~2 MiB of REAL
+      (non-comment) code AND carries an import:
+
+      | build | peak RSS (3 runs) |
+      |---|---|
+      | `e800129`, before the graph existed | 26.4 – 27.8 MB |
+      | retaining `Code` | 421 – 439 MB |
+      | `Code` dropped, specifiers aliased | 558 – 628 MB |
+      | both fixed | 31.4 – 33.3 MB |
+
+      🔴 **THE FIXTURE SHAPE IS THE MEASUREMENT.** A tree padded with COMMENTS
+      leaves almost nothing after stripping, and a tree whose leaves carry no
+      imports has no specifier to alias — both report a reassuring number while
+      exercising neither hazard. That is exactly how the first round certified
+      "39.5 MB, one file at a time" from a fixture the depth bound truncated to
+      ten files. State the fixture with the number.
+      `TestEntryGraphRetainsNoContents` pins the `Code` half structurally, and
+      says in its own body that it CANNOT see the aliasing half —
+      `strings.Contains` reads the logical string, not the backing array.
     - 🔴 **A reference to a file that is NOT THERE is a GAP, not a decided
       absence.** It is tempting to read it as "the browser 404s that, so it
       cannot be the ack" — but a project that presumably builds having an
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -485,8 +485,16 @@ neither one's.
       question answered for an AUTHOR's project and had answered it with a
       whole-tree grep instead — see item 20 for the false pass that produced.
       Guard A now calls `blockproto.ReadyAckWiring`; its control corpus moved
-      with the predicate, into `internal/blockproto/entrygraph_test.go`. Nothing
-      about Guard A's strength changed, and the depth bound it relies on
+      with the predicate, into `internal/blockproto/entrygraph_test.go`.
+      🔴 **Guard A's REJECTION set is unchanged, but its ACCEPTANCE set moved —
+      state both, because "nothing changed" was the claim an audit falsified.**
+      It now also accepts an emitter imported by an INLINE
+      `<script type="module">` in index.html (a real browser load it used to
+      miss, and missing it manufactured a finding at a correct project), and the
+      unquoted / no-`./` HTML spellings a browser resolves identically. It also
+      briefly accepted an extensionless `src="./civitai-host"`, which was a BUG
+      rather than a widening — a 404 on a no-build template — and is reverted;
+      see item 20. The depth bound it relies on
       (`readyAckWiringDepth = 2`: index.html's `<script src>` entries plus their
       DIRECT imports) is now pinned by a corpus case — widening it to 99 was a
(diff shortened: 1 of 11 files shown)


ZacxDev:
## Audit round 1 — three 🔴 confirmed, one root cause, fixed in `518e514`

I reproduced all three before touching anything, plus the 🟡s. **The audit is right and my PR body was wrong.** Corrections first, since the falsified claims are the important part.

### Claims this PR made that were false

| claim | where | reality |
|---|---|---|
| "the headline defect is fixed" | PR body, title | **False for any `index.html` that omits `./`.** `<script src="app.js">` was read as a bare module specifier → gap → presence tier → silent. Measured: `src="./app.js"` → `strict_rc=1`; `src="app.js"` → `strict_rc=0`, the shipped defect intact. |
| "an exhausted budget" makes the graph incomplete | `entrygraph.go:43-45`, PR body, AGENTS item 20 | **Not true of the depth budget.** `MaxDepth` truncated the walk and left `Complete` true, so the **strong** tier fired on a partial graph. A correct project with its ack ten imports below `index.html` warned with `--strict` rc=1. |
| the incomplete bucket is "a bundler alias or other bare specifier that is not a declared dependency" | PR body, AGENTS item 20 | Under-described to the point of being misleading — nobody maps that phrase to ordinary HTML. |
| "Guard A's strength is unchanged" | PR body | Its **rejection** set was unchanged; its **acceptance** set moved twice. One was a genuine improvement (inline `<script type="module">` imports), one was a bug (extensionless `src`). Both are now stated in AGENTS item 11. |
| `TestEntryGraphBudgetsAreGaps` / "the depth bound is a gap only when there is more to see" | test name | The test **asserted nothing about gaps**. The name claimed a property the code lacked. |

### The root cause

`<script src>` was parsed and resolved with JavaScript module-specifier rules. That single conflation produced all three 🔴 — two of them *opposites*, which is why they read as unrelated:

```
src="app.js"            bare specifier -> gap -> SILENT at a broken app
src=./civitai-host.js   unquoted, regex required quotes -> tag re-read as an
                        INLINE script with an empty body -> reference vanished
                        with Complete still TRUE -> WARNED at a correct app
src="./civitai-host"    extension-guessed to .js -> ACCEPTED, on a no-build
                        template where the browser 404s that URL
```

`refSyntax` now splits them. Under **URL** syntax: not scheme-qualified and not protocol-relative ⇒ document-relative; no bare specifiers; no extension or directory-index guessing. Under **module** syntax: bare ⇒ package, bundler extension resolution applies. The scheme regex is anchored — the scheme check runs *before* the `?`/`#` strip, so an unanchored pattern discards `./x.js?from=https://cdn` (that mutant survived the first sweep; it now dies).

Depth truncation now records a gap, but only for imports that are **not** declared packages — counting `import 'react'` at the bound would make every project as deep as the bound unobservable.

### Live matrix (before = `e800129`, after = `518e514`)

*(Note: my original matrix already used `e800129` as "before". `0ce0025` was used only to **scaffold genuinely pre-#206 projects**, never as a comparison binary — the audit's read of that was the one place I'd push back.)*

```
FIXTURE                        BEFORE (e800129)              AFTER
control-noemitter              WARN         rc=0 strict=1    WARN/missing   rc=0 strict=1
static-unwired                 silent       rc=0 strict=0    WARN/unwired   rc=0 strict=1
static-unwired-nodotslash  🔴1  silent       rc=0 strict=0    WARN/unwired   rc=0 strict=1
correct-unquoted-src       🔴2  silent       rc=0 strict=0    silent         rc=0 strict=0
correct-nodotslash              silent       rc=0 strict=0    silent         rc=0 strict=0
correct-deep               🔴3  silent       rc=0 strict=0    silent         rc=0 strict=0
static-orphan                   silent       rc=0 strict=0    WARN/unwired   rc=0 strict=1
vite-unwired                    silent       rc=0 strict=0    WARN/unwired   rc=0 strict=1
static-wired / vite-wired       silent                        silent
fresh static / page-vite / page-money   silent                silent
```

`correct-unquoted-src` and `correct-deep` warned with `--strict` rc=1 on the previous head of this branch; both are silent now. `control-noemitter` is the harness positive control (must warn on **both** binaries).

### Memory (🟡 6)

Same 200-module fan-out fixture, every file inside `maxAckGraphFiles`/`maxAckFileBytes`:

| | peak RSS | wall |
|---|---|---|
| `e800129` (no graph) | 39.5 MB | 1.26 s |
| this PR, before the fix | **410 MB** | 1.87 s |
| this PR, after | **39.5 MB** | 2.67 s |

Back to the pre-graph baseline exactly. `EntryFile` no longer carries `Code`; contents go to `EntryGraphOptions.Inspect` while in hand, so the walk holds one file at a time like the tree scan. The residual ~1.4 s is the entry walk itself on a 400 MB tree. `TestEntryGraphRetainsNoContents` pins it structurally rather than by benchmark, since a benchmark would drift.

### Mutation battery — 42 mutants

**40 killed** with attributed failures, **2 null mutants survived** as required, **0 broken**. Restore is by file-copy snapshot with sha256 verified in both directions, each run timeout-bounded so a hanging mutant is reported rather than wedging the battery.

Every mutant the audit reported surviving now dies:

| audit mutant | now |
|---|---|
| E11 unquoted-src widening | `B1-quotes-required` KILLED — `TestHTMLSrcIsAURLNotAModuleSpecifier/silent:_unquoted_src` |
| V11 unwired headline → "something is off" | `F1-unwired-headline` KILLED — `…StateTheirOwnStrength`: *the unwired advisory no longer says "DOES contain"* |
| V20 delete the operative disclosure | `F2-drop-operative-disclosure` KILLED — *no longer says "will silence this warning"* |
| fine-grained dependency prefix scan | `G1-deps-prefix-fine` KILLED — `…/a_bare_specifier_that_merely_starts_like_a_dependency` |
| Guard A depth 2→99 | `H14-depth-unbounded` KILLED — `…/reject: the emitter is two import levels below the entry` |
| cap raised to `1<<40` (hung the suite twice) | `G9-cap-raised` KILLED — `TestReadyAckCapValues`, fails in <1 s |

New kills for the root cause and the branches the sweep found unpinned: `A1-src-as-bare-specifier`, `A2-no-url-scheme-check`, `A3-scheme-boundary` (survived the first sweep, now pinned by a query-string-containing-a-URL fixture), `B2-src-first-group-only`, `C1-depth-not-a-gap`, `C2-depth-gap-always`, `D1-guess-ext-for-urls`, `D2-no-ext-resolution-at-all`, `E1-retain-contents`, `E2-inspect-index-only`, `F3/F4/F5` message content per tier, `G2-parent-dir-arm`, `G3-html-terminator` (`-->`→`->`), `G4-backtick`, `G5-escape`, `G6-block-comment-lossy`, `G7-dynamic-import`, `G8-module-exts-css`.

Three mutants came back `BROKEN-MUTANT` on the first pass — regex needles mangled by re-escaping a Go raw string inside a Python literal. That is a *false survivor* generator, so they were re-run through a line-anchored harness that takes the needle out of the file itself; all three then killed.

### 🟡/🟢 also done

**7** strength guard rewritten (literals unique to one tier + inverse controls). **8** branches pinned as listed. **9** `StripJSComments` no longer swallows the tail of an unterminated `/*` — same rule and same reason as the HTML half, which is what the asymmetry should have been all along. **12** the `v0.1.90` citation is gone. **13** caps pinned by value, plus a bound in the cap fixtures so an absurd constant fails fast instead of allocating.

### Not done / disagreements

- **Windows path-case** — out of scope per the audit, pre-existing.
- **`0ce0025` as a "before" binary** — I never used it as one; it only scaffolded pre-#206 *projects*. Flagging in case that assumption propagated elsewhere.
- **The extensionless-`src` project now gets silence, not a warning.** After the fix that reference resolves to nothing, which is a gap → presence tier → the tree holds the literal → quiet. Honest (we stopped modelling) rather than the previous false claim that the emitter was wired, but it is still a false negative. Asserted explicitly in `TestHTMLSrcIsAURLNotAModuleSpecifier` including the `Complete == false` check, so nobody reads the silence as approval.
- **`MaxDepth`/`MaxFiles` left at 8/200.** Now that truncation is a gap, a genuinely large app falls to the presence tier more often than before. That is the correct direction (quiet, disclosed) and the #206 population is small apps, but it does mean the strong tier covers less than the numbers suggest. Raising them is a separate, measurable change.

### Gate

`make ci` green. `go test ./... -v -count=1`: `=== RUN 2065`, `--- PASS 2061`, `--- FAIL 0`, `--- SKIP 4` (the pre-existing env-gated ones), 0 timeout panics, 18/18 packages ok, `gofmt -s -l .` silent. Guard B run locally with `CIVITAI_CHECK_SCAFFOLD_RUNTIME=1` — passes with both its own controls. `git status --porcelain` empty; every mutation reverted and checksum-verified.

ZacxDev:
## Audit round 2 — two regressions and a half-fix, corrected in `6288c88`

All six findings reproduced locally before I changed anything. **Two were regressions I introduced in `518e514`**, one in the exact class that commit had just closed. Corrections first.

### Claims from `518e514` that were false

| claim | where | reality |
|---|---|---|
| the graph "holds ONE file at a time exactly like the tree scan" | `entrygraph.go:110-116`, `readyack.go:224-231`, `AGENTS.md:1030-1041` | **False.** Dropping `EntryFile.Code` was half the fix. A Go substring shares its parent's backing array, so an un-cloned import specifier pinned its whole 2 MiB file — held in the BFS queue *and* `EntryFile.Spec`. Measured **558–628 MB**. |
| "39.5 MB, matching the pre-graph baseline" | `AGENTS.md:1041` | **Measured on the wrong fixture** — a chain the depth bound truncated to 10 files, padded with *comments* so nothing survived stripping. Neither hazard was exercised. |
| pre-graph baseline "~40 MB" vs "~17 MB" | AGENTS vs the source comments | Inconsistent, and both were guesses at a number I had not measured on a comparable fixture. My measurement: **26.4–27.8 MB** at `e800129` on the 200-module fixture. |
| the scheme-check / `?#`-strip **order** is load-bearing | `AGENTS.md`, `entrygraph_test.go:267` | **False**, as both your sweeps found. A scheme-qualified specifier stays scheme-qualified once its query is removed. Only the `^` **anchor** matters. Corrected in both places. |
| "an exhausted budget makes the graph incomplete" (depth) | fixed in `518e514` | Now true — but the fix **over-fired**, see N3. |

### N2 — `\b` is not an HTML name boundary *(reproduced; I'd call this 🔴 too)*

`\b` matches between `-` and `s`. Measured on a **correct** `static` scaffold:

```
ctlgood      (untouched)                  strict_rc=0  findings=0
datasrc      data-src added before src    strict_rc=1  findings=1   FALSE WARNING, strong tier
xsrc         x-src added before src       strict_rc=1  findings=1   FALSE WARNING
brokenctl    emitter removed              strict_rc=1  findings=1   positive control
```

RE2 has no lookahead, so: `(?:^|[\s/])src\s*=` for the attribute, and a required `[\s/>]` after `<script` for the tag. After the fix all three correct projects are silent and `brokenctl` still warns. The inverse control is pinned too — a `data-src` carrying the **only** reference to the emitter must still warn (`TestAttributeNameBoundary/warns:_the_emitter_is_referenced_only_by_data-src`), or "ignore `data-src`" would be indistinguishable from "ignore every `src`".

### N3 — the truncation gap over-fired and silenced the headline defect

Confirmed exactly as reported. At the bound I gapped for any non-package specifier without the `g.FileAt(file) >= 0` dedupe the normal path does:

```
                    HEAD@518e514        after 6288c88
diamond             silent  rc=0        UNWIRED  rc=1
selfcycle           silent  rc=0        UNWIRED  rc=1
control (nothing below the bound)   UNWIRED rc=1  (both)
```

The gap message was literally false — it named a file sitting in `g.Files`. `truncatedBelow` now answers "no" for a declared package **and** for an already-walked target. **A gap that over-fires is not conservative; it is the false pass wearing a different hat.** A negative control pins the other direction: a genuinely unwalked module below the bound must *still* gap, or the dedupe has simply disabled the guard.

### N1 — memory, re-measured on a fixture that actually reaches 200 files

Every module ~2 MiB of **real** (non-comment) code **and** carrying an import — the shape both hazards need. Graph reach verified at 200/200 before measuring; 3 runs each:

| build | peak RSS |
|---|---|
| `e800129`, before the graph existed | **26.4 – 27.8 MB** |
| retaining `EntryFile.Code` | 421 – 439 MB *(your figure)* |
| `Code` dropped, specifiers aliased (`518e514`) | **558 – 628 MB** |
| both fixed (`strings.Clone`) | **31.4 – 33.3 MB** |

The residual ~5 MB over the tree scan is the graph reading its files at all; the property that matters is O(one file), not O(graph). The fixture shape is now stated beside every number, in the source and in AGENTS.md, because that is what went wrong the first time.

`TestEntryGraphRetainsNoContents` now says in its own body that it **cannot** see the aliasing half — `strings.Contains` reads the logical string, not the backing array — so it can't be cited as coverage again. I could not make that half structural; the guard is the live measurement, and the test says so.

### N4 — the third tier's opening diagnosis

Reproduced: all three of the presence tier's asserted literals lived in its trailing disclosure, so replacing its opening survived. The shared *"nothing in this project's source posts"* sentence is now asserted as **set membership** — present in exactly the two tiers that report an absence, absent from the one that fires because the message *is* present. `N4a-presence-headline` now dies.

### N5 / N6

Pinned: the `rel == ".."` containment arm (reachable via directory-index resolution — `import '../..'` with an `index.js` in the parent), `srcAttrValue`'s empty-`src` verdict (a tag with a `src` is external, so its body must not execute), single-letter and uppercase URL schemes, and the JS `*/` terminator (the HTML half had a dedicated case, the JS half didn't). Doc claims corrected as tabled above.

### Battery — 52 mutants

**49 killed** with attributed failures, **2 null mutants survived** as required, **1 documented survivor**.

Regressions and previously-unpinned branches, each attributed to a test added here:

| mutant | killed by |
|---|---|
| `N2a-attr-boundary-b` (`\b` back on the attribute) | `TestAttributeNameBoundary/silent:_data-src_before_src` |
| `N2b-tag-boundary-b` (`\b` back on the tag) | `…/reject: a <script-loader> custom element is not a script tag` |
| `N3a-truncation-no-dedupe` | `TestDiamondAtTheDepthBoundStillWarns/diamond:…` |
| `N3b-truncation-never` | `…/negative control: a real truncation still gaps` |
| `N4a-presence-headline` | `…StateTheirOwnStrength`: *no longer opens by saying WHAT WAS FOUND* |
| `N5a-containment-dotdot-arm` | `TestEntryGraphRootContainment` |
| `N5b-empty-src-inline` | `…/reject: a body import inside a tag that has an (empty) src` |
| `N5c-js-terminator` | `TestStripJSComments/a_star_inside_a_block_comment…` |
| `N5d-scheme-class-narrow` | `…/reject: a single-letter scheme is still a URL` |
| `A3-scheme-boundary` (survived round 2) | `…/accept: a local src whose query string contains a URL` |

**The documented survivor: `N1a-no-clone`.** Dropping `strings.Clone` passes the entire unit suite, and no unit test can see it — a backing array is invisible to every assertion Go gives you here. Reported as a survivor rather than papered over; the guard is the RSS table above.

Five needles came back `BROKEN-MUTANT` because the code they targeted moved in this round — `C1`/`C2` are superseded by `N3a`/`N3b` (same branch, new implementation), and `B1`/`F2`/`G7` were re-run through the line-anchored harness and all three killed. Every mutation is checksum-verified as applied and as reverted; `git status --porcelain` is empty.

### Live matrix (before = `e800129`)

Re-run in full. Every fresh scaffold stays silent:

```
control-noemitter          WARN rc=0 strict=1  →  WARN/missing  rc=0 strict=1
static-unwired             silent            →  WARN/unwired  rc=0 strict=1
static-unwired-nodotslash  silent            →  WARN/unwired  rc=0 strict=1
static-orphan              silent            →  WARN/unwired  rc=0 strict=1
vite-unwired               silent            →  WARN/unwired  rc=0 strict=1
correct-unquoted-src       silent            →  silent
correct-nodotslash         silent            →  silent
correct-deep               silent            →  silent
static-wired / vite-wired  silent            →  silent
fresh static / page-vite / page-money        →  silent
```

Every probe ran behind a **fixture-existence control** (8/8 index.html present, emitter counts asserted per fixture) — your wiped-directory incident is exactly the shape where a `grep -c` of 0 reads as "no warnings".

### ⚠️ No remote CI signal

**GitHub Actions is stalled repo-wide** — checks on this branch have sat `queued` with 0 s elapsed for hours, and runs on `main` are queued too; no runner has picked anything up. **I have not seen a green check for this branch and am not claiming one.** Local only:

`make ci` green. `go test ./... -v -count=1`: `=== RUN 2086`, `--- PASS 2082`, `--- FAIL 0`, `--- SKIP 4` (the pre-existing env-gated ones), 0 timeout panics, 18/18 packages ok, `gofmt -s -l .` silent. Guard B run locally with `CIVITAI_CHECK_SCAFFOLD_RUNTIME=1` — passes with both its own controls. A missing check is not a passing check; these all need re-running once runners return.

ZacxDev:
## Audit round 3 (final) — F1 fixed in `fd8cf9b`, residuals now written down

F1 reproduced exactly as reported, and the fix is a shape change rather than another condition.

### F1 — the dedupe asked `g.Files` a question it could not answer yet

Files enter `g.Files` when **popped**, not when **queued**, so a target already enqueued looked absent. Measured on real scaffolded `static` projects with an orphaned emitter, all identical but one import line:

| project | `518e514` | `6288c88` | `fd8cf9b` |
|---|---|---|---|
| deepest module imports nothing (control) | warns | warns | **warns** |
| diamond | silent | warns | **warns** |
| deepest imports its **not-yet-walked** sibling | silent | **SILENT** ❌ | **warns** ✅ |
| same graph, imports **textually swapped** | silent | warns | **warns** |

**The import-order pair, both orders, flat and nested — all four now identical:**

```
flat-ab     before(e800129) rc=0  |  HEAD rc=1  UNWIRED
flat-ba     before(e800129) rc=0  |  HEAD rc=1  UNWIRED
nested-ab   before(e800129) rc=0  |  HEAD rc=1  UNWIRED
nested-ba   before(e800129) rc=0  |  HEAD rc=1  UNWIRED
```

(Fixture control reported alongside: 4/4 carry the orphaned emitter.)

The fix **defers** the truncation check to after the walk drains, when `g.Files` is final — so the answer cannot depend on ordering at all, rather than depending on it correctly. Consulting the queue as well would have worked; deferring removes the class instead of the instance.

### F2 / F3 / F4 — pinned

- **F2**: `</script >` must close an inline module (M9), `</scriptx>` must not end one early (M10), and a `<script-loader>` body is not a script (M8). ⚠️ **My first draft of the `</scriptx>` case asserted the opposite and was wrong about HTML** — a browser keeps parsing past a bogus close tag, so it is an **accept**, and the accept is what discriminates the widening. Corrected before commit.
- **F3**: `refUnresolved` (M16) and `resolveRefPath`-error (M17) arms both pinned — a bundler alias or dangling import at the bound must still gap, or the strong tier fires at a correct project. M29 pinned by a **nested** fixture (`src/deep/`); every previous fixture was root-flat, where root and module dir coincide.
- **F4**: dropped the `^` alternative rather than documenting an unreachable branch. `reScriptOpen`'s capture is `([\s/][^>]*)?`, so an attribute list is either empty or starts with `[\s/]`.
- **F5**: every memory row now names its fixture. The `e800129` row is the **whole-tree grep**, so it scales with the tree, not the graph — your 17.6–17.8 MB and my 26.4–27.8 MB are both right about their own fixture. The durable fact is the ratio, and the **~18×** figure is now in the `N1a` disclosure as you asked.

### Residuals, now recorded in AGENTS.md item 20

Written down because every one was found by an audit rather than by me, and a residual nobody records is indistinguishable from a bug nobody noticed:

1. **Extensionless `src`** → gap → presence tier → quiet. Correct, but a false negative, not approval.
2. **Large graphs fall to the presence tier more often than the caps suggest** — with truncation a gap, anything deeper than `MaxDepth` or wider than `MaxFiles` is presence-only. The strong tier covers small and pre-fix apps, which is the #206 population, not big ones.
3. **The `strings.Clone` property is guarded by a measurement, not a test** — dropping it passes the entire suite at **~18×** peak RSS. Declared survivor; the test says in its own body that it cannot see that half.
4. **` src=` inside an attribute *value* hijacks the reference** (`<script data-x="foo src=./nope.js" src="./x.js">`). Pre-existing and byte-identical before/after this work — fixing it needs a real attribute parser, which is a bigger change than this check earns.
5. **Computed `import(expr)` is unmatched** — the one residual that makes the graph smaller *without* setting a gap.

### Battery — 57 mutants

**54 killed** with attributed failures, **2 null mutants survived** as required, **1 declared survivor**.

| mutant | killed by |
|---|---|
| `F1a-truncation-inline-again` | `…/the bound is order-independent (import A then B)` |
| `F2a-close-tag-strict` | `…/accept: an inline module closed by `</script >`` |
| `F2b-close-tag-wide` | `…/accept: an inline module continues past a bogus `</scriptx>`` |
| `F2c-inline-tag-boundary` | `…/reject: an import inside a <script-loader> body is not a load` |
| `F3a-unresolved-not-truncation` | `…/a bundler alias at the depth bound is a gap` |
| `F3b-danglingref-not-truncation` | `…/a dangling import at the depth bound is a gap` |
| `F3c-truncation-wrong-basedir` | `…/the bound resolves against the importing module's directory` |
| `N3a` / `N3b` (re-pointed at the deferred shape) | `TestDiamondAtTheDepthBoundStillWarns` + its negative control |

`N1a-no-clone` remains the sole declared survivor. Four needles moved when `reSrcAttr` changed for F4 and were re-run line-anchored (`B1`, `N2a`, `F2`, `G7`) — all four killed. Every mutation checksum-verified as applied and as reverted; `git status --porcelain` empty.

### Live matrix (before = `e800129`) — every fresh scaffold silent

```
control-noemitter          WARN rc=0 strict=1  →  WARN/missing  rc=0 strict=1
static-unwired             silent  →  WARN/unwired  rc=0 strict=1
static-unwired-nodotslash  silent  →  WARN/unwired  rc=0 strict=1
static-orphan              silent  →  WARN/unwired  rc=0 strict=1
vite-unwired               silent  →  WARN/unwired  rc=0 strict=1
correct-unquoted-src / correct-nodotslash / correct-deep   silent → silent
static-wired / vite-wired                                  silent → silent
fresh static / page-vite / page-money                      silent → silent
```

Source-fixture existence control reported beside it: 5/5 present. Memory re-checked on `mem200b`: HEAD 31.8–33.1 MB vs `e800129` 26.9–32.7 MB.

### ⚠️ Local-only gate

**Remote CI is still stalled repo-wide** — no runner has picked up this branch or `main`. **I have not seen a green check and am not claiming one.**

`make ci` green locally. `go test ./... -v -count=1`: `=== RUN 2094`, `--- PASS 2090`, `--- FAIL 0`, `--- SKIP 4` (pre-existing env-gated), 0 timeout panics, 18/18 packages, `gofmt -s -l .` silent. Guard B passes locally with `CIVITAI_CHECK_SCAFFOLD_RUNTIME=1`. All five required contexts still need a real run.

PR #220, opened 2026-08-06
feat(login): opt into generation with `civitai login --scopes generate`

ZacxDev:
## 🔴 DO NOT MERGE OR RELEASE YET — hard ordering constraint

**This PR must not be merged or released until the `civitai/civitai` migration widening the `civitai-cli` OauthClient's `allowedScopes` from `100663297` to `100777985` is confirmed applied to PRODUCTION.**

The server's device-flow scope validation is **all-or-nothing**: requesting any bit the client does not allow rejects the **whole** login with `invalid_scope` (400). That migration is **manual-apply** and is **not live in production** as of 2026-08-06. Ship this ahead of it and `civitai login --scopes generate` 400s for every user who tries it. (Plain `civitai login` is unaffected — its request is bit-for-bit unchanged.)

This mirrors the convention already documented at `internal/appapi/oauth.go` for the previous widening (AppBlocksDevTunnel / bit 26).

- Issue: civitai/civitai#3681
- Context: https://github.com/civitai/civitai/issues/3681#issuecomment-5208310465
- Server-side PR 1 (the widening) is being implemented in parallel.

**Merge checklist**
- [ ] `civitai-cli` OauthClient `allowedScopes` = `100777985` confirmed applied to prod
- [ ] `civitai login --scopes generate` exercised end-to-end against prod auth
- [ ] Undraft

---

## What this does

`civitai login` yields a token that cannot run `civitai generate` (#3681). The device request pinned a fixed `DeviceScope = "100663297"`, which omits `AIServicesWrite` **by design** — a plain login must not silently carry Buzz-spend authority — with no way to ask for it.

### 1. `civitai login --scopes generate` (opt-in, additive)

The requested scope is now **computed**, not fixed:

| Invocation | Requested mask |
|---|---|
| `civitai login` | `100663297` — **unchanged**, load-bearing |
| `civitai login --scopes generate` | `100777985` = `100663297 \| 114688` |

`114688` = `AIServicesRead` (1<<14) `\| AIServicesWrite` (1<<15) `\| BuzzRead` (1<<16).

**Additive, not replacing**, because login yields ONE credential: a user who opts into generate must not lose app-submit / dev-tunnel ability.

- `--scopes` takes a **named set**, never a raw bitmask — nobody should type bit arithmetic to log in, and a named set lets the granted mask change server-side without changing this interface. New sets are added in one place (`deviceScopeSets` in `internal/appapi/oauth.go`); the flag help, the validation error and the resolver all read from it.
- An **unknown set name** is a hard error listing the valid ones, raised **before any network call** (verified: 0 device-init hits).
- **Surfaced at the point of use**: `civitai login --help` documents the set and that it grants Buzz-**SPEND** authority, and the login output echoes it while the browser approval is on screen.

**Composition with the existing `--token` scheme.** `internal/cmd/login.go` deliberately sets `SetInterspersed(false)` and has a position-aware `--token` parsing scheme with a documented footgun history. `--scopes` is a value-taking flag with **no** `NoOptDefVal`, so pflag consumes the following argument as its value and it produces **zero positionals** — it can never be mistaken for the `--token <value>` positional-recovery form, and the `Args` guard is untouched. The `SetInterspersed(false)` trade-off still applies in the other direction and is now pinned by tests: in `login --token <key> --scopes generate` the positional `<key>` halts flag parsing, so `--scopes` arrives as a stray positional and the whole invocation is **rejected** rather than silently running with the default scope. `--scopes` with `--token` is rejected outright with a usage error (a personal key's scopes are fixed when it is minted).

### 2. Error mapping for the ordering trap

`invalid_scope` from device-init is now mapped to `*appapi.InvalidScopeError` with an actionable message rather than a raw OAuth code. It branches on whether the request was widened:

- **widened request** → names what was requested vs the default, says the server does not permit them yet, and tells the user plain `civitai login` still works and to retry `--scopes generate` once the widening is live.
- **default request** → different meaning entirely (the CLI is ahead of the auth server it points at), so it offers the personal-key route instead of telling the user to re-run the thing they just ran.

A negative control test proves the mapping is selective: `invalid_client` still yields the generic `device init failed` error.

### 3. `dev-token --spend` — explicit request, and the decision on the default

**Decision: `--spend` is purely ADDITIVE. The default does NOT narrow.**

`civitai app dev-token --spend` unions `ai:write:budgeted` into the `scopes` array POSTed to the mint route. Without `--spend`, the outgoing body is **byte-identical to today's** (the local `block.manifest.json` scopes verbatim, or the key omitted entirely when there are none).

**Why not narrow by default**, when the point is to stop a `--scopes generate` login silently arming real-Buzz `dev:live`:

1. **It is not expressible.** `scopes` is an **intersection allow-list** whose only sentinel is "absent ⇒ server decides". There is no encoding for "everything you would grant me, minus spend". Narrowing by default would force the CLI to enumerate a grant set it does not know, silently stripping storage/collections/etc.
2. **It regresses the users it works for today.** A full-scope personal API key with **no local manifest** gets budgeted spend right now precisely because the body carries no `scopes` key. A default narrowing removes that with no error. Same for a manifest that legitimately declares `ai:write:budgeted` — narrowing would contradict the manifest the developer wrote and will be reviewed against.
3. Closing the inference hole belongs on the **server**, which knows the full grant set. Server-side PR 3 flips the default to explicit-request-only; this PR ships the **client half** (the explicit request) so PR 3 can land without breaking anyone. **This PR does not depend on PR 3.**

**Blast radius of the chosen behaviour: zero change to any existing invocation.** `--spend` only ever ADDS a scope. Both paths are tested by asserting the exact `scopes` array on the outgoing request body, including a raw-bytes check that the key is absent (a decoded struct cannot distinguish absent from empty).

One honest side effect, handled: `--spend` with **no** local manifest scopes does narrow the request to `["ai:write:budgeted"]` alone. The command sends it and prints a stderr notice saying so, pointing the user at running from their App project dir.

### 4. Corrected now-false claims

- `internal/appapi/oauth.go` — the `DeviceScope` doc block (which explained why the CLI deliberately does *not* request `AIServicesWrite`) and the "What a login token can do … It CANNOT run a real generation" block, rewritten to state **default vs generate-scoped** and the resulting `dev:live` consequence, read off what the code now does.
- `internal/cmd/whoami.go` — the `CanSpendBuzz`-false guidance pointed at minting a personal API key as the only route. It now names the fix matching the credential in hand: an OAuth login is re-runnable, so `--scopes generate` leads; a personal key's scopes are fixed at creation, so the web-UI route leads there.
- `internal/cmd/app_dev_token.go` — `readOnlyTokenWarning`'s OAuth branch asserted OAuth "can't spend Buzz"; now offers `login --scopes generate`. The credential-can-spend branch now offers `--spend` alongside the manifest fix.
- `internal/cmd/buzz.go` — `BuzzRead` ships inside the generate set, so the balance hint no longer claims a personal key is the only fix.
- `README.md` — the "Submit & auth" scope paragraph (which also stated a **stale `33554433`**; the real default is `100663297`), the "Which credential can spend?" table, the `dev:live` credential callouts, the `civitai generate` credential note, and the `login` / `whoami` / `buzz` / `dev-token` command-table rows.

`internal/cmd/login_device_test.go`'s `100663297` pins at lines ~48/104/193 are still **correct** — they cover the default path, which is unchanged — so they are left alone.

## Tests

`go build ./...` clean · `go vet ./...` clean · `go test ./...` — **2162 run / 2158 pass / 0 fail / 4 skip**, 17 packages `ok` (baseline at `b403f59`: 2111 / 2107 / 0 / 4). No `panic: test timed out`.

### Red at base (`b403f59`) → green at HEAD

Measured by copying the new test files into a clean `origin/main` worktree.

| Test | at base | why |
|---|---|---|
| `TestLoginScopesGenerateSendsWidenedScope` | 🔴 FAIL | `unknown flag: --scopes` |
| `TestLoginScopesEqualsFormSendsWidenedScope` | 🔴 FAIL | `unknown flag: --scopes` |
| `TestLoginUnknownScopeSetFailsBeforeAnyNetworkCall` | 🔴 FAIL | error was `unknown flag`, not the valid-set list |
| `TestLoginInvalidScopeFromServerIsActionable` | 🔴 FAIL | message missing `invalid_scope` / fallback |
| `TestLoginHelpDocumentsScopeSets` | 🔴 FAIL | help had no `--scopes` |
| `TestLoginScopesFlagParsingMatrix` (6 of 9 subcases) | 🔴 FAIL | see below |
| `TestAppDevTokenSpendSendsBudgetedScope` | 🔴 FAIL | `unknown flag: --spend` |
| `TestAppDevTokenSpendNoManifestNarrowsAndSaysSo` | 🔴 FAIL | `unknown flag: --spend` |
| `TestAppDevTokenSpendHelpNamesTheConsequence` | 🔴 FAIL | help had no `--spend` |
| `TestReadOnlyTokenWarningPointsAtTheNewFixes` | 🔴 FAIL | warning asserted OAuth "can't spend Buzz" |
| `TestWhoAmISpendGuidanceForOAuth` | 🔴 FAIL | no `--scopes generate` in guidance |
| `TestWhoAmISpendGuidanceForPersonalKey` | 🔴 FAIL | only one route offered |
| `TestBuzzScopeHintNamesBothRoutes` | 🔴 FAIL | stale unconditional OAuth claim |

**Labelled INVARIANT GUARDS, not regression coverage** (they pass at base, by design — they pin behaviour the bug never violated and that this PR must not change):

- `TestLoginDefaultRequestsUnwidenedScope` — plain `civitai login` still sends exactly `100663297`. This is the guard that matters most: it is what stops a future change from widening the default.
- `TestLoginScopesFlagParsingMatrix/{no flags, --token <value> then --scopes, stray positional then --scopes}`
- `TestAppDevTokenDefaultDoesNotNarrow`, `TestAppDevTokenDefaultNoManifestOmitsScopesKey`, `TestAppDevTokenNoSpendFlagPrintsNoNarrowingNotice` — the item-3 blast-radius guards.

**Cannot be red at base:** every test in `internal/appapi/oauth_scopes_test.go` exercises API introduced by this PR (`ResolveDeviceScope`, `OAuthClient.Scope`, `InvalidScopeError`), so it does not compile against `b403f59`. Same for `TestDevTokenRequestScopes` / `TestSpendNarrowsToBudgetedOnly`. These are **specification** tests. Their behavioural claims are separately covered red-at-base end-to-end through the `login`/`dev-token` command tests above, and all are mutation-checked below.

### Mutation checks

Each guard was broken deliberately at HEAD and the **specific** intended assertion was confirmed to fire (not a neighbouring guard's).

| Mutation | Assertion that fired |
|---|---|
| drop `AIServicesWrite` from the generate set | `ResolveDeviceScope(generate) = "100745217", want "100777985"`; `generate adds 81920, want exactly 114688` |
| make `ResolveDeviceScope` REPLACE instead of OR | `generate mask 114688 is NOT a superset of the default 100663297` |
| unknown set silently ignored | `returned no error — a typo must not silently degrade to the default` |
| `StartDevice` sends the constant, ignoring `c.Scope` | `device-init form scope = "100663297", want "100777985"` |
| remove the `invalid_scope` branch | `error is *errors.errorString, want *InvalidScopeError` (and `TestStartDeviceLeavesOtherErrorsAlone` stayed **green** — proving the mapping is selective) |
| make the dev-token default NARROW | `default sent scopes = [user:read:self], want the manifest verbatim [user:read:self ai:write:budgeted]` |
| make `--spend` a no-op | `--spend sent scopes = [...], want [... ai:write:budgeted]` |
| remove the `--scopes`/`--token` conflict guard | `args [login --scopes generate --token k]: want an error, got none` |
| remove the narrowing notice | `expected the narrowing notice on stderr` (and the negative-control test stayed green) |
| don't pass the resolved scope to the client | `--scopes generate sent scope="100663297", want "100777985"` |
| don't echo the scope set at login | `login output must name the generate set AND that it grants Buzz spend` |

All 11 reachable — no earlier validation short-circuits any of them.

**Instrument validation:** the runner was watched red 12 separate times above (11 mutations + the base run), and the total `=== RUN` count moved 2111 → 2162 with the new files, so the additions are being collected rather than silently skipped. Counts are read from `--- PASS`/`--- FAIL`/`=== RUN` lines in a redirected log, not from an exit code, and the log was grepped for `panic: test timed out` (0).

## Not verified

Nothing here has been exercised against a real Civitai auth server — every assertion is against `httptest`. In particular the claim that `100777985` is the mask the widened `civitai-cli` client will accept comes from the issue thread, not from a live login. That is exactly what the draft gate above is for.


ZacxDev:
## Status — the DO-NOT-MERGE gate in the title is now CLEARED

The title still says "DO NOT MERGE until the server scope widening is live". **That gate is satisfied.** Leaving the title stale is how a PR either sits untouched or gets merged by someone who assumes the warning is out of date in the other direction, so recording it here.

The `civitai-cli` OAuth client's `allowedScopes` is now `100777985` in production, and civitai/civitai#3699 (the migration + guards) is merged to `main`.

Verified live against `auth.civitai.com`, with controls rather than a single reading:

| Request | Result | What it establishes |
|---|---|---|
| `scope=100777985` | HTTP 200, device code issued | the widening is live |
| `scope=100777987` (one bit outside the ceiling) | HTTP 400 `invalid_scope` | the endpoint still gates — the 200 above is not an instrument that accepts anything |
| `scope=100663297` (current shipped default) | HTTP 200 | no regression for the released CLI |

**This PR is still a draft on purpose.** Two things remain, below.

## 1. `--spend` does not yet do the job this PR credits it with

The PR presents `--spend` as what stops a `--scopes generate` login from silently arming real-Buzz `dev:live`. On the primary path it is inert:

- `internal/scaffold/templates/page-money/block.manifest.json.tmpl:7` — the default money-path scaffold already declares `"scopes": ["ai:write:budgeted"]`.
- `internal/cmd/app_dev_token.go` — `devTokenRequestScopes(manifestScopes, spend=false)` returns the manifest **unchanged**, so that scope is sent either way.

Server-side, `clampDevScopes` strips `ai:write:budgeted` only when `!keyCanSpend`, and `keyCanSpend` is the bearer's `AIServicesWrite` bit — now set on a generate-scoped login. So:

```
civitai login --scopes generate
civitai app dev-token my-block     # real-Buzz dev token, no --spend, no notice
```

Before this PR that sequence was read-only, because no OAuth credential could carry `AIServicesWrite`. This PR creates the first one that can.

A fix is in progress on this branch: the CLI will filter `ai:write:budgeted` out of its own declared list unless `--spend` is passed, with a notice on that path.

**Scope of that fix, stated precisely so it is not over-read:** it closes the case where a manifest **declares** scopes. With no manifest the CLI sends no `scopes` field and the server decides from the bearer — genuinely not expressible client-side, because narrowing is an intersection allow-list whose only sentinel is "absent ⇒ server decides", so "everything except X" cannot be written. The server-side closure is civitai/civitai#3703.

**Behaviour change to be aware of:** a full-scope personal API key plus a scaffolded money-path app gets budgeted spend by default today and will need `--spend` afterwards. Deliberate — the failure mode is the existing `block lacks ai:write:budgeted scope` 403, which names its own fix.

## 2. Nobody has completed a real login yet

Everything on this PR is verified against `httptest` and, for the ceiling, the live device-authorization endpoint. What has **not** been exercised: the device-token **mint** (which re-validates against `allowedScopes` a second time) and an actual `civitai generate` call over an OAuth bearer.

Whether the orchestrator's tRPC procedures accept an OAuth bearer carrying `AIServicesWrite` — as opposed to a personal key — is a server fact that cannot be established from the CLI test suite. One real `civitai login --scopes generate` followed by one `civitai generate` covers it, and that should gate undrafting.

## Already fixed since the audit

Commit `cdf777a`:

- The `--scopes` + `--token` conflict guard was partial — `login --token --scopes generate` and `login --scopes generate --token` both exited 0 and silently dropped `--scopes`. Now total across all four spellings, each exiting 2 with nothing stored.
- **A pre-existing credential leak is fixed.** `login --token <key> --no-browser` echoed the API key into stderr via cobra's `unknown command "<key>"`. Confirmed against both binaries: base `b403f59` leaks it, `cdf777a` redacts it. Not introduced here, but this PR pinned that exact invocation without noticing.
- Two guards were spelled rather than structural and had surviving mutants — deleting the `--scopes` help block, and swapping the two `whoami` guidance lines — both now die to their own assertions.
- The "corrected now-false claims" sweep was incomplete; ~20 further sites fixed, including scaffold templates that ship into every new App project.

ZacxDev:
## Undrafted — the end-to-end gate is met

Ran the full flow against production on this branch (`3c10afc`), not against `httptest`. Every bit in the widened mask is now exercised for real.

| Leg | Scope | Result |
|---|---|---|
| device authorization | — | 200, code issued |
| **consent screen** | — | renders all six labels, **including the three new ones** |
| device-token mint (re-validates the ceiling a 2nd time) | — | succeeded |
| stored credential | — | `scope: "100777985"`, `auth_kind: oauth`, mode `0600` |
| `whoami` | UserRead | `Spend Buzz (AI Services): yes` |
| `buzz` | BuzzRead (bit 16) | balance returned |
| `workflows list` | AIServicesRead (bit 14) — orchestrator | 200 |
| `generate --dry-run` | AIServicesRead — orchestrator whatIf | priced at 8 Buzz |
| **`generate`** | **AIServicesWrite (bit 15)** | **submitted, charged 8, `succeeded`, 1/1 outputs** |

```
workflow 11025902-20260806220433833 — scheduled → processing → succeeded (1m45s)
Saved 11025902-20260806220433833-1.jpg (166.7 KB, verified 1024x1024 JPEG)
Buzz 206 → 198, matching the estimate exactly
```

**Instrument validated where an empty result could have lied.** `workflows list` returned an empty list, which is indistinguishable from a call that silently failed — so I re-ran it with a corrupted token and got a **401**. The empty list was therefore a genuine 200, and the orchestrator really does accept an OAuth bearer. That was the one thing the audit called untestable from here.

So `civitai login --scopes generate` produces a token that generates, with no personal API key anywhere in the flow. That is civitai/civitai#3681's actual goal.

## What changed since the audit

- **`--spend` was inert** on the primary path — the default `page-money` scaffold declares `ai:write:budgeted`, so it shipped whether or not the flag was passed. Now filtered out of the CLI's own declared list unless `--spend`.
  A literal filter would have been **inert too**: the scaffold declares that scope as its *only* one, so filtering yields an empty slice, `omitempty` drops the key, and `clampDevScopes` narrows only `if (requestedScopes.length > 0)` — the server would have re-granted spend. The request now carries `user:read:self` (force-granted unconditionally post-clamp) as a sentinel, keeping the narrowing expressible while adding no capability.
- **The `--scopes` + `--token` guard was partial** — two of four spellings exited 0 and silently dropped `--scopes`. Now total; all four exit 2 with nothing stored.
- **A pre-existing credential leak is fixed.** `login --token <key> --no-browser` echoed the API key into stderr via cobra's `unknown command "<key>"`. Confirmed against both binaries: base leaks it, this branch redacts it.
- **Two guards were spelled rather than structural** and had surviving mutants — deleting the `--scopes` help block, and swapping the two `whoami` guidance lines. Both now die to their own assertions.
- ~20 further "only a personal API key can generate" claim-sites corrected, including scaffold templates that ship into every new App project.

## Scope — stated precisely so it isn't over-read

The `--spend` change closes the case where a manifest **declares** scopes. With **no manifest** the CLI sends no `scopes` field and the server decides from the bearer — genuinely not expressible client-side, because narrowing is an intersection allow-list whose only sentinel is "absent ⇒ server decides". Verified directly: default + no manifest still yields a spend-capable token.

The server-side closure is civitai/civitai#3703, which has a reviewed design and an implementation in progress. Nothing here claims to close it.

**Behaviour change to be aware of:** a full-scope personal API key plus a scaffolded money-path app gets budgeted spend by default today and needs `--spend` after this. Deliberate — the failure mode is the existing `block lacks ai:write:budgeted scope` 403, and a pre-mint notice now warns before you hit it.

## Not verified

No CLI revoke path exists, so a credential granted this way persists for the refresh-token TTL (~30 days, scope preserved across refresh); rotation is via the account page. Worth a follow-up.

PR #221, opened 2026-08-06
feat(generate): surface silent model substitutions before and after spending

ZacxDev:
## What and why

`orchestrator.generateFromGraph` / `whatIfFromGraph` do **not** reject a `model.id` that is invalid for a `modelLocked` ecosystem. They substitute the ecosystem's default checkpoint, run the job, and bill for **what actually ran** — and the reply used to be byte-indistinguishable from success. A nonexistent version id came back `200 OK` at the default price (context: civitai/civitai#3665).

The server now records each swap and returns it. This PR consumes that record and puts it in front of the user.

Today the CLI mitigates only the demonstrated case: `--checkpoint`/`--lora` are resolved against the public API before submitting, turning a nonexistent id into a hard local 404. That **cannot** catch same-ecosystem substitution — a real version id scoped to the wrong workflow, or one that is gated — which is exactly what the server record exists for.

## The wire contract consumed

`{ requested, applied, reason }`, `reason` ∈ `{wrong-workflow, unrecognized, gated}`. Verified against the server source rather than inferred — three carriers, and they differ:

| Path | Carrier |
| --- | --- |
| `whatIfFromGraph` | **top-level only** — nothing is persisted, so the reply is the only copy |
| `generateFromGraph` | **top-level AND `metadata`** |
| later read (`getWorkflow`) | **`metadata` only** |

`SubmitResult.Substitutions()` prefers the top-level copy (the server builds it from the validation it just performed) and falls back to metadata. It must never concatenate, or one swap reports as two.

## What it does

**Warns before spending.** The estimate runs unconditionally on every path, so one call site there covers `--dry-run`, `--dry-run --json` and the full submit flow. On an interactive run this prints **before** the `[y/N]` confirmation, so the user can still back out.

**Reports after submit**, from the submit reply's own record — the receipt for what was billed. Deliberately placed at the reply, not in a renderer: there are **two** submit renderers (`printSubmitResult` for `--no-wait`, `printSubmitted` for the waiting path) and this repo already has a whole test file because `Charged:` was added to only one of them.

**Reports on a later read** (`civitai workflows get`). `--no-wait` prints an id and sends the user there, so for that documented flow this is the only surviving evidence.

**Per-reason advice** — the three cases want different fixes. The server's `reason` token is printed **raw** per AGENTS.md item 8, with advice on a separate line. The id line is **tensed by phase**: `will run` before the money moves, `ran` after.

**`--json`**: the existing raw passthrough already carries the record on stdout, so scripts get it verbatim. All human output goes to **stderr** in every mode, so `--json` stdout stays parseable — asserted by decoding stdout, not substring-matching.

## Design decision: warning by default, `--fail-on-substitution` to opt in

**Default is informational, exit 0.**

- The substitution is a **deliberate graceful degradation**. A script pinned to a version that was later retired keeps producing images instead of breaking. Making this a hard CLI failure by default would override a server-side product decision and break working automation on a CLI *upgrade*.
- The real protection for a human already exists without touching exit codes: the warning prints **before** the confirmation prompt.
- Scripts that want to branch on it can — the record is in `--json`.

For callers who would rather fail than get a different model, **`--fail-on-substitution`** refuses, evaluated against the **estimate**, so nothing is submitted and nothing is charged. Tagged `ErrModelSubstituted` so it is `errors.Is`-assertable rather than message-matched.

🔴 **It is not a spend guard, and now says so in its own `--help`.** It can only refuse what the server *reports*: against a deployment that omits the field it is **silently inert** — exit 0, submitted, charged — and nothing distinguishes that from "nothing was substituted". It also does **not** re-fire on the submit reply: by then the money is gone, and aborting would strand outputs the caller paid for and still needs to collect. A post-submit-only substitution is reported and present in `--json`, but exits **0**. Both limits are in the README and AGENTS.md 20(d).

## Backward compatibility

The key is **omitted** when nothing was substituted, so an older server decodes exactly as before.

🔴 **Absence is ambiguous** — "no substitution" *or* "a server predating the field" are the same bytes. The CLI prints **nothing at all** rather than a reassuring "no substitutions", and a test asserts it never claims the negative. This is also why the field is a plain slice, not a pointer (contrast AGENTS.md item 9's `AppAnalytics.Views`): the server never sends `[]`, so there is no third state, and no Go type can resolve an ambiguity inherent to the protocol.

An **unrecognised** reason from a newer server still warns with the ids intact.

---

## Verification

🔴 **This section supersedes the original "17/17" claim, which was misleading.** That first matrix was **renderer-scoped**: every mutant targeted the internals of `reportModelSubstitutions` / `substitutionRefusal` / `Substitutions()`, so the **arguments** passed to them, the **third phase**, and the message **operand order** were structurally outside it and could not have been caught by it. Two independent audits followed. AGENTS.md 20(g) now records the lesson.

### Round 2 — audit found 11 survivors (commit `67bd9eb`)

The behaviour was correct on every path the auditor constructed; the gaps were in my testing.

- The phase guard was **spelled, not structural**: `Contains(got, "charged")` was satisfied by the *quote renderer's* unrelated line, so mutating the estimate call site to `substitutionAfterSubmit` kept the suite green while `--dry-run` announced "HAS BEEN CHARGED". Fixed with `substitutionLead(phase)` + `assertPhase`, which also requires the other phases' leads absent.
- `substitutionOnRead`'s lead was asserted by nothing — emptied, duplicated and pre-spend-framed variants all survived.
- The refusal message's operands were unpinned: swapping `Applied`/`Requested` stated the substitution **backwards on the money-refusal line**; `subs[0]`→`subs[len-1]` and an `(and %d more)` off-by-one also survived.
- Also pinned: the `substitutionFlagHint` call site, `len()>0` vs `!= nil`, and `ErrModelSubstituted` → exit 1 (plus its four sibling generate sentinels, whose "generic 1, never auth" contract existed only as a comment).

**16/16 killed for their own reason; negative control survived.**

### Round 3 — delta re-audit: one regression, one class still open (commit `1681838`)

A 26-mutant battery against a fresh copy of `67bd9eb` killed everything and confirmed the round-2 fixes. It also found:

- 🔴 **A regression I introduced**: the phase-tensed verb made the estimate emit `will run`, but **README's transcript still showed `ran version`** — so the README displayed, verbatim, the self-contradiction the change existed to remove. Nothing compared documented console blocks to real output. Now `TestREADME_DryRunTranscriptMatchesRealOutput` renders the documented fixture through the real renderer at the documented phase and requires the id line in README byte for byte, and requires the stale form gone.
- 🔴 **The lead's money claim was still keyword-guarded** — the same spelled-guard class, on the one path I hadn't anchored. `Contains(lead, "CHARGED")` is satisfied by a lead saying the *opposite*: `"…and NOTHING was CHARGED — the run is free."` survived on `onRead`, and would tell someone reading an already-billed workflow that it was free. `assertPhase` cannot see this by construction — it proves *which* lead was selected, never what it *says*. Each lead is now pinned as an **exact literal**.
- **The renderer was only ever exercised with one record**, so `range subs[:1]` survived — a renderer silently dropping every substitution after the first was untested. Added multi-record coverage at the renderer and end-to-end, with an exact id-line count.
- **A test claiming more than it proved**: `…IsInertAgainstAnOldServer` was byte-identical to the no-substitution test beside it — necessarily, since old-server and no-substitution are the same bytes. Merged into one honest test stating what it proves and what no fixture can.
- **A comment named a guard that does not exist** (`substitutionLeadsAreDistinct`). `TestSourceComments_NameOnlyIdentifiersThatExist` now resolves every Go-shaped identifier in this file's comments against real code repo-wide — deliberately **not** scoped to `Test*`, because the real bug had no such prefix and a first version keyed on `Test*` was walked straight past by the actual mutation.

| # | Round-3 mutation | Verdict |
| --- | --- | --- |
| F1a | README reverted to stale `ran version` line | KILLED |
| F1b | renderer verb changes, README not updated | KILLED |
| F1c | README reason token drifts | KILLED |
| F2a | `onRead` lead inverted to "NOTHING was CHARGED" | KILLED |
| F2b | estimate lead gains "Your card HAS BEEN CHARGED." | KILLED |
| F2c | `afterSubmit` lead inverted | KILLED |
| F4 | comment names a nonexistent guard | KILLED |
| F5a | renderer drops all but the first record | KILLED |
| F5b | renderer drops the first record | KILLED |
| F5c | renderer drops advice lines | KILLED |
| F5d | estimate path drops all but first (end to end) | KILLED |
| **NC** | **negative control** — cosmetic advice wording | **SURVIVED (expected)** |

**11/11 killed, each for its own guard's message.**

### Gate (current tree)

`go vet`, `gofmt -s` clean. Suite **counted**: **1308 top-level PASS / 2147 incl. subtests / 0 FAIL / 4 SKIP**, no timeout panics. The 4 skips are pre-existing env-gated live guards (`CIVITAI_CHECK_*`), untouched.

`golangci-lint` **0 issues** — and the instrument was **validated before its verdict was believed**: run from inside the module with `./...`, then fed a planted defect, which it flagged at the exact line (`generate_substitution.go:214`). A `0 issues` from a linter never shown to go red is a claim about the command line.

Re-verified on the **real binary** against a mock on a port checked free first: `--dry-run` emits the exact line README documents; all three records render; `workflows get` shows the post-spend framing; the refusal reads *"would run version 2436219 instead of the version 999999999 you asked for (and 2 more)"* with a real (unpiped) **exit 1**.

### Open question from the audit, answered

**Can a CLI submit hit the `workflowMetadata` falsy branch** (`orchestration-new.service.ts:1574`), leaving `workflows get` nothing to report? **No.** `workflowMetadata` is `isWhatIf ? undefined : removeEmpty({...})`; only `isWhatIf` makes it falsy, and that path never submits. `removeEmpty` is lodash `omitBy` (`utils/object-helpers.ts:3`), which always returns an object — worst case `{}`, which is truthy.

The formatter's `hasWfMeta` gate does exist but is **irrelevant here**: `civitai workflows get` calls `orchestrator.getWorkflow`, a raw passthrough, so the CLI never goes through the formatter.

### Not verified

**Never exercised against live civitai.com.** All end-to-end runs use a local mock; the submit spends real Buzz. The wire shape comes from the server source and its tests, not a live capture.

Also note this repo's GitHub Actions runs are currently stuck `queued` fleet-wide (including pre-existing runs on `main`), so the local gate above is the only executed verdict.

Based directly on `main`, not stacked.


ZacxDev:
## Audit follow-up — call-site and third-phase gaps closed (`67bd9eb`)

An independent adversarial audit built its own 20-mutant battery (harness validated first) and got **8 killed / 11 survived**. It found no 🔴 in the shipped behaviour — every path it constructed behaved correctly, including divergent carriers, empty array, old server and post-submit-only.

**The survivors were a gap in my testing, not in the behaviour — and I want to be explicit about why my earlier 17/17 was misleading.**

🔴 **That matrix was RENDERER-SCOPED.** Every mutant I built targeted the internals of `reportModelSubstitutions` / `substitutionRefusal` / `Substitutions()`. The **arguments passed to** those functions, the **third phase**, and the refusal message's **operand order** were structurally outside the battery, so they could not have been caught by it no matter how many mutants I added. This is the "verified in isolation" failure: the components were each swept clean, and the defects lived in the seam. A second battery *constructed differently* is what surfaced them — which is exactly the argument for not trusting a single sweep's shape.

### What was wrong

**The phase guard was SPELLED, not structural.** `Contains(got, "charged")` was satisfied by a pre-existing unrelated line from the *quote* renderer ("Estimate only — nothing was submitted and nothing was charged"). So mutating the estimate call site to `substitutionAfterSubmit` kept the entire suite green while `--dry-run` printed *"this generation HAS BEEN CHARGED for the model that actually ran."* Same at `workflows.go` in the other direction.

Fixed structurally: `substitutionLead(phase)` derives the expected text **from the phase constant**, and `assertPhase` additionally requires the *other* phases' leads to be **absent**. A wrong argument now moves the expectation and fails. `TestSubstitutionLead_AllPhasesNonEmptyAndPairwiseDistinct` backs it, because an empty or duplicated lead would silently disarm every one of those assertions — `Contains(x, "")` is always true, which is precisely how the emptied-lead mutant survived.

**`substitutionOnRead`'s lead was asserted by nothing** — three survivors, the worst being the *pre-spend* framing, which made `civitai workflows get` tell someone whose generation was already billed that "Nothing has been submitted or charged yet".

**The refusal message's operands were unpinned.** Swapping `Applied`/`Requested` survived and stated the substitution **backwards on the money-refusal line** — naming the model the caller wanted as the one that would run. `subs[0]` → `subs[len-1]` and the `(and %d more)` off-by-one also survived.

**Also fixed:** the id line's verb is now tensed by phase (`will run` pre-spend, `ran` after — it previously said "ran version N" under a lead saying nothing was submitted); the `substitutionFlagHint` call site is pinned (deleting it survived); `len() > 0` vs `!= nil` in `Substitutions()` is pinned; and `ErrModelSubstituted` → exit 1 is pinned in `main_test.go`, along with its four sibling generate sentinels whose "generic 1, never auth" contract existed only as a comment.

### `--fail-on-substitution` is not a spend guard — now said out loud

The auditor is right that this was under-disclosed. The flag can only refuse what the server *reports*, so against a deployment that omits the field it is **silently inert: exit 0, submitted, charged** (verified). The usage string now says `NOT A GUARANTEE … silently inert … cannot be relied on as a spend guard against an older deployment`, pinned by a test. The README's exit-code row also implied exit 1 whenever a substitution is reported; a **post-submit-only** substitution with the flag set exits **0** (AGENTS.md 20(d) said this, the README did not). Both now match.

### New mutation matrix — the exact reported survivors

Each mutant re-applied to the committed tree; failure confirmed to be **that guard's own message**.

| # | Mutation (all previously SURVIVED) | Verdict |
| --- | --- | --- |
| F1 | estimate call site → `substitutionAfterSubmit` | KILLED |
| F2 | `workflows.go` call site → `substitutionAtEstimate` | KILLED |
| F3 | submit call site → `substitutionOnRead` | KILLED |
| F4 | `onRead` lead emptied | KILLED |
| F5 | `onRead` lead identical to post-submit | KILLED |
| F6 | `onRead` lead carries pre-spend framing | KILLED |
| F7 | refusal operands swapped | KILLED |
| F8 | refusal names the last record | KILLED |
| F9 | `(and %d more)` off-by-one | KILLED |
| F10 | verb always past tense | KILLED |
| F11 | `substitutionFlagHint` call site deleted | KILLED |
| F12 | `len()>0` → `!= nil` | KILLED |
| F13 | `ErrModelSubstituted` misclassified as auth | KILLED |
| F14 | help drops the NOT-A-GUARANTEE clause | KILLED |
| R1 | `safeTerm` dropped (line my refactor touched) | KILLED |
| R2 | empty record claims "no substitutions" | KILLED |
| **NC** | **negative control** — cosmetic string | **SURVIVED (expected)** |

**16/16 killed for their own reason; control survived.**

### Gate

`go vet`, `gofmt -s` clean; `golangci-lint` **0 issues**; suite **counted**: **2143 PASS / 0 FAIL / 4 SKIP**, no timeout panics. The 4 skips are pre-existing env-gated live guards (`CIVITAI_CHECK_*`), untouched.

Re-verified on the **real binary** against a mock: `--dry-run` shows pre-spend framing with `will run`; `workflows get` shows `CHARGED` with `ran`; the refusal reads *"would run version 2436219 instead of the version 999999999 you asked for (and 2 more)"* with a real (unpiped) **exit 1**.

### Your open question: can a CLI submit hit the `workflowMetadata` falsy branch?

**No — answered, not guessed.** `workflowMetadata` is `isWhatIf ? undefined : removeEmpty({...})` (`orchestration-new.service.ts:1195`). The only thing that makes it falsy is `isWhatIf`, and that path never submits or persists anything. On a submit it is always an object: `removeEmpty` is lodash `omitBy` (`utils/object-helpers.ts:3`), which always returns an object — worst case `{}`, and `{}` is truthy. So `workflowMetadata && modelSubstitutions?.length` can never drop the record on a real submit.

One adjacent gate worth naming since it looks like the same hazard: the *formatter*'s `hasWfMeta` (requires a `params` key) does gate read-back in `formatGenerationResponse2`. It is **irrelevant to this CLI** — `civitai workflows get` calls `orchestrator.getWorkflow`, a raw passthrough of the orchestrator workflow, so the CLI reads `metadata.modelSubstitutions` directly and never goes through the formatter.

### Not fixed, deliberately

Nothing from the list was skipped. The one thing I did **not** add is a re-fire of the refusal on the submit reply — unchanged from the original rationale (the money is already gone; failing there strands a paid-for result), now documented in the README rather than only in AGENTS.md.

### Still not verified

Unchanged: **never exercised against live civitai.com.** All end-to-end runs use a local mock; the submit spends real Buzz. The wire shape comes from the server source and its tests, not a live capture.

ZacxDev:
## Live end-to-end verification against production — the feature works, zero Buzz spent

Carrying this over from #222 (closed as a duplicate of this PR). It validates the **server contract**, not one implementation, so it applies here.

Neither PR had observed a real substitution — both were verified against `httptest` fixtures built from server source. Constructing one turns out to be straightforward: pass a **real** version id that is wrong for a `modelLocked` ecosystem.

Read-only `whatIfFromGraph` only — no submit, no cancel. Balance byte-identical before and after in all four components (Blue 1339091 / Green 999980 / Yellow 1849049).

**Positive control** — `--ecosystem NanoBanana --checkpoint 2436219` (the correct id):
→ 160 Buzz, **no** `modelSubstitutions` field, **no** warning.

**Test case** — `--ecosystem NanoBanana --checkpoint 128713` (DreamShaper, a real id from another ecosystem):
→ 60 Buzz, and the server returned:

```json
{"requested":128713,"applied":2154472,"reason":"unrecognized"}
```

The CLI printed the warning with `requested`/`applied` the right way round and both ids resolved to names.

That is the hazard reproduced exactly as issue #3665 described it: a **real** id, HTTP 200, a different model runs, you are billed, and before #3692 nothing in the reply said so. Two useful details fall out of it — `reason` came back `unrecognized` (not `wrong-workflow`) for a foreign-ecosystem id, and the 160 → 60 price drop matches the earlier measurements, so cost alone would not have identified *which* model ran.

## Two findings that apply regardless of implementation

**1. Advisory name resolution does blocking, retried network I/O in front of the workflow id for an already-charged job.** The post-submit render resolves substituted ids through `ResolveModelVersion` → `getWithRetry` (`readMaxAttempts = 4`, `defaultTimeout = 30s`) *before* any path prints the workflow id. Worst case ≈2 min per id, two ids per record. The parsing side argues carefully that a malformed advisory field must never cost the user the handle to a job they have paid for; the render side reintroduces the same exposure by a different mechanism. Cheap fix: a short `context.WithTimeout` around the lookups, or print the id first and enrich after.

**2. The approval summary still names the model that will not run.** The confirm and dry-run summaries print `Checkpoint: <requested>` — so with a substitution in play, the last model line before `Generate? [y/N]` is the wrong one:

```
⚠ The server SUBSTITUTED a different checkpoint …
  Ran:        Google's Nano Banana — Nano Banana (Checkpoint, id 2154472)
…
Checkpoint:   DreamShaper — 8 (Checkpoint, id 128713)
```

AGENTS item 13 sells that echo as "approve a name rather than an integer", which cuts against showing the requested one unannotated at the moment of approval.

Both are follow-ups rather than blockers — the warning itself fires correctly, which is the thing that matters.

PR #222, opened 2026-08-06
feat(generate): surface the server's silent model substitutions

ZacxDev:
## What

On a `modelLocked` ecosystem the generator replaces a checkpoint version id it
does not recognise for the chosen workflow with that workflow's default, returns
HTTP 200, and **bills you**. `ResolveModelVersion` (item 13) already turns a
**nonexistent** id into a local 404 — but it cannot catch a **real** id that is
wrong for the ecosystem you named, and until now nothing did.

civitai/civitai **#3665** (PRs **#3692**, **#3673**) makes the server report every
swap as `modelSubstitutions` — `{requested, applied, reason}`. This PR reads it
and says so, plainly: *you asked for version N, the server ran version M,
reason R.*

```
⚠ The server SUBSTITUTED a different checkpoint — the estimate below prices the
  model it would actually run, not the one you asked for.
  Requested:  DreamShaper — v8 (Checkpoint, id 128713)
  Ran:        SDXL 1.0 (Checkpoint, id 999001)
  Reason:     unrecognized — that version is in none of this ecosystem's lists …
```

## 🔴 Three read sites that do not agree — both carriers are read

Per the reply-contract comment on `NormalizedWorkflowMetadata`
(`orchestration-new.service.ts` ~1940-1968):

| site | carrier |
|---|---|
| `whatIfFromGraph` | **top-level only** — nothing is persisted on that path |
| `generateFromGraph` | **both** top-level and `metadata.modelSubstitutions` |
| later reads (`getWorkflow` poll / re-attach / feed) | **`metadata` only** |

A client reading only the top level sees substitutions on the submit and never
again; one reading only `metadata` misses the whatIf entirely. `internal/genapi`
reads both and merges, deduping exact records so a two-carrier submit reply is
reported **once**.

## Surfaces

`--dry-run` (human **and** `--json`), the TTY confirmation, `--yes`, the
post-submit output, and `workflows get`. The pre-spend warning is emitted from
**one call site** in `runGenerate`, before every branch — the img2img caveat
shipped as an interactive-only print and so missed `--yes` (mandatory in CI) and
`--dry-run`; a single call site makes that class of miss unreachable. Everything
goes to **stderr**, so a `--json` stdout stays machine-clean and still carries
the raw field. Ids resolve to model names, and a **failing lookup prints the id
rather than suppressing the warning**.

## Warn, never refuse

Argued in `printModelSubstitutions` and AGENTS.md item 13, not defaulted:

- Substitution on a `modelLocked` ecosystem is **legitimate** graceful
  degradation — it keeps a job pinned to a since-retired version running instead
  of hard-failing. Refusing breaks flows that work today.
- The platform has **explicitly deferred** the rejection decision ("deciding
  whether any case should REJECT is a later phase, gated on what the counter
  measures"). Refusing first would vendor a policy the server has not made —
  item 13's anti-mirror rule applied to policy.
- A refusal would be **unenforceable**: absence of the field means "no
  substitution" *or* "a server predating #3665", indistinguishable on the wire.
- Two of the four surfaces are **post-spend**, where refusing buys nothing.

Same fail-soft shape as `serverQuantityClamp` (item 13b).

## Two things that look wrong and are not

1. **Every carrier is a `json.RawMessage`, parsed leniently.** A typed slice
   would let a malformed advisory field fail the whole `json.Unmarshal` of the
   **submit reply** — the message that hands back the workflow id for a job the
   user has already been charged for. A warning we cannot parse must never cost
   them the handle. Eight malformed shapes are pinned as tests.
2. **`reason` is deliberately NOT narrowed against the server's union.** The
   server narrows it because the value doubles as a bounded Prometheus label; the
   CLI has the opposite incentive — vendoring the union would DROP a reason added
   server-side and restore the exact silence this closes. Unknown reasons render
   verbatim; the gloss table is advisory and can never gate the warning.

## Corrections

AGENTS.md item 13 and `genapi/versions.go` both claimed the correction is
"invisible through a non-browser path". That was true when written and is now
**false**; both are corrected in place with the citation, and item 13 records all
three read sites and the read-both requirement. The signal is a **live server
read, not a vendored table**, so it *strengthens* item 13's anti-mirror thrust.

## Tests

`httptest` + `t.TempDir()`, no network, no mutation against a live server.

- All three carriers, plus **a no-substitution reply on each** with the populated
  case as its positive control in the same file.
- All five surfaces, with `submitCalls`/`whatIfCalls` positive controls.
- `--json` passthrough keeps the field; the warning never leaks into stdout.
- A failing `ResolveModelVersion` still warns (and the resolver-was-called
  counter proves the failure was reached).
- Exit codes pinned by `errors.Is`, never message text.
- The `Requested:`/`Ran:` assertion is **line-scoped**, not a substring search —
  a transposition would otherwise pass while inverting the warning's meaning.

**Mutation sweep: 18 semantic mutants, all killed, each by its own guard's
message** — branch inversion (`||`→`&&`, `== 0`→`> 0`), operand swap
(requested/applied transposed), operand drop (dedupe key ignores `reason`),
a one-character metadata-key typo, per-carrier removal, call-site removal, and
call-site *relocation* reproducing the historical img2img bug shape (warning
moved after the `--dry-run` exit / below the `--json` early return). An
**M0 no-op control survived**, proving the harness distinguishes a harmless edit
from a real one. Determinism: 8/8 runs, 41 PASS each.

## Gate

- `make ci` — green.
- `go test ./... -count=1 -v` **counted**: **2148 PASS, 0 FAIL, 4 SKIP** (the four
  by-design skips), no `panic: test timed out`.
- `gofmt -s -l .` — 0 files listed across 264 `.go` files (and it caught a real
  break earlier in this branch, so the tool is demonstrably live).
- `golangci-lint run` (v2.12.2) — **0 issues**, validated against a planted
  known-bad first: `ST1008` (error not last) reported, plus `unused` ×2 and
  `misspell` ×1; probes removed.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -589,12 +589,84 @@ neither one's.
     **nonexistent** checkpoint id is otherwise accepted with HTTP 200, the
     ecosystem default silently substituted, and **billed** (measured on one
     ecosystem: the correct id priced 160, while a nonexistent id, a
-    foreign-ecosystem id, and no model at all ALL priced 60 — the server's own
-    comment says this correction is visible on-site and invisible through a
-    non-browser path). And `Graph.Resources` entries **require** `model:{type}`,
+    foreign-ecosystem id, and no model at all ALL priced 60). And
+    `Graph.Resources` entries **require** `model:{type}`,
     which is not derivable from a version id — a bare id and `{id}` alone are
     both 400s, while a *wrong* type is silently accepted. The type must come
     from the lookup, never a guess.
+    🔴 **CORRECTION: the substitution is no longer invisible.** An earlier
+    revision of this item said the server's own comment calls the correction
+    "visible on-site and invisible through a non-browser path". That was true
+    when written and has been **FALSE since civitai#3665** (PRs **#3692** and
+    **#3673**). The server now REPORTS every silent checkpoint swap as
+    `modelSubstitutions`, an array of
+    `{requested, applied, reason}` (`PersistedModelSubstitution` in
+    `civitai/civitai → src/shared/data-graph/generation/model-substitution.ts`;
+    `reason` ∈ `wrong-workflow` | `unrecognized` | `gated`). This **strengthens**
+    item 13 rather than contradicting it: it is a **live server signal**, not a
+    vendored table, so it is the same category as `ResolveModelVersion` — the
+    thing this item says IS allowed — and it closes the one gap that lookup
+    structurally cannot, a **real** id that is wrong for the chosen ecosystem or
+    workflow.
+    🔴 **THERE ARE THREE READ SITES AND THEY DO NOT AGREE, so the CLI reads BOTH
+    carriers.** Per the reply-contract comment on `NormalizedWorkflowMetadata`
+    (`civitai/civitai → src/server/services/orchestrator/orchestration-new.service.ts`,
+    ~line 1940-1968):
+    - `whatIfFromGraph` → **top-level `modelSubstitutions` only**. Nothing is
+      persisted on that path, so the reply is the sole carrier — and it is the
+      most valuable one, because it is the only place a caller learns *before*
+      spending, and the only signal for a SAME-PRICE swap.
+    - `generateFromGraph` → **both** top-level (authoritative for the validation
+      just performed, and present even when the graph produced no workflow
+      metadata to round-trip) **and** under `metadata`.
+    - any later read (`getWorkflow` poll / re-attach / `queryGeneratedImages`)
+      → **`metadata` only**; the top-level copy is gone.
+    A client reading only the top level sees substitutions on the submit and
+    never again; one reading only `metadata` misses the whatIf entirely.
+    `internal/genapi/substitution.go` reads both and merges, deduping exact
+    records so a two-carrier submit reply is reported once. The metadata key is
+    the literal `modelSubstitutions`
+    (`WORKFLOW_METADATA_MODEL_SUBSTITUTIONS_KEY`), the same on the raw
+    orchestrator workflow `getWorkflow` returns and on the normalized shape.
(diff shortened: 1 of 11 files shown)


ZacxDev:
Closing as a duplicate of #221, which is the same feature and further along.

#221 was opened at 20:58; this branch was started ~1.5h later without checking for in-flight work first. That is on me — the duplication is a process failure, not a disagreement about the approach.

**#221 is the better of the two.** An independent audit of this PR found three test gaps, and #221 already closes all of them:

| gap found here | already covered in #221 |
|---|---|
| the `substitutionPhase` argument is unpinned at **both** call sites, so flipping the tense (pre-spend ↔ already-charged) survives the whole suite | `TestSubstitutionLead_ExactTextPerPhase`, `TestSubstitutionVerb_IsTensedByPhase`, `TestReportModelSubstitutions_PhaseChangesTheSpendFraming` |
| the post-submit `--json` surface is untested; a call-site relocation below the `--json` early return survives | `TestGenerate_JSONCarriesTheRecordAndStdoutStaysParseable` |
| every fixture carries exactly one substitution, so a renderer dropping all but the first is green | `TestGenerate_EstimateReportsEveryRecord` |

#221 also offers refusal as an opt-in rather than warn-only.

Worth recording that the two implementations converged **independently on the same audit findings** — #221's own follow-up describes closing a "phase guard was SPELLED, not structural" gap and a "verified in isolation" seam, which is precisely what the audit here found still open. Two separately-constructed mutation batteries landing on the same class is reasonable evidence the class is real rather than an artifact of one sweep's shape.

The one thing this branch produced that #221 did not have — **live end-to-end verification against production, with a real substitution and zero Buzz spent** — has been posted to #221, since it validates the server contract and therefore applies to either implementation.

No further work needed here.

PR #240, opened 2026-08-07
docs(generate): the `--input` gate guards a CONFIRMED audit gap, not a suspected one

ZacxDev:
## What

The `--input` txt2img gate is documented as guarding a *suspected* server-side gap. It guards a **confirmed** one. This records the evidence in the two places a future reader would consult before lifting it, and changes no behaviour.

The dangerous artifact this replaces: the session handoff for this feature recorded the question as **"Ruled out — IMPORTANT, do not re-derive: there is no known bypass."** That note is false, and a reader trusting it would lift the gate.

## Evidence

Verified by code path against `civitai@a7e0bcd668`. The audit inside `generateFromGraph` gates on:

```ts
// orchestration-new.service.ts:1460
if ('prompt' in data && typeof data.prompt === 'string' && data.prompt.trim()) {
```

`data` is rebuilt from declared graph nodes only, and two shipped ecosystems declare their text elsewhere:

**Hunyuan3D — the audit is skipped outright.** It declares no `prompt` node, only `hunyuanPrompt` (`hunyuan3d-graph.ts:71`). Its own header explains the rename: the bare names *"collide with the standard image Controllers in `GenerationForm.tsx`"* (`:12`). So `data.prompt` is absent, `:1460` is false, `auditPromptServer` is never called — and the handler then maps the name back for the generator:

```ts
// hunyuan3d-graph.handler.ts:58
prompt: hunyuanPrompt ? hunyuanPrompt : undefined,
```

**PolyGen — `texturePrompt` is audited by nothing.** 600 chars of free text (`polygen-graph.ts:162`), reaching the orchestrator (`polygen.schema.ts:143`) and rendered publicly (`pages/3d-models/[id]/[[...slug]].tsx:279`). On `img2model3d` it is the only text in the request, because that workflow's audited `prompt` node is gated `when: ctx.workflow.startsWith('txt')` and evaluates away.

**Sweep basis:** all 79 declared node keys across `src/shared/data-graph/generation/*.ts`, checking every node whose input is a bare `z.string()`. The complete free-text set is `prompt`/`negativePrompt` (audited `:1467`), `musicDescription`/`lyrics` (audited `:1510`), `hunyuanPrompt`, `texturePrompt`, and ACE Audio's `title` (display-only per its comment).

Neither gap was introduced deliberately — Hunyuan3D renamed its nodes to dodge a **UI Controller** collision and changed its **audit** status as a side effect, because the two are keyed on the same strings with nothing tying them together. That is precisely the mechanism civitai/civitai#3667 was filed to guard.

## What this does NOT claim

- **The CLI is not what holds the line.** Both fields are reachable from the first-party form (`GenerationForm.tsx:1528`, `:2014`) and from any direct tRPC caller with a personal API key. The gate stops accidents, not adversaries. Keeping it just avoids adding a second client to a confirmed unaudited path while it is open.
- **No live generation probe was run.** This is reachability by code path, not a demonstrated exploit.
- Escalated on civitai/civitai#3667 with the same evidence — the real fix is the server-side coverage guard, not anything in this repo.

## Verification

No behaviour change; no new tests (nothing new to assert).

- `go build ./...` — ok
- `go vet ./...` — clean
- `gofmt -s -l .` — 0 files, and a **planted** malformed file was flagged, so the tool was really scanning
- `go test ./internal/cmd/ -count=1 -v` — **1185 `--- PASS`, 0 `--- FAIL`**, no timeout panic (counted, not read off the exit code)

## Note

I deliberately did **not** add an AGENTS.md "intentional decisions" item. Several open PRs already touch that file and its own rule makes the second merger renumber — the `inputWorkflow` comment is where a reader deciding to lift the gate actually lands. Say the word if you'd rather it be item 22.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -160,10 +160,11 @@ func newWhoAmICmd() *cobra.Command {
 
 Items 1–3, 10 and 11 are deliberate mirrors of the platform (items 4 and 8 are
 deliberate *non*-mirrors); items 5–9 cover `civitai app metrics`, the CLI's only
-analytics read path; items 12–17 and 19 cover `civitai generate`, the CLI's only
-path that **spends the user's money irreversibly** (19 is img2img); item 18
-covers the two checks that tell an author their EXISTING app is missing the
-item-11 handshake. The durable fix for the mirroring is a server-side
+analytics read path; items 12–17, 19, 21 and 22 cover `civitai generate`, the
+CLI's only path that **spends the user's money irreversibly** (19 is img2img, 21
+is model substitution, and 22 is the one gate on that path that guards CONTENT
+rather than money); item 18 covers the two checks that tell an author their
+EXISTING app is missing the item-11 handshake. The durable fix for the mirroring is a server-side
 `civitai app validate` endpoint that calls the real `BlockManifestValidator` —
 until that exists, vendoring is on purpose.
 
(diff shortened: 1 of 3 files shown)


ZacxDev:
Added **AGENTS.md item 22** and rebased on `main` (now includes #237).

Reason for the addition, having first decided to skip it: item 13 states in bold that *"the CLI deliberately validates NOTHING about the generation graph"*, and `--input` exists precisely to pass a graph through untouched — so the txt2img refusal reads as a **contradiction of a documented rule**, which is an active invitation to delete it. The code comment alone only reaches someone already editing `generate_input.go`. Item 22 reconciles the two: item 13 forbids reproducing the server's *judgement* about which graphs are valid; this check makes no such claim, and is the same shape as item 19(b) — a flag combination the CLI owns.

Also corrected the intro range clause, which listed neither item 21 nor 22 as covering `civitai generate` (item 21 was pre-existing drift).

Re-verified on the rebased tree: **2382 `--- PASS` / 0 `--- FAIL` / 4 `--- SKIP`** (counted, not the exit code; the 4 are the documented by-design skips), `gofmt -s -l .` 0 files, `go vet` clean.

Issue #260, opened 2026-08-07
docs/UX gaps found by a blind dogfood run (undocumented commands, no README TOC, no listing-media on-ramp)


ZacxDev:
Umbrella issue for the documentation and messaging gaps found by a blind dogfood run against `v0.1.90-21-gf56aa72` — an agent given only the built binary, `README.md` and `--help`, working as a first-time app author with no account.

Code defects from the same run are filed separately. These are docs/UX: real friction, nothing broken.

---

### 1. A fresh scaffold fails its own validation, and `--help` promises the opposite

`civitai app create --help` says verbatim: *"The scaffold is immediately runnable […], test-green, and **validates clean**."*

```console
$ civitai app create my-first-app && cd my-first-app && civitai app validate ; echo $?
✗ 1 validation error(s): package.json is present but no lockfile is committed […]
1
```
Same for `page-vite`; `static` does validate clean (it has no `package.json`). Validating before installing ~103 MB of `node_modules` is the natural first instinct, so this is a minute-3 experience.

The error message itself is excellent. The problem is the promise. Either qualify the help text ("validates clean once you have run `npm install`") or have `app create`'s Next-steps block say so — it currently says *"1. cd … && npm install # writes package-lock.json — COMMIT it"* without mentioning that `validate` fails until you do.

### 2. Three commands exist that the README never documents

`civitai app --help` lists `list`, `view` and `pull`. The README's command-reference table contains **none** of them — `grep -c 'app pull' README.md` → **0**.

`app pull` is not a minor read command: its `--help` calls it *"the read side of git authoring"*, it lazily provisions a Forgejo identity, and it carries a 10-line **⚠ SECURITY — TOKEN-IN-URL LEAKAGE** warning about your token landing in `.git/config`. A user who reads only the README does not know a git-based authoring path exists, let alone its hazards.

Related contradiction: the README Quickstart says *"Reads are **anonymous** — no `login` needed"* and root `--help` says *"Reading is anonymous"*, but `app list` and `app view` both require a credential:
```console
$ civitai app list ; echo $?
Error: no token configured — run `civitai login` (or set CIVITAI_TOKEN) to browse the App store
3
```
`civitai app list --help` explains this properly ("the endpoint keys the visible catalog off your identity") — but since the README never mentions the command, that explanation is unreachable from the docs.

### 3. The README is 1617 lines / ~104 KB with no table of contents

[...]

PR #267, opened 2026-08-07
fix(scaffold): refuse to mangle a non-ASCII name into a blockId, add --slug, and echo the derived id

### LINKED ISSUE #259: slug derivation silently drops non-ASCII characters, and --dir never echoes the derived blockId
Found by a blind dogfood run against `v0.1.90-21-gf56aa72`.

## Symptom

Slug derivation **drops** non-ASCII characters instead of transliterating or rejecting them, so a leading accented letter silently vanishes from the app's permanent public identity.

```console
$ civitai app create "ÜberApp Ω" --dir ./nm-uni -y
✓ Created App "ÜberApp Ω" …
$ grep blockId ./nm-uni/block.manifest.json
"blockId": "berapp"
```

`Ü` is dropped rather than becoming `u`/`ue`, so `ÜberApp` becomes **`berapp`** — a word that is not a truncation, a transliteration, or anything the author typed.

## Why it matters

`blockId` is not cosmetic. Per the README the app is served at `https://<blockId>.civit.ai/`, and it is the identity every later command takes (`app status <slug>`, `app metrics <slug>`, `app listing …`). It is effectively permanent once submitted. Getting it silently wrong at `create` time is expensive to discover and awkward to undo.

Civitai has a large international user base, so accented and non-Latin app names are not an edge case.

## Second, compounding problem: `--dir` never echoes the derived slug

```console
$ civitai app create "Slug Check" --dir ./slugcheck -y
✓ Created App "Slug Check" (page-money)  ·  ./slugcheck/  ·  35 files
```

No line of output contains the slug. With the default directory you can infer it from the directory name; with `--dir` it is invisible unless you open `block.manifest.json`. So the one case where derivation is most likely to surprise you is the case where it is least visible.

## The adjacent behaviour is already good

The cases that *do* error are handled well, which is what makes the silent path stand out:

```console
$ civitai app create "123 Numbers"
Error: derived slug "123-numbers" is invalid (must start with a letter, be lowercase, hyphen-separated)

$ civitai app create "!!!"
Error: cannot derive a valid slug from "!!!" (need ≥3 chars; use lowercase letters/numbers/hyphens)
```

Both name the rule and the constraint. The non-ASCII path just needs to reach the same standard.

## Suggested fix

Two independent changes; either alone is an improvement.

1. **Echo the derived `blockId` in `app create`'s success output**, always — not only when it is inferable from the directory. It is one line and it is the value the author will use in every subsequent command.
2. **Do not silently drop characters.** Options, in rough order of preference:
   - transliterate to ASCII (`Ü` → `u`, `Ω` → `o`/`omega`) and show what was derived;
   - or refuse, in the style of the two errors above, naming the offending characters and suggesting an explicit slug flag.

   Whichever is chosen, the author should see the result before 35 files are written. If a `--slug`-style override does not exist, that is the natural escape hatch.

## Test coverage this needs

- `"ÜberApp Ω"` must not produce `berapp` — pin the actual expected slug, whatever the chosen policy is.
- A pure-ASCII name must derive exactly as it does today (control).
- The two existing error cases (`"123 Numbers"`, `"!!!"`) must keep their messages and exit codes (controls).
- `app create --dir <somewhere>` output must contain the derived slug — assert the string, since this is the half that makes the bug invisible.

ZacxDev:
> **Revised after an independent audit** (see the maintainer's comment below).
> Two things this description used to claim are corrected here rather than
> quietly edited away: it does **not** close #259, and its headline test claim
> was resting on an assertion that could not execute. Both are fixed in
> `f94e3d1`; the audit trail is in this section and in AGENTS.md item 25.

**This NARROWS #259; it does not close it.** It also closes items 1 and 6 of
#260 — everything on the one screen where an app's **permanent public identity**
is minted.

## 🔴 BREAKING CHANGE, read first

`civitai app create "Café App"` used to silently produce the blockId `caf-app`.
It now **exits 2** with a message naming `é` and pointing at the new `--slug`
flag. Any script passing a non-ASCII name must pass `--slug <slug>`.

The old output was wrong — a different permanent public id than the author
typed, on an identifier that cannot be renamed — so this is the correct break,
but it *is* a break. Announced in `README.md` → **The blockId**, and recorded as
a `BREAKING CHANGE:` paragraph in the commit body (there is no CHANGELOG.md;
goreleaser builds release notes from conventional-commit subjects).

## What this closes, and what it knowingly does not

The refusal predicate is an **asymmetry, not a character list**:

| rune class | verdict | why |
|---|---|---|
| anything whose `unicode.ToLower` is `< 0x80` | never lossy | exempt **by construction** — this is what keeps every ASCII derivation byte-identical |
| non-ASCII `IsSpace \|\| IsPunct \|\| IsSymbol` (on the lowered rune) | separator | an em dash, `«»`, `©`, NBSP carry no content of their own; they fold to a hyphen exactly like a space |
| everything else non-ASCII (letters, digits, marks) | **lossy → refuse** | content the author typed, with nowhere lossless to put it |

Silent-character-loss classes, and their status **after** this PR:

| class | example | before | now |
|---|---|---|---|
| non-ASCII **letters/digits/marks** | `"Café App"` | `caf-app`, rc 0 | **CLOSED** — refuses, rc 2, names `é` |
| **invalid UTF-8** | `$'caf\xe9 app'` | `caf-app`, rc 0, **and a manifest that was not valid UTF-8** | **CLOSED** — refuses, rc 2 |
| runes that **lowercase into ASCII** | `"İstanbul App"` | `istanbul-app`, rc 0 | **OPEN, deliberately** — enumerated exception, exactly 2 runes exist (`İ` U+0130, `K` U+212A) |
| **symbols / emoji** | `"Rocket 🚀 App"` | `rocket-app`, rc 0 | **OPEN, deliberately** — product decision, tracked as #272 |

The last two are documented in `scaffold.Slugify`'s header and in the README
table, not implied away. The whole-of-Unicode census behind them: **8,580**
printable non-ASCII runes take the separator branch, **140,321** the refuse
branch; and `TestSlugifyLowersIntoAsciiIsADocumentedException` re-walks Unicode
so a future Go table growing a third such rune fails the build.

## Fixes made in this round (`f94e3d1`)

**1. The refusal named characters the user never typed.** `LossyRunes`
classified *and reported* off `strings.ToLower(name)`. Measured: `"ẞE App"`
reported `"ß"` — a rune **absent from the input**, because U+1E9E lowers to
U+00DF — and `"ＡＢＣ"` reported `"ａ","ｂ","ｃ"`. Someone searching their own name
for the quoted character finds nothing. Classification stays on the **lowered**
rune (that is what makes the ASCII exemption work); the **report** is now the
original. `LossyRunes` → `LossyChars` (`[]string`), since the reported unit is
no longer always one rune.

**2. The `mustNotProduce` block was UNREACHABLE, and this description's headline
claim rested on it.** It said each row "pins the exact pre-fix output it must not
produce (`berapp`, `caf-del-mar`) rather than merely 'an error came back'". The
block sat *after* a `t.Fatalf` that had already aborted the subtest whenever
`err == nil`, so its `e == nil` condition never held. Measured: deleting the
whole block left `internal/scaffold` green.

It now lives **inside** the `err == nil` branch, and each row's expected pre-fix
string is verified against `legacySlugify`, a copy of the pre-refusal derivation
— so a row cannot name an output the old code never emitted. Proven by mutation,
both directions:

```
# refusal disabled:
slug_lossy_test.go:145: Slugify("ÜberApp Ω") = "berapp" — that is the exact
  pre-fix mangling issue #259 is about, a different permanent public id …
slug_lossy_test.go:145: Slugify("Café Del Mar") = "caf-del-mar" — …

# refusal disabled AND the mustNotProduce block deleted:
slug_lossy_test.go:144: Slugify("ÜberApp Ω") = "berapp", want a refusal
```

The message degrades from the specific `#259` claim to the generic one, which is
what shows the block executes and carries the claim.

**3. REGRESSION this PR introduced: `--slug` silently skipped template
selection.** `--slug` was wired into the `stdinIsTTY` guard on the reasoning that
it "supplies the one thing the prompt exists to collect". `runScaffoldForm`
collects a name **and a template**, so `civitai app create --slug my-app` on a
TTY silently took `page-money` with no template choice — a question the user was
asked before the flag existed. The mutant deleting `slugFlag == ""` from that
guard **survived with 0 failures**; nothing covered the suppression at all.
`--slug` now drops only the **name field**; the template select still runs.

**4. Invalid UTF-8 (both halves).** `for _, r := range` yields U+FFFD per bad
byte, and U+FFFD is a Symbol — the *separator* branch. `Slugify` now gates on
`utf8.ValidString` before anything ranges over the string. `runAppScaffold`
separately refuses an invalid-UTF-8 **display name**, which is not redundant:
`--slug` bypasses derivation entirely, so the name reached
`block.manifest.json` unchecked. The two guards are independently pinned — M5
reddens only `internal/scaffold` rows, M6 only `internal/cmd` rows.
(`internal/validate` still has no UTF-8 check of its own; out of scope, stated.)

**5. NFD produced an unreadable message.** macOS paths and some paste routes
deliver NFD, so `"Café App"` arrives as `e` + U+0301 and the message read
`"́" cannot appear in a blockId` — an accent rendered over nothing — while the
NFC form of the same visible name said `"é"`. Combining marks are now reported
with their base, so both forms report `"é"`; a base-less mark is shown on a
dotted circle. **The set of refused names is unchanged** — a mark was always
lossy — only the rendering moved.

**6. Three mutation survivors closed.** The URL operand (swapping `slug` →
`display` survived because every assertion was a `strings.Contains` and the
first `%s` still carried the slug — the mutant printed
`https://Widget Machine.civit.ai/` as the "permanent public id"); the refusal
quoting the input name; and the ASCII-exemption boundary (`<` vs `<=`).

**7. The echoed URL promised something guaranteed false at that moment.** It
printed the bare `https://<blockId>.civit.ai/` as the app's "permanent public
id" at scaffold time — a URL that **404s until approval + deploy**, as
`README.md` says and as `app status` already tells people ("Not live yet — …
only serves after the app is approved and deployed"). That is the same
false-promise class this PR fixes for "validates clean" *two lines up in the
same output block*. Now:

```
✓ Created App "My Cool Block" (static)  ·  ./my-cool-block/  ·  7 files
  blockId: my-cool-block  —  your app's permanent public id (it cannot be renamed later)
  Will be served at https://my-cool-block.civit.ai/ — only after the app is approved and deployed · `civitai app status my-cool-block`
```

## Docs (`395cefc`)

- **README** — the `app create` row now carries `--slug`, plus a new
  **The blockId** section: what the id is, that it can never be renamed, the
  breaking change (announced the way the `--json` field-notation break was), and
  the table of what still derives rather than refuses.
- **AGENTS.md item 25** — the durable claim is the exemption: *derivation is
  safe to refuse only because it exempts every rune that **lowercases into
  ASCII**, which is what keeps every pre-existing derivation byte-identical — do
  not "improve" that into a character allowlist.* Plus the four residuals and
  the four process lessons.

🔴 **Numbering: this took item 25, not 26.** #265 is adding an item
concurrently and was expected to take 25 — but `parseAgentsItems`
(`agents_xrefs_test.go`) **enforces contiguity** (the *i*-th heading must be
numbered *i*), so skipping to 26 fails that guard unconditionally on a 24-item
file. AGENTS.md's own maintenance rule covers the collision: **whichever of
#265 / #267 merges second renumbers its own new items.**

## Testing

New: `internal/cmd/app_init_echo_test.go`; rewritten
`internal/scaffold/slug_lossy_test.go`; updated
`internal/cmd/app_init_identity_test.go`.

Controls that must stay green: 9 ASCII derivations through `Slugify` plus 7
through the real command, 4 non-ASCII separator rows (em dash / guillemets /
`©` / NBSP still derive `widget-pro`), both existing dead-end messages, and the
two documented exceptions. Exit codes are asserted with `errors.Is`, never
message text (AGENTS item 7). Every non-ASCII fixture is written with explicit
`\uXXXX` escapes and byte-asserted in the test itself — NFC and NFD render
identically, so the fixtures verify their own bytes.

**Mutation matrix — 15 mutants, re-measured on the merged tree** (`04c5c57`).
Each edit is checksum-gated (target must occur exactly once; file hash must
change) so an edit that silently failed to apply cannot read as a survivor, and
restore is via `git checkout --` so an interruption is diagnosable in one
command. Leaf `--- FAIL` lines are counted from the output, never an exit code.

| # | mutant | leaf subtests reddened |
|---|---|---|
| M1 | report the **lowered** rune, not the one typed | 5 |
| M2 | revert the refusal entirely (= pre-#259; produces `berapp`) | 15 |
| M3 | `--slug` suppresses the whole prompt again | 1 |
| M4 | the form runs but is told to ask for the name anyway (half-fix) | 1 |
| M5 | drop the `utf8.ValidString` guard in `Slugify` | 4 |
| M6 | drop the display-name UTF-8 guard (manifest half) | 3 |
| M7 | build the serving URL from the **display name** | 1 |
| M8 | the refusal quotes a fixed literal, not the input | 8 |
| M9 | widen the ASCII exemption boundary `<` → `<=` | 1 |
| M10 | report a combining mark bare, detached from its base | 1 |
| M11 | print the serving URL bare, as a currently-valid id | 1 |
| M12 | make symbols/emoji refuse (invert the #272 decision) | 2 |
| M13 | classify on the original rune (removes the `İ`/`K` exception) | 1 |
| M14 | drop the dedupe | 1 |
| M15 | **null mutant (comment only)** | **0 — SURVIVES, as required** |

**Live end-to-end against the built binary**, reproducing the original symptoms:
`"ẞE App"` → rc 2 naming `"ẞ"` (not `"ß"`); `$'caf\xe9 app'` → rc 2 (was
`caf-app` rc 0); NFD and NFC `"Café App"` → the **identical** message naming
`"é"`; `"İstanbul App"` → `istanbul-app` rc 0; `"Rocket 🚀 App"` → `rocket-app`
rc 0; `"My Cool Block"` → `my-cool-block` rc 0.

`make ci` on the merged tree: **18 packages `ok`, 0 `--- FAIL`, 0
`build failed`, 0 timeout panics**; `gofmt -s -l .` clean over 303 `.go` files.

## Note on history

`04c5c57` merges `origin/main` in to resolve a `README.md` conflict caused by
#268 (which added a TOC and two new command-table rows adjacent to the
`app create` row this PR edits). Control run: the audited tip `8ffafe9` merges
cleanly with current `main`, HEAD did not — so the conflict came from this PR's
own README edit, not from drift. Nothing was rebased or force-pushed; `8ffafe9`
is still in the history.

`cb0cd51` respells two control-character test fixtures as Go escapes (`\x7f`,
`\u0080`) after staticcheck ST1018 flagged the raw bytes. Worth recording as a
process note: **`make ci` is tidy + vet + test + build and does NOT run lint**,
so a clean local `make ci` says nothing about the `lint` job — on a box without
golangci-lint, `nix-shell -p golangci-lint --run "golangci-lint run"` gets the
same v2.12.2 CI uses. That local run was instrument-checked before its
"0 issues" was believed: reintroducing one raw U+007F reddened it with the
ST1018 message, removing it returned 0. The boundary mutant is still killed
after the respelling (re-measured).

One deviation from the brief worth flagging: it said not to touch the nav/TOC
sections. I added a **single** TOC sub-bullet for my own new `### The blockId`
section, because #268 has already merged (so there is no live sibling owning
that region) and every other `###` under Command reference is listed there —
leaving mine out would make the README's own navigation incomplete. Happy to
drop that one line if you'd rather.


ZacxDev:
**Correction to this PR's description, posted publicly rather than by editing the body** — a reviewer may already have read the original.

The description claims each refusal row "pins the exact pre-fix output it must not produce (`berapp`, `caf-del-mar`) rather than merely 'an error came back'." **That claim is false.** An independent audit found the `mustNotProduce` block in `slug_lossy_test.go:64-68` is **unreachable**: the `t.Fatalf` at line 58 aborts the subtest whenever `err == nil`, so the `e == nil` branch at line 65 never executes. Measured — deleting the entire block leaves `internal/scaffold` green, while a positive control (corrupting an ASCII control row) reddens 1, so the harness *can* go red; the assertion simply never runs. The rows really are "an error came back".

The PR is not unprotected — the over-refusal guarantee comes from separate controls (`TestSlugifyAsciiDerivationIsByteIdentical`, `…NonAsciiSeparatorsStaySeparators`, `TestAsciiDerivationIsUnchanged`), which do bite: inverting the predicate reddens 12, narrowing the ASCII exemption reddens 18. But it is not protected by the thing the description advertises.

**Second correction: this PR narrows issue #259, it does not close it.** Three classes of input still silently lose a character the author typed, each reproducing the exact `caf-` shape #259 is about — all byte-identical on `main`, so pre-existing rather than introduced here:

1. **Invalid UTF-8** — `app create $'caf\xe9 app'` → blockId `caf-app`, rc 0. The written `block.manifest.json` also contains the raw `0xE9`, is not valid UTF-8, and `validate.ManifestOnly` accepts it.
2. **Two runes ≥ 0x80 that lower to ASCII** — `İ` (U+0130) and `K` (U+212A), found by enumerating all of Unicode. `app create "İstanbul App"` → blockId `istanbul-app`, rc 0. The doc comment's absolute rule is false for exactly these two.
3. **Symbols and emoji fold to a hyphen** — `"Rocket 🚀 App"` → `rocket-app`, rc 0 (8,580 runes take the separator branch, 140,323 the refuse branch).

A fix round is in progress: (1) gets a `utf8.ValidString` guard, (2) becomes a documented enumerated exception, (3) is filed separately as a product decision. The doc comment and this description will be rewritten to state the residuals rather than claim closure.

Issue #270, opened 2026-08-07
App listing media: no published dimension or aspect requirements, and no drift-proof way to learn them


ZacxDev:
Found by a blind dogfood run of the `civitai` CLI (an agent given only the built binary, `README.md`
and `--help`, working as a first-time app author). Filed from `civitai/cli`; the answer lives on the
platform side.

## The gap

An App listing **cannot be published without an icon and a cover**, and no surface anywhere states
what shape those images should be.

Measured against the CLI at `v0.1.90-36`:

| surface | says |
|---|---|
| `civitai app listing set-icon --help` | "a square-ish image" |
| `README.md` | "square-ish, ≤2 MiB" (icon) / "landscape hero, ≤4 MiB" (cover) |
| the manifest schema | nothing |
| the CLI's validation | **format + byte caps only** |

What the CLI actually enforces (`internal/cmd/app_listing.go:26-28`) is a byte cap per kind —
icon 2 MiB, cover 4 MiB, screenshot 2 MiB — plus an accepted-format check. There is **no dimension
check anywhere** in the listing path, and `grep` for pixel dimensions across the CLI, the README and
the vendored schema returns nothing.

So an author sizing an icon has "square-ish" and a byte cap, and finds out the real requirement by
uploading and looking at the result.

## Questions

1. **Are there server-side minimum, maximum or exact pixel dimensions** for the listing icon, the
   cover, and a screenshot?
2. **Is the icon required to be square**, and is the cover a fixed aspect ratio (16:9? 3:1?
   something else)? "Square-ish" and "landscape hero" are the only guidance that exists today.
3. **Is a wrong aspect or size rejected, or silently accepted and rendered badly?** If there is a
   rejection path — at upload, at content scan, or at review time — the CLI would like to state it
   before the author spends a review cycle on it.
4. **Is any crop or rescale applied on upload?** If the platform normalises, the honest advice to an
   author is very different from "match this size exactly".
5. **Adjacent, and possibly a real trap:** the CLI notes the icon rides the inline **data-URI path
   with a server decoded cap of 2 MiB** (`app_listing.go:23`). A high-dimension but
   well-compressed PNG can sit far under the byte cap and exceed that cap once decoded. Is that
   reachable in practice, and what does the author see when it happens? The CLI cannot currently
   predict or explain that failure, because the only thing it can measure is the file size.

## What we would like the answer to look like

Not a number in a reply. **The CLI deliberately does not vendor platform constants that can move**,
[...]

PR #275, opened 2026-08-07
docs(#270): publish the real listing-media requirements, scaffold assets/, fix the broken quickstart

ZacxDev:
Closes the CLI-side half of #270: the documentation and the `assets/` scaffold. The attach-order reorder is a separate lane and is not in here.

## The issue's premise was half wrong, and this PR follows the corrected version

The issue says the platform's dimension requirements "are not published". They are — they are named constants in `civitai/civitai` `src/server/schema/blocks/app-listing.schema.ts`, enforced server-side at **attach** by `validateListingImage`, with rejection messages that name the bound and the measured value. Re-read for this PR and **byte-identical on `main` and `release`**, so these are what an author hits today:

| kind | aspect (w ÷ h) | minimum | maximum | byte cap | formats |
|---|---|---|---|---|---|
| icon | 0.9 – 1.1 | 128px shorter side | 4096px longer side | 1 MiB | png / jpeg / webp |
| cover | 1.3 – 2.4 | 640px wide | — | 4 MiB | png / jpeg / webp |
| screenshot | 0.4 – 2.6 | 320px shorter side | — | 2 MiB, ≤8 per listing | png / jpeg / webp |

**One correction to those numbers, because they cannot go in the docs as written.** The icon's 1 MiB (`MAX_LISTING_ICON_SIZE_BYTES`) and the CLI's 2 MiB (`maxIconBytes`) measure *different bytes*, and conflating them would have documented a cap that does not apply to the author's file. The icon rides the inline data-URI path: the server decodes it, downscales to ≤1024px and **re-encodes to PNG**, then stores `metadata.size = png.byteLength` (`listing-meta.service.ts`) — and *that* is what `validateListingImage` compares against 1 MiB. The CLI's 2 MiB is the source file, mirroring `INLINE_ICON_MAX_DECODED_BYTES`. Cover and screenshot take the full-res path where the CLI sends `sizeBytes: len(data)`, so there the two caps do describe the same bytes. The README documents the CLI's caps for the file you pass, and the aspect/dimension table separately as server-side. AGENTS.md item 25 records why.

## What changed

- **`assets/` is scaffolded by all three templates**, with a README of the requirements and **no placeholder images**. A placeholder passes format + byte validation and uploads cleanly — that is how a stub icon reaches a public listing, whereas a missing file fails loudly at the one step that can still fix it.
- **README**: new TOC-linked **Listing media requirements** section, splitting what the CLI enforces locally (format + byte cap) from what the platform enforces at attach (aspect + minimum dimension), plus the icon rescale, the "covers/screenshots are not rescaled" fact, and the ~16 MP source that fails with the unhelpful *"That icon couldn't be read"*. "square-ish" / "landscape hero" are gone from every doc surface. The quickstart's step 7 now says the artwork is yours to supply, and `./shot.png` became `./assets/screenshot-1.png`.
- **AGENTS.md item 25** — these bounds stay prose. Stale *guidance* costs one round-trip carrying the server's current bound; a stale local *gate* refuses valid images and cannot be argued with. Please don't promote the table into `internal/cmd` or `internal/validate`.
- Template READMEs get a short **Store-listing media** section pointing at `assets/README.md`.

## Verification

**Executed the quickstart rather than reading it**, with the binary built from this branch:

```
$ civitai app create my-app          # → 36 files, page-money
$ ls my-app/assets/                  # → README.md  (and no images — the point)
$ civitai app listing set-icon ./assets/icon.png
Error: no such file: ./assets/icon.png      ← exit 2, the loud failure we want
```

Then, after saving a real 512×512 icon as `assets/README.md` instructs, the same command clears every local check and reaches the network (run against an offline base URL so nothing touched prod):

```
Error: Get "http://127.0.0.1:1/api/v1/blocks/submissions?blockId=my-app": dial tcp: connection refused
```

On `origin/main` (1eb4095) that same command exited **2** with `no such file` on a fresh scaffold of *all three* templates — that is the bug, reproduced and then removed. Positive controls on the local gate, which is what the README now claims it does: a non-image gets `unrecognized image (want png, jpeg, or webp)`, and a 2.9 MB icon gets `larger than the 2.0 MB max for a icon`.

**Red/green matrix** for the new guards — red at `origin/main` (1eb4095) with the test files copied into a clean worktree of that ref, green at HEAD:

| | origin/main | HEAD |
|---|---|---|
| new + touched tests | 12 RUN / **0 PASS / 12 FAIL** | 12 RUN / 12 PASS / 0 FAIL |
| full suite | — | 2719 RUN / **2715 PASS / 0 FAIL** / 4 SKIP |

The 4 skips are pre-existing environment-gated guards (`CIVITAI_ANTIPATTERN_SCAN_DIR`, `CIVITAI_CHECK_SUBMISSIONS_CAP`, `CIVITAI_CHECK_PUBLISHED_PINS`, `CIVITAI_CHECK_SCAFFOLD_RUNTIME`). No timeout panics.

**Mutation-verified**, each mutant killed by *its own* message, not a neighbour's:

| mutation | failure |
|---|---|
| plant `assets/icon.png` in a template | `template "page-vite" scaffolds an IMAGE at assets/icon.png` |
| README icon cap 2 MiB → 3 MiB | `README says "…≤ 3 MiB…" for icon, but the CLI enforces 2 MiB` |
| rename `assets/README.md.tmpl` away | `has assets/ but no assets/README.md (entries: [NOTES.txt])` |
| drop "server-side" from the README section | `the Listing media requirements section never says "server-side"` |
| drop "no images, on purpose" from a template README | `does not explain that the directory ships with no images on purpose` |

The one direction I did **not** mutate is changing `maxIconBytes` itself, because that lives in `internal/cmd/app_listing.go`, which another lane is editing. The guard derives its expectation from the constant, so both directions diverge by construction, but only the README side is empirically demonstrated.

Other gates: `gofmt -s -l .` clean, `go vet ./...` clean, **golangci-lint v2.12.2** (the exact CI pin, run from nixpkgs) → **0 issues**, and the offline half of `scaffold-currency` reproduced locally (`app init --template page-money` → `TestScanDirFromEnv` → *no anti-patterns found*). The npm halves of `scaffold-currency` / `template-page-vite` / `template-page-money` are untouched by adding one markdown file.

## One thing I found and did not fix

> **Update (rebase onto `4d22bda`):** item 1 below is **DONE** — #273 merged, so the `--help` wording landed in this PR as `4f7875b`. See the rebase comment for the before/after, the help-body headroom numbers, and the re-measured red/green matrix at the new base. Item 2 remains open.

### ~~1. The deferred `--help` wording~~ — done in `4f7875b`

1. **`internal/cmd/app_listing.go` still says "a square-ish image"** in `set-icon --help` (and "a landscape hero image" in `set-cover`). Deliberately left alone — another lane holds that file for the attach-order reorder. **Follow-up owed once it merges:** replace both `Short` strings with what the CLI actually enforces plus a pointer to the README section, so `--help` stops being the last surface repeating the phrase this PR removed everywhere else.
2. **`humanBytes` labels MiB values as MB** — the CLI prints `2.0 MB` for `2*1024*1024`. The README's "2 MiB" is the correct label. Cosmetic, and the fix belongs in `download.go`, which is shared with the download UI, so it is not smuggled in here.

Also worth knowing: `assets/` is **not** excluded from `civitai app submit`, so whatever an author leaves there ships in the source bundle (measured: 37 files with an icon and cover present). Left that way on purpose — a `static` app may legitimately serve images from `assets/` — and stated in `assets/README.md` rather than silently changing packaging.

Refs #270

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -166,10 +166,12 @@ is model substitution, and 22 is the one gate on that path that guards CONTENT
 rather than money); items 18 and 20 cover the checks that tell an author their
 EXISTING app is missing the item-11 handshake (20 is the reachability repair to
 18's presence-only scan); item 23 covers the SHAPE of a validation finding — the
-`field` every `--json` consumer groups on; and item 24 covers the ONE
+`field` every `--json` consumer groups on; item 24 covers the ONE
 transport-vs-filesystem predicate now shared by the CLI-wide exit-code
 classifier (which every command's published exit code funnels through) and
-`pkg/civitai`'s read-GET retry loop.
+`pkg/civitai`'s read-GET retry loop; and item 25 is a third deliberate
+*non*-mirror — the store-listing image dimension/aspect bounds, which stay prose
+in the README rather than becoming a local check.
 The durable fix for the mirroring is a server-side
 `civitai app validate` endpoint that calls the real `BlockManifestValidator` —
 until that exists, vendoring is on purpose.
@@ -1864,6 +1866,42 @@ neither one's.
         dot-imported `net`) are listed in the file's own header; there is no
         full type resolution because `golang.org/x/tools/go/packages` would be a
         new dependency, which is an "ask first" below.
+25. **The listing-media DIMENSION and ASPECT bounds live in the README as prose,
+    and must NOT become a local check.** `civitai app listing set-icon` /
+    `set-cover` / `add-screenshot` validate the **format** and the **byte size**
+    of the source file (`maxIconBytes` / `maxCoverBytes` /
+    `maxScreenshotBytes`) and nothing else. The platform additionally enforces a
+    per-kind aspect range and a minimum dimension at ATTACH time
+    (`civitai/civitai → src/server/schema/blocks/app-listing.schema.ts`,
+    `validateListingImage`), returning a `BAD_REQUEST` that names the bound and
+    the measured value. Those numbers are documented in README →
+    *Listing media requirements* and in the scaffolded `assets/README.md`, and
+    that is deliberately as far as they go.
+    - **Why prose and not a check.** This is item 4's argument, applied to a
+      different constant set: stale *guidance* costs one round-trip carrying the
+      server's current bound, while a stale *gate* refuses valid images and the
+      author cannot override it. A local dimension check would also have to
+      re-derive the icon rescale below to avoid being wrong on day one. So do
+      not add a `LISTING_ICON_ASPECT_MIN` to `internal/cmd`, and do not "fix"
+      the docs by promoting the table into `internal/validate`. The CLI already
+      decodes width/height (`appapi.DecodeImageInfo`) — the omission is a
+      decision, not a missing feature.
+    - **The icon byte cap and the server's icon byte cap measure DIFFERENT
+      bytes, and the docs must not conflate them.** `maxIconBytes` (2 MiB) is
+      the SOURCE file, mirroring the server's `INLINE_ICON_MAX_DECODED_BYTES`
+      on the data-URI path the icon rides. The listing schema's
+      `MAX_LISTING_ICON_SIZE_BYTES` (1 MiB) is checked against
+      `Image.metadata.size`, which for that path is the byte length of the
+      **re-encoded** PNG the server produces after downscaling to ≤1024 px on
+      the longer side (`listing-meta.service.ts`) — not the file the author
+      passed. Cover and screenshot take the full-res path, where the CLI sends
+      `sizeBytes: len(data)`, so there the two caps DO describe the same bytes.
+    - **Never scaffold a placeholder icon or cover.** A placeholder passes every
+      format and byte check and uploads cleanly, so it can reach a public store
+      listing; a missing file fails loudly at the step that can still fix it.
+      `assets/` therefore ships with a README and no images, and
+      `internal/scaffold/assets_dir_test.go` fails if an image file appears
+      under it.
 
 **When you change a validation rule, keep all four vendored mirrors in sync with
 the server — `schema/`, the ported Go checks in `internal/validate/` (including
--- a/README.md
+++ b/README.md
@@ -1096,14 +1096,21 @@ Three behaviours that are not obvious from the numbers:
   most **1024 px** on its longer side and re-encoded to PNG — aspect preserved,
   and **never enlarged**. So an oversized icon is harmless, but an undersized one
   is not: the 128 px floor still bites, because nothing is ever scaled up.
+  One consequence to know about: the platform also caps the **re-encoded** icon
+  at 1 MiB, and that is a different measurement from the 2 MiB the CLI applies to
+  the file you pass. A detailed photographic icon can clear the local check and
+  still be refused for the size of the PNG the server made from it — the message
+  quotes bytes, not pixels. Flat, simple artwork re-encodes far smaller.
 - **Covers and screenshots are not rescaled.** What you upload is what the store
   renders, so ship them at the size you want shown.
-- **A wrong image is rejected, not quietly accepted.** You find out at
-  `set-icon` / `set-cover` / `add-screenshot` time — before a moderator ever sees
-  it — and the message names the requirement. A source image above ~16
-  megapixels (say 5000 × 5000) is the one exception with an unhelpful message:
-  it fails the icon decoder with *"That icon couldn't be read"* rather than a
-  dimension error. Downscale it and retry.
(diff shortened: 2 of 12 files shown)


ZacxDev:
**Rebased onto `4d22bda`** (post-#274, post-#273) and picked up the `--help` follow-up this PR had deferred. New head `4f7875b`.

### The conflict, and the semantic half of it

One hunk, as expected: the bullet list in the listing section. But taking either side verbatim would have shipped a false sentence. `main`'s version (from #274) reads:

> The CLI **does not publish or enforce** them (a copied number goes stale and starts refusing valid images).

That was true when written and this PR makes the first half false — publishing those bounds as prose is the whole point of #270. Merged text keeps A's attach-before-scan ordering description *and* my section link, with the clause corrected to the distinction that actually holds:

> The CLI does not **enforce** them locally — a copied number goes stale and starts refusing valid images — but it does *document* the current bounds, as guidance rather than as a gate: see [Listing media requirements](…).

A's "blocked image never goes live" bullet is kept untouched.

### The `--help` follow-up (was deferred to lane A's file)

| | before | after |
|---|---|---|
| `set-icon` Short | `Set the listing icon (a square-ish image)` | `Set the listing icon (png, jpeg or webp, at most 2.0 MB)` |
| `set-cover` Short | `Set the listing cover (a landscape hero image)` | `Set the listing cover (png, jpeg or webp, at most 4.0 MB)` |
| `set-icon` Long opening | `the small square-ish image shown beside your app's name` | `the small image shown beside your app's name` |

Both Shorts are now **built by interpolating `listingSourceRule(kind)`**, so they cannot drift from the cap constants. Each of the three media bodies gains **one** line — `See "Listing media requirements" in the README for the platform's bounds.` — a pointer rather than a third copy of the numbers, per AGENTS.md item 25.

**Headroom against #274's guards** (budget 1400 runes, 80 columns), measured after the change rather than inferred from the pass:

| body | runes | headroom | max col |
|---|---|---|---|
| set-icon | 873 | 527 | 79 |
| set-cover | 875 | 525 | 79 |
| add-screenshot | 1304 | **96** | 80 |
| (group) | 1305 | 95 | 80 | 

`add-screenshot` is the tightest thing I touched — 170 → 96 runes of headroom. Worth knowing before the next paragraph goes in. The group body is untouched.

### New guard: `TestListingMediaShortsQuoteTheirOwnCap`

#274's `TestListingHelpQuotesTheEnforcedCaps` reads `c.Long` only, so a `Short` can quote a cap the command does not enforce and stay green — and I just made the Shorts carry caps. Scoped to icon and cover because their caps *differ*, keeping "quotes mine" and "does not quote the sibling's" independent; `add-screenshot`'s cap string is byte-identical to the icon's, i.e. #274's declared equivalent mutant.

Mutation-measured: interpolating `kindIcon` into `set-cover`'s Short reddens it with `quotes 2.0 MB, the icon cap — listingSourceRule was interpolated with the wrong kind`, **while #274's own cap test stays green** — which is the evidence that it covers a surface theirs cannot see. Reverting a Short to the old `(a square-ish image)` reddens it too.

### Re-measured matrix at the NEW base

The earlier 12-red matrix was measured at `1eb4095` and was stale. Re-measured in a clean worktree of **`4d22bda`** with the test files copied in: **12 RUN / 0 PASS / 12 FAIL**, same failures for the same reasons (`scaffolds no assets/ directory`, `README.md has no "### Listing media requirements" heading`). **No test became vacuous** under #274/#273 — nothing to withdraw.

Full suite on the rebased tree: **2752 RUN / 2748 PASS / 0 FAIL / 4 SKIP** (the 4 are the pre-existing env-gated guards), no timeout panics. `gofmt`/`vet` clean, golangci-lint **v2.12.2** → **0 issues**. Quickstart re-executed against the rebased binary: `assets/README.md` present, no images, `set-icon ./assets/icon.png` still exits 2 with `no such file` before artwork exists and clears every local check once it does.

### One thing #273 changed that this PR had to follow

The requirements section said a wrong image is "rejected, not quietly accepted" without saying *when*. Now that attach runs before the scan poll, that verdict lands in ~2s rather than after a ≤120s scan, so the section says so — it should agree with A's new prose, not merely avoid contradicting it.

Also added, because it is reachable and confusing: the platform's **1 MiB icon cap applies to the server's re-encoded PNG**, not to the file you pass. Detailed photographic artwork can clear the CLI's 2 MiB local check and still be refused for the size of the PNG the server makes from it. Documented in both the README section and the scaffolded `assets/README.md`.

The `humanBytes` "MB vs MiB" mislabel noted in the original description is still open and still out of scope here — it lives in shared `download.go`.

PR #300, opened 2026-08-10
test(mutate): measure whether a generic Go mutation tester catches vacuous guards

ZacxDev:
**Experiment, not a feature. Do not merge yet — this is for the verdict discussion.**

## Question
Would an off-the-shelf Go mutation tester have caught the nine guards that could not fail, and is its signal-to-noise good enough to run regularly? Measured on `internal/validate` at `8ec3cb0`.

## Answer
**It catches 3 of the 9 shapes, and its own numbers cannot be quoted without a correction pass.**

Full write-up with every measurement: `claudedocs/mutation-testing-experiment.md`.

## Tool
gremlins v0.6.0. **Neither gremlins nor go-mutesting is in nixpkgs** (only a VSCode extension of the same name, and `mutmut` for Python), so the `nix-shell -p` route this repo uses for golangci-lint is unavailable. Installed instead via `go install pkg@version` into a cache dir — that runs outside the module, so **`go.mod`/`go.sum` are unchanged**. **No `.github/workflows/*` file touched.**

## Instrument controls — 2 pass, 1 fails

| control | result |
|---|---|
| **Positive** | **PASS** — a covered mutation produces a real `--- FAIL` (0 build failures), scored KILLED. |
| **Negative** | **PASS at single-mutant resolution.** Deleting the one test that kills `semantic.go:87:16` flipped exactly that mutant KILLED→LIVED and nothing else. (A first attempt — deleting `warnings_test.go` — moved *nothing*: that file is fully redundant for mutation purposes.) |
| **Non-compiling mutants** | **FAIL.** `getTestFailedStatus` maps `go test` exit 1 → KILLED, and a build failure *is* exit 1. **36 of 146 reported "kills" (24.7%) do not compile**, while gremlins reports `Not viable: 0`. |

Also: at stock settings the first run reported **162 TIMED OUT of 236** against a 0.17s suite — an artifact that looks like a result.

## Numbers

| | gremlins | **corrected** |
|---|---:|---:|
| Killed | 146 | **110** |
| Lived | 23 | 23 |
| Not viable | 0 | **36** |
| Not covered | 67 | 67 |
| Efficacy | 86.4% | **82.7%** |

**Wall clock: 18.5s** end-to-end. Integration mode (`-i`) is **14m58s for zero status deltas**.

## Survivor triage — 17 real / 4 equivalent / 2 masked-pair

Real findings include three cap boundaries that **mirror server constants** (`buildCommandMaxLength`, `heightMinFloor`/`heightMaxCeiling`), the two `readyack` scan caps, and a `dedupeFindings` `continue`→`break` that **truncates a finding list to its first entry**.

Each "equivalent" verdict is discharged with an input that *would* have discriminated it. Two survivors (`readyack.go:408`/`430`) are individually inert because they mask each other, but the 2nd-order pair flips `unobservable`→`found` and nothing catches it — a gap first-order mutation testing structurally cannot report.

## Which of the nine shapes

✅ cap-test-relative-to-its-own-constant · size-cap guard that skipped itself · ranking-stability under the n=13 threshold — **all three demonstrated live, unprompted**

⚠️ strong-tier unreachable precondition

❌ `t.Skip` row · mismatched-path `Contains` · `s == s` positive control · denylist-of-one premise · bystander-carried sibling row

The five misses are all **test-code** defects; mutation testing only mutates production code. Shape (e) is unreachable by any mutation tool by construction — the mutant *is* killed, just by the wrong assertion.

## Recommendation

**Do not propose a nightly job.** It answers the wrong question (5/9 invisible), survivors need ~an hour of triage and 6/23 recur every run, and the numbers aren't stable enough to threshold on.

**Do** land the local `make mutate` (18s, on demand), and **propose separately**: a boundary-flip battery (the only mutator class that paid off — 48% of survivors, all 5 high-value findings — and operator flips on numeric comparisons always compile, so the non-viable problem disappears) plus two AST checks over `_test.go` (`t.Skip*` inside an error branch; tautological self-comparison).

## Gates
`make ci` green (18 pkgs ok, 0 `--- FAIL`, 0 `build failed`, 0 timeouts) · `gofmt -s -l .` clean · golangci-lint v2.12.2 → 0 issues.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ COMMIT  ?= $(shell git rev-parse --short HEAD 2>/dev/null || echo none)
 DATE    ?= $(shell date -u +%Y-%m-%dT%H:%M:%SZ)
 LDFLAGS := -s -w -X main.version=$(VERSION) -X main.commit=$(COMMIT) -X main.date=$(DATE)
 
-.PHONY: all build install test vet lint fmt clean tidy ci
+.PHONY: all build install test vet lint fmt clean tidy ci mutate
 
 all: build
 
@@ -40,6 +40,17 @@ tidy:
 
 # Mirror CI locally.
 ci: tidy vet test build
+
+# Mutation-test ONE package and print a report corrected for gremlins'
+# non-compiling-mutant misclassification. Local investigation tool only:
+# deliberately NOT part of `ci`, deliberately not a CI job, and it exits 0
+# whatever it finds. `make mutate PKG=internal/blockproto` to point it
+# elsewhere. Read the "what this does not cover" header in scripts/mutate.sh
+# before believing a clean run — in particular, it cannot see a defect that
+# lives in TEST code, which is most of what it would need to see here.
+# Measurements: claudedocs/mutation-testing-experiment.md
+mutate:
+	./scripts/mutate.sh $(PKG)
 
 clean:
 	rm -rf bin dist
--- a/claudedocs/mutation-testing-experiment.md
+++ b/claudedocs/mutation-testing-experiment.md
@@ -1,0 +1,346 @@
+# Experiment: would an off-the-shelf Go mutation tester have caught the nine vacuous guards?
+
+**Date:** 2026-08-09 · **Base:** `origin/main` @ `8ec3cb0` · **Package:** `internal/validate`
+· **Go:** 1.25.12 · **Host:** 24 cores
+
+**Short answer: it would have caught 3 of the 9, and it needs a correction pass
+before any of its own numbers can be quoted.** The tool is worth keeping as a
+*local, on-demand* instrument — 18 seconds, 74% of its survivors are genuine —
+but it is the wrong shape for the problem that motivated the experiment, because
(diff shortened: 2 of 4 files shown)


ZacxDev:
Re-derived one of the four §4b equivalence claims by the same route used for #301/#302 (clean `git archive` of `main`, checksum-gated, pattern asserted unique) — the one your own report named as most likely to become load-bearing.

**`readyack.go` `i > 0` → `i >= 0` is NOT equivalent.** It survives all 18 packages, and it changes user-facing output:

```
original:  "… usually the actual bug: (1) alpha."
mutant:    "… usually the actual bug: ; (1) alpha."
```

A stray separator before the first gap. Filed as #303 with the reproduction and the one-line assertion that would kill it.

This does not change the PR's conclusions — the "3 of 9 shapes" finding and the recommendation against a nightly job rest on the shape analysis, not on the equivalence triage. But it does mean **the §4b equivalence verdicts should be read as unverified**, which is what the report already says about its own method ("spot-checked, not verified"). Worth a line in the report itself so a later reader does not take the other three on trust; the same re-derivation is a few minutes each.

Issue #280, opened 2026-08-10
workflows get: a queued workflow is reported as finished, charged and unrefunded — then "has not finished" one line later


ZacxDev:
## Summary

`civitai workflows get` on a **queued** workflow prints that the job finished without producing a usable file and that you were charged and will not be refunded — then, one line later, that the workflow has not finished. Four mutually contradictory lines, and the two alarming ones are the wrong ones.

The practical harm: a user reads "finished, no output, not refunded" about a job that is running perfectly well, and submits a duplicate paid generation.

## Evidence

```
$ civitai workflows get <workflow-id>
Workflow ID:  <workflow-id>
Status:       scheduled

Outputs (0 deliverable, 1 excluded)
⚠ 1 output(s) will NOT be saved — they are not deliverable results:
    - <output-id>-0.jpg: not available (the job finished without producing a usable file)
These were still part of the workflow you were charged for — a blocked or missing output is
not refunded.
This workflow has not finished. Re-run this command to check again — it is a read and spends
nothing.
EXIT=0
```

`Status: scheduled`, *"the job finished"*, *"not refunded"*, and *"This workflow has not finished"* — all in one output. That workflow subsequently went `scheduled → processing` and ran for another 14 minutes.

## Root cause — one line of ordering

`internal/cmd/workflows.go`:

```go
180:  reportExcludedOutputs(errw, excluded)          // fires unconditionally
181:
182:  if !genapi.IsTerminalStatus(wf.Status) {
183:      fmt.Fprintln(errw, ui.For(errw).Dim(
184:          "This workflow has not finished. Re-run this command to check again — …"))
```

`reportExcludedOutputs` (`internal/cmd/generate_output.go:76-88`) and `ExclusionReason` (`internal/genapi/status.go:256-266`) are both written for a **terminal** workflow — `!o.Available` renders as *"not available (the job finished without producing a usable file)"*, which is the wrong sentence for a pending output.

Non-terminal state clearly *was* considered: `workflows.go:155-160` has a dedicated `"(none yet — the workflow has not finished)"` branch. It only covers the case where there are **zero** outputs. A queued workflow that has already materialised an output *placeholder* skips that branch entirely.

## Why the existing test cannot see it

`internal/cmd/workflows_test.go:138-150` (`TestWorkflowsGet_UnfinishedWorkflowSaysSo`):

```go
payload := `{"id":"wf_9","status":"processing","createdAt":"…","steps":[]}`
```

[...]

PR #304, opened 2026-08-10
fix(#278,#279): the CLI stops answering two questions it cannot observe

### LINKED ISSUE #278: generate: "it was charged and is not refunded" is false for failed workflows — and a test pins the wording
## Summary

`civitai generate` tells the user that a failed generation was charged and will not be refunded. In every case observed, the Buzz came back. A test pins the false wording, so this will not self-correct.

Found during a deliberately blind usability run of the CLI (42 image generations across three sessions), then independently re-derived from the raw transcript by a second reviewer before being reported.

## The claim, verbatim

`internal/cmd/generate.go:1135`

```
Error: the generation finished with status "failed" — it was charged and produced no usable
result; inspect it with `civitai workflows get <id>`
```

`internal/cmd/generate_output.go:87`

```
These were still part of the workflow you were charged for — a blocked or missing output is
not refunded.
```

And the source comment asserting the underlying belief, `generate.go:1133-1134`:

```go
// failed / expired / canceled. Say plainly that it was still charged —
// the orchestrator does not refund a job that ran and failed.
```

## Evidence

29 submits each printed `Charged 8 Buzz`. 21 succeeded, 8 finished `status: failed`. Balance was read twice, ~25 minutes apart. Reporting deltas rather than absolute balances:

| | submits charged | successes | balance delta | delta / 8 | nominal if failures stuck |
|---|---|---|---|---|---|
| checkpoint 1 | 19 | 11 | **88** | **11** | 19 × 8 = 152 |
| final | 29 | 21 | **168** | **21** | 29 × 8 = 232 |

Both reconcile exactly against the **success** count, at two independent points, with every movement confined to a single Buzz bucket. **64 Buzz the CLI said was spent and would not return, returned.**

## Why it will not self-correct

`internal/cmd/generate_wait_test.go:445-462` (`TestGenerate_NonSucceededTerminalStatusFails`) *requires* the wording:

```go
if !strings.Contains(err.Error(), "charged") {
    t.Errorf("%s error must say it was charged: %v", st, err)
}
```

The copy is not an oversight a test would catch — the test enforces it, for all three of `failed`/`expired`/`canceled`.

## Scope — stated deliberately, because a wrong accusation about billing is worse than no finding

- I can establish **net cost was zero**. I cannot distinguish "never debited" from "debited then auto-refunded inside the polling window". Both falsify the sentence as written.
- `generate.go:1135` on `status: failed` → **falsified, 8 of 8 observed cases**.
- `generate_output.go:87` fires in a second case never exercised here: a `succeeded` workflow carrying a moderation-**blocked** output. That one is **untested** and may well be true.

## Suggested fix

Stop asserting a refund policy the CLI cannot observe. Something like:

> the generation finished with status %q and produced no usable result. It may or may not have been refunded — confirm with `civitai buzz`.

Then change the test to assert the hedged form. If the orchestrator's refund policy is knowable programmatically, read it rather than restating a belief in a string.

## Ordering note

There is a related defect where `--dry-run` reports `Generatable: true` for checkpoints that then charge and fail. Its monetary harm is currently **zero only because this copy is false** — the two cancel in practice. Fixing the refund copy alone, without also fixing the dry-run false green, would make that second defect start costing users real Buzz. Worth sequencing together.

### LINKED ISSUE #279: generate --dry-run: "Generatable: true" green-lights checkpoints that then charge and fail (8 submits, 0 outputs)
## Summary

`--dry-run` prints `Generatable: true` for checkpoints that then charge and fail with no output. The documented spend-safety preflight returns a confident green on jobs that cannot run.

Found during a blind usability run of the CLI, then re-derived from the raw transcript by a second reviewer.

## Evidence

Six SDXL-family checkpoints were probed with `--dry-run`. The preflight split them cleanly into two groups — and the group it **passed** is the group that cost money.

**Rejected free by `--dry-run`** (exit 2, no charge):

```
$ civitai generate "a red apple" --checkpoint 126688 --aspect-ratio 1:1 --dry-run
Error: the server rejected the generation request (400): Some of your resources are not
available for generation: alpha2 (xl1.0)
EXIT=2

126601   Error: ... not available for generation: v1.0
782002   Error: ... not available for generation: Jugg_XI_by_RunDiffusion
```

**Passed by `--dry-run`, then charged and failed:**

```
$ civitai generate "test" --checkpoint 128078 --aspect-ratio 1:1 --dry-run
Checkpoint:    SD XL — v1.0 VAE fix (Checkpoint, id 128078)
Generatable:   true
  total       8

... later, the real submit:
✓ Generation submitted — workflow <id>
Charged 8 Buzz
  waiting… status processing after 2m42s (next check in 1m0s)
  status failed
Error: the generation finished with status "failed" — it was charged and produced no usable result
```

Same for `290640` (Pony Diffusion V6 XL) and `1759168` (Juggernaut XL Ragnarok): `Generatable: true`, ~3 minutes of `processing`, then `failed`.

**Total: 8 submissions across 3 distinct `Generatable: true` checkpoints, 0 outputs.**

Controls were run, so the discriminator is the checkpoint and not the prompt:

- trivial prompt (`"a red apple on a table"`) + checkpoint `128078` → **failed**
- full prompt + working checkpoint `2442439` → **succeeded**

## Root cause — a label problem, not a logic problem

`Generatable` is a straight echo of the server's `ready` boolean:

- `internal/cmd/generate.go:1351` — `fmt.Fprintf(tw, "Generatable:\t%t\n", q.Ready)`
- field declared at `internal/genapi/generate.go:65-67`

`ready` means *"the resources resolve"*. The CLI renders it as **"Generatable"**, which a user reads as *"this will produce an image"*. Nothing in `generate --help` caveats it.

**Test coverage:** `internal/cmd/generate_test.go:693-698` covers the `ready:false` refusal path. There is no test and no help text for the `ready:true`-then-fails case, because it is not locally detectable.

## Suggested fix

Rename the field to what it is — `Resources ready: true` — and add one line to `generate --help`:

> `ready` means the resources resolved, not that the job will succeed. A job can pass the estimate and still fail at execution.

## Sequencing note

This defect's monetary harm is currently **zero only because of a second, independent defect**: the CLI's claim that failed generations are not refunded is false (filed separately — the Buzz comes back). The two cancel in practice. **Fixing the refund copy alone would make this one start costing users real Buzz**, so they are worth sequencing together.

## Context, not a CLI defect

SDXL-family generation appeared to be failing platform-side during this window — Z-Image Turbo went 21/21 in the same sessions. That is the *cause* of the failures, but not of the false green: the preflight correctly rejected three of the six checkpoints for free, so the information needed to reject the other three plausibly existed somewhere.

ZacxDev:
Fixes #278. Fixes #279.

They land together on purpose: **#279's monetary harm is currently zero only because #278's copy is false.** Fixing the refund copy alone turns a `ready: true` green light into real Buzz.

---

### What the server source actually says

The maintainer's instruction was to read the server, not the balance arithmetic. Done, against `~/workspace/civit/civitai`:

- **The monorepo neither charges nor refunds a generation.** `TransactionType.Generation` is defined in `src/shared/constants/buzz.constants.ts` and **zero** call sites create a transaction with it. `orchestration-new.service.ts` submits the workflow, hands the orchestrator a list of spendable currencies, and then *reads back* `workflow.transactions.list`. The refund logic is in the **orchestrator service, which is not in this checkout.**
- **`orchestratorRefundableStatuses = ['failed','expired','canceled']`** exists (`src/shared/constants/generation.constants.ts`) but drives **no** money path — it is only folded into `orchestratorCompletedStatuses` for UI derived state.
- **No webhook or job reacts to a terminal workflow status.** Generation callbacks go to the signals websocket service; `src/pages/api/webhooks/` has no generation webhook. (Contrast: *training* is refunded in-repo at `src/pages/api/webhooks/resource-training.ts` — that is the shape a monorepo-side generation refund would take, and it does not exist.)
- **The strongest in-repo statement is first-party client copy, and it says REFUNDED.** `src/components/ImageGeneration/QueueItem.tsx`: *"Terminal failure states … The orchestrator auto-refunds spent buzz on these"* over `failed || expired || canceled`, rendering **"Your Buzz has been refunded."**
- **Cancel is described as PARTIAL** — *"refunded for any undelivered images"* — while `blocks.router.ts` states for the App-Blocks customComfy path that a mid-run cancel bills the **accrued** cost non-refundably. Both can be true at once.
- **A succeeded job's blocked output is not automatically refunded, but is not lost either.** The mature-content case is `canUpgrade`: *"your Buzz is held, not lost"* (`poll-iteration.ts`), with non-yellow debits credited back on unlock. A **hard** block is explicitly not refunded (`SourceImageUpload.tsx`).

### The verdict this PR ships

The old claim is **confirmed-false**. The opposite claim is **unevidenced** — full-vs-pro-rated and the per-status rule live in a service not checked out here. So, per the brief: **neutral wording.** The touched sites state the outcome and point at `civitai buzz`. No site now asserts a direction.

### Full ledger sweep — `grep -n "charged\|refund" -r internal/`, classified

| Site | Claim | Verdict | Action |
|---|---|---|---|
| `generate.go:1181` terminal failed/expired/canceled | "it was **charged** and produced no usable result" | **CONFIRMED-FALSE** — contradicted by the platform's own client + two balance measurements | **fixed** → states outcome, points at `civitai buzz` |
| `generate_output.go:87` excluded outputs | "a blocked or missing output **is not refunded**" | **CONFIRMED-FALSE for the commonest case** — mature-content is `canUpgrade`, "held, not lost"; true only of hard blocks, and the payload cannot tell them apart | **fixed** → charge half kept, refund half neutralised |
| `generate.go:1351` spend confirmation | "It cannot be undone and **is not refunded**" | **CONFIRMED-FALSE as a universal** — it covers the failed outcome the orchestrator does refund | **fixed** → "SPENDS REAL BUZZ and cannot be undone once submitted" |
| `generate.go:173` `--help` header | "there is **no cancel-for-refund**" | **CONFIRMED-FALSE** — first-party copy: cancelling refunds undelivered units | **fixed** → neutral, points at `civitai buzz` |
| `generate.go:219` `--timeout` block | "there is no cancel-for-refund, and a mid-run cancel bills the accrued cost anyway" | first clause **CONFIRMED-FALSE**; second clause **supported** (`blocks.router.ts`) | **fixed** → keeps the accrued-cost half only |
| `generate.go:1199` succeeded, 0 deliverables | "it was **charged**" | **TRUE** — the workflow ran; makes no refund claim | left, with a test comment saying why (over-fixing would be a false negative) |
| `generate.go:194` / `:437` / `:473` / `:1416` | realized charge over the estimate "is not refunded" / "with no refund" | **UNVERIFIED** — about a job that *runs*; no source read either way | **left**, deliberately. Flagged below. |
| `generate.go:1165` / `:1169` / `:1267` timeout / Ctrl-C / re-attach | "has already been charged" | **TRUE** — no refund claim | left |
| `genapi/generate.go:64`, `genapi/workflows.go:19,171` (comments) | "no refund" / "non-refundably" | **UNVERIFIED** (comments, not user copy) | left |
| `validate/warnings.go:64` | "you are charged the real price" | **TRUE**, unrelated surface | left |
| **`workflows.go:41`, `workflows_list.go:52`, `workflows_cancel.go:43,111,119,166`** | "does **NOT refund**", "no cancel-for-refund" | **CONFIRMED-FALSE in the same way** as `generate.go:173` | **NOT touched — see below** |

🔴 **Two honest gaps, stated rather than quietly left:**

1. **`workflows cancel` still carries the retracted cancel-for-refund copy.** Six sites across `workflows.go` / `workflows_cancel.go` / `workflows_list.go`. They were out of this PR's file ownership (#280 was in flight in `workflows.go` concurrently), and the cancel rule is genuinely *conditional* — accrued cost billed, undelivered units returned — so it needs its own wording pass rather than a search-and-replace. **A follow-up issue should move all six together**; a half-fixed copy set is worse than a uniformly wrong one, and right now it is half-fixed across two files.
2. **The "realized charge over the estimate is not refunded" claims are unverified, not confirmed.** No source was read either way. They are about a job that ran, so they are plausible; they are not evidenced.

---

## #279 — `Generatable: true` promised a predicate the flag cannot carry

`ready` is computed server-side (`orchestration-new.service.ts`, the whatIf reply builder) as:

```ts
for (const job of workflowStep.jobs ?? []) {
  const { queuePosition } = job;
  if (!queuePosition) continue;           // <- a job with no queue position is SKIPPED
  if (queuePosition.support !== 'available' && ready) ready = false;
}
```

That is **resource/capacity availability**, not a prediction that the job produces output — and a job with no queue position leaves the flag `true`. Dogfooding: 8 submits across 3 checkpoints that all quoted `ready: true` produced **0** outputs.

- Label: `Generatable` → **`Resources ready`**.
- `--help` gains a caveat naming exactly what the flag does and does not report.
- **`README.md` "Generation `--json`" fixed.** It told scripts `[ .ready = "true" ] || exit 1`, i.e. it presented the TRUE direction as a green light — the exact converse the dogfooding disproved. It now gates on `ready` being **false** (decisive: the server has said it cannot serve the job) and spells out that `true` is not a success predicate. Per `CLAUDE.md`, README and behaviour move in lockstep.
- The rename is human-renderer only, and that is **pinned**, not assumed: `--dry-run --json` still emits the server's raw `ready` key, asserted on a decoded map.

---

## AGENTS.md item 28 — and a correction to `agents_size_test.go`

New **item 28** covers both defects as one rule: *the CLI must not make claims about a spend it cannot observe.* Same family as items 8, 13 and 19(b). It also records that a moderation refusal is **structurally** undetectable at `--dry-run`, because `whatIfGraph` strips the prompt before the estimate by design (item 15).

**Byte ceiling.** `agentsMaxBytes` is 67,500 against 67,213 achieved — 287 bytes. An eviction was mandatory. **AGENTS.md is now 67,305 bytes, 195 under the ceiling** (measured, not "the test passed").

🔴 **The eviction candidate the size test ranks first does not work, and the reason is in that test's own commit.** `agents_size_test.go` says items 10 (7,795 B) and 19 (7,801 B) *"are byte-identical to agentsSplitBase, so their digests can be taken straight from `git show <base>:AGENTS.md`."* Measured at `5cb45f0` — **the commit that wrote that sentence** — the same compression pass had reflowed both: item 10 is now 7,769 B, item 19 is 7,724 B. Evicting either requires a **per-row base commit** in `splitItems`, because `TestSplitDigestsAreTheBaseCommitsText` re-derives every digest from the single `agentsSplitBase`. So **item 13** was evicted instead — the largest candidate that is still genuinely base-identical (3,001 B, 42 non-blank lines, digest taken from `git show <base>:AGENTS.md`). Its stub keeps the thesis; `claudedocs/decisions/13-cli-validates-nothing.md` keeps the body verbatim.

🔴 **Item 28 is INLINE, not a stub+evidence file, and that is forced by the guards.** A *new* item cannot have an evidence file today: `TestSplitTableCoversEveryEvidenceFile` requires every file in `claudedocs/decisions/` to have a `splitItems` row, and `TestSplitDigestsAreTheBaseCommitsText` then calls `baseItemBody`, which `t.Fatalf`s with `item 28 not found in the base AGENTS.md`. **Measured, not reasoned**: a probe file + row produced exactly `--- FAIL: TestSplitDigestsAreTheBaseCommitsText/item_28`. Both of these are worth a follow-up on the guard set.

---

## Test matrix — red at base `5cb45f0`, green at HEAD

Every new assertion was run against `origin/main`'s source in a scratch worktree with only the test files copied in.

| Test | at base `5cb45f0` | at HEAD | kind |
|---|---|---|---|
| `TestGenerate_NonSucceededTerminalStatusDoesNotSettleTheRefund` / `failed` | 🔴 3 failures | ✅ | regression |
| … `/expired` | 🔴 3 failures | ✅ | regression |
| … `/canceled` | 🔴 3 failures | ✅ | regression |
| `TestReportExcludedOutputs_DoesNotSettleTheRefund` | 🔴 2 failures | ✅ | regression |
| `TestDryRun_ReadinessLabelIsNotAGeneratabilityClaim` | 🔴 3 failures | ✅ | regression |
| `TestDryRun_NotReadyStillWarns` | 🔴 1 failure | ✅ | regression (label) + **invariant guard** (the warn half) |
| `TestGenerateHelp_SaysReadinessIsNotAPromiseOfOutput` | 🔴 2 failures | ✅ | regression |
| `TestDryRunJSON_RawReadyKeySurvivesTheRename` | 🟢 **passes at base** | ✅ | **invariant guard** — labelled as such, not counted as regression coverage |
| `TestRefundDirectionalClaims_CanStillFire` | 🟢 passes at base | ✅ | **positive control** over the claim ledger |
| `TestGenerate_AllOutputsFilteredIsAnError` | 🟢 | ✅ | pre-existing, unchanged behaviour |

### Mutation controls — each guard broken on purpose, each failing with **its own** message

| Mutation | Result |
|---|---|
| `--dry-run --json` emits a re-derived `{"resourcesReady": …}` document instead of the raw payload | 🔴 `the raw 'ready' key is gone from --json` (`generate_dryrun_label_test.go:114`) |
| Delete the `if !q.Ready { warn }` block | 🔴 `a ready:false quote must still warn — that direction is decisive` (`:89`) |
| Tag the terminal-status error `civitai.Tag(civitai.ErrNotFound, …)` | 🔴 `the terminal-status error is now classified ErrNotFound, which changes its published exit code` (`generate_wait_test.go:530`), on all three statuses |
| Restore the old refund copy (= the base run above) | 🔴 the two refund guards, with their own messages |

**Assertions are paired, not spelled.** A `Contains` on the new label alone passes while the old one is still printed, so every positive assertion has its negative twin: `Resources ready` present **and** `generatable:` absent (case-insensitively); the charge half present **and** the refund claim absent. `refundDirectionalClaims` is a **claim ledger** banning both directions — `not refunded` *and* `has been refunded` — because the equal-and-opposite copy is the same defect pointing the other way, and it has its own positive control proving it can match the sentence it was written to catch.

**Exit-code pin (AGENTS.md item 7).** #278 is a message-only change and every other assertion here is on message text, which says nothing about what a script sees. `errors.Is` now pins that the terminal-status error carries **no** classification sentinel (`ErrBadRequest` / `ErrUnauthorized` / `ErrNotFound` / `ErrRateLimited` / `ErrNetwork` / `cmd.ErrUsage`), so it stays on the generic exit code. Mutation-verified above.

## Gate

- `make ci` — green. `go test ./... -count=1 -v` **counted**: `RUN=3149 PASS=3145 FAIL=0 SKIP=4`, `build failed=0`, `panic: test timed out=0`, package-level `FAIL=0`.
- `make lint` — golangci-lint is not on this host's PATH, so it was run at the **CI-pinned v2.12.2** via `nix-shell -p golangci-lint`: **0 issues**. Negative control: an unused function was injected and it reported `func lintProbeUnused is unused (unused)`, so the zero is a measurement.
- `gofmt -s -l .` — 0 files.


ZacxDev:
## Adversarial audit — three claims in the PR description do not survive measurement

Posting this as a comment rather than editing the body, because a reviewer may already have read the original. Everything below was re-measured on `7bd4821` (`go test ./... -count=1` 18 ok / 0 FAIL, `gofmt -s -l .` 0, `go vet` clean, `golangci-lint 2.12.2` 0 issues — the gate claims in the description all reproduce). These three do not.

### 1. "No site now asserts a direction" is contradicted by the PR's own table

The sentence closing the #278 section reads *"The touched sites state the outcome and point at `civitai buzz`. **No site now asserts a direction.**"* The ledger four rows above it lists `generate.go:194 / :437 / :473 / :1416` as still saying the realized charge is *"not refunded" / "with no refund"*, and marks them **left, deliberately**. Both cannot be true.

This is not only a body-vs-table inconsistency, it ships as a self-contradiction inside one `--help` screen:

- `generate.go:172-176` (new): *"What happens to that charge if the run then fails or is cancelled is decided server-side … it will not tell you either way"*
- `generate.go:197` (unchanged, ~20 lines later): *"The realized charge can exceed the estimate, **and it is not refunded**."*
- `generate.go:483` (unchanged, `--timeout` usage string): *"the job continues server-side and **is not refunded**"*

Before this PR those paragraphs agreed with each other. They no longer do.

### 2. `refundDirectionalClaims` is a spelling ban, not a claim ledger — measured

The comment above it says *"This is a claim ledger, not a spelling ban"* and the description says it bans **both directions** and *"has its own positive control proving it can match the sentence it was written to catch."* Three mutants, each applied to a pristine copy of this branch and run with `go test ./internal/cmd -count=1 -v`:

| mutant | result |
|---|---|
| terminal-status message → *"…produced no usable result — **your Buzz returns to your balance automatically**, so check `civitai buzz`…"* | **survives**, 10/10 targeted tests green |
| `reportExcludedOutputs` → *"…charged for. **That Buzz comes back to your balance automatically** — check `civitai buzz`."* | **survives**, 10/10 green |
| delete `"was refunded"`, `"has been refunded"`, `"will be refunded"` from the ledger | **survives**, whole `internal/cmd` package green |

The reason the third survives is that `TestRefundDirectionalClaims_CanStillFire`'s mirror assertion never reads the ledger:

```go
if !strings.Contains(strings.ToLower("Your Buzz has been refunded."), "has been refunded") {
```

That is a tautology over two string literals — true whatever `refundDirectionalClaims` contains. Reducing the ledger to the single entry `"not refunded"` also leaves the package green. So the mirror half of the ban is unenforced *and* its stated positive control does not cover it.

Same shape in `TestGenerateHelp_SaysReadinessIsNotAPromiseOfOutput`: rewriting the caveat to *"…is **not a promise of output** speed, but it IS a reliable green light: if it says true the job will produce an image"* keeps both required substrings and **survives**.

(The deletion/restoration mutants in the description's matrix all reproduce exactly as stated, including the `ErrNotFound` exit-code pin — that half is solid. The gap is specifically the equal-and-opposite direction the ledger was introduced to cover.)

### 3. The AGENTS.md eviction rationale was overtaken by #305 (merged as `bd3595f`)

The 15-line 🔴 comment added to `agents_split_preserved_test.go` says items 10/19 cannot be evicted because *"Evicting either now needs a per-row base commit, because TestSplitDigestsAreTheBaseCommitsText re-derives every digest from the single agentsSplitBase."* #305 added exactly that per-row `base` field and evicted both. Consequences as of now:

- `gh pr view 304 --json mergeable,mergeStateStatus` → **`CONFLICTING` / `DIRTY`** (conflict in `splitItems`).
- The new comment would land describing a `agents_size_test.go` candidate list that #305 deleted, and asserting a constraint that no longer holds.
- The item 13 row needs a `base:` field or it will not compile.

Measured on a resolved merge (main's `splitItems` + an item 13 row at `base: agentsSplitBase`): `AGENTS.md` = **54,042 bytes against the new 54,250 ceiling — 208 bytes of headroom**, and the full suite is green (18 ok, 0 FAIL, `gofmt` 0, `go vet` clean). So the eviction of item 13 is still *necessary* post-#305; only the stated reason for choosing it is stale.

---

Separately, and not a description issue: `README.md` §Generate (1494-1496), §Waiting (1793-1796) and §cancel (1846-1848) still carry *"There is no cancel-for-refund"* verbatim — the sentence this PR's own ledger marks CONFIRMED-FALSE and removes from `generate --help`. `README.md` is edited by this PR, so `CLAUDE.md`'s lockstep rule applies to it here regardless of the `workflows*.go` file-ownership argument.

ZacxDev:
## Audit rework — pushed as `9339827`, rebased onto `d20ec98`

Conflict resolved, all five findings addressed. The delta to re-audit is `f480028..9339827`.

The audit was right about the thing that mattered most: **the guard was the defect.** Details below, then the re-run battery.

---

### 1. Rebase — done
Past #298, #299 and #305. The `splitItems` conflict is resolved with `base: agentsSplitBase` for item 13 (its body was untouched by the compression pass that created wave 2, so it is genuinely wave 1). **The 15-line comment describing the deleted candidate list is gone**, replaced by five lines that defer the ranking to `evictionPlaybook()` — "the ranking that matters is computed from the live file, not restated here."

### 2. README cancel contradiction — resolved by NARROWING, not by asserting

I took the first of your two options. The `--help` header now says:

> A submitted generation is charged. There is no cancel-for-refund. […] If the run then **FAILS or EXPIRES**, what becomes of the charge is decided server-side and *[the shared note]*. Cancelling is a separate question with its own answer under `--timeout` below.

and `generate.go`'s `--timeout` paragraph is **restored to main's wording verbatim**. So this PR now touches no cancel claim at all — CLI or README — and #307 owns the question. That also removes the half-fixed-copy-set problem: `workflows*.go`'s six sites and README's three stay consistent with each other, exactly as they were.

**Finding 6 (`--help` self-contradiction) falls out of the same change**, plus one more step: the four *overage* claims (`:197`, `:447`, `:483`, `:1460`) said "is not refunded" / "with no refund" about a charge nobody read the rule for. They are reframed around what `--max-cost` **can** do — "the realized charge can be higher, and `--max-cost` cannot claw the difference back — it never reaches the server." Same deterrent, no direction asserted. README's mirror (`--max-cost is an estimate check`) moves with it.

### 3. The refund guard, rebuilt structurally

Your mutant is the headline. `refundDirectionalClaims` was a substring ban, and "does this sentence answer the refund question" is not computable from text — any list loses to the next paraphrase.

Three guards now, around one shared constant `buzzLedgerUnknownNote`, and **none subsumes the others**:

| | guard | catches |
|---|---|---|
| **A** | the constant must contain `cannot see your Buzz ledger` and no directional claim | someone rewords the constant — the one mutation consolidation would otherwise propagate everywhere silently |
| **B** | every surface renders it **verbatim** (terminal error ×3 statuses, exclusion note, `--help`) | your mutant: it replaced the sentence, so the constant is simply absent |
| **C** | asserted **ledger** of call sites, failing on growth *and* shrinkage | a new surface that invents its own wording |

**A' is the control you asked for.** The old one compared two literals and could not fail. The new one runs the *real predicate* over a corpus that must be rejected — including your exact surviving mutant and five other paraphrases ("the charge is reversed for you", a hedge that still decides) — and over the real constant, which must be accepted. A predicate wired to nothing fails it in both directions.

The wording also picked up **finding 11**: `civitai buzz` reports a balance, not a history, so it cannot settle "did *this* run refund". The note says so and points at `/user/transactions`.

### 4. Confirmation prompt — pinned

`TestConfirmGenerate_SpendWarningIsPresentAndNonDirectional`. Paired: the prompt must still be a prompt, the screen must say **both** "spends real Buzz" and "cannot be undone" (state, not one sentence), and no directional claim may appear. Your `"This is free and fully refundable."` mutant is M5 below; deleting the line entirely is M6.

### 5. `ready: false` — softened to what is evidenced (finding 7)

Dropped "a real submit would be refused". The only refusal this repo can evidence is **ours**: `runGenerate` reads the flag and stops. Now: *"The server reports the resources for this job are NOT currently available (ready: false) — this CLI refuses to submit in that state."* README and item 28 carry the same correction, naming what was searched (`support !== 'available'` appears once, in the whatIf reply builder) and that #279's bad checkpoints came back as HTTP 400s.

### 6. Fails-closed snippet (finding 8)

```bash
case "$(printf '%s' "$q" | jq -r 'if has("ready") then .ready else "absent" end')" in
  false)   echo "resources unavailable" >&2; exit 1 ;;
  true)    ;;                                            # NOT a green light
  *)       echo "no readable .ready field" >&2; exit 1 ;;
esac
```
The README now names why the `*` arm exists, tying it to `views.unavailable`'s absent-vs-measured lesson.

### 7. "generatable" purged from both dry-run streams (finding 9)

The ban was keyed on `"generatable:"` **with a colon**, so the warning one line below re-taught the equation. Both streams are now banned on the bare word, and the warning, the submit refusal, the README bullet head and the exit-code table row are reworded. Mutant M8' isolates this: it keeps "NOT currently available" so no *other* assertion can fire, and only the word-ban catches it.

### 8. Also: dead branch (finding 12)

`Resources ready:\tfalse` can never match — tabwriter emits padding. Collapsed to a single whitespace-normalised assertion, with the reason in the comment.

---

## AGENTS.md — item 28 was itself the largest item in the file

`evictionPlaybook()` ranked **item 28 at 6,757 bytes**, above everything else. Cut to ~2,300: it keeps every RULE and drops the measurements, which now live in code comments at the sites they bind. Paid for by evicting **item 25** to `claudedocs/decisions/25-listing-media-bounds.md` — using **#305's per-row `base`**, pinned to a new `agentsSplitBaseWave3 = d20ec98` (the tip this branch left from), so the digest still comes from `git show <base>:AGENTS.md` and never from the file I wrote.

**52,662 bytes, 1,588 under the 54,250 ceiling** (was 208 before).

On **finding 13**: #305's per-row `base` is not half the fix — it is the whole one, and item 25 is the proof. Any item can now be evicted from any reachable commit. A brand-new item still cannot get an evidence file (no base commit holds its body *before* it is written), which is why item 28 is inline; the way out is to land it inline and evict it in a follow-up, which the wave mechanism now supports cleanly.

---

## Re-run gate — full battery, not just the delta

**Baseline:** `RUN=3188 PASS=3184 FAIL=0 SKIP=4`, `build failed=0`, `panic: test timed out=0`, `pkgFAIL=0`, `ok=18`. `make ci` green. golangci-lint **v2.12.2** (CI-pinned, via `nix-shell`): **0 issues**. `gofmt -s -l .`: 0.

**14 mutants, enumerated by semantic failure mode — 0 survivors, each red with its own test's message:**

| # | mode | mutant | caught by |
|---|---|---|---|
| M1 | paraphrase | **your surviving mutant**, re-applied verbatim | `…IsRenderedOnEverySpendOutcomeSurface/terminal_status_error` + ledger |
| M2 | paraphrase | rewrite the **constant** to the same text | `…AssertsNonObservabilityAndNoDirection`, `…GuardCanStillFire` |
| M3 | stale re-bind | exclusion note writes its own sentence | `TestReportExcludedOutputs…`, `…/excluded_outputs_note` |
| M5 | reassuring lie | `"This is free and fully refundable."` | `TestConfirmGenerate_SpendWarning…` |
| M6 | comment-out | delete the spend warning | `TestConfirmGenerate_SpendWarning…` |
| M7 | revert | label back to `Generatable:` | `TestDryRun_ReadinessLabel…` |
| M8' | **isolated** word revival (keeps the other assertion satisfiable) | warning re-adds "generatable" | `TestDryRun_NotReadyStillWarns:101` |
| M9 | **branch inversion** | `if !q.Ready` → `if q.Ready` | `TestDryRun_NotReadyStillWarns` |
| M10 | stale re-bind | `--json` emits a re-derived document | `TestDryRunJSON_RawReadyKeySurvives…` + 2 pre-existing |
| M11' | reclassification | `civitai.Tag(civitai.ErrNotFound, …)` | the `errors.Is` exit-code pin, all 3 statuses |
| M14 | stale re-bind | `--help` drops the constant for prose | `…/help_text` + ledger |
| M16 | identifier rename | rename the constant | ledger (positive control) |
| M17 | **growth** | a new surface adopts the constant | ledger — proves it fails on growth, not only shrinkage |
| M18 | deletion | `reportExcludedOutputs` returns early | 3 tests incl. the CONTROL-failure arm |

**Red at base `d20ec98`** (scratch worktree, test file copied onto main's source): `TestDryRun_ReadinessLabelIsNotAGeneratabilityClaim` 🔴3, `TestDryRun_NotReadyStillWarns` 🔴4 (including both stream word-bans), `TestGenerateHelp_SaysReadinessIsNotAPromiseOfOutput` 🔴2. `TestDryRunJSON_RawReadyKeySurvivesTheRename` passes at base and is labelled an **invariant guard**, not regression coverage.

🔴 **The refund guards cannot be compiled against base** — they reference `buzzLedgerUnknownNote`, which this PR introduces. Their fire-ability rests on the mutant battery (M1/M2/M3/M14/M16/M17 target them specifically), not on a base run. Saying so rather than quietly reporting a matrix that does not exist.

## Residuals, stated

- **The ledger counts references, so a brand-new surface that never mentions the constant is invisible to it.** Guard B only reads the three surfaces it knows. That gap is real; M17 proves the growth direction works for any site that *does* adopt the constant, which is the reviewable path.
- **Cancel semantics remain unsettled and unfixed** — six CLI sites and three README sites still assert "no cancel-for-refund". #307.
- The overage claims are now non-directional but still **unverified** in substance; no source was read either way.

ZacxDev:
## Round 3 — pushed as `f7537b7`, rebased onto `932c5e9`, single commit

All five findings addressed, plus the two 🟢s and the confirmed residual. Delta to re-audit: `932c5e9..f7537b7`.

I am not arguing with any of it. N2 in particular was a fair hit on a mistake I had already written down and then failed to apply: round 2's own comment says *"does this sentence answer the refund question" is not computable from text* — and I still shipped a phrase list as defence in depth, which is where the five additions walked in.

---

### N0 — eviction half dropped entirely ✅
Reverted `AGENTS.md` and `agents_split_preserved_test.go` to main's and deleted both of my evidence files. **`13-cli-validates-nothing.md` would have collided** with #310's `13-generation-graph-not-validated.md` — same body, same sha, different filename, two files for one item, two red ledgers. Nothing of mine survives in that half; `agentsSplitBaseWave3` is main's `6e31b4b`.

I re-measured rather than trusting either number: main is **42,635** with `agentsMaxBytes = 45_135`, so **2,500 bytes**. My item 28 came in at 2,555 — over the budget the ceiling defines as the eviction trigger — so I trimmed it to **~2,390 rather than evict**, per your steer. **45,027 bytes, 108 under.** Tighter than I'd like; the compression is real (the measurements moved to code comments at the sites they bind) and I would not want to add another sentence to it.

### N1 — cancel is in scope, and you were right that the PR already moved it ✅
The terminal-status error covers `canceled` and I had routed it through the neutral note while `workflows cancel` still said "does NOT refund anything" — two commands contradicting each other about one event, and item 28 claiming the opposite of what the code did.

Every fate-of-charge claim is now non-directional, cancel included: **the accrued-cost half is kept** ("a mid-run cancel bills the cost already accrued", "stopping it does not call that back") because it *is* evidenced; **"nothing comes back" is dropped** because it is not. Nine sites: `workflows_cancel.go` ×4 (incl. `Short`), `workflows.go`, `generate.go` ×2, `genapi` comments ×3, README ×4. Three pre-existing tests pinned the retracted wording and now pin the new claim plus the shared note. #307 owns the substance; nothing here settles it.

### N2 — golden-output pinning, not a longer list ✅
**15 golden files**, whitespace-normalised, covering every spend surface: `generate --help` (whole `Long`), both reframed flag usages, the spend confirmation, both `--dry-run` streams, the terminal error ×3 statuses, the succeeded-but-empty error, the exclusion note, `cancel --help`, the `workflows` group help, the cancel prompt and the non-TTY cancel refusal.

Your five surviving mutants are ADD1–ADD6 in the battery below. All five now die on the golden, regardless of wording — including `Nothing is refunded`, which no list would have caught. `-update` re-approves and the diff is the review; the header says so and says not to run it to make red go green.

**Accepted cost, stated:** a cosmetic reflow of any of these breaks the test. **Two things normalisation cannot see** — a pure re-wrap, and ANSI styling — are written into the header so a green here is not read as "byte-identical".

### N3 — the four uncovered sites ✅
All four reverted cleanly with the suite green; they are ADD9/ADD10/ADD11 and the `--dry-run` caveat inside `generate_dry_run_stderr`. Pinned.

### N4 — README carries the claim, pointer fixed ✅
The *Generate* section gains a **"What the LEDGER does with a charge"** callout stating the position in both directions and naming #307. `README.md:1642`'s dangling "see *Generate* above" now points at it by name. `grep -n 'ledger' README.md` returns three sites instead of one.

### N5 / N6 — both comments corrected ✅
The constant's doc comment now scopes itself to the **FATE** of a charge and says explicitly that `reportOutputCountMismatch` and the succeeded-but-empty error deliberately do *not* render the note (they claim a charge *happened*, which is not in doubt) — so the ledger's omission of them is documented rather than contradicted. It also records the new-file residual with its measurement. `rejects()` gained the third clause its comment claimed.

### 🟢 `generate_substitution.go:75` — not out of scope; it was a live defect in my own guard
You flagged this as arguably out of scope. It was not: `reportModelSubstitutions` prints on the **ESTIMATE** path, so "never generatable" lands in `--dry-run` stderr a few lines from `Resources ready`. A third sense of the word on the same screen, and my blanket ban would have false-failed on correct output. Reworded to "never offered for generation".

🔴 **And the ban that should have caught it was UNREACHABLE** — the only dry-run fixture reported no substitutions, so the banned word never rendered. Audit mutant **ADD12 survived the whole suite** on the first run of this round's battery. There is now a substitution dry-run fixture that reaches the ban *and* serves as its false-positive control, plus a golden for that stderr.

---

## Round-3 battery — built from ADDITION first

**21 mutants, 0 survivors.** 13 addition, 8 deletion/paraphrase/branch-inversion/rename (the round-2 set retained so the battery is not narrower than one that already passed).

| | mutant | caught by |
|---|---|---|
| ADD1 | `--help` + "(Your Buzz has been refunded in that case.)" | `generate_help_long` |
| ADD2 | terminal error + "gone straight back to your account" | `terminal_status_failed` |
| ADD3 | exclusion note + "In practice the Buzz comes straight back." | `excluded_outputs_note` |
| ADD4 | dry-run caveat + "If it fails, the Buzz comes straight back." | `generate_dry_run_streams` |
| ADD5 | exclusion note + "**Nothing is refunded**." (the one the list missed) | `excluded_outputs_note` |
| ADD6 | spend confirmation + "Failed runs are refunded automatically." | `generate_confirmation` |
| ADD7 | cancel prompt + "Undelivered images are refunded." | `cancel_confirmation` |
| ADD8 | cancel `--help` + "Nothing is ever refunded." | `cancel_help_long` |
| ADD9–11 | the three reframed overage claims, reverted | the two flag-usage goldens + `generate_help_long` |
| ADD12 | substitution advice re-adds "generatable" | **survived first pass**; now `TestDryRun_SubstitutionNotice…` + its golden |
| ADD13 | `workflows` group help + "non-refundably" | `cancel_group_help_long` |
| DEL1–8 | round 2's set (paraphrase, constant rewrite, label revert, `if !q.Ready` → `if q.Ready`, `--json` re-derive, reclassification, warning deletion, constant dropped) | all still caught |

## Confirmed residual — re-measured, and narrower than it looked

I re-ran your `zz_new_refund_surface.go` probe. `go test`: **survived, rc=0**, exactly as you found. But:

```
internal/cmd/zz_new_refund_surface.go:9:6: func zzNewRefundSurface is unused (unused)
```

**golangci-lint catches it** — an *unwired* new surface cannot land. The real hole is narrower: a **wired** fate-of-charge claim on a command that has no golden. That is what item 28 now records, rather than the broader "invisible to all three".

## Gate

`RUN=3228 PASS=3224 FAIL=0 SKIP=4`, `build failed=0`, `panic: test timed out=0`, `pkgFAIL=0`, **19 packages ok**. `make ci` green. golangci-lint **v2.12.2** (CI-pinned, via `nix-shell`): **0 issues** — with the probe above as its negative control this round. `gofmt -s -l .`: 0. `AGENTS.md` 45,027 / 45,135.

PR #313, opened 2026-08-10
fix(release): the daily cask check was a tree falling in an empty forest

### LINKED ISSUE #312: [cask-check] the Homebrew cask has not followed the latest published release
> ⚠️ **This is a fire drill** (`drill=lagging`), dispatched by hand against a fixture — not a real
> failure of the live pipeline. It exists to prove this reporting path works. Close it when done.

**State:** `lagging`  
**Run:** [https://github.com/civitai/cli/actions/runs/31358273581](https://github.com/civitai/cli/actions/runs/31358273581)  
**Trigger:** `workflow_dispatch`  
**Jobs:** `push-cask` = `skipped`, `verify` = `failure`

Nobody's install is broken — the version the cask names *is* downloadable. What has not
happened is the cask following a release that was **published** over a day ago, and the
only thing allowed to move the cask is the `release: published` job in this workflow.
A dropped `release` webhook, a failed push nobody read, or an expired
`HOMEBREW_TAP_GITHUB_TOKEN` all look exactly like this.

Re-run the push for a specific release with:
```
gh workflow run release-homebrew.yml -f tag=vX.Y.Z
```

<details><summary>caskcheck output</summary>

```
FAIL (lagging): the cask has not followed the latest published release.
cask version:                 0.1.90
latest PUBLISHED release:     v0.1.91 (published 2h16m0s ago, at 2026-08-10T03:04:20Z)

Nobody's `brew install` is broken by this — 0.1.90 is downloadable — so it is not the 2026-08-09 outage.
What it means is that the only thing allowed to move the cask, the `release: published` job in
.github/workflows/release-homebrew.yml, has not moved it in over 1s. Check that workflow's runs for v0.1.91:
a dropped `release` webhook, a failed push nobody read, or an expired HOMEBREW_TAP_GITHUB_TOKEN all look
exactly like this. `gh workflow run release-homebrew.yml -f tag=v0.1.91` re-runs the push for that release.
(If v0.1.91 was deliberately never shipped to Homebrew, this check has no way to know that — say so here and
raise the threshold, currently 1s, rather than leaving a permanently red run nobody reads.)
exit status 1
```

</details>

---
Filed automatically by `.github/workflows/release-homebrew.yml`. It files one issue at a time and
rewrites this body on every failing run; it comments only when the state changes, and closes this
issue when the check goes green.

ZacxDev:
#308 built a detector for the Homebrew-cask invariant and put it on a daily cron. This makes a failing daily run reach a human, because today it does not.

## What actually happens today (measured, not recited)

**The schedule is real and has never fired yet.** `release-homebrew.yml` is `state: active` (`gh api repos/civitai/cli/actions/workflows`), cron `23 13 * * *`, and `gh run list --workflow release-homebrew.yml` returns `[]` — #308 merged at 03:33Z on 2026-08-10, so the first scheduled run is due 13:23Z that day.

**A red scheduled run notifies exactly one account.** GitHub's rule ([Notifications for workflow runs](https://docs.github.com/en/actions/concepts/workflows-and-actions/notifications-for-workflow-runs)):

> Notifications for scheduled workflows are sent to the user who initially created the workflow. If a different user updates the cron syntax, in the `schedule` event in the workflow file, subsequent notifications will be sent to that user instead.

That is the whole audience: the account behind `6e31b4b` (authored and merged by `ZacxDev`), reached through that person's own per-user Actions notification setting — which is theirs to mute and which nobody else can inspect (the API exposes no other user's notification preferences; a repo `/subscription` read needs the `notifications` scope). There is no org-, team-, CODEOWNERS- or repo-level Actions failure notification to widen it. Related, and worth knowing: this is a public repo, so [scheduled workflows are disabled automatically after 60 days of repository inactivity](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows).

**And in this repo that audience has a measured track record of zero.** `bump-flake-vendorhash.yml` is a scheduled workflow here, and **every one of its scheduled runs has failed** — 2026-07-20, 2026-07-27, 2026-08-03, three for three, a genuine non-zero exit in its `recompute flake vendorHash` step (`##[error]Process completed with exit code 1`). Across those three weeks this repo's tracker accumulated 30+ hand-filed issues, and not one of them is about that. Nobody found out. That is the exact shape #308's detector was built to escape, regenerated one layer up.

## The mechanism, and why this one

**A GitHub issue in this repo**, titled `[cask-check] …`, opened by the `signal` job.

Chosen over email / Slack / a webhook because it needs **no new secret, no external service and no per-user setting**: it is visible to anyone who opens the repo, it is where this project already tracks defects, and it outlives the 90-day expiry of the run log it links to. The only new permission is `issues: write`, scoped to that one job rather than to the workflow.

**It does not spam:**

| behaviour | how |
|---|---|
| one open issue at a time | found by the `[cask-check]` title marker over a plain issue **list** — never a search query, whose index is eventually consistent and would let two runs each conclude "there isn't one yet" |
| a week of the same failure → **1 issue, 0 comments** | a repeat failure **rewrites the body**; editing a body notifies nobody, so the issue always describes the latest run without pinging anyone |
| the one notifying update | a comment, posted **only when the kind changes** (`green`→`broken`→`lagging`→…), because that is the only change that alters what a human should do. The previous kind is read from a marker in the body, not the title, so renaming the issue by hand does not trigger a spurious "it changed" |
| recovery | a green run **closes** the issue and says so. Not cosmetic — it is what resets the mechanism, so the *next* failure opens a fresh issue and notifies again instead of deduping into a stale one |

**"Failed" and "could not run" get different words.** `caskcheck` now classifies itself, keyed on error **sentinels** rather than message text, and the default arm is a visible `unclassified` state rather than a fallback — so a failure path added later cannot silently inherit another kind's wording. The workflow adds two more states it owns: `push-failed`, and `unknown` for a `verify` job that died before producing any verdict at all. `unmeasurable`, `unclassified` and `unknown` all say in the issue body that the Homebrew channel is **unverified**, never that it is fine.

The kind rides a `-kind-file`, **not the exit code**, for a measured reason: on go1.25.12 `go run` collapses every non-zero exit to 1 (`os.Exit(3)` and `os.Exit(11)` both give rc=1 through `go run`; the same program built with `go build` propagates 3 and 11). Reading a kind off the exit code would have reported every verdict as the same one.

`TestEveryKindIsHandledByTheWorkflowThatReadsIt` is a cross-file seam guard: a kind added in Go and not named in the YAML fails the build. It states its own limits — forward direction only, and appearing is not being branched on. The behavioural half is the drill.

## Proof it fires

Not a claim. Five real `workflow_dispatch` runs against this branch, driving the whole lifecycle:

| # | run | dispatch | result |
|---|---|---|---|
| 1 | [31358041942](https://github.com/civitai/cli/actions/runs/31358041942) | `drill=broken` | **opened #312** — as `unknown`. *And found a bug; see below.* |
| 2 | [31358196617](https://github.com/civitai/cli/actions/runs/31358196617) | `drill=broken` | retitled, **1 comment**: `unknown` → `broken` |
| 3 | [31358237061](https://github.com/civitai/cli/actions/runs/31358237061) | `drill=broken` | **0 new issues, 0 new comments.** Log: `#312 already reports 'broken'; body refreshed, no comment (this is the no-spam path).` |
| 4 | [31358273581](https://github.com/civitai/cli/actions/runs/31358273581) | `drill=lagging` | retitled, **1 comment**: `broken` → `lagging`, with the lag wording (explicitly *not* claiming users are broken) |
| 5 | [31358313647](https://github.com/civitai/cli/actions/runs/31358313647) | `drill=off` — the **real** tap | green (`cask version 0.1.91; 4 archive URL(s) … all publicly downloadable`, `lag: cask matches the latest published release v0.1.91`) → **closed #312** with a recovery comment |

After runs 1–3 the marker matched **exactly 1 issue** with **exactly 1 comment**, checked via the API.

🔴 **The first drill found a real defect, and it was the expensive-direction one.** Run 1 filed "the check **could not run**" for a run in which caskcheck had produced a complete 404 finding and written `kind=broken`. Cause: GitHub invokes a `run:` block as **`bash -e {0}`** — errexit is already on before the script's first line, and `set -uo pipefail` does not clear it, so `go run ./tools/caskcheck` aborted the step the instant the check found anything. `rc=$?` never executed and `kind=` was never published. The signal job was right that no verdict existed; it was just far too quiet about a live outage. Reproduced locally in both directions, running the step **extracted from this YAML** under `bash -e` exactly as GitHub does:

```
without `set +e`   step rc=1, GITHUB_OUTPUT empty,      end of script NOT reached
with    `set +e`   step rc=0, GITHUB_OUTPUT kind=...,   end of script reached
```

The kind *file* was written in both cases — caskcheck had done its job. What was lost was the workflow's chance to read it. Fixed with an explicit `set +e` (commented with the measurement) plus `always()` on the gate step. This is the entire argument for the drill existing: a reporting path nobody has watched work is the same class of defect as a guard that cannot fail, and this one was broken on arrival in the direction that understates an outage.

The **dedupe logic** was additionally rehearsed offline over a 12-day simulated lifecycle — the `signal` step extracted from the YAML, driven against a stateful fake `gh`: 7 consecutive identical failures → 1 issue / 0 comments; a kind change → 1 comment; recovery → close; a failure after recovery → a *new* issue; `push-cask` failed while the invariant is fine → the push-failed title; `verify` dead with an empty kind → the could-not-run title. The harness carries its own negative control (a deliberately-wrong expectation, which fails).

**What I could not verify, and what to watch.** Two triggers were untestable within my permissions:

- **The `schedule` trigger itself** — first fire 13:23Z. It runs the same two jobs the drills ran with `push-cask` skipped, which is the state every drill above ran in (`- push-cask in 0s`); the only difference reaching any code is the `github.event_name` string printed in the body.
- **`release: published`** — publishing a release is out of bounds. `push-cask`'s own steps are untouched by this PR; what is new on that path is `signal` observing `needs.push-cask.result == 'failure'`, whose title/body selection is covered by the offline rehearsal but not by a live run.

On the first real failure, expect: a new issue titled `[cask-check] …` within a minute of the run finishing, and the `signal` job green (it succeeds whether or not the check did — it is the reporter, not the gate).

## The lagging cask: a signal, with a defended threshold

Left green for **24h**, a finding after that — a separate kind, a separate title, and wording that explicitly says nobody's install is broken.

Why a signal at all: a lagging cask installs cleanly, so **nothing else in the pipeline notices it**. The only thing that moves the cask is the `release: published` trigger, so a cask still lagging a day later means that event never reached its job — a dropped `release` webhook (an Actions incident did exactly that to this repo's push/PR triggers on 2026-08-06, per `ci.yml`'s own comments), a failed push nobody read, an expired `HOMEBREW_TAP_GITHUB_TOKEN`.

Why 24h, measured over the 30 most recent releases:

- `published_at` → tap commit: **~1 minute** (e.g. v0.1.90 published 00:09:13Z, tap commit 00:08:21Z).
- tag → publish, a strictly *larger* window this check does not even measure: median **~2m30s**, second-worst **22m46s** (v0.1.80), worst **1h55m** (v0.1.91 — the incident itself).

So 24h is ~1400× the normal push latency and >12× the worst publish latency ever recorded here. It is also **at least the check's own period**, so "the release published after yesterday's run" can never be reported as a lag — a threshold below the interval would false-fire on the normal path. Pinned from both sides (`TestTheLagThresholdIsPinnedFromBOTHSides`), and demonstrated live against the real GitHub API in run 5 vs run 4: the *same* fixture is `green` at the 24h default and `lagging` at `-lag-threshold 1s`.

Three ways the lag question can fail to be answerable — an unreadable releases API, a payload with no `published_at`, an unknown tag — all report "not evaluated" and stay green, never a finding. `published_at` is a **pointer** precisely so a missing key cannot decode to the zero time (maximally stale) and manufacture a lag finding out of a field the server never sent; that is its own test.

The 404 outage always outranks a lag when both hold, so the softer of two simultaneous findings can never be the one that surfaces.

## Residuals, stated rather than hidden

- **This cannot see a run that never happened.** A schedule silently dropped, or disabled after 60 days of repo inactivity, produces no run, no failure and therefore no issue. Closing that needs a second watcher, which has the same problem one level up; not attempted here.
- **`gh issue list --limit 500`** bounds the dedupe search. Past 500 open issues it could miss the existing one and open a second.
- **The issue body is machine-owned** and rewritten on every failing run; the body says so and tells you to put notes in a comment.
- **Drill run 1 left #312 closed** with the `lagging` title it last held. That is normal issue behaviour after a recovery, and the closing comment explains it.
- **`TestEveryKindIsHandledByTheWorkflowThatReadsIt` proves presence, not handling** — it says so in its own doc comment.

## Checks

- `make ci`: **19 packages ok, 0 `--- FAIL`**. `make lint` separately (golangci-lint via nix-shell): **0 issues**. `gofmt -s -l .` clean, with a positive control that it found the 3 files in `tools/caskcheck`.
- `tools/caskcheck`: **19 PASS / 0 FAIL**, re-run inside a real `git clone --depth 1` of this branch (`rev-list --count HEAD` = 1), because CI checks out shallow.
- Workflow YAML parsed with `yq`; both embedded scripts extracted and passed `bash -n` and `shellcheck`.
- The workflow seam guard was watched to **fail** before the YAML existed and pass after.


ZacxDev:
Merged-tree gate: rebuilt an integration branch off `origin/main` (`ac29e09`, 2 commits ahead of the branch point) and merged this PR into it — `make ci` **19/19 ok, 0 `--- FAIL`**, `make lint` **0 issues**, and `make ci-shallow` (new on main in #311) **19/19 ok, 0 failures, 0 timeouts** in a real depth-1 clone. The only file both sides touched is `README.md`; I read the merged region rather than trusting the clean merge, and it reads coherently.

`origin/main` is now merged into the branch itself, so the checks below ran on the merged content: **12/12 SUCCESS**, 0 non-terminal, `mergeStateStatus: CLEAN`.

Drill artifact: #312 is **closed** — it was opened and closed by runs 1–5 in the table above, and closing it is what the recovery path does.

Issue #365, opened 2026-08-10
seven UX papercuts from dogfood run 3: img2img says "Workflow: txt2img" on the approval screen, no --limit on app status, doubled ✓ glyph, and four docs nits


ZacxDev:
Re-verified today against `civitai v0.1.92-10-g3156b80` (built from `main` @ `3156b80`).

Small, individually-cheap fixes. Each was re-run or re-read against current `main`; two
sub-items from the original report did **not** survive and are recorded as refuted at the
bottom rather than filed.

---

## 1. An img2img job says `Workflow: txt2img` on the screen you approve

```
$ civitai generate "a cat" --image ./probe.png --ecosystem Flux1Kontext --dry-run
  Ecosystem:  Flux1Kontext
  Image:      ./probe.png (512x512) → https://orchestration-new.civitai.com/v2/consumer/blobs/…
If this ecosystem does not support image editing, the server IGNORES the images above,
generates from the prompt alone and still charges — the estimate cannot show the difference.
…
Workflow:         txt2img
Prompt:           <not checked by --dry-run>
Resources ready:  true

Estimated cost
  …
  total       45
rc=0
```

`Workflow: txt2img` is *correct* — the server promotes the job to image-to-image from the
presence of `--image` plus `--ecosystem`, and `AGENTS.md` item 19(a) records exactly that. But
this is the confirmation summary a user reads before approving a spend, on the same screen as
a warning that says the server may "IGNORE the images above". Reading `txt2img` there is the
single most alarming thing on the screen, and nothing explains it.

**Fix:** annotate it — `Workflow: txt2img (the server promotes this to image-to-image because
--image and --ecosystem are set)` — or show the promoted name. Do **not** silently rename the
field; the wire value is `txt2img` and item 19(a) depends on that staying visible.

---

## 2. `app status` has no `--limit`, its sibling `app list` does

```
$ civitai app status
note: showing the newest 100 submissions — the API caps this listing and offers no way to page,
so older submissions may exist but are not listed. Look up a specific app with
`civitai app status <blockId>`.
BLOCK_ID                    VERSION  STATUS     DEPLOY        SUBMITTED   URL
… 100 rows …
```

Flags: `--id`, `--json`, `--help`. `app list` has `--limit` *and* `--cursor`.

The note (added by #207) is honest about *why* there is no `--cursor` — the API cannot page.
That does not explain the absence of a client-side `--limit`, which is the useful half: the
common interactive question is "what are my last five submissions", and today the answer is
[...]

PR #368, opened 2026-08-11
docs(readme): three published-contract claims a reader could not reproduce (#360, #361, #364)

### LINKED ISSUE #360: neither worked example in the README's silent-model-substitution section reproduces — one 404s, the other shows no substitution at all
Re-verified today against `civitai v0.1.92-10-g3156b80` (built from `main` @ `3156b80`).

`README.md` § [🔴 Silent model substitution](#-silent-model-substitution) exists to teach the
single most expensive surprise in `civitai generate`. It teaches it with **two** console
blocks, and neither one reproduces. Both were run verbatim today.

## Symptom 1 — the `--dry-run` block errors out instead of warning

README (`README.md:1691`) shows:

```console
$ civitai generate "a cat" --checkpoint 999999999 --dry-run
⚠ The server will NOT use the checkpoint you asked for. It has substituted a
  different model, and this estimate prices the SUBSTITUTE. Nothing has
  been submitted or charged yet.
    requested version 999999999 -> will run version 2436219  (reason: unrecognized)
      the server does not offer that version in this model family at all …
```

Actual:

```
$ civitai generate "a cat" --checkpoint 999999999 --dry-run
Error: --checkpoint 999999999: not found (404): Model not found
rc=4
```

The CLI's own `ResolveModelVersion` pre-flight 404s on the id **before** anything is priced,
so the substitution path is not reachable by this command at all.

## Symptom 2 — the `[SUPERSEDED …]` annotation does not appear either

README (`README.md:1703`) shows the confirmation line for the same scenario:

```console
Checkpoint:   DreamShaper — 8 (Checkpoint, id 128713)  [SUPERSEDED — the server will run version 2436219 instead; see the warning above]
```

Actual, with that exact id:

```
$ civitai generate "a cat" --checkpoint 128713 --dry-run
Workflow:         txt2img
Prompt:           <not checked by --dry-run>
Checkpoint:       DreamShaper — 8 (Checkpoint, id 128713)
Resources ready:  true

Estimated cost
  base        8
  …
  total       8
rc=0
```

No `[SUPERSEDED …]`, no warning, no `2436219`. The estimate prices it as an ordinary job.

## Why it matters

This is the one section whose entire purpose is to make a user recognise a silent
substitution **before** they approve a spend. A reader who tries either example concludes the
warning machinery is not real — and the section goes on to explain, correctly, that
`--fail-on-substitution` is *not* a spend guard and that "no warning" is ambiguous. A
non-reproducing demo in front of that argument is the worst place for one.

Nothing in this run found an invocation that produces the documented warning.

## Suggested fix

Either replace both blocks with an invocation that actually substitutes today, or — if none
exists against the current server — mark them as illustrative-not-runnable and say plainly
that the pre-flight resolve now rejects an unknown id with `not found (404)` before the
estimate is priced. A `--checkpoint <unknown-id>` that 404s locally is arguably the better
behaviour; the docs just have to admit it.

## Related, not duplicate

- **#345** — the `--checkpoint` example in **`generate --help`** produces a *failed, charged*
  job. Different surface, and that one gets far enough to be billed; this one never leaves
  the CLI.
- **#352** — the design question of whether `--checkpoint` should carry its ecosystem.
- Neither covers the README section's own two console blocks.

---
**Source:** blind **credentialed** dogfood run 3, 2026-08-10 (`claudedocs/handoff-dogfood-3.md`,
finding 3) — an agent given the built binary and a live token, forbidden from reading the CLI
source, `AGENTS.md` or the prior run reports. Re-verified against current `main` before filing;
symptom 2 is new and was not in the original report. Both commands above were run with
`--dry-run` only — nothing was submitted or charged.

### LINKED ISSUE #361: the Troubleshooting index attributes `is this an App project?` to app validate, which never prints it — and validate's real message is not in the index at all
Re-verified today against `civitai v0.1.92-10-g3156b80` (built from `main` @ `3156b80`).

The README Troubleshooting index promises (`README.md:2216`):

> **Look up the message you got.** Every row's left column is a fragment of a string this CLI
> really prints, so searching this page for a few words of your error should land you on the
> right row.

The row for `is this an App project?` sends you to **Validate fidelity**. `app validate`
never prints that string. And the string `app validate` *does* print is in the index nowhere.

## Symptom

The row (`README.md:2248`, under "Validating and submitting"):

| You saw | What it means | Where to read more |
| --- | --- | --- |
| `is this an App project?` | There is no `block.manifest.json` at the path you named. The path itself was fine — which is why this exits `1` and not `2`. | [Validate fidelity](#validate-fidelity) |

What `app validate` actually prints:

```
$ civitai app validate ./emptydir
✗ 1 validation error(s) in ./emptydir:
  - block.manifest.json not found at project root ./emptydir
Error: validation failed
rc=1
```

Searching this page for `block.manifest.json not found at project root` finds nothing.

What really prints `is this an App project?` — e.g. `app listing`:

```
$ civitai app listing status          # in a directory with no manifest
Error: could not resolve the app — run this from your app directory (with block.manifest.json)
or pass --slug: no block.manifest.json found in . — is this an App project? run
`civitai app init` to create one
rc=1
```

## Where the string lives

`internal/manifest/manifest.go` (`Load`, `LoadRaw`, `SetBlockID`) is the only source of
`is this an App project?`. Its non-test callers in `internal/cmd/` are:

- `app_submit.go:113`
- `app_listing.go:141`
- `app_dev_token.go:482`
- `app_dev_tunnel.go:322`

`app validate` is not among them — it produces its own finding through `internal/validate`.

## Why the invariant test does not catch it

The index is guarded by an existence check — every left-column fragment must be a string the
CLI prints somewhere. `is this an App project?` **is** printed somewhere, so the guard is
green while the row's attribution and its "Where to read more" link are both wrong. The check
is on existence, not on which command emits it.

## Why it matters

`app validate` is the first command a new author runs after scaffolding, and its
manifest-missing message is the single most likely error in the whole tool. It is the one
message the "look your error up here" index cannot resolve.

## Suggested fix

1. Add a row for `block.manifest.json not found at project root`, pointing at
   [Validate fidelity](#validate-fidelity).
2. Repoint the `is this an App project?` row at the commands that emit it (`app submit`,
   `app listing`, `app dev-token`, `app dev-tunnel`).
3. Strengthen the guard from existence to attribution: assert that each row's fragment is
   reachable from the command(s) the row's link names, so a future re-word of either side
   fails the test instead of silently drifting.

---
**Source:** blind **credentialed** dogfood run 3, 2026-08-10 (`claudedocs/handoff-dogfood-3.md`,
finding 4) — an agent given the built binary and a live token, forbidden from reading the CLI
source, `AGENTS.md` or the prior run reports. Re-verified against current `main` before filing.

### LINKED ISSUE #364: --version is documented as accepted by every command; it is root-only and exits 2 everywhere else
Re-verified today against `civitai v0.1.92-10-g3156b80` (built from `main` @ `3156b80`).

## Symptom

`README.md:2097`:

```
## Global flags

These are accepted by **every** command:

| Flag | What it does |
| --- | --- |
| `-v`, `--version` | Print the version and exit. (`civitai version` prints version + commit + build date.) |
| `-h`, `--help` | … |
| `--no-color` | … |
| `--color` | … |
| `--no-update-check` | … |
```

`--version` is registered on the root command only:

```
$ civitai app validate --version
Error: unknown flag: --version
Run 'civitai app validate --help' for the available flags
rc=2

$ civitai models search --version
Error: unknown flag: --version
Run 'civitai models search --help' for the available flags
rc=2
```

The other four rows genuinely are global — `--color`, `--no-color` and `--no-update-check`
appear in every subcommand's `Global Flags:` block, and `--help` is Cobra's own.

## Why it matters

This is a published-contract page. `--version` on a subcommand is a natural thing for a script
to do (a wrapper that pins a CLI version often probes the subcommand it is about to call), and
it exits `2` — a **usage** error, which a caller reasonably reads as "you invoked the
subcommand wrong", not "that flag is root-only".

## Suggested fix

Either make it true (`cmd.PersistentFlags()` on root, or `--version` on each subcommand) or —
preferably, since a per-subcommand `--version` is noise — narrow the sentence: say the table
is "accepted by every command **except `-v`/`--version`, which is root-only**", and move the
`--version` row out of the table or annotate it.

## Also checked, and correct — do not "fix" these

- **The colour-precedence contract** in the same section was verified fully correct during the
  run.
- **`scopeJustifications.<scope>`** (`README.md:1040`) was reported during the run as
  documented-but-never-emitted. **That report is wrong** and is not being filed. The dotted
  form *is* emitted, by the keys⊆scopes rule:

  ```
  $ civitai app validate ./f9a --json
  { "field": "scopeJustifications.user:read:profile",
    "message": "scopeJustifications key \"user:read:profile\" is not one of the manifest's declared scopes" }
  ```

  The *other* rule — a sensitive scope with no justification — emits the bare
  `scopeJustifications` on purpose (`internal/validate/semantic.go:177-186`: the field is the
  map the author must add entries to, and the server emits one message naming them all).
  Both are correct as they stand. Noting it here so the next reader of the dogfood handoff
  does not re-file it.

---
**Source:** blind **credentialed** dogfood run 3, 2026-08-10 (`claudedocs/handoff-dogfood-3.md`,
findings 8 and 9) — an agent given the built binary and a live token, forbidden from reading the
CLI source, `AGENTS.md` or the prior run reports. Re-verified against current `main` before
filing; finding 9 did not survive re-verification and is recorded above as refuted.
Finding 4 from the same run is the same theme (the Troubleshooting index attributes a string to
the wrong command) and is filed separately as the stronger case.

ZacxDev:
Closes #360, closes #361, closes #364.

Three README claims a reader could act on and get a different answer than the
page promises. Two of the three had a guard aimed at them that was green the
whole time; part of the deliverable is closing those holes.

## Re-verification first — all three still reproduced on current `main`

The issues were filed against a binary from `3156b80`. `main` is now `9c7baed`
(eight commits later) and `cebfa45` (#356) rewrote the `--checkpoint` guidance
next door, so every symptom was re-measured against `./bin/civitai` built from
`9c7baed` (`v0.1.92-18-g9c7baed`) before a line was written. **None was already
fixed.**

**#360, symptom 1** — the documented `--dry-run` block:

```
$ ./bin/civitai generate "a cat" --checkpoint 999999999 --dry-run
Error: --checkpoint 999999999: not found (404): Model not found
rc=4
```

**#360, symptom 2** — the documented `[SUPERSEDED …]` block:

```
$ ./bin/civitai generate "a cat" --checkpoint 128713 --dry-run
Workflow:         txt2img
Checkpoint:       DreamShaper — 8 (Checkpoint, id 128713)
Resources ready:  true
…  total  8
rc=0
```

No warning, no `[SUPERSEDED …]`, no substitution.

**#361**:

```
$ ./bin/civitai app validate <emptydir>
✗ 1 validation error(s) in <emptydir>:
  - block.manifest.json not found at project root <emptydir>
Error: validation failed
rc=1
```

`is this an App project?` is nowhere in that output; `block.manifest.json not
found at project root` was nowhere in the index.

**#364**:

```
$ ./bin/civitai app validate --version
Error: unknown flag: --version
rc=2
$ ./bin/civitai models search --version
Error: unknown flag: --version
rc=2
```

Every `generate` invocation on this PR was `--dry-run` only. Nothing was
submitted and nothing was charged.

## #360 — an invocation that actually substitutes

The section's job is to make a substitution recognisable *before* a spend, and
it demonstrated it with a command that 404s locally. Searching for one that
reproduces today found this, run twice with the same shape:

```
$ ./bin/civitai generate "a cat" --ecosystem Flux1Kontext --checkpoint 128713 --dry-run
⚠ The server will NOT use the checkpoint you asked for. It has substituted a different model, …
    requested version 128713 -> will run version 1892509  (reason: unrecognized)
      the server does not offer that version in this model family at all — …
To refuse a run like this instead of being told about it, pass --fail-on-substitution.
…
Checkpoint:       DreamShaper — 8 (Checkpoint, id 128713)  [SUPERSEDED — the server will run version 1892509 instead; see the warning above]
```

A real version id in the *wrong model family* is what substitutes; a
*nonexistent* one is refused by `ResolveModelVersion` before anything is priced.
Both console blocks are now that output verbatim, with two caveats added so a
reader can tell a reproduction from a mismatch: the applied id is server state
and will differ, and `--checkpoint 999999999` is now documented as the local
`404`/exit-4 it really is.

**AGENTS item 13 was read before touching this.** Nothing was added to the
generation path — no check, table, enum or default. The change is documentation
of the live lookup (`ResolveModelVersion`) that item 13 already sanctions.

## #361 — existence → attribution, and the hole that hid it

The index gains a row for the message `app validate` really prints, and the
`is this an App project?` row is repointed at `app listing`, which is what
emits it.

**The existing guard could not see any of this.** It asserts each row's
fragment exists *somewhere* in non-test source; `is this an App project?` does
exist (in `internal/manifest`), so the row's attribution and its link were both
wrong under a green test.

`TestREADMETroubleshootingRowsAreAttributedToTheEmittingCommand` is the
strengthening: for a curated ledger it drives the **real** command through
`NewRootCmd`, and asserts (a) a row exists whose first column quotes the
fragment, (b) the row's link cell names the right anchor **and** that anchor is
a heading that exists, (c) the command really printed the fragment, and (d) for
the confusable pair, that the *other* command did not.

### Red → green matrix

| tree | result |
| --- | --- |
| `README.md` at `origin/main`, test at HEAD | **RED** — 2 failures |
| both at HEAD | **GREEN** — 8 leaf subtests |

The red output, verbatim:

```
--- FAIL: …/not_found_at_project_root
    the Troubleshooting index has no row whose first column quotes "not found at project root".
--- FAIL: …/is_this_an_App_project?
    the row for "is this an App project?" links to "[Validate fidelity](#validate-fidelity)",
    but must point at #listing-media-requirements.
```

### Mutation matrix — each arm broken on purpose and watched fail with THIS guard's message

| mutation | new guard | old existence guard |
| --- | --- | --- |
| `validate.go` finding gains `— is this an App project?` (i.e. the old row made retroactively true) | **RED** at `not_emitted_by_app_validate` | PASS — structurally blind |
| `resolveListingSlug` drops the `%w` cause | **RED** at `emitted_by_app_listing_status` | PASS |
| `validate.go` message reworded to `absent from project root` | **RED** at both `emitted_by` arms | RED (also) |
| `README.md` swapped to `origin/main` | **RED** at both rows | PASS |

The `notEmittedBy` half is labelled in the code as an **invariant guard**, not
regression coverage: `app validate` did not print that string before this PR
either. Row 1 of the mutation table is why it is there.

### A second attribution error, caught by writing the test

The first draft of the row credited the string to `app submit`, `app dev-token`
and `app dev-tunnel` as well — taken from #361's own list of `manifest.Load`
callers. **None of the three prints it for a missing manifest**: `app submit`
validates first and reports the other message, and dev-token/dev-tunnel discard
the load error. Listing a caller is not measuring a message. `app submit
--package-only` is now a `notEmittedBy` row so that stays measured.

## #364 — docs, not code, and why

`--version` is Cobra's, derived from a command's `Version` field. Making the
sentence true in code means either setting `Version` on every subcommand — so
every subcommand's help grows a `--version` flag, and `civitai app validate
--version` prints a version instead of validating — or hand-rolling a
persistent flag with the same effect. That is a behaviour change to a published
contract in order to satisfy a sentence, and `civitai version` (version +
commit + build date) already answers the scripted case from anywhere. So the
sentence moves: the table now says four of the five are global and `--version`
is root-only, with the exact failure a caller gets.

## Gates — counted, not read off an exit code

```
go test ./... -count=1 -v
=== RUN   3468
--- PASS  3464
--- FAIL  0
--- SKIP  4
ok        19 packages   FAIL 0   ("no test files" 1)
grep -c 'panic: test timed out' -> 0
```

`make ci` — green (tidy + vet + test + build).
`gofmt -s -l .` — 0 files, against 353 `.go` files in the tree (positive control
on the walk).
`make lint` via `nix-shell -p golangci-lint` — **v2.12.2**, which matches the
version pinned in `.github/workflows/ci.yml`. **0 issues.** Instrument validated
with a negative control: an unused variable added to the new test file made it
report `declared and not used: x (typecheck)`, and removing it returned it to 0.

`make ci-shallow` was **not** run — this change touches no test that reads git
history (README + two `_test.go` files).

## Overlap note

A parallel PR is fixing #365, which includes four unrelated README nits. This PR
touches four README regions: the silent-model-substitution section, the Global
flags table, the Troubleshooting intro sentence, and the "Validating and
submitting" table. If #365 touches any of those, the overlap should be resolved
between the two PRs rather than by either one unilaterally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1778,26 +1778,43 @@ default checkpoint, runs the job, and bills you for **what actually ran**. Until
 recently the reply was indistinguishable from success — a nonexistent version id
 came back `200 OK` at the default price.
 
-The server now reports each swap, and `civitai generate` surfaces it:
+The server now reports each swap, and `civitai generate` surfaces it. A version
+id that is perfectly real but belongs to a *different* model family is the
+easiest way to see it — here an SD 1.5 checkpoint sent to a Flux ecosystem:
 
 ```console
-$ civitai generate "a cat" --checkpoint 999999999 --dry-run
-⚠ The server will NOT use the checkpoint you asked for. It has substituted a
-  different model, and this estimate prices the SUBSTITUTE. Nothing has
-  been submitted or charged yet.
-    requested version 999999999 -> will run version 2436219  (reason: unrecognized)
-      the server does not offer that version in this model family at all …
+$ civitai generate "a cat" --ecosystem Flux1Kontext --checkpoint 128713 --dry-run
+⚠ The server will NOT use the checkpoint you asked for. It has substituted a different model, and this estimate prices the SUBSTITUTE. Nothing has been submitted or charged yet.
+    requested version 128713 -> will run version 1892509  (reason: unrecognized)
+      the server does not offer that version in this model family at all — it may be a community checkpoint that was never offered for generation, or a version retired since this command was written. Check it with `civitai model-versions get <id>` and pin a version that is still offered
+To refuse a run like this instead of being told about it, pass --fail-on-substitution.
 ```
 
 The checkpoint line in the summary you approve is annotated too, so the model
 that will *not* run is never the last one you read before saying yes:
 
 ```console
-Checkpoint:   DreamShaper — 8 (Checkpoint, id 128713)  [SUPERSEDED — the server will run version 2436219 instead; see the warning above]
+Checkpoint:       DreamShaper — 8 (Checkpoint, id 128713)  [SUPERSEDED — the server will run version 1892509 instead; see the warning above]
 ```
 
 Your own id stays on the line: the CLI marks it, it never quietly substitutes
 the applied one.
+
+Two things about that transcript, so you can tell a *reproduction* from a
+*mismatch*:
+
+- **The applied id is the server's current answer, not a constant.** `1892509`
+  is whatever that family's default was when this was run; yours will differ,
+  and so will the price. What reproduces is the *shape* — a warning naming both
+  ids and a `reason`, and a `[SUPERSEDED …]` note on the checkpoint line.
+- **A version id that does not exist at all will NOT get you here.** `generate`
+  resolves every `--checkpoint` against the public model-version API before it
+  prices anything, so `--checkpoint 999999999` fails locally with
+  `--checkpoint 999999999: not found (404): Model not found` and exit `4` — no
+  estimate, no submit, nothing charged. That is the [deliberate live
+  lookup](#-a-checkpoint-does-not-carry-its-ecosystem) doing its job: it is the
+  *nonexistent* id it closes off, and it cannot tell you whether an id that does
+  exist belongs to the family you are generating with.
 
 It is reported **on the estimate** (`--dry-run`, and before the confirmation
 prompt on a real run — while you can still back out), **again after the submit**
@@ -2270,11 +2287,12 @@ release and prints a one-line notice; `--no-update-check` (or
 
 ## Global flags
 
-These are accepted by **every** command:
+Four of these are accepted by **every** command. `-v` / `--version` is the
+exception — it is **root-only**:
 
 | Flag | What it does |
 | --- | --- |
-| `-v`, `--version` | Print the version and exit. (`civitai version` prints version + commit + build date.) |
+| `-v`, `--version` | **On the root command only.** `civitai --version` prints the version and exits; on a subcommand it is not a flag at all — `civitai app validate --version` fails with `unknown flag: --version` and exits `2`. From a script, use `civitai version` (version + commit + build date), which works from anywhere. |
 | `-h`, `--help` | Help for any command. `civitai --help` also prints the [exit-code](#exit-codes) contract. |
 | `--no-color` | Disable all colour and styling. Also via `NO_COLOR` or `CIVITAI_NO_COLOR`. |
 | `--color` | Force colour **even when stdout is not a TTY**. Also via `CLICOLOR_FORCE` or `CIVITAI_COLOR`. |
@@ -2399,7 +2417,10 @@ fi
 string this CLI really prints, so searching this page for a few words of your
 error should land you on the right row. (A test asserts that each of these
 strings still exists in the source, so the index cannot quietly go stale the way
-a hand-written list does.)
+a hand-written list does — and for the rows most easily confused with one
+another, a second test runs the command the row names and checks it really
+prints that string, because "the string exists somewhere" was green while a row
+credited it to the wrong command.)
 
 ### Credentials and access
 
@@ -2427,7 +2448,8 @@ a hand-written list does.)
 
 | You saw | What it means | Where to read more |
 | --- | --- | --- |
-| `is this an App project?` | There is no `block.manifest.json` at the path you named. The path itself was fine — which is why this exits `1` and not `2`. | [Validate fidelity](#validate-fidelity) |
+| `… not found at project root …` | **`civitai app validate`** (and `app submit`, which validates first) found no `block.manifest.json` in the directory you named — the full line is `block.manifest.json not found at project root <dir>`. The path itself was fine — which is why this exits `1` and not `2`. | [Validate fidelity](#validate-fidelity) |
+| `is this an App project?` | A **different** message with the same cause, from a different command: `civitai app listing …`, which has to work out *which* app you mean from the working directory. `app validate` and `app submit` never print it — they report the row above. Run `app listing` from the app directory, or name the app with `--slug` / `--dir`. | [Listing media requirements](#listing-media-requirements) |
 | `no such directory — pass the path to an App project root` | The path does not exist. This is a **usage** error: exit `2`, and `--json` prints nothing at all. | [Exit codes](#exit-codes) |
 | `is not a directory — pass the App project ROOT` | You pointed at a file — often the manifest itself. Pass the directory holding it. Exit `2`. | [Exit codes](#exit-codes) |
 | `it did NOT check that the file is loaded` | The `BLOCK_READY` advisory on its **weak** tier: it could not resolve what your `index.html` loads, so it only checked whether *some* file mentions the message. The lines that follow name what it could not follow. | [The host handshake](#the-host-handshake-block_ready) |
--- a/README.md
+++ b/README.md
@@ -1232,6 +1232,20 @@ $ civitai app listing add-screenshot ./assets/screenshot-1.png --caption "Grid v
 `assets/` is scaffolded by every template, with a README of the requirements and
 **no placeholder images** — the files above are ones you supply.
 
+**Every `app listing` subcommand has to work out *which* app you mean, and it
+does that from the working directory** — it reads `blockId` out of the
+`block.manifest.json` next to you. Run from somewhere else and there is no
+manifest to read, so it stops before it builds a request:
+
+```text
+$ cd /tmp && civitai app listing status
+Error: could not resolve the app — run this from your app directory (with block.manifest.json) or pass --slug: no block.manifest.json found in . — is this an App project? run `civitai app init` to create one
+```
+
+Two flags name the app instead: **`--slug <blockId>`** skips the manifest
+entirely, and **`--dir <path>`** points at the app directory from wherever you
+are. Both work on every `listing` subcommand.
+
 Details worth knowing before you start:
 
 - **Source images** are png/jpeg/webp and are size-checked **locally before any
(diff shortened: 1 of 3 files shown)


ZacxDev:
## Correction: the PR description above overstates what the new test checked

An adversarial audit of `fc03d33` found the description's central claims about
`TestREADMETroubleshootingRowsAreAttributedToTheEmittingCommand` to be wrong.
The body is left as written so the record is legible; this comment is the
correction. **If you reviewed this PR before now, please re-read this section.**

### What the description claimed

> existence → attribution … asserts (a) a row exists whose first column quotes
> the fragment, (b) the row's link cell names the right anchor **and** that
> anchor is a heading that exists, (c) the command really printed the fragment,
> and (d) for the confusable pair, that the *other* command did not.
>
> A row can only pass by being about the command it claims.

### What it actually pinned

(a), (b), (c) and (d) were all real. **The sentence after them was not.** The
test examined `row[0]` (the fragment) and `row[2]` (the link anchor) — and
**never `row[1]`, the cell that says which command emits the string.** The
`emittedBy` / `notEmittedBy` sets were Go-side labels with no tie to the prose a
reader actually reads for attribution. So a row could pass while being about a
different command than its own text claims — which is #361 verbatim.

Measured, at `fc03d33`, with row 2452's prose rewritten to *"Printed by
`civitai app validate` … `app listing` never prints it"* and the link cell left
untouched:

```
go test ./internal/cmd/ -run 'Attributed|SymptomsExist' -count=1 -v
RUN=9 PASS=9 FAIL=0     # both new tests AND the old existence guard: GREEN
```

The red/green matrix in the description is therefore **misattributed**. It
reddened because `origin/main`'s **link cell** said `#validate-fidelity`;
attribution was never the discriminator. The `RED at not_emitted_by_app_validate`
row of the mutation table is genuine — that arm runs the real command — but it
tests the CLI's behaviour, not the README's sentence.

Also corrected: **"GREEN — 8 leaf subtests"**. There are **5 leaves** under 2 row
subtests; 8 was the total `=== RUN` count including the parent and the two row
nodes.

### The PR's thesis also failed on the PR's own new row

`fc03d33`'s new row 2452 asserts *"`app validate` and `app submit` **never**
print it — they report the row above."* Measured against a binary built from
`fc03d33`:

```
$ ./bin/civitai app submit <empty-dir> --skip-validate --package-only --out b2.zip
Error: no block.manifest.json found in <empty-dir> — is this an App project? run `civitai app init` to create one
rc=1
```

`--skip-validate` waives the validation that would otherwise report the row
above, so `manifest.Load` is reached and prints exactly the string the row says
that command never prints. The flag is documented on this same page, and
`app submit`'s own validation failure hands the reader the falsifier in one hop
("fix before submitting, or pass `--skip-validate`"). The neighbouring row 2451
— "(and `app submit`, which validates first)" — was wrong on that same path for
the same reason. This is the #361 defect class reproduced inside the fix for it.

### What it pins now (`bad829c`)

`attributionProseCheck` parses `row[1]`: the cause cell is split into clauses
(never inside an inline-code span), each clause is classified positive or
negative against a closed set of *printing* negations, and the command names in
each are matched against the ledger **by equality on the normalised span** — so
`app submit` can never stand in for `app submit --skip-validate`, which is the
distinction F1 turns on. Three directions are required:

- every `emittedBy` command is named in a positive clause and no negative one
- every `notEmittedBy` command is named in a negative clause and no positive one
- **every command the cell names at all is in the ledger** — the row cannot carry
  an attribution nothing measured

Plus: both rows now state the flag-space truth (`app submit` prints
`not found at project root`; `app submit --skip-validate` prints `is this an App
project?`; each row says so about the other); the ledger measures all four
invocations; and the row lookup returns a match **count**, so a future row whose
first cell is a superset of an earlier one cannot silently shadow it.

The audit's demonstrated mutant, re-run against the new guard:

```
--- FAIL: .../is_this_an_App_project?/attributes_it_to_the_ledger's_commands
    the row for "is this an App project?" is MEASURED to be printed by `app listing`,
    but its cause cell never says so.
    the row for "is this an App project?" credits `app validate` with printing it,
    and running that command did not.
    the row for "is this an App project?" is measured NOT to be printed by
    `app submit`, but its cause cell does not say so in a negated clause.
```

Four other findings are fixed in the same push — the README-transcript guard was
a whole-file literal search (`--checkpoint 999999998` survived it, reordered
flags survived it, and a *correct* README spelling the literal in prose failed
it); `githubAnchorSlug` dropped `_`, so `#the-host-handshake-block_ready` — a
live link used five times — read as dead; and both rows routed to sections that
do not explain the error. Full matrix in the push's commit messages.

Issue #359, opened 2026-08-10
app pull says the app does not exist; the app status it tells you to run lists it — the real cause is "not approved yet"


ZacxDev:
Re-verified today against `civitai v0.1.92-10-g3156b80` (built from `main` @ `3156b80`).

`civitai app pull` tells you the app does not exist. It does exist — `app status`, the very
command the error tells you to run, lists it. The real cause is that the app is not
**approved** yet, which `app pull --help` states and the error never does.

## Symptom

```
$ civitai app pull ./pulled --app dogfood3-20260810-money
Error: no such app for your account — check the slug with `civitai app status`: App block not found
rc=4
```

```
$ civitai app status | grep dogfood3-20260810-money
dogfood3-20260810-money     0.1.1    withdrawn  -             2026-08-10  -
dogfood3-20260810-money     0.1.0    withdrawn  -             2026-08-10  -
```

The remedy the error names disproves the error. (Measured above on a never-approved app;
the original dogfood observation was on a **pending** app — same precondition, "the app has
no approved version", and the same code path, see below.)

## Why the good message never fires

The CLI *has* the right message. `internal/cmd/app_pull.go:122`:

```go
msg = "git access is not available for this app yet — submit and get your first version approved first."
```

but that branch is only reachable when the server answers `200` with
`ForgejoCloneInfo.NotYetAvailable = true`. The server answers **404** instead, and
`cloneInfoError` in `internal/appapi/appblocks.go:938` maps every 404 to one string with no
way to tell "slug typo" from "app exists, not approved":

```go
case http.StatusNotFound:
    return fmt.Errorf("no such app for your account — check the slug with `civitai app status`: %s", msg)
```

## The README repeats the same false explanation

`README.md:2270`, Troubleshooting:

| `no such app for your account` | The slug matches none of your submissions. List them with `civitai app status`. |

That is the one thing the user has already disproved by the time they read it.

## And `app pull --help` promises the opposite

```
The repo only exists once your FIRST version has been submitted as a ZIP and
approved; before then the command tells you so instead of failing obscurely.
```

It does not tell you so. It fails obscurely.

## Compare — `app metrics` on the identical precondition

This is the house standard and it is excellent:

> has no approved App Block yet — analytics only exist once a submitted version is
> approved; check where it is in review with `civitai app status <slug>`

## Suggested fix

The CLI already knows enough to disambiguate client-side: `app pull` can look the slug up in
the caller's own submissions (the same call `app status` makes) before, or on failure of,
[...]

PR #370, opened 2026-08-11
fix(ux): six of the seven dogfood-3 papercuts — the approval screen said "txt2img" and nothing explained it

ZacxDev:
Addresses **six of the seven** items in #365. **Item 6 is deferred and the issue stays open** — rationale at the bottom. Also fixes the latent nit that issue recorded while refuting the reported one (`truncate()` slicing bytes).

## Per-item re-verification, against current `main`

The issue was filed against a binary built from `3156b80`. `main` is now `9c7baed` — **eight commits later**, including `bfbd4b2` (#357, listing icon messaging) and `cebfa45`/`ca27b53` (`generate` help + README). Each item was re-checked separately against `./bin/civitai` built from `9c7baed` **before** any edit. **None of the seven had been fixed in the interim.**

| # | item | still live at `9c7baed`? | evidence |
| --- | --- | --- | --- |
| 1 | img2img shows `Workflow: txt2img` unexplained | **yes** | `internal/cmd/generate.go` printed the bare `generateWorkflow` constant at both spend surfaces (`Workflow:\t%s` in `printGenerateQuote`; `About to generate with %s` in `confirmGenerate`). Reproduced through the command's own code path with injected seams — `About to generate with txt2img at https://civitai.com.` (see the red matrix below). **Not run against the live server: that would spend money.** |
| 2 | no `--limit` on `app status` | **yes** | `./bin/civitai app status --help` → `Flags: -h/--help, --id, --json`. `app list --help` → `--limit`, `--cursor`. |
| 3 | doubled `✓` | **yes** | `app_listing.go:569, 698, 734` all appended `✓` to a `ui.Success(...)` argument; `ui.Styler.Success` prefixes `"✓ "` at `internal/ui/ui.go:265`. Rendered: `✓ Icon set ✓`. |
| 4 | "square-ish, not exactly square" | **yes** | `README.md:1307` (line moved from 1273). |
| 5 | relative `AGENTS.md` links | **yes, and grown to seven** (issue counted six) | `grep -n 'AGENTS.md' README.md` → 1044, 1351, 1766, 1893, 2483, 2496, 2515. |
| 6 | dev port 5186 pinned, no override | **yes** | scaffolded a real project with `civitai app create` — `vite.config.ts:55-56` `port: 5186 / strictPort: true`; `.env.development:7` `…http://localhost:5186,…`; `package.json` `dev:harness`/`dev:live`/`dev:tunnel` each `--port 5186`. |
| 7 | `.env.production` bundled, undocumented | **yes** | `pkgzip.go` `keptEnvFiles` includes `.env.production`; README documented only the `.env.development*` exclusion. |

## What changed

**1 — the img2img workflow label (display only).** The wire value is untouched: `generateWorkflow` stays `txt2img` because the server promotes it itself (item 19(a)), and sending `img2img:edit` would mean vendoring which ecosystems offer it — item 13's prohibition. What was wrong is that the word sat one line under *"the server IGNORES the images above … and still charges"* with nothing explaining it. One constant `imagePromotionNote` + one helper `workflowLabel(built)`, used by the two surfaces that name the workflow. With no `--image`, both screens are byte-identical to before (pinned by `generate_workflow_label_without_image` and the untouched `generate_confirmation` / `generate_dry_run_*` goldens).

Per **item 28**, the note states the server's **rule** and never claims the promotion **fired** — the CLI cannot observe that, and item 19(b) records why no detector is possible. Rendered:

```
Workflow:  txt2img (image editing is requested AS txt2img plus --image; the server does the promotion, not this CLI)
```

`claudedocs/decisions/19-img2img-workflow-and-upload.md` gains a header recording this (the body is byte-pinned and untouched).

**2 — `app status --limit N`.** `app list`'s shape, but **display-side**: this route accepts no limit and no cursor, so nothing goes on the wire (asserted). Applies to `--json` too. Explicit `--limit 0`/negative → usage error (exit 2) before the network; `--limit` with a `blockId`/`--id` → usage error rather than silent no-op. The API cap caveat is computed from what the **server** returned, *before* the trim.

**3** — dropped the trailing glyph at all three sites. **4** — `0.9 – 1.1 — square or near-square (1:1 is fine)`. **5** — all seven links absolute. **7** — a new README subsection stating the whole dotenv allow-list and why shipping `.env.production` is safe by construction.

**Bonus** — `truncate()` now cuts on rune boundaries. The ellipsis was **never missing**; that claim was refuted in the issue and has not been "restored".

## Red → green matrix

Every guard was watched fail against `origin/main`'s code/prose and pass after.

| guard | red at `origin/main` (9c7baed) | green at HEAD |
| --- | --- | --- |
| `TestGenerateImage_DryRunWorkflowLineExplainsThePromotion` | ✅ ``…`Workflow: txt2img` with --image set is unexplained…`` | ✅ |
| `TestGenerateImage_ConfirmationWorkflowLineExplainsThePromotion` | ✅ `"About to generate with txt2img at https://civitai.com."` | ✅ |
| `TestWorkflowLabelCallSiteLedger` | ✅ (0 of 2 sites) | ✅ |
| `TestGenerate_WorkflowLineIsUnannotatedWithoutImages` | ❌ passes pre-change — **invariant guard**, not regression coverage. It exists so the annotation cannot leak onto a plain txt2img job. | ✅ |
| `TestGoldenSpendCopy/generate_workflow_label_{with,without}_image` | ✅ (no approved copy existed) | ✅ |
| `TestAppStatusLimit{TrimsTheTable,IsDisplaySideOnly,AppliesToJSON,DoesNotSuppressTheCapCaveat,AboveTheListPrintsEverything}` | ✅ ×5 (`unknown flag: --limit`) | ✅ |
| `TestAppStatusLimit{RejectsNonPositive,RefusedOnTheDetailView}` | ⚠️ **passes pre-change for the wrong reason** — cobra rejects the unknown flag with the same `ErrUsage` classification. Mutation-tested instead (below). | ✅ |
| `TestAppListingSuccessLinesCarryOneCheckGlyph` | ✅ `success line "✓ Icon set ✓" carries 2 ✓ glyphs, want exactly 1` | ✅ |
| `TestStyledHelpersAreNotHandedTheirOwnGlyph` | ✅ names all three sites; control reports 26 call sites scanned | ✅ |
| `TestTruncateCutsOnRuneBoundaries` | ✅ 3/3 subtests: `"漢\xe6…"`, `"🐈\xf0\x9f…"`, `"…ü\xc3…"` | ✅ |
| `TestTruncateLeavesShortValuesAlone` | ❌ passes pre-change — **invariant guard** against a rune-aware rewrite that measures the wrong unit. | ✅ |
| `TestREADMEDoesNotLinkContributorDocsRelatively` | ✅ seven findings; control reports 149 relative targets examined | ✅ |
| `TestREADMEIconAspectDoesNotForbidASquareIcon` | ✅ | ✅ |
| `TestREADMEDocumentsEveryDotenvFileThePackagerKeeps` | ✅ (section absent) | ✅ |

Red runs were produced by restoring `origin/main`'s `README.md` / `generate.go` under the new tests, then restoring HEAD's.

### Mutation tests for the two guards that were green for the wrong reason

| mutation | result |
| --- | --- |
| `limit < 1` → `false && limit < 1` | `--- FAIL: TestAppStatusLimitRejectsNonPositive` — `--limit 0 must be refused`, and **only** that test |
| detail-view branch → `false && (…)` | `--- FAIL: TestAppStatusLimitRefusedOnTheDetailView`, and only that test |
| cap caveat → `submissionsListTruncated(len(subs)) && limit == 0` (the "‑‑limit hides the cap" bug) | `--- FAIL: TestAppStatusLimitDoesNotSuppressTheCapCaveat`, both assertions, and only that test |

All three reverted; each failed with **that guard's own message**, not a neighbour's.

## Gate output — counted, not read off an exit code

```
go test ./... -count=1 -v
  === RUN   3484
  --- PASS  3480
  --- FAIL     0
  --- SKIP     4
  ok        19/19 packages,  0 failing
  grep -c 'panic: test timed out'  →  0
```

```
make ci          → PASS (tidy + vet + test + build)
make ci-shallow  → PASSED in a depth-1 clone: ok=19/19  failing-tests=0  failing-packages=0  timeouts=0  go-test-rc=0
gofmt -s -l .    → 0 files
golangci-lint    → 0 issues   (v2.12.2 — matches the CI pin exactly; run via `nix-shell -p golangci-lint`)
```

`make ci` is **not** a superset of CI (it omits lint), hence the separate run. `ci-shallow` was run because this PR touches a `claudedocs/decisions/` file, which `agents_split_preserved_test.go` reads git history for.

## Deferred: item 6 (the 5186 pin), and why

The pin itself is **correct** — the SDK's `IframeTransport` needs a stable dev origin, and the issue says so. The ask is a knob that moves the port, the `VITE_BLOCK_ALLOWED_PARENT_ORIGINS` value and the npm scripts **together**, plus a readable message on a collision. Measured blast radius: 14 files reference `5186`, including four `page-money` templates, the CLI's own `defaultDevTunnelPort`, the vendored Vite dotenv mirror behind item 10's parent-origin check and the `internal/scaffold/dev_embed_contract_test.go` seam guard that pins the CLI's check against the template's literal values, and three CI jobs (`template-page-money`, `template-page-vite`, `ready-ack-runtime`).

That is a feature touching a documented deliberate mirror, not a papercut. Folding it in here would make neither half reviewable. **#365 stays open for it.**

## Notes for the parallel agents

Two other agents are on #360/#361/#364 and #359/#362/#363. I stayed out of the README **Troubleshooting index** and **Global flags** sections, which the docs agent owns. One possible overlap, flagged rather than resolved unilaterally: I edited the `civitai app status` row in the **Command reference** table (added `[--limit N]`). If the docs agent also touches that table, take both edits — they are on different clauses of the same row.

Refs #365 (six of seven items; item 6 outstanding).


ZacxDev:
## Corrections to this PR's description, and the audit fixes (`e880a4a`, `ea588bf`)

An adversarial audit re-measured every gate number in the body and found them accurate. It also found **four claims in the description that are wrong or overstated**, plus eight defects. Correcting the claims here rather than editing the body silently, so the record shows what was believed and what was measured.

### The four overstatements

| claim in the body | what is actually true |
| --- | --- |
| *"all seven links absolute"* | True for the seven `AGENTS.md` links, but **the class was not closed**. `CONTRIBUTING.md` was missing from the guard's allow-list and `README.md:2535` still linked it relatively — an instance of the bug this PR fixes, four lines below one it had just fixed. The guard was also **spelled, not structural**: `./AGENTS.md`, `../AGENTS.md` and reference-style `[x]: AGENTS.md` all passed it. |
| *"control reports 26 call sites scanned"*, presented as package-wide | It is **26 of 61**. The AST walk required the receiver to be the ident `ui`, so it saw only the package-function form. The **styler** forms — `ui.For(w).Warn(…)`, `st := ui.For(w); st.Warn(…)`, and a `ui.Styler` **parameter** (`sty.Warn(…)`, a family neither the PR nor the audit had counted) — are 35 more sites and were structurally invisible. The auditor injected the exact doubled-glyph bug into `ui.For(errw).Warn(…)` at `generate.go:1050` and the guard **passed**, reporting "checked 26 styled-helper call sites", package green. |
| *"a new README subsection stating the **whole** dotenv allow-list"* | It stated the allow-list, not the **catch-all that governs**. The authoritative matcher is *drop every `.env*` except the three kept*; the section enumerated five excluded names and stopped, so a reader holding `.env.staging` concluded it ships. It does not. Worse, the new ledger test was bound to `excludedFilePatterns`, which `pkgzip.go:72-74` explicitly labels *"human-readable … for CLI messaging. The authoritative matcher is `isExcludedFile`"* — so the guard pinned the README against the **non-authoritative** list and could not see a change to the real rule. |
| decision 19's guard inventory implied both real screens were goldened with `--image` | **Neither was.** `generate_workflow_label_{with,without}_image` pin `workflowLabel()`'s return value **in isolation**. No golden pinned either composed screen, which is how the defect below shipped: reviewed output that nothing had reviewed. |

### 🔴 The one that mattered — `README.md:1200-1201` asserted a safety property nothing enforces, about a file this CLI **uploads**

The text argued `.env.production` *"by construction cannot hold a secret"* because Vite inlines every `VITE_`-prefixed variable into the client bundle.

**Vite inlines only the `VITE_`-prefixed ones.** A plain `API_SECRET=…` in that file is not inlined, is not public, and **is packaged and uploaded** — `isExcludedFile` (`internal/pkgzip/pkgzip.go:117-122`) keeps the file by NAME and never reads a byte of it. The repo also contradicts the reasoning from the other side: `.env.development` is excluded *precisely because* it may carry `VITE_LIVE_BLOCK_TOKEN` — a `VITE_`-prefixed value this repo treats as a **spending secret**. Applied consistently, the sentence called that token safe to ship.

The same false claim was in `keptEnvFiles`' doc comment and the package doc; all three are corrected, with the retraction recorded in the source so it is not re-derived. The section now says what is true: **the allow-list is by NAME, nothing checks the contents**, the *scaffolded* file holds one public origin, and do not put a token in **any** of the three kept files — `.env.example` and `.env.sample` included, since `page-money/env.example.tmpl:33-35` literally invites pasting a live `VITE_LIVE_BLOCK_TOKEN` there.

### The other seven

- **`app status --limit N` printed a stderr note contradicting stdout.** Reproduced against a 100-row at-cap response: stdout 5 rows, stderr `note: showing the newest 100 submissions…`. Computing the caveat **before** the trim is the right order — the noun was wrong. Now *"the server returned the newest 100 submissions"*.
- **`truncate`'s second guard was never executed.** `truncate("café", 4)`: `len(s)=5 > 4` passes the byte check, `len(r)=4 <= 4` returns whole — but every test either failed the byte check outright or was far under `max`, so `if false && len(r) <= max` left the package **green** while `truncate("café", 4)` returned `"café…"`. Replaced with a single rune walk (also drops the 4×len `[]rune` allocation on server-supplied descriptions) plus cases inside the `len(s) > max >= len([]rune(s))` window.
- **The promotion note was wedged between the verb and its object on the spend-confirmation screen** — 152 columns, `at https://civitai.com` orphaning onto line 2 at 80 cols, on the screen immediately before an irreversible charge. `workflowLabel` → `workflowLines` returning *(label, note)*; each caller prints the note on its own line. The sentence is now **54 columns**. Added `generate_confirmation_with_image` and `generate_dry_run_with_image_{stdout,stderr}` goldens, a column bound on the framing sentence, and adjacency assertions — goldens are whitespace-collapsed and structurally cannot see a line break, so those are three different properties, not three copies of one.

### Two observable changes this PR made and did not flag

- **`truncate` now counts runes, which is a table-layout change.** `truncate(baseModel, 24)` (`internal/cmd/images.go:246`, a tabwriter column) goes from ~16 to ~48 display cells for CJK; taglines (`apps.go:341`) and descriptions (`apps.go:374`) go 69 → 201 characters for CJK. Intended semantics — but it is wider output for non-Latin values, and nobody said so.
- **`--input` residual.** A hand-written graph carrying an `images` array still prints a bare `Workflow: txt2img` with no note, because the note keys off `built.images` and the `--input` path deliberately never parses that graph. Consistent with *"the CLI does not interpret `--input`"*, now recorded as a residual in decision 19 rather than left implied.

Also fixed: `truncate` is rune-safe but **not** grapheme-safe (a ZWJ sequence cut mid-cluster leaves a dangling joiner — valid UTF-8, no U+FFFD, so both new assertions pass on visually-broken output); the residual is now stated in the comment instead of being over-claimed. The dotenv doc test's hardcoded `filepath.Join("..","..","README.md")` is now `runtime.Caller`-based, and its section slicing stops at `####` as well as `##`/`###`.

### Verification

Every finding was **reproduced before being fixed** — including re-running the auditor's injected `generate.go:1050` mutant, which is now RED with the guard's own message and a `61 sites / 4 receiver families` control.

Mutant battery: **28 mutants, 27 RED**, each failing with *that guard's* message. The one GREEN is `len(s) <= max` → `<` in `truncate`, kept in the battery and labelled **equivalent**: a string whose byte length equals `max` cannot hold more than `max` runes, so the walk returns it whole either way.

Three of those 28 were **surviving mutants against my own audit fixes** (`ea588bf`), which is why there are two commits: a fix round resets the verification gate. They were (a) the README safety claim restorable word-for-word — now pinned whole as reviewed copy, since a banned-phrase list is the shape item 28 records losing five times; (b) dropping the `ui.Styler` typed-declaration branch costing 3 of 61 sites and sailing past `minStyledCalls`; (c) deleting `"CONTRIBUTING.md"` from the doc list reopening the hazard silently. (b) also exposed a real redundancy in my own helper — two branches each covering the other's work, so **neither could be killed**: a guard that looked thorough and was untestable.

Gates, counted from output rather than read off an exit code:

```
go test ./... -count=1 -v   3492 RUN / 3488 PASS / 0 FAIL / 4 SKIP
                            19/19 packages ok, 0 timeouts, 0 panics
make ci                     PASS (19 ok)
./scripts/ci-shallow.sh     PASSED  ok=19/19  failing=0  timeouts=0
gofmt -s -l .               0 files  (positive control: 1 when a misformatted func is injected; 357 .go files exist)
golangci-lint v2.12.2       0 issues (negative control: 4 issues on injected bait — gofmt 1, ineffassign 1, unused 2)
```

`main` was merged in (`3b14f83`) and everything above was re-measured after the merge.

ZacxDev:
### 🔴 Packaging behaviour changed in `8d4f4c5` — read before merging

`internal/pkgzip`'s file matcher was `name == ".env" || strings.HasPrefix(name, ".env.")` — **dot-suffixed only** — while the blockquote this PR added to `README.md` says *"every file whose base name starts with `.env` is excluded"*. Those disagreed on undotted names, and the file that lands in the gap is **`.envrc`**: the direnv convention, which routinely holds exported credentials. An author who read that sentence and ran `civitai app submit` uploaded it to the platform and to a human moderator reviewer.

I **widened the matcher** rather than narrowing the sentence, so `civitai app submit` now excludes more than it did on `main`.

**Measured live**, two binaries built from this tree, one scaffolded `page-money` project carrying a `VITE_LIVE_BLOCK_TOKEN` in `.envrc` and `.env-local`, `--package-only` (no network):

| binary | files in bundle | `.env*` shipped | files containing the token |
| --- | --- | --- | --- |
| pre-fix (`.env.` prefix) | 37 | `.env-local`, `.env.example`, `.env.production`, **`.envrc`** | **`.env-local`, `.envrc`** |
| fixed (`.env` prefix) | 35 | `.env.example`, `.env.production` | *(none)* |

**Contract impact, measured before choosing the direction:** no file tracked in this repo has a `.env`-prefixed base name (`git ls-files`); `scaffold.outputName` emits exactly `.env.development`, `.env.production` and `.env.example`, all `.env.`-dotted; no decision doc in `claudedocs/decisions/` describes the packaging rule. So nothing the CLI itself ships is newly excluded. Residual, accepted and stated in the code: a base name that merely begins with those four bytes for an unrelated reason (`.envoy.yaml`) is dropped too — the failure mode of the other direction is a silent credential upload, this one is a missing file the author can rename.

`excludedFilePatterns` (the list `civitai app submit --help` prints) gains a trailing `.env*` so the published message is the catch-all rather than the five names it enumerates.

PR #371, opened 2026-08-11
fix(errors): four dead ends where the CLI knew the answer and printed the wrong one

### LINKED ISSUE #359: app pull says the app does not exist; the app status it tells you to run lists it — the real cause is "not approved yet"

### LINKED ISSUE #362: app dev-token's read-only remedy block ends with the command that recreates the problem, and repeats itself around the token
Re-verified today by reading `internal/cmd/app_dev_token.go` at `main` @ `3156b80`
(`civitai v0.1.92-10-g3156b80`). The output was **measured live during the dogfood run at
`dfdcc97`**; it was not re-minted today, because minting is a real credential-issuing call —
but every line below is still present and still assembled the same way in the current source,
quoted verbatim with line numbers.

## Symptom

Scenario: `block.manifest.json` declares `ai:write:budgeted` (the `page-money` scaffold does,
out of the box), the user runs `civitai app dev-token <slug> --env` **without** `--spend`, and
their credential *can* spend. Two warnings fire and the token is printed between them:

```
⚠ block.manifest.json declares ai:write:budgeted, but --spend was not passed —
  this token does NOT request budgeted spend.
   The CLI no longer requests budgeted spend implicitly, so `npm run dev:live` will refuse to
   generate with "block lacks ai:write:budgeted scope". Re-mint with --spend to allow REAL Buzz:
     civitai app dev-token <slug> --spend --env >> .env.development.local

VITE_LIVE_BLOCK_TOKEN=<token>                                    <-- stdout

Paste into VITE_LIVE_BLOCK_TOKEN in .env.development.local, then restart `npm run dev:live`.
Short-lived (~4h); never commit it.

⚠ This token is READ-ONLY (no `ai:write:budgeted` scope) — `npm run dev:live` will NOT spend
  Buzz or generate; Generate dead-ends silently.
   Cause: your credential CAN spend, but `ai:write:budgeted` wasn't requested — re-mint with
   `--spend` (or add it to `scopes` in block.manifest.json):
     civitai app dev-token <slug> --spend --env >> .env.development.local
     civitai app dev-token <slug> --env >> .env.development.local   # re-mint, then restart dev:live
```

Three defects in those eight lines.

### 1. The last command shown reproduces the problem

`app_dev_token.go:225`:

```go
remint := "     civitai app dev-token " + slug + " --env >> .env.development.local   # re-mint, then restart dev:live"
```

`remint` is appended to **all three** branches of `readOnlyTokenWarning` (lines 235, 241, 246),
including the `canSpend` branch whose whole point is "you needed `--spend`". So the final,
most-recently-read, `# re-mint, then restart dev:live`-annotated command is the one **without**
`--spend` — i.e. it mints exactly the read-only token the paragraph above it is complaining
about. `remint` is correct for the other two branches (where the fix is the *credential*, and
the re-mint is genuinely the follow-up step); it is wrong for this one.

### 2. The parenthetical remedy is already satisfied

`app_dev_token.go:233`:

```go
"   Cause: your credential CAN spend, but `ai:write:budgeted` wasn't requested — re-mint with `--spend` (or add it to `scopes` in block.manifest.json):\n" +
```

In this scenario the *first* warning fired precisely **because** the manifest already declares
it (`spendFilteredFromManifest`, line 134). Telling the user to add a scope that four lines
earlier the CLI told them it had found and deliberately filtered out is a dead end.

### 3. The same advice prints twice, straddling the token on stdout

`spendFilteredNotice` (line 155, emitted at line 408, **before** the mint) and
`readOnlyTokenWarning` (line 222, emitted at line 439, **after** the token) fire on exactly
the same predicate set in this scenario, and both end with the identical
`civitai app dev-token <slug> --spend --env >> .env.development.local`. The token goes to
stdout at line 419/421 in between, so on a terminal the user reads warning → token → the same
warning again.

## Why it matters

The CLI's own text says the consequence: `npm run dev:live` "will NOT spend Buzz or generate;
Generate dead-ends silently". This block is the only place that dead end is preventable, and
a user who copies the last command out of it stays in it.

## Suggested fix

- Drop `remint` from the `canSpend` branch (the `--spend` line above it already *is* the
  re-mint), or annotate it so the two are not interchangeable.
- Make the "(or add it to `scopes` …)" clause conditional on the manifest **not** already
  declaring the scope — the predicate (`spendFilteredFromManifest`) is right there.
- Suppress `readOnlyTokenWarning` when `spendFilteredNotice` has already fired for the same
  cause, or merge them into one post-token block so the token is not straddled.

---
**Source:** blind **credentialed** dogfood run 3, 2026-08-10 (`claudedocs/handoff-dogfood-3.md`,
finding 5) — an agent given the built binary and a live token, forbidden from reading the CLI
source, `AGENTS.md` or the prior run reports. The output block above was measured at `dfdcc97`
during that run; today's re-verification is a source read of `internal/cmd/app_dev_token.go`
at `3156b80`, not a fresh mint.

### LINKED ISSUE #363: four errors that name no next command, in a tool whose house style always does — including one raw Cobra message and two leaked procedure names
Re-verified today against `civitai v0.1.92-10-g3156b80` (built from `main` @ `3156b80`).

This CLI's house style is that an error names the next command to run, and it is followed
almost everywhere — the lockfile message, the `app metrics` not-approved message, the
`app dev-token --budget` bounds message were all singled out during the run as best-in-class.
These four are the places it is not followed, grouped because the fix is one convention
applied four times.

---

## 1. `app listing status` before submit — three words, no next step

```
$ civitai app create "f6 probe app" --dir ./f6app
$ cd ./f6app
$ civitai app listing status
Error: no such submission
rc=4
```

Same via the other lookup path:

```
$ civitai app listing status --slug f6-probe-app
Error: no such submission
rc=4
```

The string is `internal/appapi/appblocks.go:804`, in `FindSubmission` — a `?blockId=` lookup
for a slug with no submission answers `200` with an empty list, and the client turns that into
a bare `no such submission`.

Note the listing router **already has** the good message for the adjacent case
(`internal/appapi/listing.go:439`):

> `no such store listing found for this app (404): … — a store listing is created when you run`
> `civitai app submit` and is settable while the app is pending review; submit the app first,
> then these commands will work

That is what a user gets once a submission exists but a listing does not. The
no-submission-at-all case never reaches it, and gets three words instead.

**Fix:** `— run 'civitai app submit' first; the draft listing is created at submit time`, or
route this case through the message `listingError` already ships.

The same bare string is also what `app status <unknown-slug>` returns:

```
$ civitai app status no-such-slug-xyz
Error: no such submission
rc=4
```

---

## 2. `app pull`'s argument shape is undiscoverable, and one attempt hits raw Cobra

Every other `app` subcommand takes the slug positionally. `app pull` takes it as a **required
`--app` flag** and uses the positional for the *directory*. Three natural attempts:

```
$ civitai app pull dogfood3-20260810-money
Error: required flag(s) "app" not set
rc=2

$ civitai app pull dogfood3-20260810-money ./pulled
Error: accepts at most 1 arg(s), received 2
rc=2

$ civitai app pull
Error: required flag(s) "app" not set
rc=2
```

None of the three mentions `--app`, and the middle one is a bare framework error — the only
unguided Cobra message found anywhere in the tool during the run. `required flag(s) "app" not
set` is *technically* the name, but a user who typed the slug positionally has no way to learn
that the slug goes in the flag and the *directory* goes in the argument.

**Fix:** a `RunE`-side check that, when a positional argument is present and `--app` is not,
says so explicitly — e.g. `--app is required, and the positional argument is the DIRECTORY, not
the slug: civitai app pull [dir] --app <slug>`. Custom `Args` would also catch the 2-arg case.

---

## 3. Internal procedure names leak into user-facing errors

`workflows get` with an id the server does not recognise (verified live today):

```
$ civitai workflows get not-a-real-id-xyz
Error: no such generation procedure or resource (404): Not Found
rc=4
```

"no such generation **procedure**" reads as "the CLI is calling an endpoint that does not
exist" — i.e. the tool is broken — rather than "the id you typed is not one of your
workflows". Source: `internal/genapi/errors.go:116`.

The listing router does the same with the tRPC method name. `internal/appapi/listing.go:442`:

```go
name := trpcName(path)          // e.g. "appListings.setIcon"
…
case http.StatusBadRequest:
    return fmt.Errorf("%s rejected the request (400): %s", name, msg)
```

which surfaced during the run as `appListings.setIcon rejected the request (400)`. (That one
is source-verified only — reproducing it means mutating a real store listing, which this
re-verification did not do; it was measured at `dfdcc97` during the original run.)

**Fix:** map the transport name to the user's noun. `no such workflow — check the id with
'civitai workflows list'` for the first; `the server rejected the icon (400): <msg>` for the
second. The `default:` arms can keep the procedure name — a genuinely unexpected status is
where it helps.

---

## 4. `app metrics` with no argument is a bare Cobra error

```
$ civitai app metrics
Error: accepts 1 arg(s), received 0
rc=2
```

`app dev-token` and `app dev-tunnel` in the same situation are excellent. `app metrics` should
name what the missing argument is (`civitai app metrics <slug>` — list your apps with
`civitai app status`).

---

## Why this matters as a group

An error that names its next command is what made this CLI navigable for a first-time user
during the run; the four places above are exactly where a first-time user stalls. Three of
them (`listing status`, `app pull`, `app metrics`) are the *first* command a user reaches for
in their respective flow.

---
**Source:** blind **credentialed** dogfood run 3, 2026-08-10 (`claudedocs/handoff-dogfood-3.md`,
findings 6, 7, 11, 12) — an agent given the built binary and a live token, forbidden from
reading the CLI source, `AGENTS.md` or the prior run reports. Every command block above was
re-run against current `main` before filing, except `appListings.setIcon` as noted.

ZacxDev:
Closes #359. Closes #362. Closes #363.

Three dogfood-run-3 findings that share one shape: an error that names no next
command, or names a wrong one, in a tool whose house style always names the
right one.

## Step one — re-verification against current `main` (`9c7baed`, 8 commits past the filing)

Built `./bin/civitai` (`v0.1.92-18-g9c7baed`) from `main` and re-checked each
claim. **Nothing was already fixed.** Evidence:

```
$ ./bin/civitai app metrics
Error: accepts 1 arg(s), received 0            rc=2      # #363 §4 — still raw cobra
$ ./bin/civitai app pull some-slug
Error: required flag(s) "app" not set          rc=2      # #363 §2 — never says the positional is the dir
$ ./bin/civitai app pull some-slug ./pulled
Error: accepts at most 1 arg(s), received 2    rc=2      # #363 §2 — still raw cobra
```

Source-verified at `9c7baed` (unchanged since filing):

- `internal/appapi/appblocks.go` — the `cloneInfoError` 404 arm is still
  unconditional, so `internal/cmd/app_pull.go`'s `NotYetAvailable` branch is
  still unreachable for a not-approved app (**#359**);
- `internal/appapi/appblocks.go` — `GetSubmission`'s empty-list arm still returns
  the bare `no such submission` (**#363 §1**);
- `internal/appapi/listing.go` — the 400 arm still leads with `trpcName(path)`
  (**#363 §3**);
- `internal/genapi/errors.go` — the 404 arm still says
  `no such generation procedure or resource` (**#363 §3**);
- `internal/cmd/app_dev_token.go` — `remint` is still appended to all three
  `readOnlyTokenWarning` branches, the manifest parenthetical is still
  unconditional, and `spendFilteredNotice` + `readOnlyTokenWarning` still both
  fire on the same predicate set (**#362**).

**#359's runtime repro was NOT performed against the live API** — it needs an app
with a pending, never-approved submission, which cannot be reached without
submitting into a human moderator queue. It was reproduced instead at the
**binary** level against a stand-in server that answers the clone-info query 404
and the submissions route with NULL-`appBlockId` rows (exactly the state a
pending app is in):

```
$ CIVITAI_BASE_URL=<stand-in> ./bin/civitai app pull ./pulled --app dogfood3-20260810-money
# before: Error: no such app for your account — check the slug with `civitai app status`: App block not found
# after:  Error: app "dogfood3-20260810-money" has no approved version yet — `civitai app pull` clones the
#         repo that only exists once a submitted version is approved (latest submission: 0.1.1 pending);
#         check where it is in review with `civitai app status dogfood3-20260810-money`
rc=4 (unchanged)
```

## What changed

**#359 — `app pull` denied an app `app status` lists.** The server's 404 is
unconditional for both "no such app" and "app exists, nothing approved", so
`cloneInfoError` cannot discriminate and the good message at
`app_pull.go` was dead code for every not-approved app. `explainMissingApp` now
disambiguates client-side using the caller's own submissions — the same route
`app status` reads — and only when *all* of these hold: the error is
`civitai.ErrNotFound`, the lookup succeeds, it returns rows, and none carries an
`appBlockId`. Every other case falls back to the server's own answer verbatim.
The classification is re-applied, so the exit code stays 4 (item 7). The README
Troubleshooting row that repeated the same false explanation moves with it, plus
a new row for the message that replaces it.

**#362 — the remedy block that recreated the problem.** `remint` (no `--spend`)
is dropped from the credential-CAN-spend branch and kept in the two branches
where the fix genuinely is the credential; the "(or add it to `scopes` …)"
clause is conditional on the manifest not already declaring the scope; and the
post-token `readOnlyTokenWarning` is suppressed for the ONE overlap
(`spendFiltered && canSpend`) where it repeats the pre-mint notice's cause and
remedy around the token. The declaring-manifest predicate is now one function
(`manifestDeclaresBudgetedSpend`) read by both sites.

**#363 — four errors that name no next command.** `no such submission` (both the
empty-list and the 404 spelling) names `civitai app submit` and says the draft
store listing is created there; `app metrics` with no argument names the slug and
`app status`; `app pull` names its two slots when the slug is typed positionally;
and the two leaked internal procedure names are gone from the 400/404 arms a
user's own input causes. The `default:` arms keep the procedure name — an
unexpected status is where it earns its place, and that is pinned by a negative
control.

`app pull`'s help and the README section it mirrors are updated in lockstep.

## Red → green matrix

Every guard was watched fail on pre-change code. Two techniques: for #359/#363
the source files were reverted to `origin/main` with the tests kept; for #362 the
fix was mutated back defect-by-defect (a signature change makes a whole-file
revert a compile error, not a behavioural red).

| Guard | Pre-change state | Result | At HEAD |
| --- | --- | --- | --- |
| `TestAppPullNotApprovedSaysSoInsteadOfDenyingTheApp` | `app_pull.go` @ origin/main | **FAIL** | PASS |
| `TestNoSubmissionNamesSubmit` (2 subtests) | `appblocks.go` @ origin/main | **FAIL** (both) | PASS |
| `TestGetSubmissionEmptyListIsNotFound` | `appblocks.go` @ origin/main | **FAIL** | PASS |
| `TestAppMetricsNoArgNamesTheSlug` | `app_metrics.go` @ origin/main | **FAIL** | PASS |
| `TestAppMetricsTooManyArgsStillRefuses` | `app_metrics.go` @ origin/main | **FAIL** | PASS |
| `TestAppPullPositionalSlugNamesTheTwoSlots` (2 subtests) | `app_pull.go` @ origin/main | **FAIL** (both) | PASS |
| `TestListingBadRequestDoesNotLeakTheProcedureName` | `listing.go` @ origin/main | **FAIL** | PASS |
| `TestNotFoundNamesTheWorkflowNotTheProcedure` (2 subtests) | old 404 wording restored | **FAIL** (both) | PASS |
| `TestNotFoundOnAFixedRouteIsNotBlamedOnTheUser` | old 404 wording restored | **FAIL** | PASS |
| `TestWorkflowIDProcsCoverEveryIDCarryingCallSite` | `cancelWorkflow` renamed in the map | **FAIL** with its own message (`workflowIDProcs["cancelWorkflow"] = false, want true`) | PASS |
| `TestReadOnlyWarningCanSpendSuggestsOnlySpendingRemints` | `remint` re-appended to the canSpend branch | **FAIL** | PASS |
| `TestReadOnlyWarningDropsSatisfiedManifestAdvice` | parenthetical made unconditional | **FAIL** | PASS |
| `TestReadOnlyTokenWarning/…manifest already declares…` | parenthetical made unconditional | **FAIL** | PASS |
| `TestDevTokenSaysTheSameThingOnceAroundTheToken` | suppression removed | **FAIL** | PASS |

**Negative controls** — these are invariant guards, NOT regression coverage, and
they were checked to stay GREEN under each mutation above so no guard is a
blanket:

- `TestAppPullUnknownSlugKeepsTheServerAnswer`, `…ApprovedElsewhere…`,
  `…DisambiguationFailure…`, `TestAppPullForbiddenIsNotDisambiguated` — the four
  ways the #359 disambiguation must decline to fire.
- `TestAppPullWithDirAndFlagIsAccepted` — the correct invocation still reaches
  `RunE`.
- `TestListingUnexpectedStatusKeepsTheProcedureName` — the `default:` arm keeps
  the tRPC method name (asserted via `trpcName`, so it is structural, not a
  hard-coded spelling).
- `TestNotFoundOnAFixedRouteIsNotBlamedOnTheUser` — doubles as the control that
  the workflow wording is not applied to routes the user cannot address.
- `TestReadOnlyWarningCredentialBranchesKeepThePlainRemint` and
  `TestDevTokenReadOnlyWarningSurvivesWhenTheCauseDiffers` — the plain re-mint
  and the whole warning survive where they carry information the pre-mint notice
  does not.

Two guards are deliberately **structural rather than spelled**:
`devTokenCommandLines` extracts every suggested `civitai app dev-token …` line
and requires each to contain `--spend` (a `Contains(s, "--spend")` check would be
satisfied by the sentence above the bad command), and the listing control asserts
against `trpcName(trpcSetIcon)` rather than the literal `appListings.setIcon`.

`TestWorkflowIDProcsCoverEveryIDCarryingCallSite` is a ledger over the package's
own `generateError` call sites: it scans the non-test sources, asserts the set
matches a written-out expectation (failing when it GROWS or SHRINKS), and carries
a positive control that the scan found something — an empty match set would
otherwise satisfy every assertion in it.

## Gates

Counted from the output, not read off an exit code:

```
$ go test ./... -count=1 -v
RUN: 3488   PASS: 3484   FAIL: 0   SKIP: 4
packages ok: 19   packages FAIL: 0   "panic: test timed out": 0
```

`make ci` — green (tidy + vet + test + build). `gofmt -s -l .` — clean.

`make lint` runs golangci-lint, which is not on this host's PATH, so it was run
as `nix-shell -p golangci-lint --run 'golangci-lint run ./...'`:

```
golangci-lint has version 2.12.2 built with go1.26.5 from v2.12.2
0 issues.
```

**Version matches the CI pin (v2.12.2) exactly.** The `0` was validated with a
negative control — a deliberate `fmt.Sprintf` with no directives in
`internal/cmd` produced `S1039 (staticcheck)` + `unused`, `2 issues`, and was
then removed. So the zero is a fact about the code, not about a linter wired to
nothing.

## AGENTS.md triggers consulted

- **Item 7** (exit codes pinned by `errors.Is`) — every changed message is a
  TEXT change only; each replacement re-applies its classification and the new
  guards assert `civitai.ErrNotFound` / `cmd.ErrUsage`, never the string.
- **Item 13** (the generation path mirrors nothing) — read before touching
  `internal/genapi`. `workflowIDProcs` is a ledger of THIS package's own
  `generateError` call-site labels, not a vendored claim about the server's
  routes or graph; no check, enum, cost or default was added, and the 404 arm
  still reports whatever the server said.
- **Item 26** (project-path classification) — checked and not touched: the new
  `app pull` / `app metrics` validators gate ARGUMENT SHAPE, not a project path,
  and route through the same `enforceUsageExitCodes` tagging as before.
- **Item 5** (`app metrics`' two-hop) — untouched; only the `Args` validator
  changed.

## Overlap with the parallel agents

- `internal/appapi/listing.go` — I changed **only the 400 arm's wording**. #365
  is also described as touching listing messages (the doubled ✓ glyph); flagging
  rather than resolving.
- `README.md` — I touched the `app pull` section and three Troubleshooting rows.
  #360/#361/#364 are README changes elsewhere in the same file; flagging rather
  than resolving.
- `internal/cmd/exit_classify_test.go` — one existing row was split in two
  (`app pull` with and without a positional) because the positional case now has
  a better message. Its intent (usage-tagged, exit 2) is unchanged.


ZacxDev:
## Correction to the PR body after the adversarial audit (`4d69421`)

The audit's verdict was "merge after fixing 🔴 #1". Six findings are fixed in
`4d69421`; the body above is **stale in the places listed below**. Posting rather
than editing the body so the record of what was claimed when stays intact.

### What the body overstates

1. **The README sentence the body ships was FALSE.** "It says *no such app for
   your account* only when the slug really matches none of your submissions" —
   measured false in three reachable states against a local stand-in API:
   an approved-elsewhere row (`appBlockId:"ab_1"`, clone-info still 404s), a
   submissions lookup that itself fails (500), and `--app <appBlockId>`, which
   `app_pull.go`'s own flag help advertises and which the slug-keyed
   `ListSubmissions` cannot match. The first of those is the #359 complaint this
   PR fixes, still shipped in another state. Fixed: the section and both
   Troubleshooting rows now say the fallback is what you get **whenever the CLI
   cannot prove otherwise**, and enumerate the four states.

2. **Four message texts in the "What changed" section are now different**, so
   quote the code rather than the body:
   - `no such submission (404)` no longer names `civitai app submit` — only the
     `?id=` spelling reaches that arm, where "you have not submitted" is wrong;
     it now says to check the publish-request id. The **slug** spelling
     (empty-list arm) is unchanged.
   - the listing 400 subject follows the OPERATION (read / change / upload
     mint), not the status — `listingError` also serves `trpcQuery` and
     `MintImageUpload`, and "the server rejected this store-listing change" is
     false for both.
   - the generation 404s dropped "nothing you passed can cause this" (a claim
     about server semantics that `--input`'s unchecked user graph contradicts —
     item 13) and dropped "check the id" (also told to `generate`'s status poll,
     whose id the CLI minted, after the charge moved).
   - `app dev-token`'s surviving notice is now emitted AFTER the mint. The
     suppression left it as the only advice, and a mint that renames the slug
     made that advice name a slug `block.manifest.json` no longer contains —
     copying it re-collides, i.e. the remedy did not leave the state it
     complains about.

3. **The red→green matrix is a claim about `949a399`, not about HEAD.** Every
   row in it still holds; it is simply no longer the whole matrix. The new one
   is in the commit message of `4d69421`: 14 mutations, each watched red with
   the target guard's own message and with neighbouring controls green,
   including an ordering-only mutation for the dev-token fix and a
   call-site-only mutation separating the listing read guard from its table.
   The previously-surviving `submissionSubject` precedence mutant is now killed,
   pinned against the client's own parse precedence rather than an asserted
   order.

4. **Two behaviour changes are not mentioned in the body at all**, because they
   came from the audit: a bare `civitai app pull` no longer falls through to
   cobra's `required flag(s) "app" not set` (the exit-code table row for it
   moved with it — still exit 2), and with `--app` set the two-positional error
   drops the "the app goes in `--app`" non sequitur.

### Deliberately NOT done, so the scope is on the record

- **The proc-name de-leak is scoped to the 400/404 arms, not a completed
  sweep.** `generateError`'s 401 and 403 arms still interpolate `proc`
  (`not authenticated for generateFromGraph (401)`), and 403 does cover user
  state (muted account, onboarding incomplete). Left alone: on those statuses
  the proc names which generation step was refused, which is diagnostic rather
  than misleading. If that reads as a defect it is a separate change.

- **No exit code moved.** `app metrics <slug>` (exit 1) and `app pull --app
  <slug>` (exit 4) still answer the same precondition differently. Item 7 makes
  changing one a published-contract change; the README now **documents** the
  pull row's exit code instead, which was the gap.

- **The `--app <appBlockId>` disambiguation gap is documented, not closed.**
  `ListSubmissions` filters on `blockId`, which IS the slug, so an appBlockId
  matches no row. That costs nothing: the server nulls `appBlockId` until a
  version is APPROVED, so a never-approved app has no appBlockId for the user to
  type. Pinned by `TestAppPullAppBlockIDSelectorKeepsTheServerAnswer`, labelled
  in-test as an **invariant guard, not regression coverage**.

### Gates at `4d69421`

Counted from output, not from an exit code:

```
go test ./... -count=1 -v
RUN: 3506   PASS: 3502   FAIL: 0   SKIP: 4
packages ok: 19   packages FAIL: 0   "panic: test timed out": 0
```

`make ci` green. `gofmt -s -l .` clean, over 358 `.go` files found (positive
control on the file count, not just the empty output).
`nix-shell -p golangci-lint --run 'golangci-lint run ./...'` →
`golangci-lint has version 2.12.2 … from v2.12.2` (matches the CI pin) →
`0 issues.`, validated by a negative control (a deliberate unused function in
`internal/appapi` produced `1 issues`, then removed).

### Not verified

Everything here was measured against a **local stand-in server**, never
civitai.com — the not-approved and slug-collision states need a real moderator
queue and a real foreign-owned slug. No `app submit`, no live `generate`.

Issue #351, opened 2026-08-10
generate --ecosystem NanoBanana --image fails on the DEFAULT checkpoint (server-side edit path); works with --checkpoint 2725610 and on Flux1Kontext


ZacxDev:
## Summary

`generate --ecosystem NanoBanana --image <ref>` fails instantly and produces nothing. The same ecosystem **succeeds** with no `--image`. `--help` explicitly lists NanoBanana among the ecosystems that accept reference images, so the documented capability does not work.

## Controlled evidence — three runs, one variable

| # | Invocation | Result |
|---|---|---|
| 1 | `--ecosystem NanoBanana --image ref.png --aspect-ratio 16:9 --negative-prompt "…"` | `status failed` in **0.4s**, no output |
| 2 | `--ecosystem NanoBanana --image ref.png` (nothing else) | `status failed`, no output |
| 3 | `--ecosystem NanoBanana` (**no image**) | **succeeded**, image delivered |

Run 2 strips the negative prompt and the aspect ratio, so neither is the cause. Run 3 differs from run 2 by the reference image alone. The reference is a 1600×900 PNG, well inside the documented limits, and only one image was passed (the CLI warns above 1).

Timing on the failures: `Started 22:23:09.77 → Completed 22:23:10.14`. A 0.4-second turnaround is an immediate rejection, not an attempted generation.

## The request is well-formed

`--print-input` for the failing case:

```json
{
  "workflow": "txt2img",
  "ecosystem": "NanoBanana",
  "prompt": "Re-render this exact composition as a premium studio-grade marketing banner.",
  "negativePrompt": "text, letters",
  "aspectRatio": "16:9",
  "images": [
    { "url": "https://orchestration-new.civitai.com/v2/consumer/blobs/…png?sig=…",
      "width": 1600, "height": 900 }
  ]
}
```

The blob uploads fine and carries width/height as the server requires.

## Why this is expensive to discover

`generate --help` says:

> only SOME ecosystems accept reference images at all (Qwen, Flux1Kontext, **NanoBanana**, Seedream, OpenAI, Grok and a few more do; the SD family and the default do not)

So the documentation names NanoBanana as supported. `--dry-run` then returns `Generatable: true` and prices it at 60 Buzz — the same behaviour as #279. A user following the help and trusting the preflight submits and gets nothing.

At 60 Buzz/image this is 7.5× the cost of a Z-Image generation, so a batch of five reference-guided covers is a 300 Buzz bet on a documented path that cannot work.

[...]

Issue #331, opened 2026-08-10
generate: the failure message promises `civitai workflows get <id>` will diagnose, and it cannot (dogfood F13)


ZacxDev:
**Source:** §A finding **F13** of the blind, credentialed dogfood run of 2026-08-07
(`datapacket-talos/claudedocs/cli-dogfood-findings-2026-08-07.md`). Triage classification:
**IMPLEMENT (copy-only), but it lands on spend-surface goldens and overlaps #307.**

## The finding

> The failure message names a diagnostic command that cannot diagnose.
> Severity Low — dead-end guidance. Status CONFIRMED.

`internal/cmd/generate.go` directs the user to `civitai workflows get <id>`. The run followed
that advice on a failed job:

```
Status:       failed
Outputs (0 deliverable, 1 excluded)
    - 414f4e41-…-0.jpg: not available (the job finished without producing a usable file)
```

No reason, no error class, nothing actionable. *"The user's only remaining strategy is what this
run actually did: spend 8 Buzz per hypothesis until a pattern emerges. That cost 8 attempts
here."*

The reporter is explicit that this is **partly a server-side gap** — the orchestrator supplies
no failure reason — and that the CLI-side fix is to stop promising the command will help:

> *"inspect it with `civitai workflows get <id>` (note the orchestrator often supplies no
> failure reason)."*

## The two sites

Both in `internal/cmd/generate.go`:

- the non-succeeded terminal-status error (*"the generation finished with status %q and
  produced no usable result…"*);
- the succeeded-but-zero-deliverables error (*"the generation succeeded but produced no
  deliverable outputs — it was charged; inspect it with…"*).

## Doctrine check

The change is **copy only** and takes nothing away: the workflow id and the re-attach command
must stay, because they are the user's only handle on a job they have already paid for. It adds
no claim about the charge in either direction, so
[item 28](../blob/main/claudedocs/decisions/28-no-claims-about-unobservable-spend.md) is
satisfied — it *reduces* an over-promise rather than adding one.

## ⚠ Golden collision

Both messages are pinned by the spend-surface golden set:
`internal/cmd/testdata/golden/terminal_status_failed.txt`,
`terminal_status_canceled.txt`, `terminal_status_expired.txt` and
`succeeded_no_deliverables.txt`. Item 28 records that golden-output pinning of every spend
surface is the guard that survived after two phrase-ban guards were defeated, so these must be
regenerated **deliberately** (`-update`) with the diff read, not refreshed reflexively.

[...]

PR #372, opened 2026-08-11
fix(generate): the orchestrator recorded WHY the run failed and the CLI threw it away at unmarshal (#367)

### LINKED ISSUE #367: Every failure reports the same generic reason: the server records output.errors and the CLI drops it at unmarshal
## Summary

When a generation fails, the server records **why** on the step, and the CLI throws it away. Every failure therefore reports the same generic sentence, no matter what actually went wrong.

## What the user sees

```
$ civitai workflows get <id>
Status:       failed

Outputs (0 deliverable, 1 excluded)
⚠ 1 output(s) will NOT be saved — they are not deliverable results:
    - <uuid>-0.jpg: not available (the job finished without producing a usable file)
```

That is the whole diagnosis, for every failure mode: a bad checkpoint, an unavailable resource, a provider-side rejection, a reference image that could not be fetched. All indistinguishable.

## Where it is lost

The step payload carries `output.errors[]`. `ListedStep` in `internal/genapi/workflows.go` has no field for it:

```go
type ListedStep struct {
    Type     string       `json:"$type"`
    Name     string       `json:"name"`
    Status   string       `json:"status"`
    Metadata stepMetadata `json:"metadata"`
    Output   []Blob       `json:"output"`
    Images   []Blob       `json:"images"`
}
```

No `Errors` field, so the array is silently discarded at unmarshal — the classic case where the data arrived, nothing errored, and the information simply evaporated. `grep` across `internal/genapi/` finds no reader for it anywhere.

## Why this is worth fixing

Concretely: #351 took several hours and a set of controlled generations to root-cause. The server had already recorded the reason — a reference-image fetch failing — on the very step the CLI printed. Surfacing that one string would have made it a one-minute diagnosis.

It also compounds the other two open reports. #279 (`Generatable: true`, then charged and failed) and #351 both present as "it just failed", precisely because this field is dropped. The estimate can't warn you, and then the post-mortem can't tell you either.

## Suggested fix

Add the field and print it when a step is not `succeeded`:

```go
type ListedStep struct {
    // …
    Output []Blob   `json:"output"`
    Images []Blob   `json:"images"`
    Errors []string `json:"errors,omitempty"` // shape to be confirmed against a live payload
}
```

Then in the failure branch, prefer the server's reason over the generic sentence, falling back to the current wording when the array is empty:

```
Error: the generation finished with status "failed": <server reason>
```

Two notes for whoever picks this up:

- **Confirm the wire shape against a real failed workflow** before pinning a struct — I inferred `errors[]` from the orchestration side, not from a captured CLI-visible payload, so the exact nesting and element type should be checked rather than trusted from this issue.
- Worth a test that asserts a **non-generic** message for a failed step carrying a reason. A test that only asserts "an error was returned" would pass today.

ZacxDev:
Fixes #367. Related: #339 / #331 (the caveat this reconciles with), #351.

## The defect

When a generation fails, the orchestrator records **why** on the step. `stepOutput` had no field for it, so the reason was discarded at unmarshal and every failure printed the same generic sentence regardless of cause:

```
the generation finished with status "failed" and produced no usable result; … Inspect the run
with `civitai workflows get wf_123` (the orchestrator often supplies no failure reason, so it
may not say why)
```

`civitai workflows get` — the command that error sends you to — printed `Status: failed` and nothing else.

## The wire shape

`errors` is a **`[]string` nested inside `output`**, a sibling of `images`. Not a field on the step, not an array of objects:

```jsonc
{ "steps": [ {
    "$type": "imageGen", "status": "failed",
    "output": {
      "images": [ { "id": "…", "available": false } ],
      "errors": [ "Could not generate images with the given prompts and images. …" ]
    }
} ] }
```

Confirmed against **8 real workflows** before any code was written. The key is always present; the array is what varies:

| workflows | `output.errors` |
|---|---|
| 5 × `failed` | populated |
| 1 × `failed` | **empty** |
| 1 × `canceled`, 1 × `succeeded` | empty |

All five populated samples were the same reproduction, so they prove the field is populated and CLI-visible — **not** that the wording varies by cause. Nothing built here depends on the wording.

🔴 **`ListedStep` (`internal/genapi/workflows.go`) is untouched.** That is the normalized listing shape behind `workflows list`, where `output` is an array. Whether it also carries `errors` is unverified and out of scope; the original issue text names it by mistake.

## The #339 reconciliation — why the caveat became conditional rather than deleted

`noFailureReasonNote` shipped in `efc5dcf` (#331) from a dogfood observation that followed `workflows get` on a failed job and saw no reason. That observation was made **through the struct that drops the field**, so it could not distinguish *"the server sent nothing"* from *"the CLI discarded it"*. The measurement above says the server did send one in 5 of 6 failures.

But it is **not** deleted, because the empty case is real and measured — 1 of 6 failed workflows carried `"errors": []`. So:

- `output.errors` non-empty → the reason is printed and the caveat is **gone** (it is now false for that case).
- `output.errors` empty/absent → **byte-identical to today**, caveat included.

The constant's comment records this, and `generate_dead_end_copy_test.go` gained a header saying its fixtures cover only the empty branch and must not be made unconditional.

## What changed

**`internal/genapi/status.go`**
- `stepOutput.Errors []string` (`json:"errors"`) — the new field, and the only new field.
- `Step.failureReasons()` / `Workflow.FailureReasons()` / `Workflow.FailureReasonText()`.
- `Output.StepErrors`, folded in by `Outputs()`.
- `ExclusionReason` prefers the server's account in its `!Available` branch: `not available (the server reported: …)`. The `blocked`/`hidden` branches are unchanged — they already carry their own specific cause.

**`internal/cmd/generate.go`** — one helper, `serverReasonSuffix(wf)`, selects the tail of **both** dead-end errors (one rule, one place). New shape when a reason exists:

```
the generation finished with status "failed" and produced no usable result; <fate>.
Inspect the run with `civitai workflows get wf_123`. The server reported: <reason>
```

**`internal/cmd/workflows.go`** — `workflows get` renders a `The server reported:` block, one line per reason, outside the tabwriter (safeTerm keeps newlines, which would re-align the columns above).

**`README.md`** — the `workflows get` command-reference row, and two Troubleshooting rows.

### Constraints honoured

- **Item 13** — the string is passed through **verbatim**. Not classified, not mapped to a table, no enumeration of known messages, no wording match. The only transformations are `TrimSpace`, dropping blanks, and dropping byte-identical repeats.
- **Item 28** — the reason explains the *outcome*, never the charge. The fate sentence, `buzzLedgerUnknownNote` and `workflowSettlementPrintedNote` are untouched; the new test re-runs `refundDirectionalClaims` over the new copy, and the new rendering is golden-pinned.
- **Item 7** — both dead-end errors are asserted to carry no classification sentinel (`ErrBadRequest` / `ErrUnauthorized` / `ErrNotFound` / `ErrRateLimited` / `ErrNetwork` / `cmd.ErrUsage`), so the published generic exit code does not move.
- **Untrusted text** — `safeTerm` (strips C0/C1 controls and DEL, keeps `\n`/`\t`) is applied at every terminal-bound site, and is mutation-tested.
- **Multiple steps** — every step's reasons are collected in step order and joined with `; `. Only blanks and byte-identical repeats collapse (a four-step run that died the same way would otherwise print one string four times). Documented on `FailureReasons`; pinned by two tests.

## Golden diff

🔴 **No existing golden file changed.** `git status` after `-update` showed exactly one untracked file, and running the suite against `origin/main`'s production code with this branch's tests in place showed **23/24 golden subtests PASS and only `terminal_status_failed_with_reason` FAIL** — direct evidence that the empty-reason branch renders byte-for-byte as before, so the four goldens #339 re-approved did not need re-approving again. Nothing was removed.

Added — `internal/cmd/testdata/golden/terminal_status_failed_with_reason.txt`:

```
the generation finished with status "failed" and produced no usable result; this CLI cannot see
your Buzz ledger — `civitai buzz` reports a balance, not a history, so settle it against your
Buzz transaction history (/user/transactions). Inspect the run with `civitai workflows get
wf_123`. The server reported: the model refused the request for FIXTURE reasons
```

This is a **second rendering of an already-pinned surface**, which is the hole the golden file's own header describes: without it, a sentence could be added to the reason branch while every pinned rendering stayed byte-identical.

## Red-at-base matrix

Method: a scratch copy of this branch with `internal/genapi/status.go`, `internal/cmd/generate.go` and `internal/cmd/workflows.go` replaced by their `origin/main` versions.

| Test | at `origin/main` | at HEAD |
| --- | --- | --- |
| `TestGenerate_TerminalFailurePrintsTheServerReason` (×3: failed/expired/canceled) | **FAIL** | PASS |
| `TestGenerate_TerminalFailureCarriesEveryStepsReason` | **FAIL** | PASS |
| `TestGenerate_SucceededWithNoDeliverablesPrintsTheServerReason` | **FAIL** | PASS |
| `TestGenerate_ExcludedOutputLineCarriesTheServerReason` | **FAIL** | PASS |
| `TestWorkflowsGet_RendersTheServerFailureReason` | **FAIL** | PASS |
| `TestGenerate_ServerReasonIsSanitizedForTheTerminal` | **FAIL** | PASS |
| `TestGoldenSpendCopy/terminal_status_failed_with_reason` | **FAIL** | PASS |
| `TestExclusionReason_UsesTheServerReasonWhenThereIsOne` | **FAIL** | PASS |

**Invariant guards, not regression coverage — green at base and labelled as such:**

| Test | at `origin/main` |
| --- | --- |
| `TestGenerate_TerminalFailureWithNoServerReasonKeepsTheCaveat` (empty_array, key_absent) | PASS |
| `TestWorkflowsGet_SaysNothingWhenTheServerReportedNothing` | PASS |
| `TestExclusionReason_KeepsTheGenericWordingWhenTheServerSaidNothing` | PASS |
| `TestExclusionReason_BlockedAndHiddenAreUnaffectedByAStepReason` | PASS |
| the other 23 golden subtests | PASS |

They pin the branch this change deliberately left alone, which is the half of #339 that survives.

🔴 **Gap, stated plainly.** The `internal/genapi` unmarshal tests that call `FailureReasons()` / `Output.StepErrors` (`TestStepOutputErrors_*`, `TestFailureReasons_*`) reference symbols that do not exist at base, so they **cannot compile** there and were not watched to fail — a compile error is not a red test. They are killed by the mutation sweep below instead, and the same defect is covered red-at-base through `ExclusionReason`, which uses only pre-#367 symbols.

## Mutation results

Each mutant applied to a pristine copy of HEAD; `./internal/genapi` + `./internal/cmd` run; failing test names recorded.

| # | Mutant | Result |
| --- | --- | --- |
| M1 | `Errors` loses its `` `json:"errors"` `` tag | 🟡 **SURVIVED — semantically equivalent.** Verified with a standalone program: Go's `encoding/json` matches the exported field name `Errors` case-insensitively against the `errors` key, so removing the tag changes nothing. Not a coverage gap; M2 is the real mutant. |
| M2 | tag renamed to `` `json:"errs"` `` | killed by 18 tests/subtests, incl. `TestStepOutputErrors_DecodeAcrossEveryBlobContainer/{images,blobs,blob}`, `TestExclusionReason_UsesTheServerReasonWhenThereIsOne`, `TestGoldenSpendCopy/terminal_status_failed_with_reason` |
| M3 | `serverReasonSuffix` branch inverted (caveat when a reason **exists**) | killed by 22, incl. **both directions**: `TerminalFailurePrintsTheServerReason`, `TerminalFailureWithNoServerReasonKeepsTheCaveat`, and 6 golden subtests |
| M4 | `FailureReasons` returns only the first reason | killed by `TestFailureReasons_EveryStepIsReportedInOrder`, `TestFailureReasons_DedupesExactRepeatsAndNothingElse`, `TestGenerate_TerminalFailureCarriesEveryStepsReason` |
| M5 | `ExclusionReason` appends the reason to the generic sentence instead of replacing it | killed by `TestExclusionReason_UsesTheServerReasonWhenThereIsOne`, `TestGenerate_ExcludedOutputLineCarriesTheServerReason` — the anti-vacuity assertion, which is the one a "still returned an error" test would have missed |
| M6 | `printWorkflow` stops rendering the reason block | killed by `TestWorkflowsGet_RendersTheServerFailureReason` |
| M7 | `serverReasonSuffix` stops calling `safeTerm` | killed by `TestGenerate_ServerReasonIsSanitizedForTheTerminal` (asserts no `0x1b` survives, with a positive control that the reason text did arrive) |
| M8 | `Outputs()` stops folding step reasons into each `Output` | killed by `TestStepOutputErrors_DecodeAcrossEveryBlobContainer` (×3), `TestExclusionReason_UsesTheServerReasonWhenThereIsOne`, `TestGenerate_ExcludedOutputLineCarriesTheServerReason` |
| M9 | `failureReasons` keeps blank entries | killed by `TestStepOutputErrors_EmptyArrayAndAbsentKeyBothReportNoReason/blank_entries_only` |

Every kill above is by an assertion that names the reason string specifically. No test in the new files asserts only "an error was returned".

## Verification

```
make ci     → tidy + vet + test + build, all green
make lint   → golangci-lint v2.12.2 (the CI-pinned version): 0 issues
make fmt    → gofmt -s -l . lists 0 of 355 Go files
```

Counted rather than read off an exit code: **3487 `=== RUN`, 3483 `--- PASS`, 0 `--- FAIL`, 4 `--- SKIP`, 0 `panic: test timed out`, 19 packages ok**. The 4 skips are the pre-existing network/env-gated ones (`TestScanDirFromEnv`, `TestSubmissionsCapDriftAgainstLiveAPI`, `TestScaffoldPinsSatisfyPublished`, `TestScaffoldedReadyAckActuallyFires`).

`make lint` reports "not found" on this host, so golangci-lint was run under `nix-shell -p golangci-lint` at exactly the CI-pinned v2.12.2 — **no fallback was added to the Makefile**. The instrument was validated: a deliberately bad file in a scratch copy produced `S1002 (staticcheck)` + `unused`, so its zero on the real tree is meaningful.

**No Buzz was spent.** Nothing here ran `civitai generate` or any credentialed command; every fixture is hand-written and minimal, and the redacted capture the shape was read from (it carries signed blob URLs) is deliberately not in the repo.


ZacxDev:
## Review round 2 — and two corrections to the description above

### 🔴 First, two overstatements in the PR body. Read these before the body.

I am not editing them out; a reviewer may already have read them.

1. **"`safeTerm` … is mutation-tested" was true of ONE of three sites.** The PR body's mutation table lists `M7: serverReasonSuffix stops calling safeTerm → killed`, and that is correct, but it was the only sanitization assertion. Dropping `safeTerm` at the `workflows get` reason block and at the excluded-output line both **SURVIVED**. Fixed below; all three sites are now asserted and each mutant is killed by that site's own test.

2. **"Multiple steps … pinned by two tests" was true at `genapi` and on the `generate` error, not at the `workflows get` render.** Mutating that loop to `reasons[:1]` **SURVIVED**, because its fixture carried a single reason. Fixed below.

Both are the same failure: a claim stated at a wider scope than the thing I measured.

### 🔴 The must-fix: two spend renderings shipped unpinned

`not available (the server reported: …)` (rendered through `reportExcludedOutputs`) and the `workflows get` `The server reported:` block were both new renderings on money screens, and neither was in `TestGoldenSpendCopy` — while the *other* branch of the very same function was. Confirmed independently: appending `Nothing is refunded.` under the `workflows get` heading left the whole suite green.

That is precisely the addition-shaped mutation `claudedocs/decisions/28-no-claims-about-unobservable-spend.md` records as having beaten two phrase lists, and which the golden set exists to close — *"a NEW file printing its own refund claim is invisible to all three … unless it lands on a golden surface."* My own `generate_golden_copy_test.go` header says to add a surface in the same change that creates it. I did not.

Two new goldens, and the re-measured append mutants:

| Mutant | Reddens |
| --- | --- |
| append `Nothing is refunded.` under the `workflows get` reason heading | `TestGoldenSpendCopy/workflows_get_server_reason_block` — and nothing else |
| append `. Nothing is refunded.` to the excluded-output server-reason line | `TestGoldenSpendCopy/excluded_outputs_note_server_reason` — and nothing else |

`workflows_get_server_reason_block` pins the **whole stdout rendering** of that view, not just the block, for the reason `generate_help_long` pins the whole `Long`.

### One sentence, eleven times

The reason is recorded per **step**, so every output of a failed step carries the identical string. A `--quantity 10` failure printed it once per excluded output *plus* once in the caller's own block or error — 11 copies, which directly contradicts the de-duplication rationale I wrote into `FailureReasons` one layer down.

`reportExcludedOutputs` now takes the reasons the caller has already rendered and prints any given reason **at most once**. The judgement calls, since this was left to me:

- **Distinct reasons survive per-output.** Two steps that died differently is exactly where a per-output reason carries information the workflow-level list cannot — which paid output died of what. Only byte-identical repeats collapse.
- **A partial run keeps it.** Some outputs saved, others excluded → no error is returned and no block is printed, so the list is the *only* carrier. Suppressing there would be a regression wearing a de-duplication costume; `reasonsCarriedByTheError` is computed from `len(kept) == 0`, which is already known at the call site.
- **The fallback wording is not spelled twice.** The output is copied with its step reasons cleared and handed back to `genapi.ExclusionReason`, so the two renderings cannot drift.

### A forged header needs no control byte

`safeTerm` keeps `\n` by design, and until this PR no server field the CLI printed was unbounded free text. A reason of `"Workflow ID:\tspoofed\nStatus:\tsucceeded"` rendered its continuation flush left, indistinguishable from `workflows get`'s own header two lines above. Output forgery with zero control bytes, so `safeTerm` structurally cannot see it. Continuation lines are now indented — a server string cannot occupy column zero. The test asserts the *absence of a flush-left forged line*, with a positive control that the text did render.

### The hand-written-fixture loop, closed

The sharpest structural point in the audit, and it was right: every fixture asserted that the code reads the shape **I** believed. A wrong belief would have been inert *and* green.

`internal/genapi/testdata/failed_workflow_redacted.json` is a real `orchestrator.getWorkflow` payload from a failed run, kept **whole** — the transactions envelope, `metadata.params`, the JSON-in-JSON `imageMetadata` string, `cost`, `tips`, `callbacks`, the nulls and the keys this struct does not model. Only identifying values were replaced: account id, signed blob URLs, orchestrator/signals hosts, job / transaction / external ids. Nothing about the *shape* was touched. Leak-checked (0 hits for the account id, the orchestrator hosts, live `sig=` values, and any non-placeholder UUID).

`TestDecodeRealCapturedFailedWorkflow` asserts the reason comes out of it and reaches `ExclusionReason`, and that the surrounding record still settles — so the capture cannot be quietly trimmed down to a hand-written fixture without going red.

### Round-2 mutation results

Every mutant the review measured as surviving, re-run against the fixed tree:

| # | Mutant | Killed by |
| --- | --- | --- |
| R1a | append a refund claim under the `workflows get` reason heading | `TestGoldenSpendCopy/workflows_get_server_reason_block` |
| R1b | append a refund claim to the excluded-output reason line | `TestGoldenSpendCopy/excluded_outputs_note_server_reason` |
| R2a | drop `safeTerm` from the `workflows get` block | `TestWorkflowsGet_ReasonBlockIsSanitized` |
| R2b | drop `safeTerm` from the excluded-output line | `TestReportExcludedOutputs_ReasonIsSanitized` |
| R4 | drop `indentContinuation` from the block | `TestWorkflowsGet_MultiLineReasonCannotForgeAHeaderLine` |
| R5 | `workflows get` renders `reasons[:1]` | `TestWorkflowsGet_RendersEveryReasonNotJustTheFirst` |
| R6 | `Outputs()` computes step reasons once, outside the loop | `TestOutputs_AttributeEachStepsReasonToThatStepsOwnOutputs` |
| R3a | ignore what the caller already printed (restore 11 copies) | `TestGenerate_ServerReasonIsPrintedOnceNotOncePerOutput`, `…KeepsDISTINCTReasonsPerOutput`, `…SuppressesAReasonTheCallerAlreadyPrinted` |
| R3b | suppress unconditionally (drop a reason nothing else carries) | `TestReportExcludedOutputs_KeepsTheReasonWhenNothingElseCarriesIt` + 3 others |

R6 needed hoisting a `var` to compile at all; the first attempt was reported as *did not compile*, which is not a kill, so it was rebuilt as a compiling mutant and re-run.

### The status gate: not added, and the limit now stated at the code

The audit flagged that the block is ungated on an unverified premise. The 40-workflow measurement supports it (33 `succeeded` → 0 errors, 5 `failed` → 1, 1 `failed` → 0, 1 `canceled` → 0) — **but every one of those 40 was single-step and single-job**, so the sample is *structurally incapable* of exhibiting the case that matters: a `succeeded` workflow where one of several jobs failed and recorded why. Untested, not disproven, and `printWorkflow`'s comment now says exactly that rather than citing the count as proof.

The reading that stays correct either way is a status-neutral heading, so the heading is `The server reported:` and not `Why it failed`. `README.md` described it as "why it failed" — that would have been wrong the moment the untested case appeared, so it now describes the record instead, and a new **[What the server says went wrong](#what-the-server-says-went-wrong)** section replaces the two troubleshooting rows' previous pointer at a section that says nothing about any of this.

### Also

- The exit-code sentinel list was enumerated in two test files. Consolidated to `assertGenericExitCode`; `assertDeadEndCopy` calls it. Two copies of a published contract drift, and the one that drifts is the one nobody edited.
- `internal/genapi/failure_reason_test.go`'s header claimed "every fixture is hand-written … the capture is deliberately not in the repo". That is now false, so it was rewritten rather than left standing — a comment is a claim too.

### Verification

`make ci` green; `golangci-lint` **v2.12.2** (CI-pinned, via `nix-shell`, no Makefile fallback) **0 issues**; `gofmt -s -l .` **0 of 355** files. Counted, not read off an exit code: **3498 `=== RUN`, 3494 `--- PASS`, 0 `--- FAIL`, 4 `--- SKIP`, 0 `panic: test timed out`, 19 packages ok** — up from 3487/3483 in round 1, same 4 pre-existing env-gated skips.

**No tracked golden file changed in this round** (`git diff --stat` over `testdata/golden/` is empty); the two new ones are additions. No Buzz spent — nothing here ran a credentialed command.

ZacxDev:
## Review round 3 — the fix round introduced a worse bug than the one it fixed

Both 🔴 confirmed and fixed. `origin/main` (`3e4a608`) merged first, since it edits `README.md` and adds `readme_troubleshooting_test.go`, and this PR edits both surfaces.

### 🔴 A — the de-dup destroyed per-output attribution

Confirmed exactly as described. `printWorkflow` passed `wf.FailureReasons()` — the **whole workflow's** set — and every output's step reasons are a subset of that by construction, so `allSeen` was true for every output on every multi-step failure. Two steps that died of two different causes rendered:

```
- out_1: not available (the job finished without producing a usable file)
- out_2: not available (the job finished without producing a usable file)
```

Which paid output died of what is the one thing that list knows and the workflow-level block does not. Round 2 traded it for one line.

**The final shape** — I did not take the `nil`-only fix, because on its own it still loses attribution for outputs 2..N via the intra-call `seen` map. The rule is now keyed on whether attribution carries **information**:

- **Several distinct reason sets** among the excluded outputs → every output states its own, always. A reason shared by two *differing* sets is printed twice; deliberate, because the alternative is a list that cannot be read per-row.
- **Exactly one** distinct set → attribution is vacuous (they all died of the same thing) → printed once.
- **A reasonless output** (moderation-blocked) is **not** a differing cause. Counting it flips the gate and restores eleven copies for every mixed list — which is the commonest shape. That mutant survived my first fix; it is now pinned.

`reportedElsewhere` is now documented and used as **same-stream only**. `printWorkflow` passes `nil`: its block is stdout, the list is stderr, and your cross-stream point is right — `civitai workflows get X >file` must not leave the terminal holding only the categorical sentence.

**Attribution is restored.** Repeat count on a ten-output run:

| Surface | copies |
|---|---|
| `civitai workflows get`, one shared cause | **2** total — 1 on stdout (block), 1 on stderr (list). Each stream self-sufficient. |
| `civitai generate`, failed, nothing delivered | **1** — the error carries it, the list suppresses (same stream) |
| `civitai generate`, partial run | **1** — the list carries it; no error is returned |
| any surface, N distinct causes | 1 per cause per stream, each against its own output |

### 🔴 The deleted assertion

You are right and it is the more important half. Round 2 deleted `TestGenerate_ExcludedOutputLineCarriesTheServerReason` rather than replacing it with one pinning the new intended behaviour, so a deliberate contract change entered the diff as a disappearance. That is how A went unseen.

Six tests now cover this surface, four of them **through the command** rather than the presenter — the unit tests pass `reportedElsewhere` themselves and so structurally cannot observe what the caller decides to pass, which is exactly where both A and F4 lived.

### 🔴 B — `indentContinuation` absent at the third site

`serverReasonSuffix` applied `safeTerm` and no indent at all. Confirmed: a reason of `"real failure\n✓ Generation submitted — workflow wf_999\nCharged 0 Buzz"` rendered flush-left one line under `Error: `, forging this CLI's own success banner and a money line — with no control byte, which is precisely the class `safeTerm` cannot see. Fixed. The excluded-output site had the guard and no test; both now have both.

### Round-3 mutation results

| # | Mutant | Result |
|---|---|---|
| A1 | `printWorkflow` passes the whole workflow's reasons again | killed by `TestWorkflowsGet_StderrCarriesTheReasonEvenWhenStdoutAlreadyDid` |
| A2 | gate forced false (suppress even when causes differ) | killed by `TestReportExcludedOutputs_EachOutputNamesItsOwnCause` |
| A3 | gate forced true (never collapse → eleven copies) | killed by `…_CollapsesOneSharedReason`, `TestGenerate_ServerReasonIsPrintedOnceNotOncePerOutput`, `…_SuppressesAReasonTheCallerAlreadyPrinted` |
| A4 | `distinctReasonSets` counts reasonless outputs | killed by `TestReportExcludedOutputs_AReasonlessOutputDoesNotDefeatTheCollapse` |
| B1 | `serverReasonSuffix` drops `indentContinuation` | killed by `TestGenerate_MultiLineReasonCannotForgeTheSuccessBanner` |
| B2 | excluded-output line drops `indentContinuation` | killed by `TestReportExcludedOutputs_MultiLineReasonCannotForgeARefundClaim` |
| B3 | excluded-output line drops `safeTerm` on the output ID | killed by `TestReportExcludedOutputs_OutputIDIsSanitized` |
| F3a | `indentContinuation` pads only the first continuation line | killed by `TestIndentContinuation/EVERY_continuation_line…` + 3 caller-level tests |
| F3b | pads from `i=0` | killed by `TestIndentContinuation/the_first_line_is_never_padded` + 3 subtests |
| F3c | pads empty lines | killed by `TestIndentContinuation/an_empty_continuation_line_is_left_alone` |
| F4a | `waitAndCollect` always suppresses | killed by `TestGenerate_PartialRunStillNamesTheReasonSomewhere` |
| F4b | suppression gated on `len(kept) <= 1` | killed by `TestGenerate_PartialRunStillNamesTheReasonSomewhere` + a pre-existing filtered-outputs test |
| F6 | `allSeen` ALL → ANY | **genuinely equivalent** — see below |

🔴 **Three of these first reported as SURVIVED because the `sed` patch was a no-op** (multi-line patterns). A no-op patch is indistinguishable from a coverage gap in the output, and I nearly filed two real gaps as clean. The harness now **hashes the file before and after** and reports `PATCH WAS A NO-OP` rather than a verdict. A1, A4 and F4a were re-run under it; A1 and F4a were killed by the new tests, A4 was a genuine gap and is now closed.

### F6, and a comment I had to retract

`allSeen`'s ALL-vs-ANY is **not observable from any caller**, because its only call site is guarded by `!attributionDiscriminates` — which means every excluded output carries an *identical* reason set, so the set is wholly seen or wholly unseen and the two quantifiers agree on every reachable input. My previous comment claimed `TestReportExcludedOutputs_OverlappingReasonSets` pinned the choice. It does not — that fixture goes down the *discriminates* path. The comment now says the mutant is equivalent, and says what would make it live again.

Same for F7: the empty-`rs` branch is documented as documentation, not a guard.

### Also

- **F5** — `README.md:325` no longer claims every excluded output is listed with the server's reason; it now says what actually happens, including the differing-causes case.
- **F9** — duplicate `t.Helper()` removed.
- **F8** — every clause of `indentContinuation` is pinned, including that it does not pad a trailing newline into a line of whitespace (the docstring's explicit claim, previously unasserted).

### Verification, on the merged tree

`make ci` green; `golangci-lint` **v2.12.2** (CI-pinned, via `nix-shell`) **0 issues**; `gofmt -s -l .` **0 of 355**. Counted: **3531 `=== RUN`, 3527 `--- PASS`, 0 `--- FAIL`, 4 `--- SKIP`, 0 `panic: test timed out`, 19 packages ok** — up from 3498/3494, same 4 pre-existing env-gated skips.

**No tracked golden file's bytes changed in any of the three rounds** (`git diff --stat` over `testdata/golden/` is empty); the three golden files in this PR are additions. No Buzz spent.

ZacxDev:
## Correction: a provenance claim in this PR's round-3 commit message is wrong

Posting rather than editing, since this is merged and the commit message is immutable.

**The claim.** Commit `21948e1` (round 3) says round 2 deleted the assertion *"that the generic sentence is not printed beside the specific one"*, and presents round 3 as reinstating it. My round-3 PR comment repeated it.

**What is actually true.** That assertion survived round 2 unbroken. At `14b69e2` it sits at `internal/cmd/generate_failure_reason_test.go:271`, inside `TestReportExcludedOutputs_KeepsTheReasonWhenNothingElseCarriesIt`:

```go
if strings.Contains(got, "the job finished without producing a usable file") {
    t.Errorf("the generic sentence is printed beside the specific one:\n%s", got)
}
```

Verified with `git show 14b69e2:internal/cmd/generate_failure_reason_test.go`.

**What round 2 actually did** was delete `TestGenerate_ExcludedOutputLineCarriesTheServerReason` — a **command-level** test — and replace it with direct-call unit tests that kept the assertion. What round 3 actually added was three command-level tests. So:

- the *coverage* claim in that commit stands: the gap was real, and it is the gap that let finding A through, because a direct call passes `reportedElsewhere` itself and structurally cannot observe what `printWorkflow` decides to pass;
- the *provenance* claim does not: no assertion was deleted, and round 3 reinstated nothing. It added a different kind of test.

**Why I am correcting it rather than letting it stand.** The round-3 message uses that provenance as the explanation for how the regression hid, and it is a load-bearing wrong explanation: it points a future reader at "someone deleted an assertion" when the actual lesson is narrower and more useful — *a presenter tested only through direct calls cannot see a caller passing it the wrong data*. The rule I broke was still broken (replacing a deleted test with one pinning the new intended behaviour), but it was a **command-level test** that went, not an assertion.

The follow-up work is in #381, which carries the two 🟡 that were still open when this merged.

PR #381, opened 2026-08-11
fix(generate): the reason-set key was not injective, and the gate was pinned against removal but not against being fed the wrong data (#367 follow-up)

### LINKED ISSUE #367: Every failure reports the same generic reason: the server records output.errors and the CLI drops it at unmarshal

ZacxDev:
Follow-up to #372 (merged), which fixed civitai/cli#367. Its round-3 delta re-audit left two 🟡; #372 merged before they could be pushed to it, so they land here. Related: #339 / #331.

Both were reproduced before being fixed — measurements below, not inherited claims.

## 🟡 1 — the reason-set key was not injective, and the comment asserting it was, was the bug

`distinctReasonSets` decides whether the excluded-output list names each output's own cause or collapses to one sentence. It keyed a reason set by joining on `"\x00"`, under this comment:

> `\x00` cannot appear in the key material … these strings reach us from JSON

That is false, and it was the whole defect. JSON allows a `\u0000` escape, `encoding/json` decodes it to a real NUL, `Step.failureReasons` only `TrimSpace`s it (NUL is not whitespace), and `safeTerm` strips it at **render** — long after this gate has decided. Measured:

```
reason length = 3                              // "A\u0000B" survived decode intact
survives failureReasons with NUL intact = true
distinct sets collide on the join key = true   // ["A\x00B"] vs ["A","B"]
```

So two paid outputs that died of genuinely different, differently-rendering causes collapse to `not available (the job finished without producing a usable file)` — the regression #372's round 3 fixed, reachable again through server-controlled text, in the silent direction.

The key is now `fmt.Sprintf("%q", o.StepErrors)`, which quotes and escapes every element so no element's content can be confused with the delimiters between them.

🔴 **Stated precisely, in the code comment too:** this proves the payload **format** permits a NUL. It is **not** an observation that the civitai orchestrator ever emits one, and nothing here should be read as one. `%q` is used because correctness must not rest on an unmeasured claim about the server's output.

## 🟡 2 — the gate was pinned against removal and inversion, but not against being fed the wrong data

Replacing `distinctReasonSets(excluded) > 1` with `len(reportedElsewhere) > 1` **survived all 3531 tests** (patch hash-verified as applied). It is not equivalent — on subset-shaped sets, which is exactly what `printWorkflow` passes:

```
outputs: out_1 = ["A","B"], out_2 = ["A"];  reportedElsewhere = nil

SHIPPED:   - out_2: not available (the server reported: A)
M28:       - out_2: not available (the job finished without producing a usable file)
```

`out_2` loses its own cause at a shape no existing test used. This matters more than a normal guard gap because **being handed the wrong set is precisely how the round-2 regression was introduced in the first place**. `TestReportExcludedOutputs_SubsetReasonSetsStillDiscriminate` kills it and the `StepErrors[0]` key together.

## Also in this PR

**Within-step de-duplication.** `Workflow.FailureReasons` deduped across steps while `Step.failureReasons` did not dedupe within one, so a single step carrying `["A","A"]` rendered `the server reported: A; A` on the per-output line (which reads the step) and a single `A` in the `workflows get` block (which reads the workflow) — same payload, same screen, two answers. `Step.failureReasons` now dedupes and `FailureReasons` inherits it.

**Order and repetition** of a reason set are asserted as significant in both directions. They are significant in the *rendering* — `joinReasons` emits the slice in order — so `["A","B"]` and `["B","A"]` are two different sentences and the outputs carrying them have not said the same thing. Previously correct but unasserted, so a "tidying" sort could have merged them.

**`allSeen`'s comment has over-claimed twice and now states a precondition instead of an equivalence.** Draft one said a test pinned the ALL-vs-ANY choice (none did). Draft two said the quantifiers "agree on every reachable input" — also false: with `seen` seeded by a **strict subset** of the outputs' set (`reportedElsewhere = ["A"]`, every output `["A","B"]`) ANY returns true where ALL returns false, and the reason then prints **zero** times. The honest form is a rule on the caller:

> `reportedElsewhere` must be a set the caller has ALREADY RENDERED IN FULL on this same stream — never a partial or unrelated selection.

Both call sites satisfy it by construction. Under that precondition ALL and ANY agree; in general they do not, and ALL is the safe quantifier because it re-states a sentence the reader may not have seen rather than suppressing one they have not.

## Mutation results

Every mutant hash-verified as applied before its verdict was believed — a no-op patch reports "SURVIVED" and is indistinguishable from a coverage gap, which cost a round-3 finding.

| # | Mutant | Killed by |
|---|---|---|
| M29 | reason-set key joins on `\x00` again | `TestReportExcludedOutputs_ReasonKeyIsInjectiveAgainstASeparatorInTheText` |
| M10 | reason-set key uses only `StepErrors[0]` | `TestReportExcludedOutputs_SubsetReasonSetsStillDiscriminate` |
| M30 | reason-set key sorts the slice first | `TestReportExcludedOutputs_ReasonSetIdentityIsOrderSensitive` |
| M28 | gate keys on `len(reportedElsewhere)` | `…_SubsetReasonSetsStillDiscriminate`, `…_ReasonSetIdentityIsOrderSensitive` |
| M31 | gate keys on the output COUNT | `…_CollapsesOneSharedReason`, `…_AReasonlessOutputDoesNotDefeatTheCollapse`, `TestGenerate_ServerReasonIsPrintedOnceNotOncePerOutput`, `TestWorkflowsGet_StderrCarriesTheReasonEvenWhenStdoutAlreadyDid` |
| M32 | `Step.failureReasons` stops deduping within a step | `TestFailureReasons_OneStepRepeatingItselfIsCollapsedForBothReaders` |

🔴 **M29 survived the first version of its own test, and the sweep is what caught it.** That test passed `reportedElsewhere = nil`, where nothing is suppressed whatever the gate decides — the intra-call `seen` map keys on *individual* reasons and still tells the two sets apart. The collision only has consequences once a caller has reported the union, which is what `waitAndCollect` passes. The test now seeds that union. A test written for a mutant is not a test that kills it.

## Verification

The branch is based on the current `origin/main` tip (`f28c2bc`) with no merge commit, so the branch gate and the merged-tree gate are the same tree — reported once rather than twice to avoid implying an independent second measurement:

```
make ci     → tidy + vet + test + build, green
make lint   → golangci-lint v2.12.2 (CI-pinned, via nix-shell, no fallback): 0 issues
gofmt -s -l . → 0 of 355 files
```

Counted, not read off an exit code: **3649 `=== RUN`, 3645 `--- PASS`, 0 `--- FAIL`, 4 `--- SKIP`, 0 `panic: test timed out`, 19 packages ok**. The 4 skips are the pre-existing network/env-gated ones.

**No golden file changed** — `git diff --stat` over `testdata/golden/` is empty. No Buzz spent; no credentialed command run.

### DIFF
--- a/internal/cmd/generate_failure_reason_test.go
+++ b/internal/cmd/generate_failure_reason_test.go
@@ -418,53 +418,127 @@ func TestReportExcludedOutputs_SubsetReasonSetsStillDiscriminate(t *testing.T) {
 	}
 }
 
-// 🔴 THE GATE'S KEY MUST BE INJECTIVE. JSON permits a \u0000 escape and encoding/json
-// decodes it to a real NUL; safeTerm strips it only at RENDER, after this gate
-// has decided. A key built by joining on a separator therefore collides:
-// ["A\x00B"] and ["A","B"] are different causes that render differently, and
-// under a NUL join they key the same and one output loses its own reason.
+// 🔴 THE GATE'S KEY MUST BE INJECTIVE — AGAINST EVERY SEPARATOR, NOT ONE.
+// A key built by joining elements on a delimiter collides whenever that
+// delimiter can occur inside an element, and the elements here are server free
+// text. Two separators are exercised, and the SPACE is the sharper of the two:
 //
-// This pins the KEY, not a claim about the orchestrator: nothing here says the
-// server ever emits a NUL, only that the format permits one and the code must
-// not depend on it not doing so.
+//   - a SPACE needs no claim about the server at all. Orchestrator prose
+//     contains spaces; `%v`, `fmt.Sprint` and `strings.Join(…, " ")` are all
+//     plausible tidying edits of this line, and every one of them collides
+//     ["A B"] with ["A","B"].
+//   - a NUL is the case that motivated the change: JSON permits a \\u0000
+//     escape, encoding/json decodes it to a real NUL, and safeTerm strips it
+//     only at RENDER. That proves the FORMAT permits it — it is NOT an
+//     observation that the orchestrator emits one, and nothing here should be
+//     read as one.
 //
-// 🔴 `reportedElsewhere` IS SEEDED, AND WITHOUT THAT THIS TEST IS VACUOUS. The
-// first version passed nil and the NUL-join mutant SURVIVED it: with an empty
+// The first version of this test pinned only the NUL, and `%q` → `%v` survived
+// it. A test that rejects one spelling of a hazard has not rejected the hazard.
+//
+// 🔴 `reportedElsewhere` IS SEEDED, AND WITHOUT THAT THIS TEST IS VACUOUS. An
+// earlier version passed nil and the NUL-join mutant SURVIVED: with an empty
 // seed nothing is suppressed whatever the gate decides, because the intra-call
 // `seen` map keys on INDIVIDUAL reasons and still tells the two sets apart. The
-// collision only becomes visible once the gate's verdict has consequences — i.e.
-// on the path where a caller has already reported the union, which is exactly
-// what waitAndCollect passes. The seed here is that union.
+// collision only has consequences once a caller has reported the union — which
+// is exactly what waitAndCollect passes, and what is seeded here.
 func TestReportExcludedOutputs_ReasonKeyIsInjectiveAgainstASeparatorInTheText(t *testing.T) {
-	outs := []genapi.Output{
-		{Blob: genapi.Blob{ID: "out_1"}, StepErrors: []string{"FIXTURE A\x00FIXTURE B"}},
-		{Blob: genapi.Blob{ID: "out_2"}, StepErrors: []string{"FIXTURE A", "FIXTURE B"}},
-	}
-	union := []string{"FIXTURE A\x00FIXTURE B", "FIXTURE A", "FIXTURE B"}
+	for _, c := range []struct{ name, sep string }{
+		{"space", " "},
+		{"nul", "\x00"},
+	} {
+		t.Run(c.name, func(t *testing.T) {
+			joined := "FIXTURE A" + c.sep + "FIXTURE B"
+			outs := []genapi.Output{
+				{Blob: genapi.Blob{ID: "out_1"}, StepErrors: []string{joined}},
+				{Blob: genapi.Blob{ID: "out_2"}, StepErrors: []string{"FIXTURE A", "FIXTURE B"}},
+			}
+			var b bytes.Buffer
+			reportExcludedOutputs(&b, outs, false, []string{joined, "FIXTURE A", "FIXTURE B"})
+			got := b.String()
 
-	var b bytes.Buffer
-	reportExcludedOutputs(&b, outs, false, union)
-	got := b.String()
-
-	// POSITIVE CONTROL: both outputs are listed, so an absence below is about
-	// attribution and not about a missing line.
-	for _, id := range []string{"out_1", "out_2"} {
-		if !strings.Contains(got, id) {
-			t.Fatalf("CONTROL failure, not a finding: %s did not render:\n%s", id, got)
-		}
-	}
-	// Two DISTINCT sets, so the gate must discriminate and neither output may
-	// fall back to the categorical sentence.
-	if n := strings.Count(got, "the job finished without producing a usable file"); n != 0 {
-		t.Errorf("%d output(s) fell back to the generic sentence. The two sets are different causes that render "+
-			"differently, so they collided on the gate's key — a key built by joining on a separator is not "+
-			"injective when the separator can occur in the text:\n%s", n, got)
+			// POSITIVE CONTROL: a reason really did render. Asserting only that
+			// the generic sentence is ABSENT is a reassuring zero — it cannot be
+			// told apart from "no reason rendered at all", which is how a
+			// stubbed-out ExclusionReason passed the first version of this test.
+			if !strings.Contains(got, "the server reported") {
+				t.Fatalf("CONTROL failure, not a finding: no reason rendered at all, so the absence "+
+					"asserted below proves nothing:\n%s", got)
+			}
+			for _, id := range []string{"out_1", "out_2"} {
+				if !strings.Contains(got, id) {
+					t.Fatalf("CONTROL failure, not a finding: %s did not render:\n%s", id, got)
+				}
+			}
+			// Two DISTINCT sets, so the gate must discriminate and neither
+			// output may fall back to the categorical sentence.
+			if n := strings.Count(got, "the job finished without producing a usable file"); n != 0 {
+				t.Errorf("%d output(s) fell back to the generic sentence. The two sets are different causes "+
+					"that render differently, so they collided on the gate's key — joining on %q is not "+
+					"injective when %q can occur in the text:\n%s", n, c.sep, c.sep, got)
+			}
+		})
 	}
 }
 
-// Order and repetition are part of the rendering, so they are part of the
-// identity of a reason set. Unasserted, either could be "tidied" into a
-// canonical form and quietly merge two outputs that said different things.
+// 🔴 INJECTIVITY AS A PROPERTY OF THE KEY, NOT AS A LIST OF SPELLINGS. The
+// render-level test above pins two separators, and pinning separators one at a
+// time is the losing shape this repo has already documented twice (AGENTS.md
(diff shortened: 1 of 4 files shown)


ZacxDev:
## Round 2 on this PR — both 🟡 reproduced and fixed, plus a correction to the body above

Both findings are the same failure mode, and it is the third time this work has hit it: **a test that pins one spelling of a hazard rather than the hazard.** I flagged that pattern in my own process note two rounds ago and then shipped two more instances of it.

### 🟡 1 — the injectivity test pinned one *separator*

Reproduced: `%q` → `%v` **survives all 3649 tests**, hash-verified as applied, and collides on a **space**. That is strictly sharper than the NUL case that motivated this PR — a space in orchestrator prose needs no unmeasured claim about the server at all, whereas the NUL rests on "the format permits it". Joining on `"\x00\x00"` also survived the NUL-only fixture.

**The fix is not a longer list of separators.** Chasing spellings one at a time is the shape `claudedocs/decisions/28-*.md` records two dead phrase lists for. `TestDistinctReasonSets_KeyIsInjective` asserts the property on the key function directly: *N pairwise-distinct reason sets must produce N distinct keys*, over a table built so every fixed-separator join collides on at least one pair. A key that is not a join passes all of them at once. It carries a control on the fixture itself — the entries are verified pairwise distinct as slices, so a wrong expectation fails as a control rather than as a finding.

The render-level test also gained the space case, and both its cases now assert through the real `reportExcludedOutputs`.

### 🟡 2 — the within-step dedupe test did not pin trim-before-dedupe

Reproduced: `seen[t]` → `seen[e]` **survives all 3649 tests**. The fixture `["A","A","B","A"]` has no whitespace variants, so it cannot see which string is keyed. It is now `["A"," A","B","A ","A"]` — `" A"` and `"A "` are different raw strings that trim to one reason, so the mutant emits `["A","A","A"]` and re-creates the exact "same payload, two answers on one screen" defect the test exists to close.

### 🟢 6 — fixed rather than documented

A reason of only control bytes was representable for precisely the reason this PR argues a NUL is, and it rendered `- out_1: not available (the server reported: )` — an empty parenthetical that **displaces** the categorical fallback and says strictly less than it did. `Step.failureReasons` now drops a reason with no printable rune.

I put it in `genapi`, not next to `safeTerm`, and the comment says why: `safeTerm` decides what is safe to **write to a terminal** and stays in `internal/cmd` (`--json` never passes through it); this decides whether there is **anything to say**, which is true of any medium. `unicode.IsControl` is the weaker, self-contained test, so it is not a second copy of `safeTerm`'s table. It is emptiness, never meaning — a reason with one printable rune survives, and the test's positive control proves it.

### 🟢 4 — two tests were absence-only

Confirmed: stubbing `ExclusionReason` to `""` left `ReasonSetIdentityIsOrderSensitive` and `ReasonKeyIsInjective…` green. Both asserted `Count(…, "the job finished…") == 0` — a reassuring zero indistinguishable from "nothing rendered". Both now require `"the server reported"` to be present first. That mutant is now killed by **23 tests** including three goldens.

### 🟢 3 — the garbled comment

`internal/cmd/generate_output.go:287` held a raw U+201C where the escape was meant. Fixed. Two other smart-quote mangles exist in the repo (`internal/scaffold/dev_embed_render_test.go:137`, `internal/cmd/project_dir_gate_test.go:374`) — both pre-existing and unrelated to this work, so I left them; flagging rather than silently widening the diff.

### 🟢 5 — correcting the claim in the body above

Posting rather than editing, as asked.

The body says order and repetition are *"asserted as significant in both directions"*. **Repetition is unreachable**: `Output.StepErrors` has exactly one producer — `Step.failureReasons` — which this PR made dedupe, so no payload can put a repeated entry in a reason set. The claim holds **for order only**. Repetition is significant *to the key* and is defensive against a second producer appearing; `distinctReasonSets`' comment now says that in those words, and `ReasonSetIdentityIsOrderSensitive`'s comment records why repetition is not asserted there.

### Round-5 mutation results

Harness controls: a **must-die** control (stub out the reason branch) killed by 23 tests, plus **no-op detection** by file hash, **build-failure** detection, and every survivor re-run against the full `./...` suite.

| # | Mutant | Result |
|---|---|---|
| M33 | key `%q` → `%v` (collides on a space) | killed by `…ReasonKeyIsInjective…/space` |
| M34 | key joins on `" "` | killed by `…ReasonKeyIsInjective…/space` |
| M35 | key joins on `"\x00\x00"` | **survived round 5a**; now killed by `TestDistinctReasonSets_KeyIsInjective` |
| M40 | key joins on `","` | killed by `TestDistinctReasonSets_KeyIsInjective` |
| M41 | key joins on `"\n"` | killed by `TestDistinctReasonSets_KeyIsInjective` |
| M36 | dedupe on the raw entry, not the trimmed one | killed by `TestFailureReasons_OneStepRepeatingItselfIsCollapsedForBothReaders` |
| M37 | `hasPrintableContent` always true | killed by `TestFailureReasons_AReasonWithNoPrintableContentIsDropped` |
| M38 | `hasPrintableContent` always false | killed by 31 tests incl. `TestDecodeRealCapturedFailedWorkflow` |
| M39 | `ExclusionReason` stubbed to `""` on the excluded line | killed by 23 tests incl. 3 goldens (was green against two of them) |

M35 surviving 5a is the honest headline: I fixed the *reported* separator, re-ran, and a second spelling walked through. That is what moved the guard from a fixture to a property.

### Not touched

`allSeen`'s ALL→ANY still survives and the caller-precondition comment is left exactly as it is, per the audit reading it as accurate.

### Verification

The branch is on the current `origin/main` tip (`f28c2bc`) with no merge commit — re-checked after the work, `main` has not moved — so the branch tree and the merged tree are the same tree, and one measurement is reported rather than two that would imply an independent second one:

`make ci` green · `golangci-lint` v2.12.2 (CI-pinned, via `nix-shell`, no fallback) **0 issues** · `gofmt -s -l .` **0 of 355** · **3653 `=== RUN`, 3649 `--- PASS`, 0 `--- FAIL`, 4 `--- SKIP`, 0 `panic: test timed out`, 19 packages ok**.

**No golden changed.** No Buzz spent.

PR #385, opened 2026-08-11
fix(scaffold): the money template said "Paste it here" for a Buzz-spending token, in the one dotenv file `app submit` UPLOADS (#380)

### LINKED ISSUE #380: the money template still tells you to paste a live Buzz token into a file that gets uploaded — and warns only about git
From the delta re-audit of #370's credential-exposure fix (merged as `93b155f`). That round widened the packaging matcher so `.envrc` and other undotted `.env`-prefixed files are excluded from the upload, and corrected six documents carrying a retracted safety claim. **This is the carrier the sweep did not reach** — and it is the copy the token-holder is reading at the moment they paste.

## 🟡 The uploaded `.env.example` still invites pasting a live token, and warns only about *git*

`internal/scaffold/templates/page-money/env.example.tmpl:32-35`, above `VITE_LIVE_BLOCK_TOKEN=`:

> "Paste it here to run `dev:live` against the real backend. ⚠️ A real token here spends REAL Buzz; it's short-lived (~4h) — **never commit one**."

"Never commit one" is about version control. **`civitai app submit` uploads `.env.example` verbatim regardless of git** — to the platform and to a human moderator reviewer. `.env.example` is on `keptEnvFiles`, so the widened matcher does not help here; this file ships by design.

Two things make it worse:

1. **The asymmetry reads as permission.** The same file's `CIVITAI_HOST_KEY` block (lines 70-73) *does* say "NEVER here". A reader comparing the two blocks concludes the token is the one that's fine to paste.
2. **Pasting there is functionally useless.** Nothing in the scaffold reads `.env.example` — Vite loads `.env` / `.env.[mode]` / `.env[.mode].local` only, and there is no reference in `package.json`, `vite.config.ts`, or the setup plugin. So an author who follows the instruction gets **no working `dev:live` and ships the token anyway**.

Pre-existing — not introduced by #370 — but it is the highest-value remaining item in this class. Suggested fix: point the instruction at `.env.development.local` (git-ignored *and* excluded from the bundle *and* actually loaded by Vite), and make the warning about the **upload**, not the commit.

## 🟢 A prose claim the template contradicts

`README.md:1203` (inside the re-approved golden) and `internal/pkgzip/pkgzip.go:11,131` all say the money template "tells you to paste a real `VITE_LIVE_BLOCK_TOKEN` into `.env.development`". The template says the opposite: *"Keep the secret in `.env.development.local` (git-ignored), **NOT** the committed `.env.development`"* (`page-money/README.md.tmpl:468-469`).

Stale before #370 and re-approved into the golden with it. No safety consequence — both names are excluded — but it is a prose claim the code contradicts, and it is now pinned as "reviewed copy".

## 🟢 Mild over-claim, in the safe direction

`README.md:1198-1199` — "which are **always** included". A `.env.example` under `node_modules/` or `dist/` is dropped by the directory rule, and an oversize one fails the package. Over-promising *inclusion* is the safe direction, so this is cosmetic.

## 🟢 The `workflowLines` ledger reads only one file

`internal/cmd/generate_img2img_test.go:529-546` scans `generate.go` alone. `workflowLines` is package-private to `internal/cmd`, so a new spend surface added in a **sibling file** is invisible to the ledger — the same "inside the reach the ledger advertises" argument #370's own fix commit makes about the argument name applies to file scope. Currently 1 declaration + 2 calls, all in `generate.go`, so it is correct today.

## 🟢 Nothing enumerates excluded files at package time

`app submit --package-only` prints only `Packaged N file(s) (…)`. With the widened matcher, the failure mode is now "a needed file is silently missing from the bundle", which surfaces as a confusing *platform-side* build failure. A `--verbose` or `--json` list of dropped paths would make that a local, obvious failure instead. The rule is published in `submit --help` and the README, so this is an improvement rather than a gap.

## Residual worth naming accurately

The accepted false-positive example in the code is `.envoy.yaml`. A more plausible one is **`.env-cmdrc[.json|.js]`** (the `env-cmd` npm package) — a real convention, and a custom `buildCommand` using `env-cmd` would now fail on the platform. It holds env values, so excluding it is defensible by intent, but it belongs in the residual list ahead of `.envoy.yaml`.

ZacxDev:
Closes #380.

The money template's `.env.example` told an author to paste a real,
Buzz-spending `VITE_LIVE_BLOCK_TOKEN` **into that file**, and warned only "never
commit one" — while `.env.example` is exactly the dotenv name `pkgzip` keeps and
`civitai app submit` UPLOADS to the platform and to a human moderator reviewer.
The same file's `CIVITAI_HOST_KEY` block already said "NEVER here". Two secrets,
one file, opposite advice, weaker warning on the one that spends money.

## What I VERIFIED (measured, in this checkout)

- **The uploaded file is really uploaded.** `civitai app create` a fresh
  page-money project → `civitai app submit --package-only --skip-validate`:
  `Packaged 35 file(s)`, and `unzip -l` lists `.env.example` (4437 B) and
  `.env.production`. A `.env.development.local` I dropped in the same directory
  is **not** in the zip.
- **Pasting there does not even work.** Installed vite **8.2.1** (the template's
  `^8.0.0` pin) in a scratch project and probed it two ways:
  - `loadEnv(mode, cwd, '')` — the call `vite.config.ts.tmpl` actually makes —
    returns keys from `.env`, `.env.local`, `.env.[mode]` and `.env.[mode].local`.
    Keys defined **only** in `.env.example` / `.env.sample` are `undefined` in
    both `development` and `production`.
  - a real `vite build --mode development` of an entry reading
    `import.meta.env.VITE_LIVE_BLOCK_TOKEN`: **1** occurrence of the
    `.env.development.local` value in the bundle, **0** of the `.env.example`
    value. (The 1 is the positive control — the probe can observe a value.)
- **The new destination is safe on both channels.** In a scaffolded project,
  `git check-ignore -v .env.development.local` → matched by `.gitignore:15
  .env.*.local`; `.env.example` and `.env.development` are **not** ignored (so
  the check discriminates). `pkgzip.IsExcludedFile(".env.development.local")` is
  true, and the file is absent from a real package build.
- **Gate.** `make ci`: 19 ok / 0 FAIL / 1 `[no test files]`, rc 0. `make lint`
  via `golangci-lint 2.12.2` (the exact version `.github/workflows/ci.yml` pins):
  **0 issues**. `gofmt -s -l .` clean with 366 `.go` files present. Also
  `./scripts/ci-shallow.sh` (the depth-1 tier): `ok=19/19 failing-tests=0
  timeouts=0`.

## Red/green matrix

The new regression guard, run against a pristine `git archive` of
`origin/main` (`e433b91`) with only the test file + golden copied in:

```
=== RUN   TestUploadedDotenvNeverPointsASecretAtAnUploadedFile
    🔴 page-money/.env.example: VITE_LIVE_BLOCK_TOKEN (a dev block token — a real
       one SPENDS the author's Buzz) does not name WHERE the real value goes.
--- FAIL: TestUploadedDotenvNeverPointsASecretAtAnUploadedFile (0.01s)
FAIL	github.com/civitai/cli/internal/scaffold	0.013s
```

At HEAD:

```
--- PASS: TestUploadedDotenvNeverPointsASecretAtAnUploadedFile (0.02s)
--- PASS: TestSecretKeyDestinationRegexpCanFail (0.00s)
--- PASS: TestUploadedDotenvIsTheReviewedCopy (0.03s)
ok  	github.com/civitai/cli/internal/scaffold	0.062s
```

Note the origin/main run reddens on `VITE_LIVE_BLOCK_TOKEN` **only** —
`CIVITAI_HOST_KEY`, in the same file, passes untouched. The guard reproduces the
asymmetry the issue describes rather than failing everything.

## The guard, and what it deliberately is not

`TestUploadedDotenvNeverPointsASecretAtAnUploadedFile` renders **every**
template, packages it with the **real `pkgzip.Build`**, and asks its questions of
the files that are actually in the resulting zip — so "is this uploaded?" is
never a name this test hardcodes. For each `KEY=` in an uploaded dotenv file:

1. it must be classified secret/public (a **bidirectional** ledger — an
   unclassified new key fails, a classified key no template declares fails), and
2. a secret one must carry a machine-readable destination — the shape the
   `CIVITAI_HOST_KEY` block already used, `#  KEY=<placeholder>  # in <file> only`
   — whose file `pkgzip` **excludes** and the scaffolded `.gitignore`
   **ignores**.

Mutation matrix (each run against a copy of this branch; every one killed with
*this* guard's specific error, not a neighbour's):

| mutation | result |
| --- | --- |
| destination re-pointed at the uploaded `.env.example` | FAIL "…which `civitai app submit` UPLOADS" + "IN THE PACKAGE this render produced" |
| a new `VITE_SECOND_TOKEN=` added, unclassified | FAIL "…is in NEITHER ledger in this file" |
| `.env.*.local` deleted from `gitignore.tmpl` | FAIL "…the scaffolded .gitignore does NOT ignore" (for **both** secret keys) |
| the destination annotation line deleted (the #380 shape) | FAIL "does not name WHERE the real value goes" |
| `.env.example` removed from `keptEnvFiles` | FAIL at the matcher **CONTROL** (not the count floor — corrected after review) |
| templates stop emitting an uploaded dotenv file | FAIL at the **count floor** |
| `!.env.development.local` appended to `gitignore.tmpl` | FAIL — `git check-ignore … NOT ignored` |
| a real-looking token pasted into `VITE_LIVE_BLOCK_TOKEN=` | FAIL — `NON-EMPTY value … in a file civitai app submit UPLOADS` |
| a new `env.sample.tmpl` holding a key | FAIL — `NEITHER ledger` (and it renders DOTTED) |
| `outputName`'s dotenv rewrite reverted to its enumeration | FAIL — `UPLOADED with an UNDOTTED name` |

**What it cannot see, measured rather than assumed:** it reads the destination
annotation, not the prose. I appended *"(or just paste it here for a quick
test)"* to the fixed block, leaving the annotation intact — the structural guard
**passed**. That is why the second test, `TestUploadedDotenvIsTheReviewedCopy`,
pins the *words* of every uploaded dotenv file: the same mutation reddens it. Its
own controls are measured too — a pure re-wrap of a comment line **passes**
(markers are stripped and whitespace collapsed before comparison), an added
sentence **fails**. My first version of that normalisation did not strip the `#`
markers and a pure re-wrap failed; the comment claiming otherwise was wrong until
the code was fixed to match it.

Item 28's finding drove this shape: no banned-phrase list is attempted, because
two of them lost to paraphrases on the money copy.

## Prose the code contradicted (issue's 🟢 item)

`pkgzip.go` (3 places) and the README's dotenv table both said the money template
"tells you to paste a real `VITE_LIVE_BLOCK_TOKEN` into `.env.development`". It
never did: the scaffolded `.env.development` carries no such instruction — the
file that carried it was `.env.example`, the allow-listed one. Corrected, and the
README dotenv golden re-approved with `-update` (the diff is exactly those two
table cells; the new text asserts nothing about a file's *contents* and says
outright that the routing is documentation, not a check).

## Round 2 — three surviving mutants from the blind audit (`4b0ced2`)

- **The git leg was a reimplementation and it lied.** `path.Match` over the
  `.gitignore` lines ignores negation, anchoring and last-match-wins, so
  `!.env.development.local` left the guard asserting "ignored" about a file real
  git calls NOT ignored. Now **`git check-ignore`** against the rendered tree,
  with global/system config neutralised and inherited `GIT_DIR`/`GIT_WORK_TREE`/
  `GIT_INDEX_FILE`/`GIT_COMMON_DIR`/`GIT_CEILING_DIRECTORIES` scrubbed. `git`
  missing is a **failure, not a skip**. The scrub has its own pair: hostile
  `GIT_DIR` → passes as written, fails with the scrub removed.
- **Nothing checked the uploaded declaration is EMPTY.** A pasted token with the
  annotation intact passed everything; only the golden reddened, and the
  golden's instruction asked about *sentences*. A non-empty secret declaration in
  an uploaded file is now an error — the one deterministic check that makes a
  token in an upload impossible rather than discouraged.
- **"Add a dotenv file and this follows automatically" was false.**
  `scaffold.outputName` dot-prefixed three enumerated names, so `env.sample.tmpl`
  would have rendered UNDOTTED, invisible to `pkgzip` *and* to this guard's own
  filter — while `.env.sample` sat in `keptEnvFiles` for a name the scaffold
  could not emit. The rewrite is now the rule, and the guard fails on any
  undotted `env.*` file in the package.

Re-gated after each: `make ci` 19 ok / 0 FAIL, `ci-shallow` 19/19,
`golangci-lint 2.12.2` 0 issues, `gofmt` 0 files, guard still RED at `e433b91`
for the same single reason.

## 🔴 Disclosure: existing projects keep the old text, and nothing detects it

Templates apply at `civitai app create` only. Every already-scaffolded project
keeps the *"Paste it here"* copy — possibly with a real token already in
`.env.example` — and the CLI cannot see it: `internal/validate` has **0** dotenv
references, `internal/antipattern` has **0**, and `app submit` prints only what
it EXCLUDES (`app_submit.go:35-36`), never naming the three dotenv files it
uploads. So `submit` will still ship a pasted token silently for those projects.
**No detector is built here** — that is filed on #380 with candidate designs. The
README now carries hand-check + re-mint advice (`README.md:1217-1225`).

## What I BELIEVE but did not verify

- That a moderator reviewer would in fact read a leaked token — I verified the
  file is uploaded, not what happens to it afterwards.
- That no *other* surface still invites the paste. I grepped the templates, the
  README, and every command `Long`; what I found is fixed
  (`env.example.tmpl`, `dev-transport.ts.tmpl`). A grep is not a proof of
  absence, and prose inside the scaffolded `README.md.tmpl` — which is also
  uploaded — is covered by neither new guard.
- That the `-update` re-approval of the README golden is *correct* rather than
  merely deliberate: I read the diff against the rule its test states, which is
  the process, not an independent check.

Gating note: **`template-page-money` reports but does not gate** on `main`
(required contexts measured by the reviewer: `pins-vs-published`,
`scaffold-currency`, `build-test`, `ready-ack-runtime`, `template-page-vite`).
Relevant because this PR edits that template. I did not re-measure it myself.

## Not fixed here (left in #380)

The remaining 🟢 items: the README's "always included" over-claim (safe
direction), `generate_img2img_test.go`'s `workflowLines` ledger reading only
`generate.go`, the absence of a per-package list of *excluded* files, and
`.env-cmdrc` deserving the residual list ahead of `.envoy.yaml`. None of them
share a surface with this change.


ZacxDev:
Follow-up commit `2f0ecc4` (self-audit of the guard, before review):

The classification ledger's key matcher was `^[A-Z][A-Z0-9_]*=` — dotenv
CONVENTION, not a rule. A key spelled `vite_side_token=` in the uploaded
`.env.example` was invisible to **every** check in the file, including the ledger
whose entire job is to notice a credential that was added. That is the "can it
pass while the hazard exists in a different shape" question answered wrongly, so
it is fixed rather than filed as a residual.

Measured both directions after widening to `^[A-Za-z_][A-Za-z0-9_]*=`:

- lower-case `vite_side_token=` appended to the scaffolded `.env.example` →
  `🔴 … declares vite_side_token, which is in NEITHER ledger in this file`
- the same file passed with the old matcher
- the origin/main red is unchanged: still exactly one failure, on
  `VITE_LIVE_BLOCK_TOKEN`

Re-gated after the change: `make ci` rc 0, 19 ok / 0 FAIL; `golangci-lint 2.12.2`
0 issues. The first push's checks were all green (12/12 terminal, all pass)
before this commit landed, so this push re-runs them.

ZacxDev:
## Audit round 2 — three surviving mutants fixed (`4b0ced2`), each re-verified

Each fix was re-run against the **whole** battery, not just its own mutant. Every
mutant below is applied to a fresh copy of the branch and discarded after.

### 1. The git leg was a reimplementation, and it lied (M19)

`gitIgnores` was a `path.Match` loop over the `.gitignore` lines. That shape
cannot be made correct — `.gitignore` has negation, anchoring, directory-only
patterns, `**`, and last-match-wins ordering. Replaced with **`git check-ignore`
run against the rendered tree**, with `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM`
neutralised and the inherited `GIT_DIR`/`GIT_WORK_TREE`/`GIT_INDEX_FILE`/
`GIT_COMMON_DIR`/`GIT_CEILING_DIRECTORIES` **scrubbed** from the child env.
`git` missing is a **failure, not a skip** — a skipped safety leg is exactly the
vacuous green this file exists to prevent.

```
M19  !.env.development.local appended to gitignore.tmpl
--- FAIL … 🔴 VITE_LIVE_BLOCK_TOKEN is sent to ".env.development.local", which
    `git check-ignore` says is NOT ignored in the scaffolded project
    (fires for CIVITAI_HOST_KEY too)
```

The env scrub has its own pair, because a guard against the environment is
itself a claim: with a hostile `GIT_DIR=<other repo>/.git GIT_WORK_TREE=<other
repo>` the test **passes** as written, and **fails** with the scrub removed
(oracle answering about the wrong repository). So the scrub is load-bearing, not
decoration.

### 2. The uploaded declaration was never checked EMPTY (M9)

Added the one deterministic check in the file: a **non-empty value for a secret
key in an uploaded dotenv file is an error**. Not implied by anything else — with
the annotation intact, a pasted token passed every other assertion.

```
M9   VITE_LIVE_BLOCK_TOKEN=eyJhbGciOiJSUzI1NiJ9.REAL.LOOKING
--- FAIL … 🔴 declares VITE_LIVE_BLOCK_TOKEN=… — a NON-EMPTY value for a dev
    block token — a real one SPENDS the author's Buzz in a file `civitai app
    submit` UPLOADS …
```

The golden's re-approval instruction is fixed too: it asked about *sentences*,
and a bare `KEY=<token>` is not a sentence. It now asks about values, and says
the value question is enforced rather than left to the reader.

### 3. "Add a dotenv file and this follows automatically" was false (M14)

`scaffold.outputName` dot-prefixed **three enumerated names**, so `env.sample.tmpl`
would render as the undotted `env.sample` — invisible to `pkgzip.isExcludedFile`
*and* to my own `.env` filter. Same gap meant `.env.sample` sat in `keptEnvFiles`
for a name the scaffold could not emit. The rewrite is now the rule
(`env.*` → `.env.*`), **and** the guard fails on any undotted `env.*` file that
reaches the package, so a regression at either end is caught at the other.

```
M14  a new env.sample.tmpl holding SOME_SIDE_TOKEN=
--- FAIL … 🔴 page-money/.env.sample declares SOME_SIDE_TOKEN, which is in
    NEITHER ledger in this file
     (and it now renders DOTTED: golden line reads `page-money .env.sample`)

M14b the enumeration restored in outputName, same new template
--- FAIL … 🔴 page-money: env.sample is UPLOADED with an UNDOTTED name.
    `scaffold.outputName` must rewrite `env.*` to `.env.*` …
```

### Full battery after all three fixes (11 mutants, each killed by its own error)

| mutant | verdict |
| --- | --- |
| M0 unmutated | **ok** |
| M19 gitignore negation | FAIL — `git check-ignore … NOT ignored` |
| M9 token pasted into the declaration | FAIL — `NON-EMPTY value` |
| M14 new `env.sample.tmpl` | FAIL — `NEITHER ledger` |
| M14b `outputName` enumeration restored | FAIL — `UPLOADED with an UNDOTTED name` |
| M1 destination → `.env.example` | FAIL — `which civitai app submit UPLOADS` |
| M2 unclassified lower-case key | FAIL — `NEITHER ledger` |
| M3 `.env.*.local` removed from `.gitignore` | FAIL — `NOT ignored` (both keys) |
| M4 annotation deleted (the #380 shape) | FAIL — `does not name WHERE` |
| M5 `.env.example` off `keptEnvFiles` | FAIL — matcher CONTROL |
| M6 templates stop emitting uploaded dotenv | FAIL — COUNT FLOOR |
| G1 golden, pure re-wrap | **ok** |
| G2 golden, appended invitation | FAIL |

**Correcting my own earlier evidence:** I previously attributed M5's kill to the
count floor. It is not — the matcher control at the top fires first, as the
auditor said. The count floor is reachable by the *other* emptying (M6, deleting
`env.example.tmpl` + `env.production.tmpl`), which is now measured above and
recorded in a comment at that line.

### Gate, re-run after every change

`make ci` rc 0 — **19 ok / 0 FAIL / 1 no-test-files**; `./scripts/ci-shallow.sh`
**19/19, 0 timeouts**; `golangci-lint 2.12.2` **0 issues**; `gofmt -s -l .` → 0
files. Guard still **RED at `e433b91`** for the same single reason
(`VITE_LIVE_BLOCK_TOKEN … does not name WHERE the real value goes`), GREEN here.

### Disclosure: existing projects keep the old text (no detector in this PR)

Templates apply at `civitai app create` only. Every already-scaffolded project
keeps the *"Paste it here"* copy — possibly with a real token already sitting in
`.env.example` — and **nothing in the CLI detects it**: `internal/validate` has
no dotenv reference, `internal/antipattern` has no rule, and `app submit` prints
only what it EXCLUDES, never naming the three dotenv files it uploads. So
`civitai app submit` will still ship a pasted token silently for those projects.
Per the review, no detector is built here; it is filed back onto #380. The README
now carries the hand-check + re-mint advice (`README.md:1217-1225`).

### README lines touched in this round (for the integration-branch re-run)

- `README.md:1205` — the `.env.example` table cell: "Nothing enforces that" →
  "A test pins the *scaffolded* line empty; nothing checks the copy in **your**
  project".
- `README.md:1217-1225` — new ⚠️ paragraph, the migration/rotation advice.
- `internal/pkgzip/testdata/readme_dotenv_section.golden.txt` re-approved with
  `-update`; the diff is exactly those two edits.

### Nits

Taken: the `alsoUploaded` leg is now labelled corroboration-only in the code; the
count floor carries its reachability note; the file moved to `package
scaffold_test` (it uses only exported API, and now proves it). Left: the
annotation's single-line rigidity — it errs toward crying wolf, and the failure
message prints the exact required line.

Noted on gating: **`template-page-money` reports but does not gate** on `main`,
which is worth knowing since this PR edits that template. I did not re-measure
the required contexts myself this round — that number is the auditor's.

PR #386, opened 2026-08-11
fix(appapi): a listing 400's subject came from the HTTP VERB, so two ingest routes claimed a change that never happened (#374)

### LINKED ISSUE #374: listingOp is keyed on HTTP verb, not on what the route does — and the mutation-400 mislabel is invisible to all 1993 tests
From the delta re-audit of #371. All measured, not inferred. Two of that audit's findings are being folded into #371 itself (the `app metrics` twin, and a vacuous invariant guard); these are the remainder.

## 🟡 The `listingOp` trichotomy is keyed on HTTP verb, not on what the route does — two ingest routes land on the wrong side

`internal/appapi/listing.go:255` (`IngestAssetFromDataURI`) and `:331` (`persistAssetImage`) go through `trpcMutation` → `listingOpChange`, so a 400 reads *"the server rejected this store-listing change (400) … fix the value and retry; `civitai app listing status` shows the listing as it stands"*. **Neither touches a listing** — they create an Image row; `setIcon`/`setCover`/`addScreenshot` is what attaches it. By the classification's own stated criterion the message is false.

Sharpest case, `IngestAssetFullRes` (`:311-334`): **one** user action runs `MintImageUpload` → PUT → `persistAssetImage`. A 400 at step 1 says *"no listing was changed"*; a 400 at step 3 says *"fix the value … `app listing status`"*. Same action, two different stories.

The doc comment at `:413-418` states the classification is exhaustive (*"serves three kinds of request"*). It is not.

**Frequency:** the data-URI icon path caps around 2 MiB server-side and rasterizes, so an oversize or unsupported icon on `app listing set-icon` is a first-attempt failure, not an exotic state.

## 🟡 Two of the three `listingOp` call sites have no reachability test — and the dangerous direction is invisible

Measured, both mutants **SURVIVED**:

- `MintImageUpload`: `listingOpUpload` → `listingOpChange` — survived the whole `appapi` suite.
- `trpcMutation`: `listingOpChange` → `listingOpRead` — **survived both `./internal/appapi` (236 tests) and `./internal/cmd` (1757 tests)**. That mislabels every `setIcon`/`setCover`/`addScreenshot`/`submitListingRevision` 400 as *"the server rejected this store-listing lookup … nothing was changed"* — telling a user nothing changed about a **mutation that may have partially applied** — and the entire suite is blind to it.

`TestListingBadRequestSubjectFollowsTheOperation`'s "change" case passes the constant to `listingError` **directly**, so it cannot see a call-site error. `TestListingReadBadRequestIsReachableAsARead` (`:113`) is the right pattern — it just exists for one op of three.

## 🟢 Smaller

- **`pullReviewAdvice`'s `strings.ToLower` is unpinned.** Dropping it survives, because every fixture is already lowercase — a server sending `Rejected` would fall through to the in-review default, untested. The `ToLower` is correct (it matches `appblocks.go:616`); note `app_status.go:161` compares `s.Status == "rejected"` exactly, a pre-existing split worth reconciling.
- **The neither-selector 404 advice names the command the user just ran.** `internal/appapi/appblocks.go:844` advises *"list what you have submitted with `civitai app status`"*, and the only caller with no selector is bare `civitai app status` (`app_status.go:76`). Low frequency — a 404 there means the route is gone — but it is the "remedy does not leave the state it complains about" shape.
- **The read arm's advice is specific to one of the three reads it serves.** `internal/appapi/listing.go:468` says *"check the app you named"* — right for `getMyListingForApp`, wrong for `getMyListingForEdit` (takes a listingId) and `getAssetScanStatuses` (takes imageIds). Low frequency: both take CLI-derived ids.
- **README over-precision.** `README.md:2470` says *"a version **is** approved"*; the code returns the server's answer when any submission has **ever** carried an appBlockId — i.e. *was* approved. (Line number predates #368's merge; re-locate.)
- **`AppBlockID != nil && != ""` → `!= nil` survives** — pre-existing, outside that PR's delta, noted because it is one of the four states the README now enumerates.

## Load-bearing assumption worth an independent check

*"The server nulls `appBlockId` until approval"* is what makes the whole `app pull` disambiguation sound, and it is a **vendored server claim with only an in-repo citation** (`resolveAppBlockID`, `app_metrics.go:208-211`) — the same assumption restated, not independent evidence. It could not be verified against a real server (no reachable moderator queue). If it is ever false, the disambiguation misfires rather than degrading.

ZacxDev:
Fixes #374.

## The defect

`listingOp` was picked at each of the three call sites, from the **HTTP verb** — `trpcQuery` → read, `trpcMutation` → change, `MintImageUpload` → upload. Both asset-ingest routes are POSTs, so both landed on "change":

- `appListings.ingestAssetFromDataUri` and `appListings.persistAssetImage` create an **Image row** and touch no listing; `setIcon` / `setCover` / `addScreenshot` is what attaches it. A 400 there said *"the server rejected this store-listing change … `civitai app listing status` shows the listing as it stands"*.
- `IngestAssetFullRes` runs mint → PUT → persist as **one** user action, and told two different stories depending on which step 400'd.

The mirror direction is the dangerous one and was equally unpinned: flip `trpcMutation`'s constant to `listingOpRead` and every `setIcon` / `setCover` / `addScreenshot` / `submitListingRevision` 400 becomes a *lookup* that "changed nothing" — about a mutation that may have **partially applied**.

## The fix

The op travels **with the route**: `listingRoute{path, op}`, stated once where the route is declared, instead of being re-derived at three call sites. `listingOpUpload` → `listingOpIngest`, now covering the mint, the data-URI ingest and the persist.

The zero op is `listingOpUnclassified`, deliberately **not** `read`: `listingRoute{path: "…"}` is a keyed literal Go accepts with no op, and if the zero meant `read`, a route added with the op forgotten would ship silently claiming "nothing was changed" — #374's dangerous direction re-created by omission. Its 400 arm asserts nothing about the listing.

## What I VERIFIED (measured, with output)

**1. I re-ran the issue's two mutation claims myself on `origin/main` (e433b91), before changing anything.** Both survived, as claimed:

| mutant (on origin/main) | suites | result |
| --- | --- | --- |
| `trpcMutation`: `listingOpChange` → `listingOpRead` | `./internal/appapi` + `./internal/cmd` | **SURVIVED** — 2119 `--- PASS`, 0 `--- FAIL`, rc=0, no timeout panic |
| `MintImageUpload`: `listingOpUpload` → `listingOpChange` | `./internal/appapi` | **SURVIVED** — 235 `--- PASS`, 0 `--- FAIL` |

(The issue said 236 appapi tests; I measure 235 today. Counted from `--- PASS` lines, not an exit code.)

**2. Red at `origin/main`, green at HEAD.** The new tests cannot compile against `origin/main` (the signatures changed), so I ported their behavioural claims — same client methods, same 400s, same literal expected phrases — onto `origin/main`'s API and ran them there:

- **Red at `origin/main`**: `TestBasePortSubjectIsReachablePerRoute` → 11 subtests PASS, **2 FAIL** (`ingestAssetFromDataUri`, `persistAssetImage`); `TestBasePortIngestFullResTellsOneStory` → **FAIL**. Actual output:
  ```
  --- FAIL: .../ingestAssetFromDataUri
      missing "image-upload request"; got: the server rejected this store-listing change (400): refused: data uri too large — fix the value and retry; `civitai app listing status` shows the listing as it stands
  --- FAIL: TestBasePortIngestFullResTellsOneStory
      one action, two stories:
        mint:    the server rejected the image-upload request (400):  — no listing was changed; check the image and retry
        persist: the server rejected this store-listing change (400):  — fix the value and retry; `civitai app listing status` shows the listing as it stands
  ```
- **Green at HEAD**: `./internal/appapi` 253 `--- PASS` / 0 `--- FAIL`; `go test ./...` 19 packages `ok`, 0 FAIL.

**3. Mutation-tested my own guard: 15 mutants, 15 killed, each with a specific error.** Control (pristine) = 253 PASS / 0 FAIL. Every mutant was applied to a per-agent scratch copy, never the branch.

| # | mutant | killed by | the specific error |
| --- | --- | --- | --- |
| M1 | `setIcon` change→read | 6 FAIL | `route …setIcon is declared read, but it does a change` |
| M2 | data-URI ingest→change | 4 FAIL | `route …ingestAssetFromDataUri is declared change, but it does a ingest` |
| M3 | persist ingest→change | 3 FAIL | `persist-step 400 missing "image-upload request"` |
| M4 | mint ingest→change | 5 FAIL | `route /api/v1/image-upload is declared change, but it does a ingest` |
| M5 | `getMyListingForEdit` read→change | 2 FAIL | `sent as GET but classified change` |
| M6 | `submitListingRevision` change→read | 2 FAIL | `is declared read, but it does a change` |
| M7 | `setCover`'s **path** swapped to `setIcon`'s | 3 FAIL | `case "setCover" is wired to route ".../setIcon" … want proc "appListings.setCover"` |
| M8 | the ingest arm deleted from `listingError` | 8 FAIL | `missing "image-upload request"` ×3 routes |
| M9 | a new route declared with no reachability case | 1 FAIL | `route "….someNewThing" is declared but no case … drives a 400 through it` |
| M10 | **instrument negative control**: AST walker matches nothing | 1 FAIL | `parser found only 0 listingRoute declarations … it is not reading what it thinks` |
| M11 | one case deleted from the table | 2 FAIL | `route "….setCover" is declared but no case …` |
| M12 | **double flip** (source + this file's `wantOp`) on `setIcon` | 6 FAIL | killed by the *other*, independently-written test in `procname_leak_test.go` |
| M13 | **double flip** on `submitListingRevision` (no second test covers it) | 2 FAIL | `classified a lookup but the CLI sends it as POST — a POST can change data` |
| M14 | zero op made `read` again | 1 FAIL | `the zero listingOp must not be 'read' …` |
| M15 | `setIcon` declared with its op forgotten | 7 FAIL | `route …setIcon is declared unclassified, but it does a change` |

Two of these **survived my first draft** and I want that on the record: M5 and M6 passed the full 252-test package because the test derived its expectation from `route.op` — the very field under test. The table now carries a hand-written `wantOp` per case, plus a cross-check that owes nothing to either (the verb the client actually sends: a POST is never a lookup), which is what kills M13.

**4. Instrument validation.** `golangci-lint` **v2.12.2** (the CI pin) reports `0 issues` on the tree — and I proved it can go red here first: a planted `declared and not used` gave `1 issues: * typecheck: 1`. `make ci` rc=0, 19 packages ok. The AST ledger's positive control (`≥13 routes`, must include `setIcon` and `ImageUploadPath`) is what M10 trips.

**5. Exit codes do not move.** `TagStatus` maps on status alone, so re-wording a 400 cannot change its class; `errors.Is(err, civitai.ErrBadRequest)` (→ exit 2) is now asserted for all 13 routes (AGENTS item 7).

**6. README.** Grepped the whole file for every surface naming this behaviour. No prose quoted the old message, and the exit-code table / `exitcodes_doc.go` are untouched because the classification did not move. I added **two Troubleshooting rows** — one per 400 subject — because that section is the "look the error message up here" index. `TestREADMETroubleshootingSymptomsExistInTheSource` (the drift detector) and `…RowsAreAttributedToTheEmittingCommand` (ledger-driven; I did **not** edit it) both pass.

## What I BELIEVE but did not verify

- **That the three ops are the right classification of each route** — that `persistAssetImage` writes only an Image row, that `getMyListingForEdit`'s shadow-revision side effect should still read as a lookup. This mirrors `civitai/civitai`'s procs and the existing in-repo comments; I did not run any of it against a real server, and no test here can tell a correct classification from a confidently wrong one that both sides of the ledger agree on.
- **That "may have partially applied" is true of a `setIcon` 400.** It is the safe assumption the wording is built on, not something I observed.
- The base-tree red/green matrix is a **port** of the new tests onto `origin/main`'s API, not the same file — the new file cannot compile there.

## Found, not fixed (all from #374's 🟢 list, out of this PR's scope)

- **The read arm's advice is specific to one of the three reads it serves.** `"check the app you named"` is right for `getMyListingForApp`, wrong for `getMyListingForEdit` (takes a listingId) and `getAssetScanStatuses` (takes imageIds). This PR fixes the *subject* (which op), not the *advice*; the advice is a second axis and both remaining reads take CLI-derived ids.
- `pullReviewAdvice`'s unpinned `strings.ToLower`, the neither-selector 404 advice naming the command just run, the README "is approved" over-precision, and `AppBlockID != nil && != ""` — all outside `internal/appapi/listing*.go` and outside this PR.
- **No AGENTS.md decision item added.** Keying the op on the route rather than the verb does not "look wrong", which is what that list is for, and three sibling agents are in flight against files that would collide there. The rationale lives in `listing.go`'s comments and this PR.

## Residual I could not close

A **coordinated** edit of both the declaration in `listing.go` and this file's `wantOp` is a spec change, not a mutant, and for a route no *other* test independently pins it would go unnoticed for the change↔ingest pair (M13 is only caught because read↔change disagrees with the verb). Closing that would need a signal outside this repo — the server's own route semantics.

---

## Correction to a claim I made in this PR's audit-response comment

In the round-2 comment I wrote that I *"removed the 'edit the other place' framing from the failure text."* **That is wrong, and it was a claim about my own diff.** No such string existed in `bf6e2b0` — its failure message reads `route %s is declared %s, but it does a %s`. I was paraphrasing the round-1 audit's *characterisation* of the guard ("your guard's failure message names both places that must agree, so the cheapest way for a future agent to green it is to edit `wantOp`") and wrote it as if it were a literal string I had deleted.

What I actually did was **additive**: the new `TestListingRouteOpsAreExactlyThisSpelledSet` doc comment says *"this ledger is not the thing to fix first"* and directs the reader to the wording tests. The substance holds; the description of how it was achieved did not. Caught by the delta re-audit, not by me.


ZacxDev:
## Audit round 2 — all five 🟡 addressed (9664c95)

The audit was right on every count, and finding 1 was a defect in my own guard, not a documentation gap. Re-ran the whole battery; a fix resets the gate.

### 1. The coordinated re-label: 6 of 7 change routes were open, now 0

Reproduced exactly as reported. `wantOp` lives three lines from `route.op`, so a two-line edit re-labels a route; only `setIcon` died, because `procname_leak_test.go`'s wording table was the one place naming a route independently — and it named only `setIcon`.

Two fixes, both in a **different file** from `wantOp`:
- `TestListingBadRequestSubjectFollowsTheOperation` now carries **all 13 routes**, each with a spelled-out sentence (`setCover attaches a cover to the listing`, …) and per-case literal `want`/`notWant`.
- New `TestListingRouteOpsAreExactlyThisSpelledSet`: the AST reads the op **identifier as written in the source** and compares it to a hand-written set, bidirectionally.

A re-label now has to be written down three times, in two files, and each site says why it matters. I also removed the "edit the other place" framing from the failure text — it now says: fix the wording tests, confirm the new subject is *true* of the route, and only then record it here.

**All six previously-surviving coordinated flips now die** (control = 263 PASS / 0 FAIL):

| coordinated flip (declaration + `wantOp`) | result | error |
| --- | --- | --- |
| `setCover` → ingest | 3 FAIL | `procname_leak_test.go:173: missing "store-listing change"; got: the server rejected the image-upload request (400): … — no listing was changed` + `listing_op_test.go:489: route ".../setCover" is declared listingOpIngest at listing.go:52, but this ledger says listingOpChange.` |
| `addScreenshot` → ingest | 3 FAIL | same pair, `listing.go:53` |
| `removeScreenshot` → ingest | 3 FAIL | same pair, `listing.go:54` |
| `reorderScreenshots` → ingest | 3 FAIL | same pair, `listing.go:55` |
| `beginListingRevision` → ingest | 3 FAIL | same pair, `listing.go:56` |
| `submitListingRevision` → ingest | 3 FAIL | same pair, `listing.go:57` |
| `setIcon` → ingest (was already dying) | 5 FAIL | + `listing_test.go:291` |

### 2. The ledger was blind to sibling files

Confirmed. `declaredListingRoutes` now parses **every non-test `.go` file in the package** (`os.ReadDir` + `ParseFile` per file — not the deprecated `ParseDir`), with a positive control on the file sweep itself (`< 2 files` → fatal, naming the cwd). Consts resolve package-wide.

Your repro — `listing_extra.go` with `listingRoute{path: "…siblingProc"}` plus a real client method — now **dies, 2 FAIL**:
```
listing_op_test.go:430: route ".../siblingProc" (declared at listing_extra.go:5) has no case in listingRouteCases() driving a 400 through it — its op is unpinned
listing_op_test.go:501: route ".../siblingProc" (declared (no op given) at listing_extra.go:5) is not in this ledger — classify it here, and give it a case in listingRouteCases()
```
The comment that claimed otherwise is rewritten to say what the code does.

### 3. The ledger reddened on correct Go

Fixed: keyed literals are read **by key** (`path`, `op`), and a missing `op` key is what leaves the op name empty — which is exactly what the ledger should report, rather than a parse failure. Every remaining parse error now names the remedy and says *do not delete the ledger*.

**Negative control, and it must stay green**: `listingRoute{path: "…setCover", op: listingOpChange}` → **263 PASS / 0 FAIL, rc=0**. That's the one entry in the battery that is supposed to pass, and it does.

### 4. The PUT step — README narrowed, fix filed as #388

I re-measured rather than taking it on report. Against an `httptest` server 400ing the presigned PUT:
```
PUT-step error: image upload PUT failed (HTTP 400): EntityTooLarge
errors.Is(err, civitai.ErrBadRequest) = false   → exitCode()'s default arm → exit 1
```
Not fixed here, as you asked. **README `#2604-2607`** (the four rows in *Validating and submitting*) now:
- narrows the ingest row to "**No listing was changed** by any of those three" and points at the PUT row;
- adds a row for `image upload PUT failed` stating plainly that it exits **`1`, not `2`**, and that this is tracked in **#388**;
- adds the missing **read-subject** row (`the server rejected this store-listing lookup (400)`) — the PR body had promised one row per subject and shipped two of three;
- adds `beginListingRevision` to the change row's enumeration.

Both README guards pass, including the drift detector that requires each symptom to exist verbatim in source.

### 5. `getMyListingForEdit` — rule restated, measurement filed as #389

Restated the rule as **what the caller asked the route for**, which is the option you offered and the one that keeps the common failure correct: a rejected `listingId` must not be answered with *"fix the value and retry; `app listing status` shows the listing as it stands"* about a listing that was never resolved. The residual is now stated in the `listingOp` doc rather than implied — including that the CLI **cannot see** whether the server opened a revision before rejecting, so "nothing was changed" is a claim about the *request*, not a proof about the database. **#389** names the measurement that would settle it and says not to re-label before it exists.

### Nits taken

- `listing.go` no longer claims the predicate was "wrong at two of three sites" — it was wrong at **one** site (`trpcMutation`) answering for two routes, and the comment now says *that is why nobody spotted it: two of the three sites read correctly*.
- The unclassified 400 arm now names the next step: `— this is a CLI bug (the route is unclassified); please report it at https://github.com/civitai/cli/issues`.
- The ledger-shrink half you flagged as unreported is now in the battery: deleting a declared route → **3 FAIL**, `parser found only 12 listingRoute declarations … it is not reading what it thinks` and `positive control: expected the case table to cover every route, got 12`.

### Nits NOT taken (deliberate)

- **"fix the value and retry" on all 7 change routes.** Every one of them does carry a user-supplied or user-named value (a screenshot id, an ordered id set, a changelog, a listing id), so the sentence is true, if bland. Re-wording it splits the change arm and moves a published README string; not worth it in this PR.
- **`const` → `var` for the routes.** They are unexported, nothing mutates them, and `const` cannot hold a struct. A `func` returning them would buy immutability at the cost of readability.

### Re-verified after the fix (the gate reset)

- `./internal/appapi`: **263 `--- PASS`, 0 `--- FAIL`**.
- `make ci`: rc=0, **19 packages ok**.
- `golangci-lint v2.12.2`: **`0 issues`**.
- Mutation battery: **22 runs — 1 control green, 20 defect mutants all killed, 1 negative control green.** Every anchor asserted to match exactly once; counts read from `--- PASS`/`--- FAIL` lines, never an exit code; checked for `panic: test timed out` (none).

Your merged-tree numbers predate this commit — worth a re-run on the integration branch. README lines touched: **2604–2607** (four rows in the *Validating and submitting* Troubleshooting table); no other file in README moved.

PR #392, opened 2026-08-11
fix(workflows): `list` had the failure reason in hand and rendered `failed … 0/1` (#382)

### LINKED ISSUE #382: workflows list drops the failure reason too — it is at steps[].errors there, not steps[].output.errors, and the string names the provider
`#367` fixed the failure reason on `workflows get` and `generate`. It explicitly ruled `ListedStep` — the `workflows list` shape — out of scope as "unverified", through four review rounds. That was wrong: **the listing endpoint carries the reason too, at a different path, and the CLI drops it.**

## Measured

`workflows list --json` emits the raw payload, so this is a free read. Over 30 real workflows:

| status | steps with a populated `errors` |
|---|---|
| 5 × `failed` | 1 each |
| 1 × `failed` | 0 |
| 1 × `canceled` | 0 |
| 23 × `succeeded` | 0 |

Same distribution as the `getWorkflow` endpoint.

## The path is DIFFERENT, which is why it was missed

```
workflows get   →  steps[].output.errors     (output is an OBJECT)
workflows list  →  steps[].errors            (output is an ARRAY — errors is a SIBLING of it)
```

Confirmed: in the listing shape `output` is an array at every step sampled, so `output.errors` cannot exist there — the normalizer lifts `errors` up to the step. The listed step's own keys:

```
["$type","completedAt","errors","images","metadata","name","output","queuePosition","status","timeout"]
```

`ListedStep` (`internal/genapi/workflows.go`) models `$type`, `name`, `status`, `metadata`, `output`, `images` — and nothing reads `errors` anywhere in the listing path.

## The listing string is strictly MORE informative

The two endpoints do not return the same text for the same workflow:

```
workflows get :  "Could not generate images with the given prompts and images. Please try again with different inputs."
workflows list:  "Google Gemini: Could not generate images with the given prompts and images. Please try again with different inputs."
```

The listing form **names the provider**. On the `#351` reproduction that cost hours to root-cause, `Google Gemini:` is a material part of the diagnosis — and it is the half the CLI currently cannot show at all, on either surface.

## Consequence

`civitai workflows list` renders a failed workflow as `failed … 0/1` with no reason, while the server has handed it one in the same response. A user scanning the list for why a run failed sees nothing, and the per-workflow command they are sent to next returns the *less* specific string.

## Work

1. Add `Errors []string \`json:"errors"\`` to `ListedStep` — **step level, not under `output`**; verify against a real payload rather than assuming symmetry with the `get` shape.
2. Decide what `workflows list` shows. It is one line per workflow, so the full sentence does not fit; a marker plus the reason on `workflows get` may be right, or a truncated form. This is a UX call, not just a struct change.
3. Consider whether `workflows get` should surface the provider prefix the listing endpoint has and the get endpoint lacks — that is a question about which upstream field the CLI reads, not about reformatting a string.

## Constraints

Same as `#367`: pass the string through verbatim (AGENTS.md item 13 — no classifying, no mapping, no matching on wording), `safeTerm` it before it reaches a terminal, say nothing about the charge (item 28), and pin exit codes if any error path changes (item 7). `internal/cmd/generate_output.go`'s key-injectivity and within-step de-duplication work from `#381` is the reference for how repeated/near-identical reasons are handled — reuse it rather than writing a second rule.

Reproduction is free in both directions: reading existing failed workflows costs nothing, and failed generations are refunded.

ZacxDev:
Fixes #382. Prior art: #367 (`47d2f71`) and its follow-up #381 (`36cd097`).

`civitai workflows list` rendered a failed workflow as `failed … 0/1` with no
reason, while the server had handed one back **on the same response**.
`ListedStep` had no field for it, so it was dropped at unmarshal.

## The wire shape — measured, not assumed

Verified against a live payload before the struct was pinned (a read; **spends
nothing**, 2026-08-11):

```
workflows get   ->  steps[].output.errors   (output is an OBJECT)
workflows list  ->  steps[].errors          (output is an ARRAY -- errors is a SIBLING of it)
```

Measured step keys: `["$type","completedAt","errors","images","metadata","name","output","queuePosition","status","timeout"]`.

Over 30 real workflows: 5 `failed` carried one reason each, 1 `failed` carried
none, 1 `canceled` and 23 `succeeded` carried none. **A failed workflow with no
reason is a measured branch**, so the renderer prints nothing rather than an
empty label.

🔴 **One correction to the issue's write-up: "none" arrives as `null`, not as an
empty array.** `formatGenerationResponse2` emits
`errors: errors.length > 0 ? errors : undefined`, and all 25 unpopulated steps in
the sample were the literal `"errors": null`. All three presentations (`null`,
`[]`, key absent) decode to a nil slice and are pinned; the empty array is pinned
as a shape that must behave, never as the shape to expect.

`internal/genapi/testdata/listed_workflows_redacted.json` is a real page, kept
whole with only identifying values replaced (account id, signed blob URLs,
orchestrator host, transaction ids, resource names), so the hand-written fixtures
are not a closed loop — the failure mode that let #367 rule this surface out.
Leak-checked against the raw capture.

## Layout

```console
WORKFLOW ID                STATUS     CREATED                   COST  OUTPUTS
8753561-20260810224136984  succeeded  2026-08-10T22:41:36.984Z  104   1/1
8753561-20260810223715659  failed     2026-08-10T22:37:15.659Z  0     0/1
    Google Gemini: Could not generate images with the given prompts and images.
    Please try again with different inputs.
```

Full text, wrapped, never truncated, emitted only for rows that carry one.
Strictly-one-line-per-workflow is deliberately given up: `--json` is the
scripting contract, the table is not.

**The rows are aligned in a buffer and the reason lines spliced into the flushed
text.** A tabwriter line carrying no tab *ends* the column block, so interleaving
directly would re-align every group of rows independently and the columns would
jump around the failures. The splice pairs blocks to rows by **line count, not
index**: `safeTerm` keeps `\n` in every server-origin cell, ids included, so one
row can occupy two lines and index pairing would hand a workflow's reason to the
next workflow's row.

## 🔴 The hazard this layout creates

Server free text under a table of rows can **forge a row**. Two mechanisms, both
closed:

- a `\n` in the reason → `indentContinuation` pads every line after the first,
  and the `Fprintf` pads the first. Nothing but a real row can occupy column
  zero.
- a line longer than the terminal → the **terminal** soft-wraps it back to column
  zero, with no newline in the string for `indentContinuation` to see. New
  `wrapServerText` (`internal/cmd/safeterm.go`) hard-wraps to the same
  `findingWrapWidth` the CLI's other long-prose surface uses, reusing
  `wrapTokens` rather than hand-rolling a second filler. It preserves the
  string's own line breaks and collapses only intra-line whitespace runs; the
  words are untouched.

Pinned with a **4-line** reason by asserting the *state* — the only lines at
column zero are the header and the real rows — not the absence of one spelling.
A raw CSI/OSC introducer is pinned separately (`safeTerm`), and a `\t` in a
reason is safe because the reason never passes through the tabwriter.

## De-duplication: #381's rule, not a second one

`dedupeReasons` is now **one function** in `internal/genapi/status.go`, called by
both shapes. `TestBothWorkflowShapesApplyTheSameReasonRule` asserts they answer
identically for identical reason material — the #381 defect (one payload, two
answers on one screen) at a new seam.

**The listing path does NOT need an injective reason-set key, and here is why.**
`distinctReasonSets` exists because `reportExcludedOutputs` compares two outputs'
reason *sets* and suppresses one as already-reported; a colliding key made an
output lose its own cause. Nothing on the listing path compares two rows' sets or
suppresses one — each row renders its own workflow's reasons unconditionally. The
conclusion is pinned rather than asserted:
`TestWorkflowsList_EachRowKeepsItsOwnReasonsUnderSubsetOverlap` uses the
subset-shaped triple that killed the #381 gate, so if a cross-row gate is ever
added, it reddens.

Cross-**step** de-duplication *is* needed and is not defensive: the server
dedupes within a step (`extractStepErrors` ends in `Array.from(new Set(...))`)
and not across them. The 30-workflow sample was all single-step, so the
multi-step case is untested against a real payload — stated, not disproven.

## Provider prefix — INVESTIGATED, NOT FIXED (as instructed)

**`workflows get` is not failing to read a field. The provider name is not in its
payload at all.**

Measured on the same workflow (`8753561-20260810223715659`), both `--json`:

| | string |
|---|---|
| `list` | `Google Gemini: Could not generate images…` |
| `get` | `Could not generate images…` |

The `get` payload contains **zero** case-insensitive occurrences of `google`, and
exactly one `gemini`: `steps[0].input.engine`.

The mechanism, from `civitai/civitai`:
`orchestration-new.service.ts:2447-2449` →
`const engine = params.engine ?? step.input?.engine;` →
`extractStepErrors(step).map(msg => sanitizeProviderError(msg, engine))`.
`sanitizeProviderError` (`provider-errors.ts`) looks the engine up in a
**16-entry `providerNameMap`** (`gemini → "Google Gemini"`) and returns
`` `${provider}: ${msg}` `` unless an alias is already mentioned.
`orchestrator.getWorkflow` hands the raw orchestrator workflow straight back and
runs none of this.

**Two consequences worth recording before anyone "fixes" `workflows get`:**

1. Reproducing the prefix would mean vendoring `providerNameMap` plus the
   double-branding rule — a server table keyed on engine, which is exactly what
   AGENTS.md item 13 forbids on this path.
2. The difference is **not only a prefix**. `sanitizeProviderError` also
   *replaces* any message failing `isLikelySafeMessage` (URLs, paths, serialized
   structures, stack frames, infra/DB names, credentials, >300 chars, multi-line)
   with a generic *"… reported a system error"*. So the two endpoints can differ
   in either direction, and `get`'s raw string can be the more informative one.

`workflows get` is untouched. The README now states the difference and attributes
it to the server.

## Test matrix

**Red at `origin/main`, green at HEAD.** The shipped tests name new symbols and
cannot compile at base — a build failure is a non-verdict, not a red — so the
matrix was measured with a base-compatible probe naming no new symbol:

| | `origin/main` | HEAD |
|---|---|---|
| the reason reaches the screen | **FAIL** (row rendered, reason absent) | PASS |
| every reason line is indented | **CONTROL failure** — no server text renders at base, so the hazard does not exist there | PASS |

The second row is an **invariant guard on new behaviour**, labelled as such, and
is not counted as regression coverage.

New tests — `internal/genapi/listed_failure_reason_test.go`,
`internal/cmd/workflows_list_failure_reason_test.go`,
`TestWrapServerText`, and a `TestGoldenSpendCopy` subtest:

- **Decode**: populated / `null` / `[]` / key absent / two steps two causes / two
  steps one cause / order + shared entry / blank + control-only entries / no
  steps; independence from which blob array the step filled (4 shapes); a reason
  nested under `output` is *not* read (the symmetry assumption that cost #367
  this surface); the real capture, with a control asserting the fixture really
  puts `errors` beside an array-valued `output`.
- **Render**: the reason appears under its own row; nothing (and no legend) for
  the four "none" shapes; a reason stays with its row when an id spans two lines;
  repeated causes across steps said once while the differing one survives;
  verbatim pass-through including the provider prefix; long reasons wrapped and
  not truncated.
- **Forgery**: 4-line reason, column-zero set asserted; control bytes stripped.
- **Alignment**: two independent halves — every row agrees on the column
  positions (catches per-group re-alignment) *and* the rows are byte-identical to
  the same table with no reason in it (catches a row whose width depends on
  having a reason). Neither half sees the other's defect; that was measured.
- The `TestGoldenSpendCopy` set gains `workflows_list_with_reason` — this view
  prints a COST column and gained a new sentence, the residual item 28 names.
  Its comment records that whitespace normalisation makes it blind to the indent,
  so nobody reads it as the forgery guard.

**Mutation results.** Harness controls first, because two instrument failures
have burned this lineage:

| control | verdict |
|---|---|
| must-die | KILLED |
| deliberate no-op (identical text) | **PATCH WAS A NO-OP** (distinct non-verdict) |
| deliberate build break | **BUILD FAILED** (distinct non-verdict) |

| mutant | verdict |
|---|---|
| json key `errors` → `error` | KILLED by `TestDecodeRealCapturedWorkflowPage` |
| first-step-only | KILLED by `…_EveryPresentation` |
| list path skips the shared rule | KILLED by `…_EveryPresentation` |
| list/get shapes disagree | KILLED by `TestBothWorkflowShapesApplyTheSameReasonRule` |
| dedupe removed | KILLED by `…_EveryPresentation` |
| emptiness check inverted | KILLED by `…_EveryPresentation` |
| untrimmed value collected | KILLED by `TestBothWorkflowShapes…` |
| continuation lines unindented | KILLED by `…CannotForgeARow` |
| first line unindented | KILLED by `…CannotForgeARow` |
| no hard wrap | KILLED by `…ALongReasonIsWrapped…` |
| `safeTerm` dropped | KILLED by `…CannotCarryTerminalControlBytes` |
| reason printed before its row | KILLED by `…UnderItsRow` |
| splice paired by index, not line count | KILLED by `…WhenAnIDSpansTwoLines` |
| legend gate inverted | KILLED by `…NoReasonPrintsNothingAndNoLegend` |
| row width depends on having a reason | KILLED by `…ReasonsDoNotDisturbTheTable` |
| per-row tabwriter flush (each row its own block) | KILLED by `…ReasonsDoNotDisturbTheTable` |
| own newlines flattened | KILLED by `TestWrapServerText` |
| width budget widened | KILLED by `TestWrapServerText` |

Three rounds were needed and the first two rounds' *misses were the harness's,
not the code's*, so they are reported rather than hidden:

- `tabwriter.Debug` was a **bad proxy** for per-group re-alignment — it degrades
  the baseline identically, so the property test could not see it. Replaced with
  a per-row flush, which the widened test kills.
- the naive-interleave mutant was **killed for the wrong reason** (the ordering
  tests fire before the alignment test can see it), which is why the per-row
  flush variant exists.
- one mutant read SURVIVED under a filtered `-run` because it was **aimed at the
  wrong test**: it broke `Step.failureReasons`, which `Workflow.FailureReasons`
  no longer routes through. Re-run against the whole suite it is killed by the
  pre-existing #367/#381 guards, and re-aimed it is killed by the intended test.
  No surviving mutant remains.

## Gate

- `make ci` — **passed**. Verbose run counted directly, never from an exit code:
  `=== RUN 3692`, `--- PASS 3688`, `--- FAIL 0`, `--- SKIP 4`, `panic: test timed
  out` **0**, `build failed` **0**. The 4 SKIPs are the pre-existing
  network-gated ones (`TestScanDirFromEnv`,
  `TestSubmissionsCapDriftAgainstLiveAPI`, `TestScaffoldPinsSatisfyPublished`,
  `TestScaffoldedReadyAckActuallyFires`).
- `make ci-shallow` — **PASSED**, 19/19 packages, 0 failures, 0 timeouts.
- `golangci-lint run ./...` (v2.12.2, the pinned version) — **0 issues**.
- `gofmt -s -l .` — empty, over 202 Go files in the two touched trees.
- `origin/main` had not moved (`e433b91`); the branch is a direct descendant, so
  the merged tree is this tree.

## README (three surfaces, per CLAUDE.md)

Command table row; the *Listing and cancelling workflows* section (with the
worked example and a provenance note); *What the server says went wrong* (both
commands named, plus the endpoint-difference paragraph); and the Troubleshooting
symptom index gains a row keyed on the legend sentence — the only CLI-authored
string on the new surface, since the reason itself is server text and cannot be
indexed.

## Scope notes / disagreements

- **I wrapped the reason, which means the text is not byte-verbatim in
  whitespace.** `printWorkflow` deliberately keeps whitespace verbatim (#367).
  Wrapping is what makes the indent hold against the terminal's own soft wrap
  under a *table*, which is a hazard `workflows get` does not have; the words are
  untouched and the string's own line breaks survive. Flagging it because it is a
  knowing divergence from a documented #367 decision.
- **Residual, not fixed, flagged rather than acted on:** the same soft-wrap
  vector exists in `printWorkflow`'s reason block — a single long reason there
  can still spill to column zero under the `Workflow ID:` header.
  `wrapServerText` is now available to close it, but `workflows get` just took
  four audit rounds and this brief scoped changes to it out.
- **Could not verify:** the multi-step listing case against a real payload (all
  30 sampled workflows were single-step), and whether the orchestrator populates
  `errors` on a non-failed workflow. Both are asserted only against synthetic
  fixtures, and the code takes the status-neutral reading for the second.
- No error path changed, so no exit code moved (item 7). Nothing here says
  anything about the charge in either direction (item 28).


ZacxDev:
## Audit response — F1 fixed, F2 fixed, 🟢 recorded

Pushed `7b9d7a7` + `2a6a18b`. **I reproduced the F1 counterexample myself before
changing anything**, and it is exactly as described.

### 🔴 F1 — confirmed, and it was worse than a doc bug

Probe against the previous tip (`9b4c6ab`), with the audit's payload:

```
U+2800 IsSpace=false IsControl=false IsPrint=true
strings.Fields(token) -> 1 token(s)
safeTerm strips any rune of it? false
EMITTED LOGICAL LINE of 266 runes exceeds the 79-column budget
```

266 runes against a 79-column budget, from a payload the server's own
`isLikelySafeMessage` passes verbatim (under 300 chars, single line, matches no
unsafe pattern). The audit's reading of the guard is right too: `columnZeroLines`
inspects *emitted logical* lines, every one of which starts with the indent, so
it is complete for the `\n` vector and structurally blind to the overflow one.

**(1) Hard-split adopted — I agree with the reasoning and have nothing to argue
against it.** `hardSplitOverlong` breaks any token wider than the budget. The
framing that convinced me is the one the audit gave: an over-long token *already*
overflows, so the choice was never "clean line versus corrupted id", it was
"silently overflow into a forgeable column versus a visibly broken token". The
old `safeterm_test.go:163` assertion is reversed, and its replacement carries the
counterexample so the reversal is not re-reversed by someone reading only the
original rationale.

One scoping decision worth flagging: the split is applied in `wrapServerText`,
**not** in `wrapTokens`. `wrapTokens` is shared with `printFinding`, where
`findingTokens` deliberately keeps a quoted span whole so remedy text stays
pasteable — that surface renders the CLI's *own* findings and does not sit under
a table. Splitting there would corrupt text a user is meant to copy, to close a
hazard that does not exist on that path.

**(2) All three claims softened — agreed that (1) does not make this optional.**
Confirmed independently: `x/term.GetSize` appears nowhere in this repo, so the
CLI wraps to a fixed budget and never learns the terminal's width. What holds is
narrower and is now what each site says: *no line the CLI emits starts at column
zero or exceeds the budget*; a terminal narrower than the budget re-wraps and the
spill reaches column zero regardless. Changed at `internal/cmd/workflows_list.go`,
`internal/cmd/safeterm.go` and the README Troubleshooting row.

**(3) Guards are now property-shaped.** Two, at both levels:

- `TestWrapServerText/no_emitted_line_exceeds_the_budget,_for_any_input` —
  generated over 13 adversarial atoms × 4 separators × 5 widths.
- `TestWorkflowsList_NoEmittedLineCanOverflowIntoColumnZero` — the same invariant
  on the **rendered command surface**, which is the contract that matters, over 8
  generated reason shapes.

Both include U+2800 and an over-long unbroken token, and both are written as
*"no emitted line exceeds the budget"* rather than *"this payload is handled"* —
the shape the first version got wrong.

**Display-width deferred as instructed, and stated as a residual** at
`wrapServerText` and in the README row: `len([]rune)` is not display cells, so 38
wide runes make a 76-column line. It needs a width table (new dependency or
hand-rolled) — operator's call, filed separately.

I also found and closed one thing the audit did not ask for: the `width < 1`
clamp was **unreachable**, so a mutation removing it was unkillable. It now has a
test, and removing it is killed (by hang — reported honestly below).

### Red at the previous tip, green at HEAD

`internal/cmd/safeterm.go` checked out at `9b4c6ab`, new tests unchanged:

| assertion | `9b4c6ab` | HEAD |
|---|---|---|
| `TestWrapServerText/a_token_longer_than_the_width_is_hard-split` | **FAIL** | PASS |
| `TestWrapServerText/no_emitted_line_exceeds_the_budget,_for_any_input` | **FAIL** | PASS |
| `TestWorkflowsList_NoEmittedLineCanOverflowIntoColumnZero` | **FAIL** | PASS |

### Mutation — new guards

| mutant | verdict |
|---|---|
| `hardSplitOverlong` no-ops | KILLED by `…/no_emitted_line_exceeds_the_budget…` (+2 others) |
| hard split not called | KILLED by `TestWorkflowsList_NoEmittedLineCanOverflowIntoColumnZero` |
| off-by-one loop bound | KILLED by `…/no_emitted_line_exceeds_the_budget…` |
| off-by-one chunk size | KILLED by `…/no_emitted_line_exceeds_the_budget…` (+2 others) |
| first chunk only | KILLED by `…/no_emitted_line_exceeds_the_budget…` (+2 others) |
| tail discarded instead of split | KILLED by `…/a_token_longer…` + `…/a_non-positive_width…` (truncation assertions — correct, it is a truncation defect, not an overflow one) |
| width clamp removed | **KILLED-BY-HANG** — the only honest verdict: without the clamp `wrapServerText` never terminates, and no assertion can express that except a timeout. Reported as its own verdict rather than folded into KILLED. |
| wrap budget widened (re-anchored) | KILLED by `TestWrapServerText` |
| render budget widened (`listReasonWrapWidth`) | KILLED by `TestWorkflowsList_NoEmittedLineCanOverflowIntoColumnZero` |

**Harness honesty, since the round-1 sweep was re-run over the fix.** Two of its
entries came back as **non-verdicts, not survivals**, and both are real harness
behaviour rather than noise:

- `width budget widened` → `PATCH-NOT-UNIQUE (0 occurrences)`: the F1 fix moved
  the line its patch anchored on. Re-anchored in round 7 and KILLED. **This is
  the failure mode the harness exists to expose** — a stale anchor silently
  applies nothing and reads as SURVIVED to anything counting only `--- FAIL`.
- `rule not applied on the get path` → SURVIVED, unchanged and still the
  wrongly-aimed mutant documented in the original body (killed by the pre-existing
  #367/#381 guards under the full suite; the re-aimed version is killed by the
  intended test).

The harness also gained a `KILLED-BY-HANG` verdict and unbuffered output after
the clamp mutant hung the first sweep — a hang was previously indistinguishable
from a long run. Its three controls (must-die → KILLED, no-op → `PATCH WAS A
NO-OP`, break → `BUILD FAILED`) were re-validated after that change.

### 🟡 F2 — fixed

The README now states both directions and tells the reader what to do about it:
the same sanitiser *replaces* any message it cannot vouch for (URL, path, stack
frame, infra name, >300 chars, multi-line) with a generic *"… reported a system
error"*, so `list` can be strictly **less** informative than `get` — *"when a
failure is worth chasing, read both."*

### 🟢 Recorded, not fixed

The `Step`≡`Workflow` seam test now carries a comment saying it is fixture-shaped
rather than property-shaped, that it does fire, and why the property version was
not written (the two shapes now call the *same* function, so divergence is a
refactor away rather than an input away).

### Gate, re-run in full

- Counted from `-v`, never an exit code: `=== RUN 3695`, `--- PASS 3691`,
  `--- FAIL 0`, `--- SKIP 4`, `panic: test timed out` **0**, `build failed` **0**.
  Same 4 pre-existing network-gated SKIPs.
- `make ci` passed · `make ci-shallow` 19/19, 0 failures, 0 timeouts ·
  `golangci-lint` v2.12.2 **0 issues** · `gofmt -s -l .` empty.
- `origin/main` **has not moved** (still `e433b91`) and the branch is a direct
  descendant, so the merged tree is this tree.
- Re-checked live against the real endpoint after the change (a read; spends
  nothing) — the rendering is unchanged for real reasons.

Issue #186, opened 2026-07-30
No CLI path to set store-listing media (icon/cover/screenshots) — CLI-authored apps are born below the publish floor


ZacxDev:
## Summary

There is **no CLI path to set an app's store-listing media** (icon / cover / screenshots). The `civitai` CLI is otherwise the canonical authoring tool — `app create → validate → submit` covers the whole lifecycle — but listing media is **web-upload only**, so every author has to break out of the CLI and finish in the browser.

The practical result: an app submitted purely via the CLI is **born below the publish floor**. Since #3392 the floor is **icon + cover required** (screenshots optional), so a CLI-only author gets their app approved and deployed, then discovers the store listing can't publish until they upload two images through the web UI.

## Evidence

`block.manifest.json` has **no** media fields (checked against `schema/app-block.manifest.schema.json` on `main`):

```
top-level: $schema, blockId, version, name, tagline, type, contentRating, category,
           renderMode, trustTier, scopes, scopeJustifications, minApiVersion,
           buildCommand, outputDir, publicSettingsKeys, assetBundleUrl, iframe, page, targets
media-ish: NONE
```

`page.icon` exists but is an **icon-name string** (a glyph identifier for the run-page chrome), not listing artwork. And there are no `--icon` / `--cover` / `--screenshot` flags anywhere under `cmd/` or `internal/`.

Note the `screenshots/` bundle convention is a **different surface** — it populates `app_blocks.screenshots` (the block detail-page gallery), not `AppListing` icon/cover/screenshots.

## Why the obvious workarounds don't close it

Both server-side attach paths are reachable only from a browser session in practice:

- `appListings.persistAssetImage` takes an **already-CF-uploaded** image key — it presumes the browser upload widget ran first.
- `appListings.ingestAssetFromUrl` takes a **public HTTPS URL**, and its SSRF guard (`safe-fetch.ts`: https-only + `isPrivateIp`-blocked, redirect-revalidating) correctly rejects `localhost` / RFC1918. So a CLI can't hand it a local file, and "just serve it locally" is deliberately blocked.

So this isn't only a missing flag — **it likely needs a server-side companion endpoint** that accepts image bytes from a bearer-authenticated CLI and routes them through the standard `createImage` ingest + scan pipeline (same as the web path — no scan bypass), returning an image id the existing `setIcon` / `setCover` / `addScreenshot` procs already consume.

## Suggested shape (non-prescriptive)

Something like:

```
civitai app listing set-icon   ./assets/icon.png
[...]

Issue #295, opened 2026-08-10
app listing set-icon: echo the pixel dimensions it already decoded


ZacxDev:
Split out of #286, which was closed because its premise was wrong. This part is real.

## The problem

A **5000×5000 icon at 187,599 bytes — 8.9% of the CLI's 2 MiB cap** — passes every local check and is then rejected server-side with:

```
That icon couldn't be read. Try uploading it manually.
```

The real cause is sharp's `limitInputPixels` at **16 MP**: `4096²` is the last size that passes, `4097²` throws, and the `catch` in `listing-meta.service.ts` swallows a distinguishable libvips error into that generic sentence.

So an author with a perfectly ordinary large icon gets a message that names neither the quantity that failed nor a number to aim at — while the CLI is holding the answer.

## The cheap half, CLI-side

`loadAndValidateImage` already calls `DecodeImageInfo` and has `Width`/`Height` **header-only, at zero cost** — then discards them on the icon path. Print them alongside the byte count in the existing confirmation output (`app_listing.go:465`).

🔴 **This asserts no threshold and vendors no constant.** `AGENTS.md` item 25 is explicit that listing-media dimension/aspect bounds stay prose in the README and must **not** become a local check, and nothing here argues against that. Showing the author `5000×5000` next to `187.6 KB` lets them compare against the README's documented bound themselves. It is the difference between a silent number and a legible one.

## The real fix is platform-side

Branch on the libvips pixel-limit error rather than catching everything into one sentence, and name the bound. Filed here for the CLI half only.

## Prior art

README → *Listing media requirements* (from #275) documents the bound. This is about the CLI surfacing the quantity it already has, not about re-stating the limit.

PR #403, opened 2026-08-12
fix(listing): `set-icon` reported a failure having just succeeded, on every live app below the publish floor (#400)

### LINKED ISSUE #400: set-icon exits 1 after successfully staging the icon, on the first media command of every CLI-authored app
`set-icon` **stores the icon and then exits 1**, on what #186 establishes is the first media command of every CLI-authored app.

## What happened

A live listing with no icon and no cover (`sensei`):

```
$ civitai app listing set-icon icon-sensei-upload.png --slug sensei --changelog "..."
⚠ This listing is live — your change opens a revision for moderator re-review.
Uploading icon (18.5 KiB, 1024×1024)…
  scanning image… (0s)
  … (10s)
Error: the server rejected this store-listing change (400): Listing needs at least an
icon and cover before it can be published (missing: cover). — …
```

Reads as a failure. It wasn't — the icon was stored:

```
$ civitai app listing status --slug sensei
Icon:            set ✓
Cover:           MISSING (required)
```

Setting the cover next then opened one revision carrying **both** assets, which is the desired outcome. So the only thing wrong is that the successful first half is reported as an error.

## Why this isn't an edge case

#186's own framing is that **CLI-authored apps are born below the publish floor**. Clearing that floor requires two commands, and the floor is by definition not yet met when the first one runs — so *every* app authored through the CLI hits this on its first media command. The natural scripted form is the one that breaks:

```bash
civitai app listing set-icon icon.png && civitai app listing set-cover cover.png
#                                     ^^ aborts here, having succeeded
```

An operator who trusts the exit code concludes the icon failed and stops, or retries `set-icon` — which can never clear the condition, because the missing thing is the cover.

## Exit code

`cmd/civitai/main.go:47-50` sends any non-nil error to `os.Exit(exitCode(err))`. The change arm at `internal/appapi/listing.go:580` returns a bare `fmt.Errorf` carrying none of the sentinels `exitCode` classifies, so it falls through to `exitGeneric` = **1**. (Read from `origin/main`; the observed `Error:` prefix matches `main.go`'s handler. I did not capture `$?` at the time — the exit path is from source, the rest is observed.)

## Not a duplicate

- **#391 → #395** fixed the *wording* of this same arm — the remedy used to say *"fix the value and retry"*, which is wrong here since the value was fine. `main` now says *"the change may have partially applied; `civitai app listing status` shows the listing as it stands"*, which is accurate. **This report is about the exit code, not the message** — the corrected wording still arrives as an error on a successful attach.
- **#388** is the presigned PUT being the only sub-step that exits 1. Same family, different step.
- **#186** added the feature and named the born-below-floor condition; it didn't cover this consequence.

## Suggested fix, cheapest first

The CLI already knows the listing is below the floor before it attaches — it fetches listing state to print the *"this listing is live — your change opens a revision"* banner, and `status` renders the floor from the same data. So it can distinguish "the revision could not open **because the floor is not met yet**" from a genuine rejection, and report the former as **progress**:

```
✓ Icon staged. Cover still required before this can publish — run
  `civitai app listing set-cover <file>` to open the revision with both.
```

…exiting 0, since nothing failed. A narrower variant: keep the message, exit 0 only when `status` confirms the asset landed.

The alternative is server-side — accept the attach and return success with "cover still required" rather than 400 — which is arguably the truer root cause but a larger change.

## Environment

`civitai e433b91` (built 20260811053815). Re-read `origin/main` at `53984ca` while writing this; the arm and the exit path are unchanged there apart from #395's wording. **n=1** — `sensei` was the only listing on the account below the publish floor, so I could not re-run it without disturbing a pending revision.

ZacxDev:
Fixes #400.

## The bug

On a **live** listing that is still below the publish floor, `set-icon` did all of its work and then reported a failure:

```
⚠ This listing is live — your change opens a revision for moderator re-review.
Uploading icon (18.5 KiB, 1024×1024)…
Error: the server rejected this store-listing change (400): Listing needs at least
an icon and cover before it can be published (missing: cover). — …
```

Nothing failed. The icon was ingested, attached to the shadow revision and scanned clean. The only step that refused was the **submit**, which could not have worked yet: a revision cannot go to a moderator until the listing has both an icon and a cover, and clearing that floor takes two commands. #186's framing is that CLI-authored apps are born below the floor, so this is the ordinary first media command on such an app once it is live — and the natural scripted form is the one that breaks:

```bash
civitai app listing set-icon icon.png && civitai app listing set-cover cover.png
#                                     ^^ aborts here, having succeeded
```

## After

```
⚠ This listing is live — your change opens a revision for moderator re-review.
Uploading icon (211 B, 64×64)…
✓ Icon staged on an open revision — not submitted for review yet.
Your live listing is unchanged; nothing reaches a moderator until the publish floor is met.
Still required before publishing: cover.
Add the rest with civitai app listing set-cover <file> — it reuses this revision and submits it once the floor is met.
```

…and exit `0`. It deliberately does **not** say `pending moderator review` or print a request id: there is no review. The revision is open, holds the icon, and is unsubmitted — which is exactly the state the next command needs, since `beginListingRevision` is idempotent and reuses it.

## The discriminator is state, never prose

Matching the 400's message text was the cheap fix and it is the wrong one. It goes quietly false the day the server rewords, and it fails in the **reassuring** direction — a genuine rejection would start exiting `0`.

So on a submit failure the CLI re-reads the listing and asks two structural questions, on the same `Present()` predicate `listing status` and the floor line already use:

1. Did the asset **this command** attached actually land?
2. Is the floor **still unmet** — i.e. can the floor even be the reason?

Progress is reported only if both hold. A read failure reports nothing and the 400 survives: with no re-read there is no evidence the attach landed, and claiming what it cannot see is the defect this PR removes.

**Residual, stated rather than hidden:** a submit that fails for some *other* reason while the floor is *also* unmet is reported as progress, and its reason is not shown. It is bounded rather than silent — the next `set-`/`add-` command resubmits the same shadow, so once the floor is met that reason surfaces as a normal error.

## Two corrections to the issue's analysis

- **It exits `2`, not `1`.** `listingError` tags every 400 through a **deferred** `civitai.TagStatus`, so the error carries `civitai.ErrBadRequest` even though the arm returns a bare `fmt.Errorf` — `TagStatus` leaves `Error()` byte-for-byte unchanged (AGENTS item 7), which is exactly why reading the return statement misses it. Measured via `errors.Is`, not read off the source. This makes it slightly worse than reported: exit 2 is documented as *a mistake about the invocation*, so the CLI was telling scripts the author's command was wrong.
- **The failing call is `submitListingRevision`, not the attach or `beginListingRevision`.** The `scanning image…` lines in the report come *after* the attach, by #270's deliberate attach-before-scan ordering. Also worth noting `listing status` reads the **shadow** on an approved parent, so the reporter's `Icon: set ✓` was the shadow's state — the live listing was still empty, which is why the success message must not say "Icon set".

## Verification

- **Red at `a228062`** on the reported case; the two guard cases pass there, so this is a regression test for one behaviour and an invariant guard for two.
- **Mutation-tested.** Dropping the floor-met check, dropping the attach-landed check, and flipping the read-failure guard to fail-open each turn exactly one case red, with that case's own message.
- 🔴 **The fail-open mutant survived the entire `internal/cmd` package** before the fourth case was added — the guard was unreachable from any test, so the other three certified a branch nothing exercised. That case exists because its absence was measured, and re-running the mutation now kills it.
- **Both CI tiers**: full clone (`make ci`) and depth-1 (`make ci-shallow`, 20/20 packages). `make lint` at the CI-pinned golangci-lint v2.12.2 reports 0 issues — and that zero is validated by a positive control (an injected `ineffassign` in this same file is caught).
- The existing `TestListingChangeRouteAdviceLedger` `submitListingRevision` case is **unchanged and still green**: its stub reports an empty listing, so the attach does not land there and the 400 correctly survives.

## README

Moves in both places that state this contract: the `app listing` section, and the Troubleshooting row for the store-listing-change 400 — which until now claimed that row covered submitting a revision unconditionally.

## Known, left alone

The consent banner still says *"your change opens a revision for moderator re-review"* before the CLI can know the submit will be withheld. That overstates in the **conservative** direction (it asks for more consent than the command ends up needing), and the lines that follow correct it explicitly. Making the banner conditional would mean pre-checking the floor and gating the submit on a vendored copy of the server's rule — the design this PR deliberately rejects.


ZacxDev:
## Audit round 1 — three findings fixed in `fac64a1`, and a correction to this PR's own body

An adversarial audit found three ways the new success path could still report success while something had gone wrong. All are closed in `fac64a1`.

**1. It proved presence, not identity.** `attachLanded` asked whether the icon/cover *slot* was non-empty — which the image an author is **replacing** already satisfies. Measured: an attach of image `501` whose re-read still showed `999` in the slot printed `✓ Icon staged` and exited 0. It now compares image ids, preferring the id the attach **echoed** (the server's own statement of what it wrote) over the ingested one, so a re-encoded derivative stored under a new id still proves landed and anything else fails closed.

**2. It swallowed any submit failure, not just a refusal.** Entry was "the submit returned an error at all", so a `500`/`503`/`429` exited 0 during an outage — on the *busiest* path this command has, since below-floor is by #186's premise the normal state of every CLI-authored app's first media command. Entry is now gated on `errors.Is(err, civitai.ErrBadRequest)`, which is structural (`TagStatus` attaches it from the HTTP status, never the body), so it costs nothing that the refusal-to-match-prose was protecting.

**3. `--help` still asserted an unconditional moderator re-review** on all three attach commands. That paragraph was stated three times; it is now one constant.

### 🔴 Correction: the mutation evidence in the body above was worthless

The body claims three mutants were killed. That claim was made against a **red package** and I did not notice: extending the help text silently blew `TestListingHelpStaysWithinTheBudget`'s 1400-char cap, and a mutant battery run against an already-failing package "kills" every mutant for free.

Re-run from a genuinely green baseline, **2 of 11 mutants survived** — the empty-screenshot-id guard, and preferring the echo over the ingested id. Both were *equivalent* under fixtures where the two ids held the same value, which is exactly the collapse that identical-valued fixtures always hide. Both now have a case with distinct values.

Current battery, from a verified-green baseline: **11 killed, 0 survived**, including the four the audit independently found surviving.

| mutant | |
| --- | --- |
| floor-check `&&` → `\|\|` | killed |
| read-failure fails open | killed |
| `attachLanded` negated | killed |
| 400 gate inverted (swallow everything) | killed |
| 400 gate commented out (text still present) | killed |
| next-command choice inverted | killed |
| `kindCover` reads the icon slot (operand swap) | killed |
| screenshot id match negated | killed |
| empty-screenshot-id guard dropped | killed |
| `slotHolds` presence-only (the audit's finding 1) | killed |
| `attachedID` prefers ingested over echo (wrong bind) | killed |

Test cases grow **4 → 14**: both other media kinds (`set-cover` and `add-screenshot` were not exercised at all before), identity mismatch per floor slot, an empty-id row that must not match an empty id, the echo-wins case, and both 5xx shapes — `500` (deliberately unclassified → exit 1) and `503` (`ErrNetwork` → exit 5, the code the README tells scripts to retry on), because a gate could plausibly pass one and swallow the other.

Both CI tiers re-run green (`make ci`, `make ci-shallow` 20/20); `golangci-lint` v2.12.2 → 0 issues.

### Two open questions the audit raised that I could not settle from this repo

- **`listing status` on the new resting state.** It prints *"A revision is currently under moderator review."* whenever `hasPendingRevision` is set. This path deliberately leaves an **open, unsubmitted** shadow. If the server sets that flag for a shadow that exists but was never submitted, `status` will contradict what `set-icon` just printed. Not introduced here — a failed submit already left the shadow open — but this turns that from an error state into the normal one, so it wants a check against the real API.
- **Shadow seeding.** The floor check reads the *shadow's* assets. If a shadow were ever seeded empty rather than copied from the live listing, a fully-live listing updating its icon would read as below-floor. #400's own transcript is consistent with copy semantics, but nothing pins it.

ZacxDev:
## Audit round 2 (delta re-audit of `22e1948..fac64a1`) — five findings, all fixed in `bf6b734`

Round 1's fixes were re-audited blind against the delta. Verdict was **safe to merge, no 🔴** — but five 🟡/🟢 claims did not hold up, and one of them decided whether this PR fixes #400 *in production* at all.

**1. The identity check could refuse the very fix it enables.** Round 1 preferred the attach's echoed `iconId`/`coverId` over the ingested id, justified by a comment claiming it "stays correct if the platform stores a re-encoded derivative under a new id". That only holds **if the server echoes the new id**. If it echoes back the id it was *given* — the ordinary tRPC shape — while storing a derivative under a new one, the re-read reports an id this CLI has never seen, the check fails closed, and **#400 reproduces verbatim**. This repo already documents that icons *are* re-encoded server-side (#295/#344, `app_listing_reencode_test.go`), so the headline fix was resting on unmeasured server behaviour while the comment read as though it were handled.

Now **both ids count** — each names an image this command put there, and a pre-existing image matches neither — and the comment states the residual plainly instead of asserting it away. Failing closed stays the safe direction: the user gets today's error, never a false success.

**This is the one thing worth a real-server check before merge** (see the open question below).

**2. A stale doc claim, twenty lines above the code it described.** Still said "two structural questions … on the same `Present()` predicate", when there are now three and the first is no longer `Present()`-based. A maintainer reading only that paragraph would conclude `attachLanded` is presence-only — the exact bug round 1 fixed.

**3. README moved in one of its two places, again.** The Troubleshooting row listed all three conditions; the `app listing` bullet still promised exit `0` unconditionally — false for a 500, a 503, an unreadable re-read, and an identity mismatch. It also said the image is matched "by image id", true for icon/cover but not for a screenshot (an `alsc_…` row id).

**4. Three mutants survived round 1's battery** — all from fixtures that collapse distinct implementations into identical output:
- the cover slot had no echo-wins case, so `resCoverID` could read `res.IconID`;
- no case omitted the echo, so the ingested-id fallback **and the entire `imageID` parameter** could be replaced by a literal `0`;
- a nil echo substituting a placeholder id could match a slot holding that id.

Fixtures now vary per kind and per candidate. Battery is **15 mutants, 0 survivors**, from a baseline verified green first.

**5. Round 1's own deliverable was unpinned.** Folding three help paragraphs into one constant was the point, and deleting the below-floor clause left the package green. It is now asserted on the **rendered** body of all three commands — not on the constant, which would pass while a command stopped using it — and watched to fail: removing the clause reddens all three. The clause also moved to the LIVE branch of the sentence, since a below-floor **draft** neither stages nor opens a revision.

Also: the help-budget comment carries the measured margin (~28 runes on `add-screenshot`, paid by three bodies at once) instead of a stale "~100", and a leftover empty-string concatenation is gone.

Test cases **14 → 17**. Both CI tiers green (`make ci`, `make ci-shallow` 20/20); `golangci-lint` v2.12.2 → 0 issues.

### The one thing I'd want checked against a live listing before merge

Does `setIcon`'s echoed `iconId` equal what `getMyListingForEdit` then reports in the slot? If the server re-encodes into a **new** image row *and* echoes the input id, this fix silently does not apply — the author still sees the original #400 error. It fails closed, so nothing gets worse, but it would mean #400 stays open. One credentialed run against a below-floor live listing settles it.

Two smaller unknowns, unchanged from round 1: whether `hasPendingRevision` is set for an open-but-unsubmitted shadow (if so, `listing status` will contradict what `set-icon` just printed), and whether a shadow is seeded from the live listing rather than empty.

ZacxDev:
## Measured against the live platform — the fix does apply

The fix merged here proves an attached asset landed **by image id**, and shipped with one honest gap: if the platform stored a re-encoded **derivative** under a **new** image id while echoing back the id it was *given*, the CLI would see an id it had never met, fail closed, and this issue would silently stay broken.

Ran that measurement today. One `set-icon` against a live approved listing, reading the wire:

| step | image id |
| --- | --- |
| `ingestAssetFromDataUri` minted | `139517683` |
| `setIcon` echoed as `iconId` | `139517683` |
| `getMyListingForEdit` then reported in the slot | **`139517683`** |

All three agree. The `Image` row is created at **ingest** and the slot references that same row; the re-encode is a **URL transform** (`…/width=320/…` in the asset URL), not a second row. So the identity check matches and **the fix works in production**.

Method, since it touches a live listing: a local logging proxy forwarded everything to civitai.com **except `submitListingRevision`**, which it answered locally with the below-floor 400. No revision reached a moderator (3 submit attempts, 3 refused locally, 0 forwarded) and the live listing was untouched throughout.

Two related unknowns the same run settled:

- **`hasPendingRevision` means SUBMITTED, not "a shadow exists"** — measured `false` on a listing carrying an open, media-bearing, never-submitted shadow. That is exactly the state this fix leaves behind, so `app listing status` will not claim a moderator is reviewing something that was never sent.
- **Shadows are seeded from the live listing, not empty** — it reported the live icon and cover ids on a read that had just opened it. The floor check depends on that.

It also confirmed the exit-code correction end-to-end: the run exited **2**, and the fix correctly *declined* to swallow the 400 because that listing's floor is met.

Recorded in the code at #404 (comments only), so the next reader gets the measurement rather than the caveat.

Issue #412, opened 2026-08-14
app submit accepts a version BELOW what is live (silent regression on approval); app status never surfaces the drift


ZacxDev:
## Summary

`civitai app submit` accepts any version string, including one **lower than the version currently live**. Submitting `0.4.1` while `0.5.2` is serving is accepted without comment, and on approval it replaces the newer deployment with older code — while the version number reads as an ordinary forward bump, so nothing downstream looks wrong.

Separately, `app status` never compares the local manifest version against the published one, so the condition that leads to this is invisible until someone checks by hand.

Both are cheap to fix and neither needs a server change.

## The near-miss

While preparing tagline releases for six first-party apps I was about to bump `custom-generators` from its repo version `0.4.0` to `0.4.1` and submit. The live version is **`0.5.2`** — the repo is behind (see the companion issue #411 (submit provenance)). That submit would have been accepted, and on approval would have regressed a live app by five patch versions of real work.

The only thing that stopped it was manually running `civitai app status custom-generators` and noticing the mismatch. Nothing in `validate` or `submit` mentions it.

Four more apps are in the same state right now (`gen-matrix` 0.7.3 vs 0.8.2, `model-benchmarking` 0.1.3 vs 0.2.2, `playable-collections` and `app-requests` both 0.2.0 vs 0.2.2), so this is a live trap, not a hypothetical.

## Proposed fix

**1. Monotonic-version guard in `app submit`.** Before packaging, fetch the app's highest **approved** version and refuse when the submitted version is not strictly greater:

```
Error: refusing to submit custom-generators@0.4.1 — 0.5.2 is already approved and live.
Submitting an older version replaces the newer deployment on approval.
Your repo may be behind what was last released. Pass --allow-downgrade if this is deliberate.
```

Notes on the edges:

- Compare with **semver ordering**, not string comparison — `0.10.0` must beat `0.9.0`.
- Compare against the highest **approved** version, not the newest row by `submittedAt` — those differ (see #390, and the withdrawn-row behaviour where a duplicate can outrank an approved one by timestamp).
- Equal versions should also refuse: resubmitting the live version is almost always an accident, and it is what a repo-behind-live situation produces naturally.
- `--allow-downgrade` exists because deliberate rollback is legitimate.

**2. Drift line in `app status`.** When run inside an app directory, compare the local `block.manifest.json` version against the published one and say so:

```
Block ID:  custom-generators
[...]

Issue #390, opened 2026-08-11
the newest-first row pick is unpinned at three more sites — `ownedSubmission` prints a verbatim status off a one-row fixture (#378 residual)


ZacxDev:
Found by the blind adversarial audit of #384 (which fixed the same defect for `pullReviewAdvice`'s two callers). #384 deliberately did **not** widen its ledger to cover these — its scanner is keyed on `pullReviewAdvice(` call sites, and widening it to the row-pick *shape* would be a spelled guard turning on a variable name. So the residual is enumerated in `newestFirstAdviceSites`' doc comment and filed here.

Each site below reads **one row** out of a list documented as newest-first, and each has a fixture that **cannot see which end it read**.

## 🟡 1. `internal/cmd/apps.go` — `ownedSubmission` (the real one)

`ownedSubmission` returns the first row whose `BlockID` matches the slug, and `appViewOwnedAdvice` prints that row's `Status` and `DeployState` **verbatim** to the user. Its fixture, `ownedSubmissionBody` (`internal/cmd/app_view_owned_test.go:59`), is **one row** — so `subs[0]` and `subs[len(subs)-1]` are the same row and the choice is untested.

This is #378's defect under a different name, in the same package, and it is arguably worse: `app pull` at least printed a self-contradiction a reader could notice, whereas a wrong `Status`/`DeployState` here reads as plain fact. An author whose newest submission is `pending` over an older `approved`/`live` one would be told the wrong deploy state of their own app.

**Fix shape:** give `ownedSubmissionBody` a second, differently-stated row (newest first), then assert the printed status is the **newest** one. Verify by mutating the row pick to `subs[len(subs)-1]` and watching it go red — the auditor measured it fully green today (3673 RUN, 0 FAIL).

## 🟡 2. `internal/cmd/app_metrics.go` — the first-non-null `appBlockId` pick

Five lines above the call #384 pinned, `resolveAppBlockID` scans for the first non-null `appBlockId`. `TestAppMetricsPicksNewestNonNullAppBlockID` asserts **"newest"** in its name, but its fixture has only **one** non-null row — so it pins "skips the null row", which is real, and says nothing about *newest*. It is vacuous for the property its name claims.

**Fix shape:** add a second non-null row (an older approved version with a different `appBlockId`) and assert the newest non-null one is what reaches the tRPC query. Rename the test if it ends up pinning something narrower than its name.

## 🟢 3. `internal/appapi/appblocks.go:609` — "newest-first, so the first match is latest"

[...]

Issue #410, opened 2026-08-14
app status <slug>: a withdrawn newest row hides an approved+deploying one — "Not live yet" during a live deploy (#390 follow-up)


ZacxDev:
Follow-up to #390 / #396. **Not a claim that the newest-first pick is wrong** — it is deliberate, and #396 pinned it on purpose. The gap is that the rule reproduces the exact symptom #390 set out to eliminate when the newest row is `withdrawn`.

`ownedSubmission` (`internal/cmd/apps.go:234`) takes the first slug-matching row and `appViewOwnedAdvice` prints its `Status` / `DeployState` verbatim. The comment there states the hazard being avoided:

> an older row's state reads as plain fact about the newest submission: an author whose newest version is pending over an older approved/live one would be told the wrong deploy state of their own app, with nothing on screen to mark it wrong

A `withdrawn` newest row produces the same class of wrong answer, in the opposite direction.

## What I saw

Immediately after a moderator approved `sensei@0.1.1`, with the deploy already `building`:

```
$ civitai app status sensei
Block ID:         sensei
Version:          0.1.1
Publish request:  pubreq_01KZZ8H20KJ1M753MR4S20SDDV
Status:           withdrawn
Deploy state:     -
Submitted:        2026-08-13 23:29 CDT

Not live yet — sensei.civit.ai only serves after the app is approved and deployed (deployState 'live').
```

That reads as "your submission was withdrawn and nothing is happening." The truth, from `app status --json` at the same moment:

| id | submittedAt | status | deployState |
|---|---|---|---|
| `pubreq_01KZZ8H2…` | 04:29:48.436Z | `withdrawn` | `null` |
| `pubreq_01KZZ78V…` | 04:07:50.846Z | **`approved`** | **`building`** |

It reached `live` ~2 minutes later. The row the CLI showed was never wrong about *itself* — it was simply the least informative row available.

## Why the withdrawn row was there at all

It was not authored deliberately. Approving one request appears to **auto-withdraw** the other pending request for the same app: `reviewedAt` on the approved row is `04:29:49.684Z` and `updatedAt` on the withdrawn row is `04:29:49.689Z` — **5 ms apart**. So the platform created the row that then outranked everything by `submittedAt`.

That is the part worth weighing: the tie-break surfaced a row representing no user intent, and hid the one that did.

## Suggested direction

Not "pick the oldest" — that reintroduces #390. Some options, roughly in order of how little they disturb the pinned behaviour:

[...]

Issue #378, opened 2026-08-11
the newest-first submission assumption survives mutation at both call sites, and README moved in two of three places (#371 follow-ups)


ZacxDev:
From the delta re-audit of the fold-in commit on #371 (merged as `7f780fb`). Neither blocked the merge — the code is correct today; what is missing is coverage and one doc sentence.

## 🟡 A surviving mutant: the newest-first assumption is unpinned at both call sites

`internal/cmd/app_metrics.go:243` and `internal/cmd/app_pull.go:195` both read `subs[0].Status`.

**Measured: changing BOTH to `subs[len(subs)-1].Status` passes the entire suite — 1782 RUN, 0 FAIL.**

Under that mutant `app pull` prints a **self-contradictory** message:

```
(latest submission: 0.1.1 pending); that submission was WITHDRAWN, so nothing is in review
```

— because `latest submission:` comes from `subs[0]` and the advice from the last row, and nothing notices.

The seam guard added in the same commit (`TestPullAndMetricsGiveTheSameNextStepForTheSameState`) **structurally cannot see it**: it uses a one-row body, so first and last are the same row. This is also the shape where a shared predicate hides a defect — both sites move together, so the seam guard stays green while both are wrong.

**One fixture closes it:** two never-approved rows, newest `pending` / oldest `withdrawn`, asserting (a) the advice matches the **newest** row and (b) `latest submission:` and the advice name the **same** state.

(The `app pull` half predates the fold-in — it arrived with the PR's earlier commits — so the fold-in widens this gap rather than creating it.)

## 🟡 README was moved in two of three places

`README.md:1544` — the **`app metrics` command section** still says only *"an app still in review reports that instead of an empty dashboard"*, while the parallel `app pull` section (`:1461-1466`) spells out the rejected/withdrawn wording. The exit-code bullet (`:2419`) and the errors table (`:2506`) both moved; the command's own section did not.

`CLAUDE.md` makes `README.md` the published contract that must move in lockstep with behaviour. One sentence. (Line numbers are post-`7f780fb`; re-locate by content.)

## 🟢 The exit-code claims ledger under-describes what exit 1 now promises

`internal/cmd/exitcodes_claims_test.go:58-62` — the code-1 claim's `phrases`/`why` still frame the whole state as *"still in review"* / *"wait for approval"*. Not published text, but it is the ledger a future reader consults for what exit 1 means, and it now covers less than the code does.

## Context worth keeping

[...]

PR #413, opened 2026-08-14
feat(app status): warn when the local manifest is BEHIND the highest approved version (#412)

ZacxDev:
Closes nothing on its own — `Refs #412`. This is **piece 2** of the issue (the `app status` drift line). The monotonic-version guard in `app submit` is a separate PR; #412 needs both.

## What it does

`civitai app status <slug>` — and the `--id` spelling — now compares the local `block.manifest.json` against the caller's highest **approved** version of the same app, and warns on stderr when the repo is behind:

```
Block ID:  custom-generators
Version:   0.7.0
...
⚠ local block.manifest.json is 0.4.0 — BEHIND the highest APPROVED version of custom-generators, which is 0.5.2.
  An approved version is what gets deployed, so submitting from this repo would replace newer code on approval.
  Sync the released code (civitai app pull --app custom-generators) or raise the local version above 0.5.2 before civitai app submit.
```

Pure client-side; no server change. **The exit code is unchanged in every case** — this is a warning, not a refusal, and there is no new exit path.

---

### 1. Compared against the HIGHEST APPROVED version, not the newest row

The issue asks for this explicitly and #390 is why. The two genuinely differ: the newest row by `submittedAt` can be a `pending` resubmission, or a `withdrawn` duplicate that outranks the approved row by timestamp. Neither is code anyone is running, so neither is a thing a repo can be *behind*.

It is also the reference the companion `app submit` guard refuses against. If `status` quoted a different number the two commands would contradict each other about the same repo on the same day.

One degree of freedom the issue does not name: *highest* approved vs *newest* approved. Those differ under a rollback (a lower version approved later). This takes the **highest**, again for agreement with the submit guard. Pinned by `TestAppStatusDriftUsesHighestApprovedNotNewestApproved`; the trade-off is that a deliberate rollback over-reports (it will say you are behind the highest version ever approved), which is the conservative direction and stays literally true because of decision 4.

`highestApprovedVersion` is also **order-independent** — a maximum over a filtered set, unlike every other read of this route in the package (see `newest_row_pick_test.go`). Asserted over all 24 permutations of a four-row fixture rather than claimed in a comment.

### 2. Three-way outcome; only one branch speaks

| relation | rendered |
|---|---|
| local **BEHIND** published | the warning above |
| local **AHEAD** | nothing — the normal state of a repo about to release |
| **EQUAL** | nothing — the healthy state right after a release |

`app submit` separately refuses an *equal* resubmit; that is a decision about an action, not a description of a repo. Copying it here would print a warning on every healthy repo the day after a release, and noise is how the BEHIND case gets ignored.

### 3. Everything unknowable degrades to SILENCE

No local manifest · unreadable/malformed manifest · a manifest for a **different** app · an unparseable version on **either** side · the drift listing 403/500/429s · nothing approved yet · an empty listing. Every one of those returns without writing a byte and without touching the exit code.

Two specifics worth calling out:

- `compareVersions` (shared with the self-update check, **not modified**) treats an *unparseable current* as OLDER, so that any real release surfaces as available. Inheriting that default here would turn a typo in a manifest into a confident "your repo is BEHIND". Both sides are gated on `isParseableVersion` first, so that branch is unreachable from here.
- The manifest gate runs **before** the network. A caller not standing in the matching app directory makes exactly as many requests as before this PR — asserted, not assumed (`want 1`).

### 4. Wording says what was actually compared

The issue's mock reads "your repo is BEHIND what is live". This says **"BEHIND the highest APPROVED version of X, which is 0.5.2"**, because that is what was measured: an approved row's `deployState` may be `building` or `failed`, and this CLI did not check it. The message also quotes the **raw** manifest string, never the numeric triple it was reduced to for comparison.

**Known limitation, recorded not papered over:** `parseSemver` drops `-prerelease`/`+build`, so a local `0.5.2-rc1` reads as EQUAL to a published `0.5.2` and stays silent. That is the safe direction (it can under-report, never fabricate), and sharpening it means changing `parseSemver`, which `civitai version` also depends on. Pinned in both directions by `TestAppStatusDriftPreReleaseComparesByTheNumericTriple`.

---

New tests: `internal/cmd/app_status_drift_test.go` (new file — nothing else in `internal/cmd` was touched besides `app_status.go`).

| tree | scope | RUN | PASS | FAIL |
|---|---|---|---|---|
| **HEAD** (`bc9d074`) | all 45 new cases incl. subtests | 45 | **45** | 0 |
| **HEAD** | full `./internal/cmd` (`-v`) | 2103 | 2103 | 0 |
| **HEAD** | `make ci` (tidy+vet+test+build, all packages) | — | all `ok` | 0 |
| **`origin/main`** (`e49ddc6`) | 34 behavioural cases, unit cases removed | 34 | 19 | **15** |

The 11 unit cases (`TestHighestApprovedVersion`, `…IsOrderIndependent`, `TestVersionDriftWarningIsThreeWay`, `TestWarnLocalVersionDriftIgnoresAnErroredListing`) **cannot run at base** — they name symbols that do not exist there, so the package fails to compile. That is a compile-level red, reported as such and not counted as a watched failure.

Red at base (15): the whole positive half — `WarnsWhenLocalManifestIsBehind`, `WarnsOnTheIDPathAndKeepsJSONPure`, `UsesHighestApprovedNotNewestApproved`, both `OrdersBySemverNotByString` subtests, `PreReleaseComparesByTheNumericTriple/git-describe`, `HelpDocumentsTheDriftWarning` — **plus** `SilentWhenAheadOrEqual` (3 subtests), `SilentWhenNothingIsApproved` and `SilentWhenTheListingIsEmpty`, which are *negative* controls that are red at base only because each asserts the drift listing was actually fetched (`want 2 requests`). Without that reachability precondition they would pass against a drift check that had been deleted outright.

Green at base (19) and honestly so: `SilentWithoutALocalManifest`, `SilentOnAnUnreadableManifest` (×3), `SilentOnUnparseableVersions` (×4), `SilentForADifferentApp` (×2), `SilentWhenTheListingFails` (×3), `SilentOnTheListView`, `PreReleaseComparesByTheNumericTriple/rc`. These assert silence, and base is silent. Their non-vacuity comes from the mutation battery below, where each of them goes red.

## Mutation battery

`make mutate PKG=./internal/cmd` was **not** run: gremlins runs one full suite per mutant and this package's suite is ~24 s over ~60 files, which is not a bounded run. Instead the comparison was hand-broken eight ways, each mutant applied alone against the unmutated control (46 RUN / 46 PASS / 0 FAIL), with the **specific failing assertion** recorded — not just "a test failed".

| mutant | FAIL | killed by | with |
|---|---|---|---|
| **M1** invert the comparison (`>= 0` → `<= 0`) | 13 | `WarnsWhenLocalManifestIsBehind` | `the drift warning is missing or reworded` |
| **M2** drop the `status == "approved"` filter | 12 | `HighestApprovedVersion` | `= ("0.7.0", true), want ("0.5.2", true)`; also `SilentWhenNothingIsApproved` |
| **M3** semver → string compare (both sites) | 6 | `OrdersBySemverNotByString/0.10.0…` | `warn=true, want false. "0.10.0" < "0.9.0" as STRINGS` |
| **M4** drop the manifest-slug match | 3 | `SilentForADifferentApp/some-other-app` | `a drift warning was printed on stderr, but this case must stay SILENT` |
| **M5** newest approved instead of highest | 4 | `UsesHighestApprovedNotNewestApproved` | `want the HIGHEST approved version (0.5.2)` |
| **M6** drop the local `isParseableVersion` gate | 3 | `SilentOnUnparseableVersions/local_is_not_semver` | `must stay SILENT` |
| **M7** drop the listing `err != nil` gate | 1 | `WarnLocalVersionDriftIgnoresAnErroredListing` | `a listing that ERRORED was read anyway` |
| **M8** drop the manifest-load gate | 5 | `SilentWithoutALocalManifest` | `must stay SILENT` |

🔴 **M7 survived the first round.** `ListSubmissions` returns `nil, err` on every failure path today, so the downstream "no approved rows" branch absorbed the error and the guard never mattered — the entire behavioural suite stayed green with it deleted. It is now pinned at the seam by a direct call with a stub lister returning rows **and** an error, with a nil-error positive control in the same test so the silence is attributable. That control also independently kills M1.

## Anti-vacuity

`driftRows` **refuses to build** a fixture whose rows share `id`, `version` or `submittedAt` — a one-row (or field-sharing) listing cannot tell "highest approved" from "newest row", which is exactly how #390 shipped green with four unpinned row picks. The canonical fixture puts the approved row *third*, under a newer `pending 0.6.0` and a newest `withdrawn 0.7.0`, and the positive control asserts neither of those two versions is named.

## Scope / what is deliberately not here

- **The bare `civitai app status` list view does not warn** (`TestAppStatusDriftSilentOnTheListView` records the decision): it is many apps at once with no single "the version" line to attach to.
- `internal/cmd/app_submit.go`, `internal/cmd/exitcodes_doc.go` and the README exit-code blocks are untouched — the submit-side guard is a separate PR.
- `update_check.go`'s `parseSemver`/`compareVersions` are **called, not modified**.
- `newest_row_pick_test.go`'s ledger prose was left alone to stay file-disjoint from that PR; `cmd/app_status.go` is already ledgered, and the new read's order-independence is pinned by a permutation test rather than asserted in a comment there.

## Gates

- `make ci` — green (tidy, vet, test, build).
- `golangci-lint run ./...` (2.12.2 via `nix-shell`) — **`0 issues.`** Instrument validated: a deliberately-planted `declared and not used` made the same invocation report `1 issues: typecheck: 1`, then was reverted.
- `gofmt -s -l .` — clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,7 @@ contract, and **packages/submits** it for review.
   - [After you submit: review → approve → deploy](#after-you-submit-review--approve--deploy)
   - [Listing media requirements](#listing-media-requirements)
 - [Submission status](#submission-status)
+  - [Is your repo behind what you shipped?](#is-your-repo-behind-what-you-shipped)
   - [Deployed is not the same as listed in the store](#deployed-is-not-the-same-as-listed-in-the-store)
 - [Pull your app's repository (`app pull`)](#pull-your-apps-repository-app-pull)
 - [Browse the App store](#browse-the-app-store)
--- a/README.md
+++ b/README.md
@@ -317,7 +317,7 @@ README. For the end-to-end walkthrough, see
 | `civitai app submit [dir] [--yes] [--package-only] [--out f.zip] [--skip-validate]` | Validate + package the source tree + upload it with your stored token (or, with no token, write the bundle + print next steps). **A submit that would really upload asks for confirmation, and in a non-interactive shell it refuses without `--yes`** — `civitai app submit --yes` is the CI form. (The refusal is reached only when there is a token to upload with: `--package-only`, and the no-token fallback that just writes the .zip, never submit and so never ask.) |
 | `civitai app pull [dir] --app <slug\|appBlockId>` | **Clone (or sync) the canonical git repository behind one of your approved Apps** — the read side of git authoring. ⚠ The clone URL embeds your access token, and a fresh clone persists it into `.git/config`. See [Pull your app's repository](#pull-your-apps-repository-app-pull). |
 | `civitai app listing status\|set-icon <file>\|set-cover <file>\|add-screenshot <file>\|rm-screenshot <id>\|reorder <id...>` | **Attach the store-listing media your App needs before it can be published** — an **icon and a cover are mandatory** (screenshots are optional, up to 8). `listing status` prints what is attached vs. what the publish floor still requires. The CLI checks format + byte size locally; **dimensions and aspect ratio are checked by the platform at attach**. See [After you submit](#after-you-submit-review--approve--deploy) and [Listing media requirements](#listing-media-requirements). |
-| `civitai app status [blockId] [--id <pubreq>] [--limit N] [--json]` | Check the review/deploy status of **your own** submissions. No arg lists them all; `--limit N` shows only the newest N (display-side — this route cannot page); a `blockId` (app slug) or `--id` shows one in detail (rejection reason if rejected, live URL once deployed). See [Submission status](#submission-status). |
+| `civitai app status [blockId] [--id <pubreq>] [--limit N] [--json]` | Check the review/deploy status of **your own** submissions. No arg lists them all; `--limit N` shows only the newest N (display-side — this route cannot page); a `blockId` (app slug) or `--id` shows one in detail (rejection reason if rejected, live URL once deployed). Run from **inside** an app checkout it also warns on **stderr** when your local `block.manifest.json` is **BEHIND** your highest approved version — advisory only, the exit code never changes. See [Submission status](#submission-status) and [Is your repo behind what you shipped?](#is-your-repo-behind-what-you-shipped). |
 | `civitai app metrics <slug> [--from <d>] [--to <d>] [--json]` | **Owner-only analytics for one of your Apps** — installs, runs + Buzz spent, Buzz purchased, and API engagement. Always prints the window the **server** served (it defaults to 30 days and clamps to 366), so a zero is never ambiguous. Needs a **personal API key** (an OAuth login is refused). See [App metrics](#app-metrics). |
 | `civitai app withdraw [pubreq-id] [--id <pubreq>] [--yes]` | **Withdraw your own pending submission** (the `pubreq_…` id from `civitai app status`). Frees the slug so a fresh `civitai app submit` can replace it. **Also deletes a first-version app's store listing — icon, cover and every captioned screenshot**, so it asks first and needs `--yes` in a script. Idempotent for the submission only; only a `pending` request can be withdrawn. See [Submission status](#submission-status). |
 | `civitai generate "<prompt>" [--negative-prompt <p>] [--quantity <n>] [--aspect-ratio <r>] [--checkpoint <version-id>] [--lora <version-id>[:strength]] [--image <path-or-url>] [--ecosystem <key>] [--input <file>] [--print-input] [--dry-run] [--json] [--max-cost <buzz>] [--fail-on-substitution] [--yes] [--no-wait] [--timeout <dur>] [--out-dir <dir>] [--out-name <template>] [--no-download] [--force] [--external-id <key>]` | **Generate images from a text prompt — this SPENDS REAL BUZZ.** Prices the job with the server's estimator, shows the cost + your balance, asks before spending, submits, then **waits and downloads** the results. `--dry-run` prices it and exits without submitting; `--max-cost` is an **estimate check, not a spending cap**. Needs the AI Services scopes — `civitai login --scopes generate` or a full-scope **personal API key**; a **default** OAuth login is refused. See [Generate](#generate) for the wait/download flags, image-to-image, raw graphs, and [silent model substitution](#-silent-model-substitution). |
@@ -1516,6 +1516,60 @@ silently ignoring the flag would be worse than saying so.
 "run `civitai app submit`" hint; with no token it points you at `civitai login`.
 Notes like the cap caveat go to stderr, so `--json` stdout stays pure and the
 exit code stays 0.
+
+### Is your repo behind what you shipped?
+
+A checkout can fall **behind its own live deployment** — you (or a teammate)
+released 0.5.2, then came back to a working copy still on 0.4.0. Submitting from
+there is accepted, and on approval it **replaces newer code with older code**
+while the version number reads like an ordinary forward bump.
+
+So when you ask about a single submission **from inside that app's directory**,
+`civitai app status` compares the local `block.manifest.json` against your
+**highest approved** version of the same app and says so on **stderr**:
+
+```text
+$ civitai app status custom-generators
+Block ID:         custom-generators
+Version:          0.7.0
+...
+
+⚠ local block.manifest.json is 0.4.0 — BEHIND the highest APPROVED version of custom-generators, which is 0.5.2.
+  An approved version is what gets deployed, so submitting from this repo would replace newer code on approval.
+  Sync the released code (civitai app pull . --app custom-generators) or raise the local version above 0.5.2 before civitai app submit.
+```
+
+The remedy is `civitai app pull . --app <slug>` — the `.` matters. Without a
+`[dir]`, `app pull` clones into `./<slug>` (see [Pull your app's
+repository](#pull-your-apps-repository-app-pull)), which from inside the checkout
+would create a **second copy nested in your repo** and leave the checkout itself
+just as far behind. The other way out is legitimate too: bump the local version
+above the published one, which is what you want when the local work really is
+newer.
+
+What it does **not** do:
+
+- **It never changes the exit code.** It is a warning, not a refusal — nothing
+  that scripts `app status` starts failing because a repo is out of date.
+- **It goes to stderr**, like the cap caveat above, so `--json` stdout stays a
+  pure, parseable payload on both renderings.
+- **It says nothing unless it is sure.** No local manifest, an unreadable one, a
+  manifest for a *different* app, a version on either side this CLI cannot order,
+  a listing that 403s/500s/429s, or nothing approved yet — every one of those is
+  silence. A false "your repo is behind" would send you to re-pull released code
+  for no reason, so the check only ever speaks when it has both numbers.
+- **It only speaks when you are BEHIND.** Being *ahead* is the normal state of a
+  repo about to release, and being *equal* is the healthy state right after one.
+- **It is scoped to the detail view** (`app status <blockId>` or `--id`). The
+  bare listing is many apps at once and has no single version line to attach to.
+
+The reference is your **highest approved** version, which is deliberately not the
+same thing as the newest row: the newest row can be a `pending` resubmission or a
+`withdrawn` duplicate, and neither is code anyone is running. It is also the
+version `civitai app submit` measures against, so the two
+commands never quote different numbers for the same repo. (Pre-release suffixes
+are ignored in the comparison — a local `0.5.2-rc1` reads as equal to a published
+`0.5.2` and stays quiet.)
 
 ### Deployed is not the same as listed in the store
 
--- a/internal/appapi/appblocks.go
+++ b/internal/appapi/appblocks.go
@@ -789,16 +789,56 @@ func (c *Client) ListSubmissions(ctx context.Context, blockID string) ([]Submiss
 
 // GetSubmission returns a single submission. Exactly one of id (a pubreq id) or
 // blockID (an app slug) should be set; id takes precedence if both are given.
+//
+// It is the narrow spelling of GetSubmissionRows: identical request, identical
+// row pick, and the rest of the narrowed listing discarded.
 func (c *Client) GetSubmission(ctx context.Context, id, blockID string) (*Submission, error) {
+	s, _, err := c.getSubmissionRows(ctx, id, blockID)
+	return s, err
+}
+
+// GetSubmissionRows is GetSubmission plus the rows it read to answer.
+//
+// 🔴 IT EXISTS SO A CALLER DOES NOT HAVE TO ASK TWICE. The `?blockId=` spelling
+// of this route answers with the app's WHOLE narrowed listing and GetSubmission
+// throws all of it away but Submissions[0] — so `app status <slug>`'s drift
+// check used to re-issue the byte-identical GET (submissionsURL("", blockID))
+// just to see the rows this call already held. Same URL, same auth, same page;
+// only the latency and the rate-limit budget were extra.
+//
+// The second return is the row set BEHIND the answer, and it is nil — not empty
+// — whenever there is no such set to hand back. That distinction is the whole
+// contract: the `?id=` spelling answers with a single-row envelope and no
+// listing at all, so a caller that needs every row for the app must still fetch
+// it itself. A nil return therefore means "I did not read a listing", never
+// "the listing was empty"; the empty listing is a real, distinct answer (a slug
+// with no submissions) and it is reported as a not-found error above, not as
+// nil rows.
+//
+// Rows are returned as read (newest first, per the route) and are NOT filtered
+// or reordered here — highestApprovedVersion and friends do their own filtering
+// and must see exactly what the server sent.
+func (c *Client) GetSubmissionRows(ctx context.Context, id, blockID string) (*Submission, []Submission, error) {
+	return c.getSubmissionRows(ctx, id, blockID)
+}
+
+// getSubmissionRows is the single implementation both spellings above delegate
+// to. It is unexported ON PURPOSE: newest_row_pick_test.go's ledger derives
+// "every EXPORTED func that reaches submissionsURL" and follows unexported
+// chains, so keeping the body private makes the derivation report BOTH exported
+// boundaries — GetSubmission and GetSubmissionRows — instead of silently
+// dropping whichever one became a wrapper. A wrapper is still an accessor to
+// every caller that uses it.
+func (c *Client) getSubmissionRows(ctx context.Context, id, blockID string) (*Submission, []Submission, error) {
 	build := func() (*http.Request, error) {
 		return http.NewRequestWithContext(ctx, http.MethodGet, c.submissionsURL(id, blockID), nil)
 	}
 	status, raw, err := c.authedDo(ctx, build)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 	if status != http.StatusOK {
-		return nil, submissionsError(status, raw, id, blockID)
+		return nil, nil, submissionsError(status, raw, id, blockID)
 	}
 	// An `?id=` lookup returns {submission: {...}}; an `?blockId=` lookup returns
 	// {submissions: [...]} (narrowed list) — handle both so either selector works.
@@ -806,7 +846,9 @@ func (c *Client) GetSubmission(ctx context.Context, id, blockID string) (*Submis
 		Submission *Submission `json:"submission"`
 	}
 	if err := json.Unmarshal(raw, &single); err == nil && single.Submission != nil {
-		return single.Submission, nil
+		// No listing was read on this path — nil rows, deliberately (see the doc
+		// comment): a caller needing every row for the app must fetch it itself.
+		return single.Submission, nil, nil
 	}
 	// 🔴 THE `?blockId=` FORM IS A NARROWED LIST, NOT A SINGLE ROW, and the row
 	// taken below is `Submissions[0]` — NEVER `Submissions[len-1]`. Every
@@ -825,7 +867,7 @@ func (c *Client) GetSubmission(ctx context.Context, id, blockID string) (*Submis
 		Submissions []Submission `json:"submissions"`
 	}
 	if err := json.Unmarshal(raw, &list); err != nil {
-		return nil, fmt.Errorf("unexpected /api/v1/blocks/submissions response: %s", string(raw))
+		return nil, nil, fmt.Errorf("unexpected /api/v1/blocks/submissions response: %s", string(raw))
 	}
 	if len(list.Submissions) == 0 {
 		// A `?blockId=` lookup for an unknown slug answers 200 with an EMPTY list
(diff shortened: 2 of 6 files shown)


ZacxDev:
## ⚠️ #413 and #414 are each green alone but do not compose — do not merge both without consolidating

Both PRs implement #412 and both add a `highestApprovedVersion` to package `cmd`. The git merge is textually clean; the merged tree **does not compile**.

Verified by merging both branches onto `main` (`e49ddc6`) in a scratch worktree and building:

```
internal/cmd/app_submit_version_guard.go:92:6: highestApprovedVersion redeclared in this block
        internal/cmd/app_status.go:351:6: other declaration of highestApprovedVersion
internal/cmd/app_submit_version_guard.go:163:10: assignment mismatch: 1 variable but highestApprovedVersion returns 2 values
```

### Renaming one of them would NOT be enough — the two predicates disagree

They differ on approved rows carrying a pre-release / build suffix:

- **#413** gates on `isParseableVersion` + `compareVersions`, which truncate at the first `-`/`+`.
- **#414** introduces the stricter `comparableVersion`, which declares such a version **not orderable** and routes it to warn-and-proceed.

Probed on unmodified `main`:

```
parseSemver("0.6.0-rc1")     -> {major:0 minor:6 patch:0} ok=true ; isParseableVersion=true
parseSemver("0.6.0+build.1") -> {major:0 minor:6 patch:0} ok=true ; isParseableVersion=true
compareVersions("0.6.0-rc1","0.5.2") = 1
```

So with an approved `0.6.0-rc1` alongside an approved `0.5.2`, `app status` (#413) picks `0.6.0-rc1` as the highest approved version, while `app submit` (#414) skips it and lands on `0.5.2`. `status` would report the repo BEHIND `0.6.0-rc1` while `submit` accepts `0.5.3` without complaint — the exact contradiction #413's own doc comment warns against ("the two commands have to name the same number or they contradict each other").

### Suggested resolution

Whichever PR lands second consolidates to a single implementation. #414's `approvedPeak` is the better survivor: it is stricter on pre-release ordering and additionally carries the unorderable-rows list and the `live` deploy state. #413's `versionDriftWarning` then consumes `peak.version` / `peak.found`, and #413's `("0.5.2", true)` assertions adapt to the struct.

Both are currently blocked by `pins-vs-published`, which is **unrelated and pre-existing**: it compares scaffold template pins against live npm (`@civitai/app-sdk ^0.31.0` vs published `0.34.0`, `@civitai/blocks-react ^0.39.0` vs `0.42.0`) and fails identically at unmodified `main`. It is green on main's last runs only because those predate those npm releases. #408 is the fix.

ZacxDev:
## Audit round — 5 findings + 1 nit fixed (`d1df4f5`)

| # | Finding | Fix | Pinned by |
|---|---|---|---|
| 1 🟡 | Remedy was **wrong for the only case the warning can appear in** | `civitai app pull . --app <slug>` | `TestDriftRemedyCommandIsRunnableFromTheWarningsOwnDirectory` (+ e2e `…RemedyReachesTheUser`) |
| 2 🟡 | Advisory call ran **before** the command's own output | render first, + a 10s deadline of its own | `TestAppStatusRendersTheAnswerBeforeTheAdvisoryDriftCheck` (2 arms) |
| 3 🟡 | Byte-identical **duplicate GET** | `appapi.GetSubmissionRows` — slug path reuses the rows | `TestAppStatusDriftReusesTheRowsTheDetailLookupAlreadyRead` |
| 4 🟡 | No-false-BEHIND lived in the **caller**, not the reusable function | guard inside `versionDriftWarning` | `TestVersionDriftWarningRefusesUnparseableVersionsItself` |
| 5 🟡 | README not updated | new `### Is your repo behind what you shipped?` + table row | — |
| 6 nit | `Status != "approved"` case-sensitive | `strings.EqualFold(strings.TrimSpace(…), "approved")` | `TestHighestApprovedVersionNormalisesTheStatus`, `…WarnsOnACapitalisedApprovedStatus` |

### 1 — the remedy was not runnable from where it printed

The warning only prints when **cwd IS the app checkout** (`driftManifestDir` is `"."`). `civitai app pull --app <slug>` with no `[dir]` targets `./<slug>` (`app_pull.go:269`), so run verbatim from where the line printed it **clones a second copy of the app nested inside the repo** and leaves the checkout exactly as far behind. The author follows the advice, sees a success line, and submits the downgrade anyway — the outcome #412 exists to prevent. (It also put a duplicate copy of the app source inside the next submission's ZIP; `.git` is excluded by `pkgzip`, the *source* is not.)

The old tests only `strings.Contains`'d the literal, which agrees with a wrong command as readily as a right one. The new test **parses the printed string against the real cobra tree** — command path, `--app`, the positional `[dir]`, and the command's own `Args` validator — and carries a control that runs the *pre-fix* string through the same parser and asserts it comes back with **no `[dir]` at all**, so the check is known to distinguish the two forms.

### 2 — ordering, not stream

`--json` stdout purity is unaffected: that is about *which stream*, this was about *when*. Two arms: one points both streams at a single sink so their relative order is observable at all (the normal harness has separate buffers, which cannot see order); the other has the server record what had been written **at the moment the advisory request arrived** and asserts the answer was already there. Second arm runs on `--id`, the only path that still makes an advisory request.

### 3 — one request on the slug path, two on `--id`

`GetSubmission` and `GetSubmissionRows` now share an **unexported** body. Both remain exported accessors, so `newest_row_pick_test.go`'s derived accessor ledger reports both — it caught the split on the same run, which is exactly what it is for. Request **URLs** are asserted, not just counts.

🔴 **Bookkeeping this forced, stated rather than buried:** the slug path's reachability control *was* the second request (`want 2`). With the rows reused it is 1 whether the check ran or not, so those assertions became non-discriminating. Every relocated control says so in place, and where a count still discriminates it moved to the `--id` path (`TestAppStatusDriftManifestGateRunsBeforeTheNetwork`). `TestAppStatusDriftSilentWhenTheListingFails` now drives `--id` too — left on the slug path it would have kept passing while asserting silence about a request that no longer happens.

### 4 / 6

4 is zero behaviour change today (`:314` still gates first) — the point is that the *function* is the reusable half and the `app submit` guard will inherit from it. 6 aligns the predicate with `appblocks.go` / `app_pull.go` and shrinks the consolidation delta against #414's copy: a raw `==` cannot fail loudly, it just goes silently inert if the server ever answers `Approved`.

---

## Re-run mutation battery — full, not just the new tests

Control: **2129 RUN / 2129 PASS / 0 FAIL**, each mutant applied alone against it.

| mutant | verdict | killed by |
|---|---|---|
| M1 invert the comparison | **KILLED** (24 FAIL) | `WarnsWhenLocalManifestIsBehind` |
| M2 drop the approved filter | **KILLED** (19) | `WarnsWhenLocalManifestIsBehind`, `HighestApprovedVersion` |
| M3 semver → string (`highestApprovedVersion`) | **KILLED** (2) | `TestHighestApprovedVersion` |
| M3b semver → string (render site) | **KILLED** (4) | `OrdersBySemverNotByString`, `VersionDriftWarningIsThreeWay` |
| M4 drop the manifest-slug match | **KILLED** (3) | `SilentForADifferentApp` |
| M5 newest approved, not highest | **KILLED** (4) | `UsesHighestApprovedNotNewestApproved` |
| M6 drop the local `isParseableVersion` gate | **KILLED** (2) — *see below* | `ManifestGateRunsBeforeTheNetwork` |
| M7 drop the listing `err != nil` gate | **KILLED** (1) | `WarnLocalVersionDriftIgnoresAnErroredListing` |
| M8 drop the manifest-load gate | **KILLED** | `SilentWithoutALocalManifest` |
| M8a drop **only** the `err != nil` half of it | **SURVIVED — equivalent** | — |
| M9 remedy loses its `[dir]` | **KILLED** (3) | `DriftRemedyCommandIsRunnable…`, `RemedyReachesTheUser` |
| M10 advisory check back before the render | **KILLED** (3) | `RendersTheAnswerBeforeTheAdvisoryDriftCheck` |
| M11 never reuse rows | **KILLED** (11) | `ReusesTheRows…`, + every slug-path count |
| M12 "reuse" even when rows are nil | **KILLED** (12) | `WarnsOnTheIDPath…`, `ManifestGate…`, `ListingFails` |
| M13 drop the in-function parseability guard | **KILLED** (4) | `VersionDriftWarningRefusesUnparseableVersionsItself` |
| M14a case-sensitive status compare | **KILLED** (4) | `NormalisesTheStatus`, `WarnsOnACapitalisedApprovedStatus` |
| M14b untrimmed status compare | **KILLED** (4) | same two |

🔴 **The fix round did break a guard, and this is the report of it.** **M6 survived the first re-run.** Fix 4 makes `versionDriftWarning` refuse an unparseable version itself, which made the caller's `isParseableVersion` gate redundant *for output* — deleting it changed no printed byte anywhere and the whole suite stayed green. It is not redundant in fact: it is what keeps an uncomparable manifest from spending a request. A case was added to `TestAppStatusDriftManifestGateRunsBeforeTheNetwork` (unparseable local version, `--id` path, **want 1 request**) and M6 now dies for its own reason. Redundant guards need a killing reason each, or one of them is untested.

**M8a is an equivalent mutant, not a gap.** `manifest.Load` returns `nil` on every error path (`internal/manifest/manifest.go:57-71`), so `err != nil ||` in `if err != nil || m == nil` is unreachable-by-construction today and no input can kill it. Dropping the gate *entirely* (M8) does die.

## Gates

- `make ci` — green (tidy, vet, test, build).
- `go test -v -count=1 ./internal/cmd ./internal/appapi` — **2409 RUN / 2408 PASS / 0 FAIL / 1 SKIP**, no `panic: test timed out`. The SKIP is the pre-existing opt-in live guard `TestSubmissionsCapDriftAgainstLiveAPI`.
- `golangci-lint run ./...` (via `nix-shell`) — **`0 issues.`** Instrument re-validated this round: a planted unused variable made the same invocation report `1 issues: typecheck: 1`, then was reverted.
- `gofmt -s -l .` — clean.

## Knowingly deferred

Audit nits 8–11 are **not** fixed and are recorded rather than silently dropped: the writer-that-never-writes shape of `versionDriftWarning`'s API; the "highest version *you* have had approved" wording; cwd-only scope vs `app submit [dir]`; and three versions on one screen (open issue #410).

⚠️ **Merge note for #414:** the accessor ledger in `internal/cmd/newest_row_pick_test.go` had to be updated here — the derived-set guard fails otherwise — so this PR now touches that file too. The edit is `submissionsRouteAccessors` (~L504); #414's is a `submissionsListReaders` entry (~L471). Different declarations, non-overlapping hunks, but the file is no longer disjoint between the two PRs. The `highestApprovedVersion` redeclaration in #414 is unchanged and still resolves when whichever lands second rebases.

ZacxDev:
## Delta re-audit leftovers closed — `845cc4b`

All four findings from the re-audit that verdicted this PR safe to merge. No production behaviour changed: the only non-comment production edit is `const driftLookupTimeout` → `var`, so a test can shrink it.

### 1 🟡 `GetSubmissionRows` documented a row ORDER that nothing enforced

The doc comment promises rows "returned as read (newest first) … NOT filtered or reordered", and `newest_row_pick_test.go` ledgers it *because* it "exposes the ordering itself to its caller" — yet **reversing the returned slice left the whole suite green**. Its only consumer (`highestApprovedVersion`) takes a maximum and is order-independent by construction, and no `appapi` test called `GetSubmissionRows` at all. Exactly the shape that shipped #378/#390.

- **`TestGetSubmissionRowsHandsBackTheRowsInServedOrder`** (`internal/appapi/submissions_test.go`) — three rows, pairwise distinct on ID/version/status/submittedAt (mechanically asserted, not eyeballed) so neither a reversal nor a rotation can coincide with the expectation; the expected order is written out by hand, never read back from what the implementation returned. Also asserts the two returns agree about which end is newest.
- **`TestGetSubmissionRowsReturnsNilRowsOnTheIDPath`** — the other half of the contract: `nil` means "I did not read a listing", never "the listing was empty". That is the bit `driftLister` branches on.
- Ledgered as the third pin on `appapi/appblocks.go`, and the entry's `reads:` now names all three picks.

**Mutants watched die:**
| mutation | result |
|---|---|
| reverse the rows returned by `getSubmissionRows` | **only** `TestGetSubmissionRowsHandsBackTheRowsInServedOrder` fails, across all of `./internal/...` — `row 0 = {id:pubreq_01H …}, want {id:pubreq_03H …}` |
| return `[]Submission{}` instead of `nil` on the `?id=` path | `TestGetSubmissionRowsReturnsNilRowsOnTheIDPath` fails on its own message (+ 7 pre-existing cmd tests) |

### 2 🟢 `driftLookupTimeout` had no killing mutation in the REMOVAL direction

Setting it to `1ns` kills 6 tests — that proves the value is *wired*, not that the deadline exists. Deleting the `context.WithTimeout` and handing `warnLocalVersionDrift` the command's own ctx **survived everything**.

**`TestAppStatusDriftLookupGetsADeadlineOfItsOwn`** — a drift listing that never answers, an advisory budget shrunk to 50 ms, bounded on **both** sides (a fast return alone would also be produced by a check that declined to run), plus reachability controls (2 requests issued; the handler reached its hang; the client, not the server, ended it).

🔴 **It could not be pinned the way the audit suggested, and the comment says why:** `app status` builds its own `ctx := context.Background()` and never reads `cmd.Context()`, so a parent deadline passed to `root.ExecuteContext` is inherited by nothing — measured while writing this, the mutant ran **30 s against a 4 s "parent"**, the two wall clocks simply being independent. What actually bounds the request once the dedicated deadline is gone is the appapi client's **30 s ANSWER budget**, so that is the reference the test measures against. `driftLookupTimeout` became a `var` solely so the test can shrink it — a deadline is not observable from the server side of an `httptest` handler.

An intermediate version of this test died on a *racy* reachability guard instead of on the timing bound (red for the wrong reason); that guard is now polled and runs after the discriminator.

**Mutant watched die:** delete the `context.WithTimeout` at the drift call site → `TestAppStatusDriftLookupGetsADeadlineOfItsOwn` fails at **30.03 s against its 2 s ceiling**, with its own message, and is the **only** failure in `./internal/...`.

### 3 🟢 README ToC

Added `Is your repo behind what you shipped?` under `## Submission status`, next to its sibling. (`readme_nav_test.go` gates `##` headings only — noted below.)

### 4 🟢 The `isApprovedStatus` comment was making a false claim

It said "every other reader of the same field in this binary already folds it". Verified: the two *named* precedents do (`appblocks.go:635` `strings.ToLower`, `app_pull.go:233` `ToLower+TrimSpace`) — but `app_listing.go:303` and `:487` compare `ref.Status == "approved"` raw, and `app_status.go:247` compares `"rejected"` raw. The comment now says what is true, including the concrete consequence: if the server ever answers `"Approved"`, this predicate keeps working while `app_listing.go:487`'s `live :=` flips the wrong way. `app_listing.go` is deliberately **untouched** — that consolidation is not this PR's scope.

### Left alone, as scoped

- `app_status.go:370-373` `!ok` guard — re-confirmed **SURVIVES** deletion, as documented. Unkillable by construction (`highestApprovedVersion` returns `("", false)` iff `best == ""`).
- `app_status.go:126-132` writeJSON-error-before-drift-check (EPIPE only).
- The `TestHighestApprovedVersionNormalisesTheStatus` name collision with #414 — whichever lands second resolves it.

### Gates

- `make ci` — exit 0
- `go test ./... -v -count=1` — **4019 `=== RUN` / 4015 `--- PASS` / 0 `--- FAIL` / 4 `--- SKIP`**, 0 package-level `FAIL`, no `panic: test timed out`
- `make ci-shallow` (depth-1 clone, committed tip) — **ok=20/20, 0 failures, 0 timeouts**
- `golangci-lint run ./...` — **0 issues**

### Full mutation battery re-run after the fix (12 mutants, whole `./internal/...` suite each)

Baseline: 0 failures. `M1 timeout=1ns` → 6 died · `M2 timeout removed` → 1 (the new test) · `M3a case-sensitive` / `M3b space-sensitive` approved compare → 2 each · `M4 drop the approved filter` → 10 · `M5 drift before render` → 1 · `M6 never reuse rows` → 5 · `M7 always reuse rows` → 7 · `M8 !ok guard deleted` → **0, survived as documented** · `M9 manifest gate deleted` → 1 · `M10 GetSubmission picks oldest` → 5 · `M11 rows reversed` → 1 · `M12 empty-not-nil rows` → 8. Nothing that was killing before this round stopped killing.

### Two things noted but NOT changed

1. **`app status` ignores `cmd.Context()`** (`app_status.go:98` builds its own `context.Background()`), so the command is uncancellable from its caller. Pre-existing, unrelated to #412, and out of scope here — but it is why finding 2's pin is shaped the way it is.
2. **The new ToC entry has no gate.** `TestREADMETableOfContentsCoversEverySection` covers `##` only, so a missing `###` entry can rot again exactly as this one did. Widening it to `###` is a separate change (it would need every `###` in the file audited first).

`pins-vs-published` is red at base and unrelated (#408 is its fix).

ZacxDev:
## Integration re-check at final tips — three collisions, not one

Re-measured after both PRs' audit-fix rounds, merging `#413 845cc4b` and `#414 b81626f` onto `main` (`e49ddc6`). Both merges are **textually clean** — including `internal/cmd/newest_row_pick_test.go`, which both PRs now edit in non-overlapping hunks. The merged tree still does not build:

```
internal/cmd/app_submit_version_guard.go:144:6: highestApprovedVersion redeclared in this block
        internal/cmd/app_status.go:410:6: other declaration of highestApprovedVersion
internal/cmd/app_submit_version_guard.go:265:10: assignment mismatch: 1 variable but highestApprovedVersion returns 2 values
```

Swept every top-level declaration in the merged package. Three things collide (`detachAndStart` also appears twice but is pre-existing and build-tag separated — `spawn_unix.go` / `spawn_windows.go` — so it is fine):

1. **`highestApprovedVersion`** — declared in both, incompatible signatures (`approvedPeak` vs `(string, bool)`).
2. **`TestHighestApprovedVersionNormalisesTheStatus`** — both fix rounds independently added a test with this exact name, one in each new test file. Two agents working blind converging on the same function name *and* the same test name is a strong signal this is one predicate written twice.
3. **The two predicates still disagree on policy**, so a rename alone leaves a live contradiction:

| | #413 (`app_status.go:414`) | #414 (`app_submit_version_guard.go:133`) |
|---|---|---|
| pre-release | `isParseableVersion` — truncates at `-`/`+`, so `0.6.0-rc.1` ranks **above** `0.5.2` | `comparableVersion` — declares it **not orderable**, skips it |
| slug | `s.BlockID != blockID`, exact | `sameSlug()`, `EqualFold`+`TrimSpace` |
| liveness | n/a | `rowIsServing()` — `liveUrl` leads, `deployState=="live"` fallback |

With an approved `0.6.0-rc.1` alongside an approved `0.5.2`, `app status` reports the repo BEHIND `0.6.0-rc.1` while `app submit` accepts `0.5.3` — the contradiction #413's own doc comment warns against.

### Suggested land order

`#408` (unblocks `pins-vs-published`, which is red at unmodified `main` and unrelated to either PR) → **#413** → **#414** rebased, consolidating all three collisions in that rebase.

#414's implementation is the better survivor for the merged predicate: stricter on pre-release ordering, normalises the slug, and carries the unorderable-rows list plus the `rowIsServing` liveness that #414's delta audit showed is required to avoid telling an author a serving app is not live. #413's `versionDriftWarning` then consumes `peak.version` / `peak.found`, its `("0.5.2", true)` assertions adapt to the struct, and one of the two identically-named tests is renamed or merged.

Separately worth its own issue (deliberately not fixed in either PR, since it spans files both own): the slug predicate is now open-coded **three** times — `sameSlug` normalises, while `apps.go:262` and `appblocks.go:632` still compare raw.

Issue #411, opened 2026-08-14
app submit records no provenance and accepts an uncommitted tree — five first-party apps are now behind their live version


ZacxDev:
## Summary

`civitai app submit` packages whatever is in the directory and records **nothing about where it came from**. A bundle built from a tree that was never committed submits exactly as cleanly as one built from a tagged release, and afterwards there is no way to tell which was which.

The result is silent, permanent drift between what is deployed and what is in the source repo — and because no provenance is stored, the drift cannot be diagnosed after the fact, only observed.

## Evidence this happens in practice

Five first-party apps are currently behind their live deployed version:

| app | source repo | live |
|---|---|---|
| `custom-generators` | 0.4.0 | **0.5.2** |
| `gen-matrix` | 0.7.3 | **0.8.2** |
| `model-benchmarking` | 0.1.3 | **0.2.2** |
| `playable-collections` | 0.2.0 | **0.2.2** |
| `app-requests` | 0.2.0 | **0.2.2** |

`custom-generators` has **never** held 0.5.2 anywhere in its git history — the file goes 0.3.3 → 0.3.4 → 0.4.0 — and the newer source is not in any local clone either.

The publish-request timestamps identify the mechanism. These are `pubreq` records, i.e. real `app submit` calls, not listing-media revisions:

```
2026-07-29 23:33:34  app-requests        0.2.2
2026-07-29 23:33:36  gen-matrix          0.8.2
2026-07-29 23:33:41  custom-generators   0.5.2
```

Three apps in **7 seconds** — a scripted batch — with the same shape earlier that day (20:36–20:47) and on 07-27 (21:55:0x). Each app's repo diverges at exactly those batches.

Nothing here is unusual or wrong to want to do; batch-submitting is reasonable. The problem is that it leaves no trace.

## Why it can't be diagnosed today

`Submission` (`internal/appapi/appblocks.go:65-81`) carries:

```
ID, BlockID, AppBlockID, Version, Status, RejectionReason, ApprovalNotes,
DeployState, DeployDetail, DeployUpdatedAt, SubmittedAt, ReviewedAt,
UpdatedAt, CreatedAt, LiveURL
```

There is no commit field, no tree hash, no dirty flag. Given a deployed version, there is no way to answer "which commit is this?" — the information was never captured. That is the part that makes recovery archaeological: the bundle does land in Forgejo `civitai-apps/<slug>`, but nothing links it back to a source revision.

## Proposed fix

[...]

PR #414, opened 2026-08-14
fix(app submit): refuse a version at or below the highest APPROVED one (#412 piece 1)

ZacxDev:
Piece 1 of #412: the **monotonic-version guard** on `civitai app submit`. (Piece 2 — the drift line in `app status` — is a separate PR, so this is `Refs #412`, not `Closes`.)

Before packaging, and only on the path that would really upload, `app submit` now reads the app's own submissions, computes the **highest APPROVED** version, and refuses when the manifest version is not strictly above it:

```
$ civitai app submit --yes
Error: refusing to submit custom-generators@0.4.1 — 0.5.2 is already approved and live.
Submitting an older version replaces the newer deployment on approval.
Your repo may be behind what was last released. Pass --allow-downgrade if this is deliberate
```

Entirely client-side: `GET /api/v1/blocks/submissions` already carries every row's `version` and `status`, so no server change is involved.

Two adaptations to the issue's wording, both to avoid making a claim the data does not support:

- **"and live" is conditional** on the winning row's `deployState` really being `live`. An approved row still building says only "is already approved".
- **The middle line differs for the EQUAL case** — nothing is being replaced by something older there, so it reads "Resubmitting the version that is already live is almost always an accident."

---

### 1. Semver ordering — reuse the existing comparator, no new dependency

There is no semver dependency in `go.mod` and this does not add one. `internal/cmd/update_check.go` already had a minimal `parseSemver` + `compareVersions`; the numeric ordering is now extracted into `func (a semver) compare(b semver) int` and **shared** by both callers, while their *policies* stay separate — `compareVersions`' `-1`/`0` fallbacks for unparseable input are tuned for "should we advertise an update", which is the wrong default for "may this submit replace what is live". That is the "one rule, one place" split: share the ordering, not the policy.

`0.10.0 > 0.9.0` is asserted in both directions (see the mutation table: a string compare dies to `TestVersionGuardOrdersBySemverNotByString`).

**On `0.5.0-beta.1`** — checked explicitly rather than assumed: `parseSemver` truncates at the first `-`/`+`, so it reports `0.5.0-beta.1` and `0.5.0` as the **same** version. Real semver orders `0.5.0` *above* `0.5.0-beta.1`, so ordering on that lossy compare would refuse a legitimate release-over-its-own-prerelease submit and call it a regression. Rather than vendor a full semver implementation for a shape the manifest schema does not encourage, `comparableVersion` declares any version carrying pre-release or build metadata **NOT COMPARABLE** and routes it to decision 2.

### 2. Unparseable / unorderable versions — warn and proceed, never silently

Both directions are handled and both are tested:

- **local version unorderable** → warn (`cannot order <slug>@<v> against the approved <x> …`), proceed. Refusing would block any app on a pre-release tag.
- **some approved rows unorderable** → warn naming them (`ignoring N approved version(s) … <list>`), and still compare against the highest row it *can* order. Silently dropping them would be a fail-open the author never sees.
- **every approved row unorderable** → warn, proceed. There is no verdict to reach.

Nothing here is an accident of control flow: each branch is a named `return nil` with a comment, and each has both a behavioural test and a mutant (M10, M12, M13) that reddens a specific assertion.

### 3. Network failure / no prior submissions

- **`ListSubmissions` errors** → **warn and proceed.** This guard is an accident preventer, not an authorization check. Hard-failing would mean a transient API blip blocks *every* submit in the fleet — strictly worse than the failure it prevents. The warning carries the underlying error, so it is a *loud* fail-open, not a silent one. (M07 = hard-fail, M12 = silent fail-open; both die.)
- **no prior submissions at all** → proceed **silently**. A first submit is the normal case with nothing to compare against, and a warning there would be noise on the happy path.
- **rows exist but none approved** (only pending / rejected / withdrawn) → proceed silently. Nothing is serving, so nothing can be regressed.
- **`--allow-downgrade`** short-circuits **before** the network read — a deliberate rollback should not pay for an answer it would discard, and a release script must not fail on it during an outage.

### 4. Exit code — **1**, not 2

Read `internal/cmd/exitcodes_doc.go` in full. `2` is documented as a mistake about the **invocation**; when this fires, every flag, argument and path in `civitai app submit --yes` is well-formed. What is wrong is the **project** relative to what is published — the same shape as an invalid manifest, which code 1 already publishes as "a validation verdict". So the refusal is deliberately left **untagged** for `exitCode`'s `default`.

Because "untagged" is indistinguishable from "forgot to tag", the decision is pinned rather than assumed:

- a new sentinel `cmd.ErrVersionRegression`, attached with the established `civitai.Tag` pattern (`pkg/civitai/errkind.go`) so `errors.Is` reports the kind while `err.Error()` is byte-identical;
- `cmd/civitai/submit_version_guard_exitcode_test.go` asserts `exitCode(...) == exitGeneric` **plus a negative control** (`ErrBadRequest → 2`) so the assertion is not a fact about a function that always answers 1, and asserts the new sentinel matches none of the five API kinds (with a positive control on the `errors.Is` walk itself);
- every test asserts the **sentinel via `errors.Is`**, never a substring of the message.

The refusal path is enumerated in the docs, so `exitcodes_doc.go` gained a bullet under code 1 and README.md was regenerated from it — `TestREADMEExitCodeTableIsGenerated` / `TestREADMEExitCodeSectionsAreGenerated` are green. A Troubleshooting row and the command-reference row were updated too.

---

The new test files reference symbols that do not exist on `main`, so they cannot compile there — that is red, but uninformative. So a **symbol-free probe** of the same regression was written to compile against `origin/main` and watched to fail, in a scratch worktree at `e49ddc6`:

| Test | at `origin/main` (e49ddc6) | at HEAD |
| --- | --- | --- |
| `SubmitMustRefuseARegression` (probe: `app submit` 0.4.1 under an approved+live 0.5.2 must not reach the submit route) | **FAIL** — "`app submit` accepted 0.4.1 …"; the submit route was hit and the server answered `pubreq_new … pending` | **PASS** |
| `ForwardBumpStillSubmits` (negative control, same wiring, 0.6.1) | **PASS** | **PASS** |

The negative control passing at base is what makes the failure attributable to the guard rather than to the harness. The probe was deleted after measuring; its behaviour ships as `TestAppSubmitRefusesARegressionAndNeverUploads` + `TestAppSubmitAllowsAForwardBumpEndToEnd`.

The shipped test files at `origin/main`: `vet: internal/cmd/app_submit_version_guard_test.go:77:9: undefined: checkVersionNotRegression` and `vet: cmd/civitai/submit_version_guard_exitcode_test.go:33:25: undefined: cmd.ErrVersionRegression`.

At HEAD, the guard's own subset is **RUN=36 PASS=36 FAIL=0**; `make ci` and `make ci-shallow` are green (20/20 packages, 0 timeouts); `golangci-lint run ./...` reports **0 issues**.

---

## Anti-vacuity: a fixture that survived a mutant, and the fix

🔴 **Worth reading even if you skip the rest.** The first version of `regressionFixture` had five rows that disagreed on newest-row, highest-version, and other-app — and the mutant *"take the first approved row of a newest-first list instead of the maximum"* **SURVIVED the entire file**. Because the list is newest-first and the highest approved row also happened to be the newest approved row, "newest approved" and "highest approved" returned the same answer, and every case passed under either predicate. Exactly the #390 shape.

The fixture now carries `pubreq_rollback` — an approved `0.4.9` submitted **after** the approved `0.5.2`, with `deployState: failed` (a rollback whose deploy failed) — so the two predicates diverge, and M04 now dies to five tests. `TestHighestApprovedVersionIsNotTheNewestRow` also asserts the divergence itself, so a later edit that flattens the fixture fails there rather than quietly making the file vacuous again.

Every row's id, version, status, timestamp and deploy state is pairwise distinct, so a wrong-field read cannot coincide with the right answer. The refusal test also asserts the message contains **none** of `0.6.0 / 0.7.1 / 0.9.9 / 9.9.9 / 0.4.9` — the five distinct wrong answers.

---

## Mutation table — 14 hand-mutants, 0 survivors

Each mutant was applied to the real source, the guard's test subset re-run, and the tree restored from a byte-checked copy (`git stash` is repo-global here and was not used). Control: RUN=36 PASS=36 FAIL=0. **Tree restored byte-identical after every mutant.**

| # | Mutant | Killed by (first failing assertion) |
| --- | --- | --- |
| M01 | comparison inverted to allow **EQUAL** | `TestVersionGuardRefusesEqualVersion` — "resubmitting the approved version 0.5.2 must be refused" |
| M02 | comparison inverted to allow **LOWER** | 7 tests, incl. `TestVersionGuardRefusesBelowHighestApproved` — "submitting 0.4.1 while 0.5.2 is approved+live must be refused" |
| M03 | `approved` status filter dropped | 8 tests; `TestHighestApprovedVersionIsNotTheNewestRow` — `highest approved = "0.9.9", want 0.5.2` |
| M04 | first approved row (**newest**) instead of **highest** | 5 tests; `TestHighestApprovedVersionIsNotTheNewestRow` — `highest approved = "0.4.9", want 0.5.2` |
| M05 | slug filter dropped | 8 tests; `TestHighestApprovedVersionIgnoresOtherApps` — `another app's approved row must not answer …, got "9.9.9"` |
| M06 | semver compare → **string** compare | `TestVersionGuardOrdersBySemverNotByString` — "0.10.0 > 0.9.0 by semver and must be allowed (a string compare refuses it)" |
| M07 | network failure becomes a **hard fail** | `TestVersionGuardWarnsAndProceedsWhenTheLookupFails` — "a failed submissions lookup must not block the submit" |
| M08 | `--allow-downgrade` no longer short-circuits | `TestVersionGuardAllowDowngradeSkipsTheCheckEntirely` + `TestAppSubmitAllowDowngradeUploadsAnyway` |
| M09 | guard **not wired** into `app submit` | `TestAppSubmitRefusesARegressionAndNeverUploads` — "`app submit` must refuse 0.4.1 under an approved 0.5.2" (the reachability proof) |
| M10 | pre-release treated as orderable (lossy truncation) | 3 tests; `TestVersionGuardWarnsOnUnorderableLocalVersion` — "an unorderable local version must not hard-fail, got: refusing to submit …@0.5.0-beta.1" |
| M11 | sentinel tag dropped | 3 tests; "refusal must carry ErrVersionRegression, got `&errors.errorString{…}`" |
| M12 | fail-open warning **silenced** | `TestVersionGuardWarnsAndProceedsWhenTheLookupFails` — "the fail-open must be announced" |
| M13 | skipped-approved-version warning silenced | `TestVersionGuardWarnsAndStillComparesWhenSomeApprovedRowsAreUnorderable` — "the skipped approved version must be reported, not silently dropped" |
| M14 | `live` claimed unconditionally | `TestVersionGuardDropsTheLiveClaimWhenNotDeployed` — "deployState is \"building\", so the message must not claim live" |

Every kill is attributed to a **named assertion of this guard's own**, not merely "a test failed".

---

## Ordering, stated because it is a trade

The guard runs **after** `confirmSubmit` and **before** packaging. Putting it first would avoid prompting for a submit we are about to refuse — but it would also make a bare `civitai app submit` in a non-TTY shell (the accidental-footgun case, which refuses for lack of `--yes`) reach the API first, and `TestAppSubmit_NonTTYRefusesWithoutYes_NoNetworkCall` pins that invocation as touching nothing. The cost of this order is one confusing sequence on an interactive TTY — prompt, `y`, then the refusal — which is better than a confirmation gate that quietly acquired a network dependency. The reasoning is in the code, not just here.

## Seam ledger

`internal/cmd/app_submit.go` now reads the submissions route, so `TestSubmissionsListReadersAreLedgered` (from #390) demanded an entry — it is the ledger's **one deliberately non-newest-row reader**, and the entry says why: only an approved row can be replaced on approval, so a later pending/withdrawn/rejected row must not answer. The pinning test is the end-to-end `TestAppSubmitRefusesARegressionAndNeverUploads`.

## Not verified

`make mutate PKG=./internal/cmd` (gremlins) was **not** run. That package's suite takes ~24 s per invocation and gremlins would need hundreds of mutant runs; the tool's own header calls it a one-package investigation tool and defaults to `internal/validate` (0.17 s). The 14-mutant hand battery above is what was run instead, and unlike gremlins it can mutate function arguments, struct fields and message construction — most of what this change is.


ZacxDev:
CI note: `pins-vs-published` is **red for a reason unrelated to this PR**, and the discriminating control was run rather than assumed.

That job queries **live npm** (`CIVITAI_CHECK_PUBLISHED_PINS=1`) and reports `@civitai/app-sdk ^0.31.0` vs published `0.34.0` and `@civitai/blocks-react ^0.39.0` vs published `0.42.0` in `templates/page-money/package.json.tmpl` — files this PR does not touch.

Control: the same test run in a **clean worktree of unmodified `origin/main` (e49ddc6)** fails identically, with the same two packages and the same versions. It is green on the last five `main` runs only because those ran on 2026-08-12, before those npm releases. #408 is the fix.

Every other check is green: `build-test`, `lint`, `schema-drift`, `ready-ack-runtime`, `scaffold-currency`, `template-page-money`, `template-page-vite`, and all four CodeQL analyses.

ZacxDev:
## ⚠️ #413 and #414 are each green alone but do not compose — do not merge both without consolidating

Both PRs implement #412 and both add a `highestApprovedVersion` to package `cmd`. The git merge is textually clean; the merged tree **does not compile**.

Verified by merging both branches onto `main` (`e49ddc6`) in a scratch worktree and building:

```
internal/cmd/app_submit_version_guard.go:92:6: highestApprovedVersion redeclared in this block
        internal/cmd/app_status.go:351:6: other declaration of highestApprovedVersion
internal/cmd/app_submit_version_guard.go:163:10: assignment mismatch: 1 variable but highestApprovedVersion returns 2 values
```

### Renaming one of them would NOT be enough — the two predicates disagree

They differ on approved rows carrying a pre-release / build suffix:

- **#413** gates on `isParseableVersion` + `compareVersions`, which truncate at the first `-`/`+`.
- **#414** introduces the stricter `comparableVersion`, which declares such a version **not orderable** and routes it to warn-and-proceed.

Probed on unmodified `main`:

```
parseSemver("0.6.0-rc1")     -> {major:0 minor:6 patch:0} ok=true ; isParseableVersion=true
parseSemver("0.6.0+build.1") -> {major:0 minor:6 patch:0} ok=true ; isParseableVersion=true
compareVersions("0.6.0-rc1","0.5.2") = 1
```

So with an approved `0.6.0-rc1` alongside an approved `0.5.2`, `app status` (#413) picks `0.6.0-rc1` as the highest approved version, while `app submit` (#414) skips it and lands on `0.5.2`. `status` would report the repo BEHIND `0.6.0-rc1` while `submit` accepts `0.5.3` without complaint — the exact contradiction #413's own doc comment warns against ("the two commands have to name the same number or they contradict each other").

### Suggested resolution

Whichever PR lands second consolidates to a single implementation. #414's `approvedPeak` is the better survivor: it is stricter on pre-release ordering and additionally carries the unorderable-rows list and the `live` deploy state. #413's `versionDriftWarning` then consumes `peak.version` / `peak.found`, and #413's `("0.5.2", true)` assertions adapt to the struct.

Both are currently blocked by `pins-vs-published`, which is **unrelated and pre-existing**: it compares scaffold template pins against live npm (`@civitai/app-sdk ^0.31.0` vs published `0.34.0`, `@civitai/blocks-react ^0.39.0` vs `0.42.0`) and fails identically at unmodified `main`. It is green on main's last runs only because those predate those npm releases. #408 is the fix.

ZacxDev:
## Audit findings 2, 3 and 4 fixed — `28b649a`

All three reproduced first against the code as it stood, then fixed, then re-verified. No code from #413 was touched and no function was renamed; the duplicate-`highestApprovedVersion` compile break stays for whichever PR lands second.

---

### 🟡 2 — the equal-version refusal asserted "already live" when it was not

**Reproduced** against `{0.5.2, approved, deployState: "failed"}` at local `0.5.2`, exactly as reported:

```
refusing to submit custom-generators@0.5.2 — 0.5.2 is already approved.
Resubmitting the version that is already live is almost always an accident.
Your repo may be behind what was last released. Pass --allow-downgrade if this is deliberate
```

Line 1 says not live, line 2 says live, line 3 says behind a version the repo is exactly at.

**Fixed.** `versionRegressionError` now branches on `peak.live` **and** equal-vs-lower, and every line says only what is known. Same input now:

```
refusing to submit custom-generators@0.5.2 — 0.5.2 is already approved.
That version is approved but not live, so this may be a deliberate retry of a deploy that never landed.
Bump the version in your manifest, or pass --allow-downgrade to resubmit it unchanged
```

It **still refuses** — #412's contract is that the escape hatch is explicit — but it names the retry instead of denying it. `--allow-downgrade` is the documented way to resubmit unchanged.

**Pinned by** `TestVersionGuardEqualVersionThatIsNotLiveNeverClaimsItIsLive` (the combination no fixture in the file had: equal **and** not live, across `failed`/`building`/`deploying`), plus the two arms that keep the switch from collapsing: `…EqualVersionThatIsLiveStillCallsItAnAccident` and `…LowerVersionKeepsTheBehindWording`.

Also updated the README Troubleshooting row: it was keyed on `Your repo may be behind what was last released`, which is now only the lower-version case. It names all three second lines.

---

### 🟡 3 + 4 — re-run mutant battery

Control **2090 RUN / 2090 PASS / 0 FAIL / 0 timeouts** (`go test ./internal/cmd/ -v`, counted, not read off an exit code). Two deliberately-broken controls confirmed the harness can go red **before** any mutant was scored: `ctl-noop` (guard never refuses) → 13 FAIL; `ctl-first` (peak = first slug row) → 7 FAIL. Every mutant asserts its target string occurs **exactly once** before it is applied, and the tree is restored from pristine copies between runs.

| # | Mutation | Verdict | Killed by / message |
|---|---|---|---|
| a | status `EqualFold+TrimSpace` → `s.Status != "approved"` | **DIES** | `TestHighestApprovedVersionNormalisesTheStatus` — `Status=" Approved " must count as approved — an exact compare here silently disables the guard` (×4 spellings) |
| b | tie-break `<= 0` → `< 0` | **DIES** | `TestHighestApprovedVersionKeepsTheFirstOfTwoEqualApprovedRows` — `live-first: peak = "v0.5.2" live=false, want "0.5.2" live=true` **and** `failed-first: peak = "0.5.2" live=true, want "v0.5.2" live=false` (both orders, opposite expectations) |
| c | `peak.live = …` → `peak.live \|\| …` | **DIES** | `TestHighestApprovedVersionLiveFlagBelongsToTheWinningRow` — `the refusal should name 0.6.0 as approved-not-live, got: … 0.6.0 is already approved and live.` |
| d | drop `TrimPrefix(…, "v")` in `comparableVersion` | **SURVIVES — equivalent mutant** | see below |
| e | deployState `EqualFold+TrimSpace` → `== "live"` | **DIES** | `TestHighestApprovedVersionNormalisesTheDeployState` — `deployState="Live": the refusal should claim live, got: … 0.5.2 is already approved.` |
| f | `lister(ctx, slug)` → `lister(ctx, "")` | **DIES** | `TestVersionGuardNarrowsTheListingToThisApp` — `the guard read the submissions route with blockId="", want "custom-generators"` |
| g | warn writer `cmd.ErrOrStderr()` → `cmd.OutOrStdout()` | **DIES** | `TestAppSubmitGuardWarningsGoToStderrNotStdout` — `the guard warning leaked onto STDOUT ("2.0.0-rc.1") — stdout is the machine-readable channel for this command` |
| h | slug filter (see finding 4) | **DIES** | `TestHighestApprovedVersionNormalisesTheSlug` — `blockId="Custom-Generators" must match slug "custom-generators" — an exact compare here disables the guard silently` |
| i | *new:* delete the finding-4 announcement | **DIES** | `TestVersionGuardAnnouncesWhenTheListingHeldNoRowForThisApp` — `rows came back and NONE was for this app — that is a blockId mismatch, not a first submit` |
| j | *new:* revert the finding-2 message fix | **DIES** | `TestVersionGuardEqualVersionThatIsNotLiveNeverClaimsItIsLive` — `deployState="deploying": this is the deploy-failed RETRY path and the message should say so rather than calling it an accident` (+ the README drift guard, correctly) |

Each mutant fails **exactly one** guard test (j reddens two: its own test and the README symptom-index guard, which is the right second reader). No mutant died to an unrelated check's error.

#### (d) survives, and it is not a coverage gap

`parseSemver` (`update_check.go:229-231`) already does `TrimSpace` **and** `TrimPrefix(s, "v")`, so `comparableVersion`'s own `TrimPrefix` is redundant: removing it changes the answer for no realistic input. Measured directly — both spellings run over a 39-string corpus (plain, `v`-prefixed, padded, empty, pre-release, build-metadata, 4-component, non-numeric, negative): **0 diverge**. The probe is not wired to nothing: the positive control `"vv0.5.2"` **does** diverge (`real=({0 5 2},true)` vs `mutant=({0 0 0},false)`), which is the only input class that differs.

So there is no honest test that kills (d) without a nonsense fixture. What I did instead: `TestVersionGuardOrdersAVPrefixedApprovedVersion` pins the observable contract both sites implement together — a `v`-prefixed approved version is orderable rather than routed to warn-and-proceed, orders equal to its bare form, is quoted verbatim in the refusal, and a `v`-prefixed local bump is allowed. That reddens if the prefix handling is removed from **both** sites. The redundancy is now stated in the code so it does not read as load-bearing, and the test says in its own doc comment what it does and does not prove.

---

### 🟡 4 — decision: **both**, and here is why

1. **Normalise** — `sameSlug` compares `EqualFold(TrimSpace(…))`, matching how `Status` and `DeployState` are already compared.
2. **Announce** — when the listing came back **non-empty** and **no row matched the slug**, the guard now warns instead of proceeding silently.

Reasoning. The asymmetry was the smaller half of the problem; the silence was the larger. Every other non-refusal branch announces itself, and this one does not — by design, because "no approved rows" is what a genuine first submit looks like. That makes a slug mismatch **indistinguishable from the happy path**, so the #412 trap returns with zero signal on exactly the app the guard exists to protect. Normalising removes the *likely* mismatch (casing, stray whitespace). It cannot remove the residue, because — as the audit says and I could not settle either — nothing in this CLI pins how the server cases `blockId` on that route. An unverifiable assumption should fail **loudly**, so the residue is announced.

The announcement is deliberately narrow: `len(subs) > 0 && peak.slugRows == 0`. Rows that *are* for this app but none approved yet stay silent — that is an ordinary not-yet-approved app, and `TestVersionGuardStaysSilentOnAGenuineFirstSubmitWithRows` is the negative control that stops the condition being widened to "no approved rows", which would put a warning on the happy path. `TestHighestApprovedVersionNormalisesTheSlug` carries its own negative control too (`gen-matrix`, `custom-generators-2`, `custom`, `""` must still **not** match), so the normalisation cannot be widened into "everything matches".

---

### Gates

| Gate | Result |
|---|---|
| `make ci` | rc=0, all 21 packages `ok` |
| `make ci-shallow` | `PASSED in a depth-1 clone: 20 package(s) ok, 0 failures, 0 timeouts` |
| `golangci-lint run ./...` | **`0 issues.`** (read from the output, not the exit code — the nix-shell banner shares stdout) |
| `go test ./... -v`, counted | **3979 RUN / 3975 PASS / 0 FAIL / 4 SKIP / 0 `panic: test timed out`** |

The 4 skips are pre-existing network/live-API tests (`TestScanDirFromEnv`, `TestSubmissionsCapDriftAgainstLiveAPI`, `TestScaffoldPinsSatisfyPublished`, `TestScaffoldedReadyAckActuallyFires`).

CI's `pins-vs-published` is red at base and unrelated (stale npm scaffold pins; #408 is the fix) — not chased.

**Not verified, stated plainly:** how the server actually cases `blockId` on `GET /api/v1/blocks/submissions`. That is the open question behind finding 4, and it is the reason the fix is normalise **plus** announce rather than either alone.

ZacxDev:
## Delta re-audit fixes — `b81626f`

All findings addressed except the two explicitly out of scope. Every mutant below was **watched to fail with its own message**; the tree was restored and re-verified green after each.

---

### 🔴 1 — the `equal:` arm made a false claim and then recommended the #412 footgun

**Confirmed at the source, both sides.** `isApprovedAndServing` (`civitai:src/shared/constants/app-block-deploy.constants.ts:106-119`) returns **true** for an approved row with `deployState === null` when `deployUpdatedAt` is set, `reviewedAt` is null, or `reviewedAt < DEPLOY_STATE_TRACKING_EPOCH_MS`. `shapeRow` (`civitai:src/pages/api/v1/blocks/submissions.ts:132,148`) emits `liveUrl` from **exactly** that predicate — so `liveUrl != null` *is* the server saying "serving", and for legacy rows it is the **only** field on this route that says it. `app_status.go:216` gates `Live at:` on `s.LiveURL != nil && *s.LiveURL != ""` — the same field, which is why the two commands contradicted each other.

**Fix:** liveness moved into `rowIsServing(s)` — `liveUrl` leads, `deployState == "live"` retained as a **fallback** so a server that stopped emitting `liveUrl` degrades to the old (understated) answer rather than claiming nothing is ever live. A blank `liveUrl` does not count as serving.

**Reproduced against the real binary**, same stub row (`deployState:null`, `deployUpdatedAt` set, `liveUrl:"https://custom-generators.civit.ai/"`), `app submit` at `0.5.2`:

```
# 28b649a (pre-fix)
Error: refusing to submit custom-generators@0.5.2 — 0.5.2 is already approved.
That version is approved but not live, so this may be a deliberate retry of a deploy that never landed.
Bump the version in your manifest, or pass --allow-downgrade to resubmit it unchanged

# HEAD
Error: refusing to submit custom-generators@0.5.2 — 0.5.2 is already approved and live.
Resubmitting the version that is already live is almost always an accident.
Bump the version in your manifest, or pass --allow-downgrade if this is deliberate
```

**Pins + mutants**

| mutant | result |
|---|---|
| `liveUrl` read deleted (the shipped defect) | **killed** — 5 fails, incl. *"the server reports this row as serving (liveUrl is set), so the refusal must not tell the author it is not live"* |
| blank-check dropped (`LiveURL != nil` alone) | **killed** — *"an empty liveUrl is not the server saying the row is serving"* |
| `rowIsServing` always true | **killed** — 12 fails |
| `deployState` fallback deleted | **killed** — 15 fails |
| **`json:"liveUrl"` → `json:"live_url"`** | **killed by the e2e case only** |

🔴 **The e2e case was VACUOUS on its first cut and I only found that by mutating it.** `submitGuardServer` marshals the *same struct* the CLI unmarshals, so renaming the tag applied symmetrically and the round trip still worked — green. It now serves a **literal JSON body** spelled the way `shapeRow` spells it (`rawSubmitGuardServer`), and the tag mutant reddens. This is the seam every unit test in the file is structurally blind to, because they all build `appapi.Submission` in Go.

---

### 🟡 2 — the announcement fired with an EMPTY slug

Returns **before** the network read when `slug` is blank, with a message about the real cause. Measured on the wire, same manifest with no `blockId`:

```
# pre-fix — server log: GET /api/v1/blocks/submissions   (NO blockId — unnarrowed)
⚠ the submissions listing returned 1 row(s), none of them for  — the version guard has nothing to
  compare against. If  is already published, this is a blockId mismatch rather than a first submit.

# HEAD — no submissions GET at all
⚠ this manifest declares no blockId, so the version guard has no app to compare against — submitting without it.
```

Dropping the unnarrowed read is a second win: `submissionsURL` omits an empty `blockId` (`appblocks.go:753-765`) and the server only narrows `if (blockId)` (`submissions.ts:302`), so that read asked for **every** app's rows against a capped route — the exact failure `TestVersionGuardNarrowsTheListingToThisApp` exists to prevent.

Mutant *(branch deleted)* → **killed**, and the failure reproduces all four reported artifacts including `"none of them for  "` and `"If  is already published"`.

---

### 🟢 3 — comments corrected

Verified your reading: `submissions.ts:302` sets `where.slug = blockId` (exact Prisma match) and `:135` echoes `blockId: row.slug` from a non-nullable column — so every returned row matches byte-for-byte, and mis-casing yields **zero** rows, not mis-cased ones. Both comments (`sameSlug`, and the announce branch) now say **defence in depth against an undocumented server change, not a live hazard**, and the announce branch additionally records that it is *unreachable* against the route as documented. The checks are kept.

---

### 🟢 5 — line order pinned

`TestVersionRegressionMessageLineOrderIsPinned` asserts exactly 3 lines and pins lines 2 and 3 **by equality** for all four arms. Mutant M17 (swap `middle`/`tail` in the `fmt.Errorf` args) → **killed**, 5 fails: *"line 2 (the README's \"which case you are in\" line) = …, want …"*.

---

### 🟢 6 — one-row announcement boundary

`TestVersionGuardAnnouncesWithExactlyOneForeignRow`. Mutant `len(subs) > 0` → `> 1` → **killed** (survived the two-row fixture, as you said).

---

### 🟢 7 — residual overstatements, and a fourth arm

`versionRegressionError` now has **four** arms — `(equal|lower) × (live|not)`. The lower-version case was one arm asserting *"replaces the newer deployment on approval"* whatever `peak.live` held: the same false-claim class the equal arm had just been fixed for, left standing one branch over.

| arm | line 2 |
|---|---|
| equal + live | `Resubmitting the version that is already live is almost always an accident.` |
| equal + not live | `That version is approved but not live, so this may be a deliberate resubmit of a deploy that has not landed.` |
| lower + live | `Submitting an older version replaces the newer live deployment on approval.` |
| lower + not live | `That version is approved but not live, so approving an older one deploys code older than the highest approved version.` |

- `never landed` → `has not landed` (false while `building`/`deploying` — a deploy in progress).
- `resubmit it unchanged` → `submit this version again` (#411: the CLI packages the working tree and cannot know).

🔴 **My own first draft of the lower+not-live arm was itself false** and I caught it re-auditing the delta: it said *"supersedes it as the highest approved version"*, which is wrong by this guard's own definition — `highestApprovedVersion` takes the **maximum**, not the newest, so approving `0.4.1` under an approved `0.5.2` leaves `0.5.2` the highest approved. Corrected to a claim about **code**, which is what approval actually deploys.

Two claims are deliberately **not** made anywhere: that nothing at all is deployed for the app (a *lower* approved version can be serving while the peak is not — this route does not say and the guard does not look), and anything about the bundle's contents.

`TestVersionRegressionArmsClaimOnlyWhatIsKnown` sweeps all four with required *and* forbidden strings. Mutants: lower+not-live claims a deployment → **killed** (*"this arm cannot know \"replaces the newer\" is true — it is printed for a lower+not-live row"*); lower arms inverted → **killed**, 10 fails; both wording reverts → **killed**, 5 fails each.

**README updated in the same commit** (the symptom-index drift guard is genuinely wired — I re-ran its corpus check plus its negative control). The cause cell now describes four cases; all three first-column fragments still resolve to live source strings.

---

### Mutation battery — 12/12 killed, **0 survivors**

Script + full output in the session scratchpad. Baseline and post-restore both `RUN=98 FAIL=0`.

### Gates

- `make ci` — **exit 0**, `ok github.com/civitai/cli/internal/cmd 24.1s`
- `go test ./... -count=1 -v` — **RUN=3996 / PASS=3992 / FAIL=0 / SKIP=4 / timeouts=0**. The 4 skips are the pre-existing env-gated ones (`TestScaffoldPinsSatisfyPublished` = `pins-vs-published`, plus `TestScanDirFromEnv`, `TestSubmissionsCapDriftAgainstLiveAPI`, `TestScaffoldedReadyAckActuallyFires`) — not chased, per #408.
- `golangci-lint run ./...` — **full repo** this time, `0 issues.`

### Not done, deliberately

- 🟢 **4** (three open-codings of the slug predicate: `sameSlug` normalises, `apps.go:262` and `appblocks.go:632` still use `!=`) — consolidating touches files #413 owns. Left for an issue.
- `highestApprovedVersion` / `TestHighestApprovedVersionNormalisesTheStatus` collision with #413 — nothing renamed.
- M25, N4–N7 left as proven-equivalent.

### What I could not verify

**No DB access** — the *mechanism* for finding 1 is confirmed at both source sites and reproduced end-to-end against a stub, but I make **no claim about how many production rows are in the `deployState:null` + `liveUrl` state**. That population is unmeasured.

ZacxDev:
## Integration re-check at final tips — three collisions, not one

Re-measured after both PRs' audit-fix rounds, merging `#413 845cc4b` and `#414 b81626f` onto `main` (`e49ddc6`). Both merges are **textually clean** — including `internal/cmd/newest_row_pick_test.go`, which both PRs now edit in non-overlapping hunks. The merged tree still does not build:

```
internal/cmd/app_submit_version_guard.go:144:6: highestApprovedVersion redeclared in this block
        internal/cmd/app_status.go:410:6: other declaration of highestApprovedVersion
internal/cmd/app_submit_version_guard.go:265:10: assignment mismatch: 1 variable but highestApprovedVersion returns 2 values
```

Swept every top-level declaration in the merged package. Three things collide (`detachAndStart` also appears twice but is pre-existing and build-tag separated — `spawn_unix.go` / `spawn_windows.go` — so it is fine):

1. **`highestApprovedVersion`** — declared in both, incompatible signatures (`approvedPeak` vs `(string, bool)`).
2. **`TestHighestApprovedVersionNormalisesTheStatus`** — both fix rounds independently added a test with this exact name, one in each new test file. Two agents working blind converging on the same function name *and* the same test name is a strong signal this is one predicate written twice.
3. **The two predicates still disagree on policy**, so a rename alone leaves a live contradiction:

| | #413 (`app_status.go:414`) | #414 (`app_submit_version_guard.go:133`) |
|---|---|---|
| pre-release | `isParseableVersion` — truncates at `-`/`+`, so `0.6.0-rc.1` ranks **above** `0.5.2` | `comparableVersion` — declares it **not orderable**, skips it |
| slug | `s.BlockID != blockID`, exact | `sameSlug()`, `EqualFold`+`TrimSpace` |
| liveness | n/a | `rowIsServing()` — `liveUrl` leads, `deployState=="live"` fallback |

With an approved `0.6.0-rc.1` alongside an approved `0.5.2`, `app status` reports the repo BEHIND `0.6.0-rc.1` while `app submit` accepts `0.5.3` — the contradiction #413's own doc comment warns against.

### Suggested land order

`#408` (unblocks `pins-vs-published`, which is red at unmodified `main` and unrelated to either PR) → **#413** → **#414** rebased, consolidating all three collisions in that rebase.

#414's implementation is the better survivor for the merged predicate: stricter on pre-release ordering, normalises the slug, and carries the unorderable-rows list plus the `rowIsServing` liveness that #414's delta audit showed is required to avoid telling an author a serving app is not live. #413's `versionDriftWarning` then consumes `peak.version` / `peak.found`, its `("0.5.2", true)` assertions adapt to the struct, and one of the two identically-named tests is renamed or merged.

Separately worth its own issue (deliberately not fixed in either PR, since it spans files both own): the slug predicate is now open-coded **three** times — `sameSlug` normalises, while `apps.go:262` and `appblocks.go:632` still compare raw.

ZacxDev:
## Integrated with `main` — and the two halves of #412 no longer contradict each other

`main` moved under this branch (#413 squashed as `3a53c2f`, #408 as `ca9880a`). **The merged tree did not compile**, and `git merge` reported no conflict — the collision is semantic, so a clean `git` result was not evidence about anything:

```
internal/cmd/app_submit_version_guard.go:144:6: highestApprovedVersion redeclared in this block
        internal/cmd/app_status.go:410:6: other declaration of highestApprovedVersion
internal/cmd/app_submit_version_guard.go:265:10: assignment mismatch: 1 variable but highestApprovedVersion returns 2 values
```

### Integration strategy: MERGE, not rebase

`git merge origin/main` into the branch (`41d3d77`), pushed as a **fast-forward**. No history rewrite, no force-push, so the review thread above stays anchored to commits that still exist. The repo squash-merges anyway (both #413 and #408 landed that way), so the merge commit costs nothing on `main`.

### The three collisions

**1. `highestApprovedVersion` — incompatible signatures.** Consolidated into a new file, `internal/cmd/approved_version.go`, holding the one copy plus everything it is built from (`approvedPeak`, `sameSlug`, `isApprovedStatus`, `rowIsServing`, `comparableVersion`). #414's shape survives — stricter pre-release ordering, normalised slug, `rowIsServing` liveness, the skipped-rows list; #413's `(string, bool)` pick is deleted and `versionDriftWarning`'s path consumes `peak.version` / `peak.found`.

A *new* file rather than either command's: a shared predicate parked in `app_submit_version_guard.go` would leave `app_status.go` depending on a file named after the other command — which is roughly how the second copy came to be written.

**2. Duplicate test name — MERGED, not renamed, with no assertion dropped.** With one predicate the two `TestHighestApprovedVersionNormalisesTheStatus` cases call the same function with the same return, so they were the same test with different fixtures. The survivor is #414's (it also drives the refusal end-to-end); #413's three fixtures it did not already cover — `"\tapproved\n"` positive, `"not approved"` and `"preapproved"` negative, the two that catch a decay into a *substring* match — were folded into it. #413's end-to-end half (`TestAppStatusDriftWarnsOnACapitalisedApprovedStatus`) stays put, and a note at the removed site records where the unit half went.

**3. The policy difference — resolved toward `comparableVersion`, and the consequence is real.**

Adopted #414's rule everywhere: a version carrying pre-release or build metadata is **not orderable**, rather than truncated at the first `-`/`+` and ranked.

The contradiction was not theoretical. With an approved `0.6.0-rc.1` beside an approved `0.5.2`, truncation ranks the rc **first**, so `app status` told the author their repo was BEHIND `0.6.0-rc.1` while `app submit` — which refuses to rank it — accepted `0.5.3` without a word. `app_status.go`'s own doc comment says the two commands *"have to name the same number or they contradict each other."* They did not.

**What a user now sees differently:** where the local version *or* the highest approved one carries a suffix, `app status` is **silent** rather than warning. Concretely, a local `0.4.0-3-gabc123` under an approved `0.5.2` used to warn and no longer does. That is a lost true-ish warning, traded for never stating an order this CLI cannot justify — the direction the whole feature is built in, where every fact it cannot establish is a silence. `app submit` stays louder in the same situation because it is about to *act*: it names what it could not order and proceeds.

The `KNOWN LIMITATION` comment on `versionDriftWarning` described the **old** policy, so it is replaced rather than left standing — as is the README's *"(Pre-release suffixes are ignored in the comparison — a local `0.5.2-rc1` reads as equal…)"*.

### Redundant guards — checked, not assumed

`isParseableVersion` / `compareVersions` are gone from `app_status.go`'s code paths but still used by the self-update family (`upgrade.go`, `update_notice.go`, `update_check.go`, `update_cache.go`); nothing is orphaned.

Two guards are now **output**-redundant, kept, and now say exactly that instead of claiming more: `versionDriftWarning`'s own orderability check (the caller gates first) and the caller's `!peak.found` return (`versionDriftWarning` rejects the resulting `""`). They are each other's mirror — deleting either is an equivalent mutant — but the first is still killable at the direct-call seam (M6b below), and only it is inherited by a future caller.

### The test neither PR had

`internal/cmd/version_412_contract_test.go`: **`app status` and `app submit` name the same version for the same rows**, over six fixtures including the `0.6.0-rc.1`-beside-`0.5.2` one and the silent-vs-silent direction. It lives in neither command's test file because it belongs to the seam — each half was verified in isolation, mutation-swept and audit-clean, and no fixture anywhere built the combined state. Both sides read the **same served JSON** (round-tripped through `appapi.Submission`) rather than two hand-built fixtures.

Watched **red** before being believed. Against the exact pre-consolidation code (#413's truncating pick restored verbatim):

```
THE TWO COMMANDS DISAGREE about the same rows: `app status` names "0.6.0-rc.1",
`app submit` refuses against "0.5.2".
```

Against a mutant reverting only the *renderer* it stayed **green** — which is what located the defect in the **pick** rather than the comparison. Its own positive control has two arms: both extractors must observe a version, and the comparison must be able to come out **unequal** (at EQUAL the two commands differ by design). Worth noting: the battery's first run went red on the harness's *own* regex, which captured `"0"` out of `"0.5.2."` — recorded in the code, because an extractor returning the wrong string is loud while one returning `""` is silent and green.

### Ledger correction

`newest_row_pick_test.go` said `app_status.go` makes **two** reads. It has made **three** since #413 — the drift check is a whole-list scan for the highest approved version, precisely the non-newest shape that ledger exists to track. Corrected, along with the neighbouring *"THE ONE ENTRY IN THIS LEDGER THAT IS NOT A NEWEST-ROW PICK"* claim (now one of two, and literally the same pick). The structural derivation only checks that a **file** reaching the route is ledgered, so a file already present can grow a new pick unnoticed — that limit is now recorded there rather than implied.

### Gates

All re-run against the **pushed** head `41d3d77`, with the working tree confirmed clean and identical to it first.

| Gate | Result |
| --- | --- |
| `go build ./...` / `go vet ./...` | rc=0, no output |
| `go test ./... -count=1 -v` | **4077 `=== RUN` / 4073 `--- PASS` / 0 `--- FAIL` / 4 `--- SKIP`**, 0 `panic: test timed out`, 0 panics of any kind |
| `make ci` | rc=0 — 20 pkg `ok` / 0 `FAIL` / 0 timeouts |
| `make ci-shallow` | rc=0 — `ok=20/20 failing-tests=0 failing-packages=0 timeouts=0 go-test-rc=0` |
| `golangci-lint run ./...` | **`0 issues.`** |
| PR checks on `41d3d77` | **12/12 present, all terminal, all SUCCESS** |

Counted, not read off an exit code. The 4 skips are pre-existing opt-in guards (`TestScanDirFromEnv`, `TestSubmissionsCapDriftAgainstLiveAPI`, `TestScaffoldPinsSatisfyPublished`, `TestScaffoldedReadyAckActuallyFires`), and 4073 + 4 = 4077 reconciles.

**`pins-vs-published` is green** — CI `SUCCESS`, and separately run locally with its network guard enabled (`CIVITAI_CHECK_PUBLISHED_PINS=1` → `--- PASS`), because the default suite *skips* it and a green local run would otherwise have said nothing about it.

Lint instrument re-validated rather than trusted: a planted unused var reported `1 issues`; tree restored byte-identical (sha256).

The CI rollup was checked by requiring every check to hold a **terminal** conclusion (`SUCCESS|FAILURE|CANCELLED|TIMED_OUT|NEUTRAL|SKIPPED|ACTION_REQUIRED|STALE`) and asserting a minimum count of 12 — an empty conclusion is "not started", not "settled", and a rollup that has registered only some of its checks settles instantly and correctly-looking.

### Mutation — BOTH batteries re-run against the consolidated tree

25 mutants (#414's M01–M14, #413's M1–M8, plus 3 new on the consolidated seams). Control `RUN=2265 PASS=2265 FAIL=0`. Tree restored from a sha256-checked copy after every mutant (`git stash` is repo-global here and was not used).

**23 killed. Nothing that previously died now survives.**

Six mutants now kill tests on **both** sides, because they land on shared code — that is the consolidation working, not a coincidence:

| Mutant | Killing tests |
| --- | --- |
| M03/M2 approved-status filter dropped | **22** |
| M04/M5 newest approved instead of highest | **12** |
| M05 slug filter dropped | **14** |
| M06/M3 semver → string compare | 3 |
| M10 pre-release treated as orderable | 6 |
| Ma status compared raw | 3 |

Spot-check of the rest, each watched red with its own message:

- **M10** → `TestAppStatusDriftDeclinesToOrderAPreReleaseVersion` — *"ranked 0.6.0-rc.1 ABOVE 0.5.2 and quoted it"*
- **M02** allow LOWER → 15 tests, incl. `TestStatusAndSubmitNameTheSameVersion`
- **M6b** `versionDriftWarning`'s own guard → `TestVersionDriftWarningRefusesUnparseableVersionsItself` — *"stated a comparison it cannot make"*
- **M09** guard not wired → `TestAppSubmitRefusesARegressionAndNeverUploads`
- **M11** sentinel dropped → 4 tests
- **M14** live claimed unconditionally → 8 tests

**2 survivors, both equivalent, both demonstrated rather than asserted:**

- **M8** — dropping the `err != nil` half of the manifest-load gate. Already documented as equivalent by #413 (as M8a); **re-verified here rather than taken on trust**: `manifest.Load` returns `nil, err` on every error path (`manifest.go:60-68`), so the surviving `m == nil` check absorbs it.
- **Mfound** — dropping the `!peak.found` return, new to this change. `peak.version` is `""` when `found` is false and `versionDriftWarning` rejects `""`, so there is no output difference. That is what the comment on that line now says.

One process note worth recording, because it is the failure mode this whole PR is about. On the re-run, `Mfound` came back **`NOT APPLIED — pattern occurs 0x`**: the comment I had added above that line broke the mutant's exact-string anchor. The harness refused to score it instead of silently counting it among the kills. Re-anchored and re-run on its own, it survives as described. A mutation table is only a claim about the mutants that were actually *applied*, so "not applied" must never be allowed to read as "killed".

### Not verified

`make mutate` (gremlins) was not run — same reasoning as the original PR. And these are local gates plus this PR's checks; nothing has been exercised against the live API.

Issue #409, opened 2026-08-14
app submit fails from a git worktree/submodule: packager includes `.git` (excludedDirs only checked for directories)


ZacxDev:
## Summary

`civitai app submit` fails from a **git worktree** (and from a **submodule**) because the packager includes `.git`. The server rejects it, and the error names Forgejo rather than anything the author can act on:

```
Error: server returned 400: could not push bundle to review repo: Forgejo 422 Unprocessable Entity:
{"message":"path contains a malformed path component [path: .git]","url":"https://forgejo.civitai.com/api/swagger"}
```

Nothing in that message mentions the CLI, the bundle, or the shape of the author's checkout, so there is no route from the error to the cause.

## Root cause

In `internal/pkgzip/pkgzip.go`, the exclusion is keyed on the filesystem **type** rather than the **name**:

- `excludedDirs` correctly lists `.git` (`:56`) and `node_modules` (`:60`).
- That map is consulted **only** at `:208`, inside `if d.IsDir()`.
- The regular-file path falls through to `isExcludedFile` (`:155`), which covers only `*.zip` and `.env*` — it never consults `excludedDirs`.

In a normal clone `.git` is a directory, so the `IsDir()` branch catches it. **In a worktree or a submodule, `.git` is a regular file** (`gitdir: …`), so it passes both gates and is packaged.

## Reproduction

```
git clone <app-repo> app && cd app
git worktree add --detach ../app-wt HEAD
cd ../app-wt && civitai app submit --yes     # 400 / Forgejo 422

cd ../app && civitai app submit --yes        # succeeds
```

The file count is the tell — same commit, 61 tracked files:

| checkout | CLI reports |
|---|---|
| worktree | `Packaged 62 file(s)` |
| clone | `Packaged 61 file(s)` |

The extra entry is `.git`, which Forgejo then names explicitly in its rejection.

## Suggested fix

Apply the name-based exclusion to regular files too, not just directories — e.g. have `isExcludedFile` also reject any base name in `excludedDirs`, or check `excludedDirs[d.Name()]` before the `d.IsDir()` branch. That covers the submodule case in the same change.

Two things worth folding in:

1. **A regular-file `.git` contains an absolute path** (`gitdir: /home/<user>/…`). It is rejected server-side today, so nothing leaks now — but the packager should not be assembling it into an upload regardless.
[...]

PR #415, opened 2026-08-15
feat(app submit): refuse a submit from a dirty git work tree (#411 piece 1)

ZacxDev:
Refs #411 — the **refusal half** only. The provenance stamp stays open; see [Scope](#scope-what-this-does-not-close) at the bottom.

`civitai app submit` packages whatever is on disk and records nothing about where it came from, so a bundle built from a tree that was never committed submits exactly as cleanly as one built from a tagged release. Five first-party apps are live at a version their source repo has never held, and `custom-generators@0.5.2` exists in no commit anywhere.

This adds the preventer: when the packaged directory is inside a git work tree, a submit that would really upload **refuses** while files that go into the bundle are uncommitted, names them, and points at `--allow-dirty`.

It **degrades rather than enforcing**, which is the constraint the issue puts first:

| state | behaviour |
|---|---|
| not a git repo, or no `git` on `PATH` | proceed, **silently** — the scaffold path |
| repo, clean | proceed |
| repo, **dirty** | **REFUSE**, naming the paths, `--allow-dirty` in the text |
| repo, clean, `HEAD` on no remote | **WARN**, proceed |

---

### 1. Which directory's repo — the tree being PACKAGED, never the cwd

`app submit [dir]` takes an optional positional, so a release script running `civitai app submit ./packages/my-block` from a monorepo root has a manifest directory and a cwd that are different places. The dirtiness question is about the bytes that go into the zip, so every git call goes through `git -C <dir>`.

The status pathspec is scoped to that subtree for the same reason: a **sibling** package's uncommitted work is not in this bundle and must not block this submit.

Pinned by `TestDirtyGuardInspectsThePackagedDirNotTheCwd`, which asserts **both arms** — dirty cwd + clean target must proceed, clean cwd + dirty target must refuse — so a guard that simply always says "clean" cannot pass. And by `TestDirtyGuardScopesToThePackagedSubtree`, which also pins the path **spelling**: reported paths are relative to the packaged directory, i.e. their path inside the zip.

### 2. What counts as dirty

> A path is dirty when git reports it as differing from `HEAD` — staged, unstaged, or untracked — **AND** the packager would put it in the bundle.

One rule with one consequence: the guard fires exactly when the bundle carries bytes that are in no commit. The two filters that rule implies are not special cases:

- Files git **ignores** never appear, because `--porcelain` without `--ignored` does not list them.
- Files the **packager excludes** are dropped, via a new `pkgzip.IsExcludedPath`. An untracked `dist/`, `node_modules/`, a `.env.local`, or the `*.zip` a previous `--package-only` run left behind are not differences between the bundle and `HEAD` — the bundle does not contain them. Firing on those is how a guard trains everyone to pass `--allow-dirty` by reflex.

Deliberately **in** scope, because the bundle really does carry them: a modified **submodule** (`pkgzip.Build` walks the filesystem and excludes only `.git`, so submodule *content* is packaged), and untracked editor residue the packager does not exclude — `notes.md`, `App.tsx.orig`, a stray `.swp`. Those genuinely ship, so refusing over them is correct rather than over-broad. The boundary is tested from **both** sides: `TestDirtyGuardIgnoresWhatTheBundleDoesNotCarry` and its negative control `TestDirtyGuardStillSeesWhatTheBundleDOESCarry`, whose cases (`.env.example`, `docs/dist.md`, `src/App.tsx.orig`) are the ones a slightly-too-broad exclusion rule would swallow.

`pkgzip.IsExcludedPath` **reproduces** `Build`'s decisions rather than sharing them (one walks a filesystem, the other takes a string), so `TestIsExcludedPathAgreesWithBuild` walks one tree through both and asserts the file sets are identical — with a positive control that the fixture actually exercises the exclusions, or the agreement would be vacuous.

Also: `--untracked-files=all`, not the default `normal`. `normal` **collapses** an untracked directory to a single `src/` entry, which breaks the guard in both directions — the refusal names `src/` instead of the file to commit, and the exclusion filter is handed a directory name when the decision belongs to the files inside it. (This was found by the tests, not by reading: the first implementation used `normal` and `TestDirtyGuardStillSeesWhatTheBundleDOESCarry` went red.)

### 3. Exit code — **1**, via an `errors.Is` sentinel

`ErrDirtyWorkTree`, attached with `civitai.Tag` and left untagged for the exit mapper's `default`, exactly as `ErrVersionRegression` is. Same verdict as #412's refusal for the same reason: exit 2 is documented as a mistake about the **invocation**, and every flag, argument and path is well-formed when this fires — it is the **project** that is wrong, here its working tree against its own history.

`exitcodes_doc.go` gains a bullet on code 1, and both README blocks it generates were regenerated (the byte-assertions `TestREADMEExitCodeTableIsGenerated` / `TestREADMEExitCodeSectionsAreGenerated` are green). `TestDirtyWorkTreeExitsGeneric` pins the process contract with a negative control that `exitCode` can return something else; `TestDirtyWorkTreeAndVersionRegressionAreDistinctSentinels` pins that the two submit refusals stay separately identifiable — `--allow-dirty` and `--allow-downgrade` are not interchangeable.

Two README **Troubleshooting** rows were added, both quoting literals the CLI really prints (checked by the existing drift guard): one for the refusal, and a **separate** one for `HEAD is on no remote`, so a reader searching for the warning is not sent to a refusal.

### 4. Guard ordering — **before** `confirmSubmit`, unlike the version guard

The version guard sits *after* the prompt because moving it earlier would give a bare non-TTY `civitai app submit` an API read it does not have today, and `TestAppSubmit_NonTTYRefusesWithoutYes_NoNetworkCall` pins that invocation as touching nothing.

This guard reads only the local filesystem, so running it first **acquires nothing** — that test's invocation still touches no network — and it stops the CLI asking "Submit for review? [y/N]" about a submit it has already decided to refuse. Pinned by `TestDirtyGuardRefusesBeforePrompting`, which runs with an interactive TTY and asserts neither prompt line is printed.

Like the version guard it runs only on the upload path: `--package-only` and the no-token fallback never reach the server.

### 5. How git is invoked

A new `gitOutputFunc` seam over `git -C <dir>`, separate from `app_pull.go`'s `gitRunner` — that one streams progress to the terminal and returns only an error, which is right for `clone`/`fetch` and useless for a question whose whole answer is on stdout.

`git` absent from `PATH` returns `errGitUnavailable`, which the guard treats **identically to "not a repo"**: silent, proceed. Pinned by `TestDirtyGuardTreatsAMissingGitLikeNoRepo`. A `git status` that fails on a tree that *is* a repo warns and proceeds — an accident preventer must not take every submit down with one broken checkout — and `TestDirtyGuardWarnsAndProceedsWhenStatusFails` asserts it is not a *silent* fail-open.

Three git calls, all read-only: `rev-parse --is-inside-work-tree --show-prefix`, `status --porcelain -z`, and (only when clean) `for-each-ref --contains HEAD refs/remotes`. `-z` because the default format C-quotes any path with a space; `-c status.relativePaths=false` because a user can turn relative paths on in their config and the prefix arithmetic must be arithmetic, not a guess about someone's `~/.gitconfig`.

---

## Red at base, green at HEAD

The shipped test file **cannot compile** at `origin/main` (`undefined: checkWorkTreeClean`, `ErrDirtyWorkTree`, `gitOutput`, `errGitUnavailable`), which is a true statement but weak evidence. So the matrix was measured with a **base-compatible probe** asserting only observable behaviour of `civitai app submit` — no new symbol, no new flag — so the identical file runs at both refs.

| probe | `origin/main` (`773c72e`) | HEAD |
|---|---|---|
| `DirtyTreeMustNotUpload` — an uncommitted `App.tsx` must not reach the submit route | **FAIL** | **PASS** |
| `CleanUnpushedWarnsAndStillUploads` — warns *and* proceeds | **FAIL** | **PASS** |
| `NoRepoStillUploads` — harness control / the scaffold path | **PASS** | **PASS** |

`base: 1 PASS / 2 FAIL · HEAD: 3 PASS / 0 FAIL`. The third row is the point: it proves the probe harness **can** go green at base, so the two reds measure the feature and not a broken probe. The base failure is verbatim:

```
--- FAIL: TestPROBE_DirtyTreeMustNotUpload
    the submit route was hit from a DIRTY work tree — an uncommitted App.tsx went to review.
    stdout:
    Packaged 3 file(s) (712 bytes compressed, 445 decompressed)
    ✓ Submitted — pubreq_p is pending moderator review.
```

The probe file is not part of this PR.

## Mutation results

16 mutants, enumerated from semantic failure modes rather than deletions. Baseline 2,323 PASS / 0 FAIL. Every kill below was checked for **which** test died and **its own** message, not merely that something went red.

| mutant | killed by | its message |
|---|---|---|
| M1 refusal condition inverted | `TestDirtyGuardRefusesAnUncommittedBundleFile` | `an uncommitted src/App.tsx goes into the bundle and must be refused` |
| M2 `--allow-dirty` ignored | `TestDirtyGuardAllowDirtySubmitsAnyway` | `--allow-dirty must permit a dirty tree, got: refusing to submit…` |
| M3 no-repo treated as dirty | `TestAppSubmitUploadsWhenTokenAndPathConfigured` | `submit upload: refusing to submit demo-block@0.1.0 from a dirty git work tree` |
| M4 packager filter inverted | `TestDirtyGuardRefusesAnUncommittedBundleFile` | `…must be refused` |
| M5 wrong directory inspected | `TestDirtyGuardRefusesAnUncommittedBundleFile` | `refusal missing "src/App.tsx"` |
| M6 warn path turned into a refusal | `TestDirtyGuardInspectsThePackagedDirNotTheCwd` | (7 tests) |
| M7 `--untracked-files=normal` | `TestDirtyGuardRefusesAnUncommittedBundleFile` | `refusal missing "src/App.tsx"` |
| M9 prefix stripping dropped | `TestDirtyGuardScopesToThePackagedSubtree` | `the path is repo-root-relative — the --show-prefix arithmetic did not run` |
| M10 rename's second `-z` field not consumed | `TestDirtyGuardRefusesEveryKindOfDifference/renamed_file` | `a rename is ONE change; its -z record's second field is the original path` |
| M11 `IsExcludedPath` always false | `TestDirtyGuardIgnoresWhatTheBundleDoesNotCarry` | `none of these reaches the bundle…` |
| M12 `IsExcludedPath` ignores directory components | `TestDirtyGuardIgnoresWhatTheBundleDoesNotCarry` | same |
| M13 refusal message drops the paths | `TestDirtyGuardRefusesAnUncommittedBundleFile` | `refusal missing "src/App.tsx"` |
| M14 wrong sentinel attached | `TestDirtyGuardRefusesAnUncommittedBundleFile` | `refusal must carry ErrDirtyWorkTree` |
| M15 refusal wording changed | `TestDirtyGuardRefusesAnUncommittedBundleFile` | `refusal missing "from a dirty git work tree"` — also the README symptom-index guard |
| M16 `gitOutput` drops `-C` (runs in the process cwd) | `TestAppSubmitUploadsWhenTokenAndPathConfigured` | `refusing to submit demo-block@0.1.0 … 4 path(s)` |

**Two mutants were findings, not passes:**

- **M10 originally SURVIVED, and it was a real gap.** The rename assertion was *spelled*, not structural: it grepped for the original filename `index.html`. A parser that fails to consume the second `-z` field does not leak a recognisable path — it reads the first two bytes of `index.html` as status letters and emits `in ex.html`, which that grep never matches. The assertion is now a **count** (`1 path(s)`), plus a new `TestBundleDirtyPathsParsesTheZRecordLayout` that drives the parser with literal records so the entry list itself is the assertion. M10 now dies.
- **M14 originally did not compile** (removing `civitai.Tag` orphaned an import), so it proved nothing. Re-expressed as attaching the *wrong* sentinel — which compiles, and is the truer mutant. It dies.

Two further mutants, added with the follow-up commit that covers the no-blockId branch:

| mutant | killed by | its message |
|---|---|---|
| M17 empty-slug arm dropped | `TestDirtyRefusalWithNoBlockIdStillReadsAsASentence` | `an empty blockId was spliced in verbatim` |
| M18 version suffix never attached | `TestDirtyGuardRefusesAnUncommittedBundleFile` | `refusal missing "refusing to submit custom-generators@0.6.1"` |

**One surviving mutant, and it is EQUIVALENT — proven, not asserted:**

- **M8, dropping the `-- .` pathspec from the status call.** The pathspec and the `strings.HasPrefix(rel, prefix)` filter select the **same set**: git's `.` pathspec means "paths under the cwd", and `prefix` is that same cwd spelled repo-root-relative with a trailing slash, so every repo-root-relative path under `dir` starts with it and no path outside `dir` does. Measured rather than reasoned: with the pathspec removed, **all 2,323 tests still pass**, including the monorepo fixtures. The pathspec is kept for **cost** (git stops walking the rest of a monorepo), the prefix filter for **correctness of the reported path**. `TestBundleDirtyPathsDropsPathsOutsideThePackagedSubtree` pins the correctness half, and its comment records the redundancy — otherwise a future reader deletes it as "already covered by the pathspec", at which point the redundancy is gone and nothing notices.

## Gates

| gate | result |
|---|---|
| `go test ./...` (verbose, counted) | `=== RUN 4114`, `--- PASS 4110`, `--- FAIL 0`, `--- SKIP 4`, no `panic: test timed out`. The 4 skips are the pre-existing network/env-gated guards (`TestScaffoldPinsSatisfyPublished`, `TestSubmissionsCapDriftAgainstLiveAPI`, `TestScanDirFromEnv`, `TestScaffoldedReadyAckActuallyFires`). |
| `make ci` | exit 0 (`tidy vet test build`); `go.mod`/`go.sum` unmoved by `tidy`. |
| `make ci-shallow` | exit 0 — `ok=20/20 packages, failing-tests=0, failing-packages=0, timeouts=0`. Run because the fixtures create git repos; they are all in temp dirs, so depth-1 is irrelevant to them, but it was run rather than assumed. |
| `golangci-lint run ./...` | **0 issues**, exit 0 — and the **instrument was validated**: a planted unused function made it report `1 issues: unused` / exit 1, so the zero is a measurement. (The first attempt was invalid — run from the wrong cwd it reported `0 issues` alongside `pattern ./...: directory prefix . does not contain main module`, i.e. it linted nothing. Re-run inside the worktree.) |

### Live probe of the real binary

The exit-code test pins `exitCode()`; this pins the **binary**. `CIVITAI_BASE_URL` points at an address that cannot be dialled, so a refusal arriving at all also proves the guard fires before any network use.

| arm | result |
|---|---|
| dirty (uncommitted `src/App.tsx` + `dist/`, `node_modules/`, `.env.local`, a stray `.zip`) | **exit 1**, refusal names **only** `?? src/App.tsx` |
| same tree + `--allow-dirty` | gets past the guard, packages, then `exit 5` on the unreachable server |
| clean, unpushed | `⚠ the work tree is clean, but HEAD is on no remote…` **1** matching line, submit continues |
| clean, `HEAD` published | **0** matching lines — the negative control on that warning |
| **no repo at all** (the scaffold path) | packages and reaches the network, unchanged |

### CI

All 12 checks on the first commit were terminal and green — `build-test`, `lint`, `pins-vs-published`, `ready-ack-runtime`, `scaffold-currency`, `schema-drift`, `template-page-vite`, `template-page-money`, and the four CodeQL runs. `pins-vs-published` in particular is green rather than assumed-pre-existing.

## Scope: what this does NOT close

**#411 stays open.** Its other half — stamping the commit SHA + dirty flag onto the publish request — needs a **server-side field**: `appapi.Submission` (`internal/appapi/appblocks.go`) carries `ID, BlockID, AppBlockID, Version, Status, RejectionReason, ApprovalNotes, DeployState, DeployDetail, DeployUpdatedAt, SubmittedAt, ReviewedAt, UpdatedAt, CreatedAt, LiveURL` and no commit, tree or dirty column. The CLI cannot stamp what the API will not accept, so that piece is blocked on an API change and is not in this PR.

Practically: this makes the *future* occurrences preventable, and leaves the existing five apps' drift as undiagnosable as it is today. Only the stamp fixes that.

**Also not claimed:** the guard's cost on a repository with a very large untracked-but-not-gitignored tree was not measured — `--untracked-files=all` enumerates it before the filter discards it. The verdict is unaffected either way; only latency is, and no fixture here is that shape.


ZacxDev:
## Audit fixes — all 9 findings addressed (`5777203`)

Every finding was reproduced first (git 2.55.0 + the built binary), then fixed, then re-measured. **The M8 "equivalent — proven" claim was FALSE and is retracted in both places it was published.**

### 🟡 F1 — `-- .` is not equivalent to the prefix filter

Reproduced exactly. `git mv packages/my-block/src/App.tsx packages/other/App.tsx`, packaged dir `packages/my-block`:

```
WITH  `-- .` :  [D  packages/my-block/src/App.tsx]
WITHOUT      :  [R  packages/other/App.tsx][packages/my-block/src/App.tsx]
```

Fixed: both comments retracted **in place** (`app_submit_dirty_guard.go` at the status call, and `TestBundleDirtyPathsDropsPathsOutsideThePackagedSubtree`) — neither now says "equivalent"; both state what was measured and why the green suite meant "no fixture crossed the boundary". New fixture `TestDirtyGuardRefusesARenameOutOfThePackagedSubtree` (real repo, with a control asserting `src/App.tsx` really is in `HEAD` and really is gone from the packaged dir), failing message: *"FAIL-OPEN: the bundle lost src/App.tsx (which is in HEAD) and the guard reports CLEAN"*.

🔴 **One correction to the brief, measured rather than assumed.** The instruction was "confirm the fixture fails with the pathspec removed and passes with it". After **F3**'s fix that is no longer true, because F3 independently closes the same hole — and hiding that would be exactly the kind of claim this finding is about. Three arms over the one fixture:

| mutant | fixture |
| --- | --- |
| `-- .` removed only | **PASSES** |
| rename-original handling removed only | **PASSES** |
| **both removed (= the pre-fix code)** | **FAILS** |

So the fixture pins the *conjunction* — i.e. it reproduces the fail-open the audit measured — and neither singleton, because the two mechanisms are now redundant. That matrix is written into the fixture's own comment so it cannot be re-derived in either direction ("the pathspec is dead" would be as wrong as "equivalent — proven"). The pathspec's non-redundant job (bounding the `--untracked-files=all` walk) is pinned structurally and labelled as structural: `TestDirtyGuardScopesTheStatusCallToThePackagedSubtree` (dies on the pathspec mutant).

### 🟡 F2 — a symlink was refused but never shipped

Reproduced against `bin/civitai` (exit 1, `?? alias.js`, zip = `['app.js','block.manifest.json','index.html']`).

New `pkgzip.IsExcludedEntry(rel, fs.FileMode)` answers the file-type half `IsExcludedPath` structurally cannot. The guard applies it **only to paths that are in no commit** (`??` / `A…`) — a tracked regular file replaced by a symlink makes the bundle *lose* committed content and must still refuse, which is pinned separately.

🔴 **The seam guard's blindness is fixed at the root, not patched around it.** `excluded_path_test.go` filtered `!d.Type().IsRegular()` out of its *own oracle* — it copied `Build`'s filter into the thing meant to check `Build`. It now walks every non-directory entry, judges with `IsExcludedEntry`, and asserts a positive control that a non-regular entry really reached the oracle.

Live binary, after: symlink-only tree → **exit 5, no refusal**; the same tree with a *regular* `alias.js` → exit 1 with the refusal (negative control on file type, not on the name).

### 🟡 F3 — a rename whose new path is packager-excluded

`git mv src/App.tsx dist/App.tsx` → `R  dist/App.tsx\0src/App.tsx`, previously CLEAN. An `R` now contributes **both** paths.

A `C` deliberately does **not** — and that is measured, not a stylistic choice. A copy leaves its original exactly as `HEAD` has it, and git's status copy detection only pairs against a source that is itself modified, for which git emits its own record: `[C  src/copy.html][src/index.html][M  src/index.html]`. Adding a copy's source would either double-count a path git already reported (an inflated "N path(s)") or refuse over a file identical to `HEAD`. Both directions are pinned (`TestBundleDirtyPathsDoesNotCountACopySOURCE`, plus de-duplication).

Live binary, after: `Error: refusing to submit probe-app@0.6.1 … 1 path(s) … R  src/App.tsx` (and `dist/App.tsx` is *not* named — negative control on the packager filter).

### 🟢 F4 — `GIT_DIR`/`GIT_WORK_TREE` override `git -C`

Reproduced (dirty tree → exit 1 refusal; same invocation with `GIT_DIR` at a clean repo → exit 5, 0 refusals).

🔴 **The suggested one-liner would have been catastrophic and is not what shipped.** `c.Env = append(os.Environ(), "GIT_DIR=", …)` sets the variable to the empty string, and git does **not** read that as unset — measured on git 2.55.0, `GIT_DIR= git status` fails with `fatal: not a git repository` naming the empty string, which this guard classifies as "no repo here" and proceeds. That one line would have failed the guard **open on every invocation** while reading as a hardening change. The variables are **removed** from the child environment instead (`gitScrubbedEnv`), and both halves are pinned: a unit test asserting removal-not-emptying, and a real-repo test whose **control arm runs an unscrubbed runner first** to prove the environment really does relocate git here (without that arm the assertion attributes nothing). The empty-string variant is itself in the mutation battery — it kills 29 tests.

User git *config* is deliberately **not** scrubbed: `status.renames` / `core.excludesFile` are the user's own answer about their tree — and F5 depends on that staying true.

### 🟢 F5 — the `C` branch is now pinned

Reproduced a real `C` record (`-c status.renames=copies`, source modified). Added to `TestBundleDirtyPathsParsesTheZRecordLayout` plus the dedicated copy-source test. `ContainsAny(code,"RC")` → `"R"` now kills 2 tests.

### 🟢 F6 — whitespace-leading directory name

Reproduced: ` spacey/app` → prefix `" spacey/app/"` → `TrimSpace` → `"spacey/app/"` → every path dropped. Now trims the line ending only (`TrimSuffix(…, "\r")`). `TestDirtyGuardHandlesAPackageDirWhoseNameBeginsWithASpace` has a positive control that git really answers with the leading space. Live binary, after: exit 1, names `src.tsx`.

### 🟢 F7 — zero-commit repository

Documented in the Troubleshooting row (it is between "no repo" and "repo, dirty", and the scaffold path is unaffected — `app create`/`app init` run no `git init`), **and pinned** by `TestDirtyGuardRefusesAFreshRepoWithNoCommits` so the documented behaviour cannot drift silently. The same row also gained the symlink and `git mv` clauses that F2/F3 changed.

### 🟢 F8 — the exit-code test builds its own fixture

`TestDirtyWorkTreeExitsGeneric` now takes the **real** error: a real git fixture + a real `cmd.NewRootCmd()` `app submit --yes`, with an httptest handler that fails the test if anything is requested (the reachability control). It no longer constructs `civitai.Tag(cmd.ErrDirtyWorkTree, …)` itself, so a wrong sentinel now dies here too, not only in `internal/cmd`.

### 🟢 F9 — "three git calls, all read-only"

`--no-optional-locks` added, so the claim in the PR description is now true rather than nearly true. It is a **behavioural no-op** and survives the battery by design — flagged rather than papered over.

### Do-not-change items

- **MU22** (`len(rec) < 4` → `< 3`) and **MU24** (dropping `-c status.relativePaths=false`) re-run against the *new* code: still equivalent, still kept.
- The MU24 comment is corrected. Measured on git 2.55.0, `--porcelain` **ignores `status.relativePaths` entirely** — unset / `true` / `false` give byte-identical output — so the old "a user CAN turn it on in their config" is wrong. The comment now says the flag buys nothing today, is kept as future-proofing, and must not be restated as a live hazard *or* deleted as dead weight.
- #409 (linked-worktree `.git`) untouched. **#411 stays open** — this is still only the refusal half; the provenance stamp needs the server-side field. `Refs #411`.

### Gates

| gate | result |
| --- | --- |
| `make ci` | **pass** (tidy + vet + test + build) |
| `go test ./... -count=1 -v` | `RUN=4132 PASS=4128 FAIL=0 SKIP=4 TIMEOUT=0` (4 skips identical to baseline; baseline was `RUN=4117 PASS=4113`) |
| `make ci-shallow` | **pass** — `ok=20/20 failing-tests=0 failing-packages=0 timeouts=0` |
| `golangci-lint run ./...` | **0 issues**, run from the module root, instrument validated first (planted an unused variable → reported 1 issue → reverted) |

### Mutation battery, re-run after the fixes

17 mutants over `internal/cmd` + `internal/pkgzip` + `cmd/civitai` (baseline `RUN=2339 PASS=2339 FAIL=0`). Every one landed where expected; each verdict is the counted `--- FAIL` lines, not an exit code.

| mutant | expected | result — killing test |
| --- | --- | --- |
| drop `-- .` pathspec | die | died — `TestDirtyGuardScopesTheStatusCallToThePackagedSubtree` |
| `len(rec) < 4` → `< 3` (MU22) | survive | **survived** (equivalent, re-proven) |
| drop `-c status.relativePaths=false` (MU24) | survive | **survived** (equivalent, re-proven) |
| `ContainsAny(code,"RC")` → `"R"` | die | died ×2 — `…DoesNotCountACopySOURCE`, `…ParsesTheZRecordLayout` |
| rename no longer contributes its original | die | died ×5 — `…CountsARenameOriginalThePackagerWOULDDrop`, `TestDirtyGuardRefusesARenameIntoAPackagerExcludedDir`, `…RefusesEveryKindOfDifference/renamed_file`, … |
| a copy also contributes its original | die | died ×2 — `…DoesNotCountACopySOURCE` |
| `TrimSuffix(prefix,"\r")` → `TrimSpace` | die | died — `…HandlesAPackageDirWhoseNameBeginsWithASpace` |
| gitOutput stops scrubbing the env | die | died — `TestDirtyGuardIsNotRelocatedByGIT_DIR` |
| scrub by setting `GIT_DIR=""` instead of removing | die | died ×**29** |
| `bundlesOnDisk` always yes (file type ignored) | die | died ×2 — both symlink tests |
| type check applied to tracked paths too | die | died — `…RefusesATrackedFileReplacedByASymlink` |
| `absentFromHEAD` drops its `A` half | die | died — `TestDirtyGuardIgnoresAStagedSymlink` |
| drop de-duplication | die | died — `TestBundleDirtyPathsReportsAPathOnce` |
| prefix filter never fires | die | died — `…DropsPathsOutsideThePackagedSubtree` |
| drop `--no-optional-locks` | survive | **survived** (documented behavioural no-op) |
| `--untracked-files=all` → `normal` | die | died ×11 |
| `IsExcludedEntry` stops excluding non-regular | die | died ×4 — incl. `TestIsExcludedPathAgreesWithBuild` (the widened seam guard) |

**14 died, 3 survived** — two proven equivalent, one a documented no-op. Every "expected to survive" mutant is named above rather than left as a silent gap.

### Not verified / deferred

- **Windows and macOS.** Everything is measured on Linux + git 2.55.0. The symlink tests `t.Skip` where `os.Symlink` fails rather than passing vacuously; the file-type rule itself is OS-independent (it reads `Lstat`'s mode, exactly as `Build` reads `DirEntry.Type()`).
- **`--no-optional-locks` was not observed to change anything here.** The auditor measured `.git/index` being rewritten by `git status`; on this box the index mtime did not move in my probe. The flag is correct and free either way, but I am not claiming to have reproduced the write.
- **Older git versions.** The rename/copy record layouts, the `status.relativePaths` no-op and the empty-`GIT_DIR` behaviour are all measured on **2.55.0 only**.

PR #418, opened 2026-08-15
fix(pkgzip): exclude `.git` when it is a FILE, not only a directory (#409)

### LINKED ISSUE #409: app submit fails from a git worktree/submodule: packager includes `.git` (excludedDirs only checked for directories)

ZacxDev:
Closes #409.

## What was wrong

The exclusion was keyed on the entry's **type**, not its **name**. `.git` was on `excludedDirs` from the start, but that map was consulted only inside `Build`'s `if d.IsDir()` branch. In a **linked worktree** and in a **submodule**, `.git` is a regular file holding `gitdir: <absolute path>` — it passed the directory gate (not a directory) and the file gate (`isExcludedFile` knew only `*.zip` and `.env*`), and got packaged.

The author saw a Forgejo 422 that named neither the CLI, the bundle, nor the shape of their checkout, while the same commit submitted from a clone succeeded.

## Reproduced against the built binary, before and after

Real `git worktree add`, real `civitai app submit --package-only`:

| binary | clone | worktree | worktree bundle |
|---|---|---|---|
| `origin/main` @ `e816594` | `Packaged 5 file(s)` | `Packaged 6 file(s)` | `.git` `.gitattributes` `.gitignore` `block.manifest.json` `index.html` `src/App.tsx` |
| this branch | `Packaged 5 file(s)` | `Packaged 5 file(s)` | `.gitattributes` `.gitignore` `block.manifest.json` `index.html` `src/App.tsx` |

The 6-vs-5 tell is the issue's own (62 vs 61 there). The packaged `.git` body was `gitdir: /tmp/i409-repro-tkkaGc/app/.git/worktrees/wt` — the developer's local filesystem layout, assembled into an upload.

## The change

`isExcludedFile` now drops the VCS metadata names — `.git`, `.hg`, `.svn` — by **exact name, at any depth**, so the rule holds whichever shape the entry has. New `vcsMetadataNames` var carries the rationale.

Two decisions worth reviewing:

- **Exact name, not a prefix.** `.gitignore`, `.gitattributes`, `.gitmodules`, `.gitkeep` and a `.github/` directory are project content. A prefix match would delete every one of them from a submission with no error anywhere — the opposite failure mode from the dotenv rule below it, which *is* a prefix on purpose because there being-too-narrow means a leaked credential.
- **Only VCS metadata is promoted, not all of `excludedDirs`** (the issue suggested the latter). That would also drop a regular file named `build`, `dist`, `out` or `coverage` — an extensionless build script is a real thing, and dropping one is silent content loss with no server-side rejection to catch it. A file named exactly `.git`/`.hg`/`.svn` is never author content.

## The shared-predicate seam (#411)

`IsExcludedPath` / `IsExcludedEntry` are what the `app submit` dirty-tree guard asks "would the packager bundle this?", so this change moves two behaviours. Both now answer the same as `Build`, and `TestIsExcludedPathAgreesWithBuild` carries `.git` + `sub/.git` in its fixture (plus `.gitignore` / `.gitattributes` / `src/.gitkeep` as content), so a fix applied to only one side goes red — that mutant is in the table below.

Checked live against both binaries, on a throwaway repo, with the guard actually reachable (a token configured, base URL at a closed port — `--package-only` skips the guard entirely, so an earlier probe through it was wired to nothing):

| arm | `origin/main` | this branch |
|---|---|---|
| uncommitted bundle file (positive control) | REFUSED | REFUSED |
| clean tree (negative control) | passed to network | passed to network |
| untracked `dist/` | passed to network | passed to network |
| untracked `.gitignore` | REFUSED | REFUSED |
| untracked `.gitfile-probe` | REFUSED | REFUSED |

Identical in every arm. `git status` never reports `.git` itself, so in practice the guard is never asked about it — the value here is that the two predicates cannot drift.

## Tests

Real `git worktree add` and real `git submodule add` fixtures, each asserting its own precondition (that `.git` really came out a regular file holding a `gitdir:` pointer) — a fixture that quietly produced a directory would exercise the branch that never broke and pass for the wrong reason. Plus a git-free any-depth fixture and predicate-level assertions. Fixtures are built in temp dirs only.

**Red at base / green at HEAD** (`internal/pkgzip`, counted from `--- PASS` / `--- FAIL`, not an exit code):

| | `=== RUN` | `--- PASS` | `--- FAIL` | `--- SKIP` |
|---|---|---|---|---|
| the 4 new bug tests at `origin/main` `e816594` | 4 | 0 | **4** | 0 |
| whole package at HEAD | 28 | 28 | 0 | 0 |

**Mutants** — all six killed, none by a compile error, each with a distinct killing assertion:

| mutant | killed by | on |
|---|---|---|
| revert the fix | 5 tests | `the bundle carries ".git" from a linked worktree (#409)` |
| prefix match instead of exact name | 5 tests, incl. `TestIsExcludedFileRule` | `IsExcludedFile(".gitignore") = true — an exact-name rule must not swallow it`; bundle loses `.gitignore` `.gitattributes` `.gitmodules` `src/.gitkeep` |
| root only, not any depth | 4 tests (worktree test correctly still passes) | `VCS metadata "a/b/c/.git" was packaged — the exclusion is keyed on the name, at any depth` |
| drop the DIRECTORY half (`.git` out of `excludedDirs`) | 9 tests | `%q is in vcsMetadataNames but not excludedDirs — a DIRECTORY by that name would still be walked into` |
| fix `Build`'s walk inline, not the shared predicate | 3 tests | `Build and IsExcludedPath disagree about this tree` (the seam guard) |
| promote only `.git`, leave `.hg`/`.svn` | 2 tests | `VCS metadata ".hg" was packaged` |

## Neighbouring exclusions — measured, and one is NOT fixed here

Run against the fixed binary, planting each name as the *other* filesystem type:

- **`node_modules`, `dist`, `build`, `out`, `coverage`, `.next`, `.venv`, `venv`, `.turbo` as regular FILES → all shipped.** Same *shape* as #409, deliberately not the same *defect*: none of them is ever VCS plumbing, an extensionless `build` script is plausible author content, and the server does not reject them. Left directory-only, and `TestVCSMetadataNamesAreASubsetOfExcludedDirs` now pins that on purpose.
- 🔴 **`.env.local/`, `.env.d/` and `artifact.zip/` as DIRECTORIES → their contents shipped**, including a planted `.env.d/db.env` containing `API_SECRET=leak`. This is the **mirror** of #409 — a name-keyed rule that fires for only one filesystem type — and for the dotenv half it carries the same credential-leak hazard the rule was written to prevent. **It is not fixed in this PR**, deliberately: it is a separate behaviour change with its own blast radius (a `.environment/` directory would start being dropped whole), and the dotenv rule is ledger-bound to a README section that documents it in terms of files. Worth its own issue rather than a rider on this one.

## Not done

The issue's second suggestion — a local fast-fail in `app validate` / `app submit` saying "`.git` is a file here, you are in a worktree" — is deliberately not implemented. With the packager fixed there is nothing to warn about: submitting from a worktree now produces a bundle byte-identical to the clone's, so a hard stop would refuse a workflow that works.

## Gates

`make ci` (20/20 packages ok, 0 failures), `make ci-shallow` (`ok=20/20 failing-tests=0 failing-packages=0 timeouts=0`), full-repo `golangci-lint run ./...` → `0 issues`. The linter was validated against a negative control first (a planted unused func → `rc=1`, `1 issues`, naming the file) so the clean verdict is a fact about the code rather than about the invocation. No `panic: test timed out` in any run.

### DIFF
--- a/internal/pkgzip/excluded_path_test.go
+++ b/internal/pkgzip/excluded_path_test.go
@@ -156,7 +156,20 @@ func TestIsExcludedPathAgreesWithBuild(t *testing.T) {
 		".env.local",
 		".envrc",
 		"nested/.env.development.local",
+		// 🔴 #409: a linked worktree's / submodule's `.git` is a regular FILE.
+		// It is here because the two sides used to AGREE about it — and agree
+		// that it ships. Both now drop it, and a fix applied to only one side
+		// (an inline check in Build's walk instead of in the shared
+		// isExcludedFile) makes this test red, which is what makes the fix
+		// observable through the seam rather than only through Build.
+		".git",
+		"sub/.git",
 	}
+	// The names that merely START with ".git" stay in the fixture as content:
+	// both sides must keep shipping them, and a prefix match would make Build
+	// and the predicate agree on a WRONG answer, which only the exact-name
+	// assertions in git_metadata_file_test.go can catch.
+	files = append(files, ".gitignore", ".gitattributes", "src/.gitkeep")
 	for _, f := range files {
 		writeFile(t, dir, f, "x")
 	}
--- a/internal/pkgzip/git_metadata_file_test.go
+++ b/internal/pkgzip/git_metadata_file_test.go
@@ -1,0 +1,332 @@
+package pkgzip
+
+import (
+	"bytes"
+	"io/fs"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"sort"
+	"strings"
+	"testing"
+)
+
+// TESTS FOR ISSUE #409: `.git` is a FILE in a linked worktree and in a
+// submodule, and the exclusion used to be keyed on the entry's TYPE rather than
+// its NAME.
+//
+// 🔴 THE DEFECT WAS NOT "we forgot .git" — `.git` was on excludedDirs from the
+// start. It was that the name was consulted ONLY inside `if d.IsDir()`, so the
+// rule held for the shape the author of it had in mind (a clone) and silently
+// did not hold for the two shapes that are equally normal (a linked worktree,
+// a submodule). In those, `.git` is a regular file holding `gitdir: <abspath>`;
+// it passed the directory gate (not a directory) and the file gate
+// (isExcludedFile only knew *.zip and .env*) and was packaged.
+//
+// What that cost: Forgejo rejects a bundle containing `.git` with
+// "path contains a malformed path component [path: .git]" — a 400 that names
+// neither the CLI, the bundle, nor the author's checkout, so `app submit` was a
+// hard stop with no route to the cause. And the file's CONTENT is an absolute
+// path on the author's machine, which the packager should not be assembling
+// into an upload whatever the server does with it.
+//
+// The fixtures below build REAL worktrees and REAL submodules rather than
+// hand-writing a `.git` file, because the claim under test is about what git
+// actually produces. Each one asserts its own precondition (that `.git` really
+// came out a regular file) — a fixture that silently produced a directory would
+// exercise the path that always worked and pass for the wrong reason.
+
+// gitEnv isolates every git invocation here from the machine's own config.
+func gitEnv(t *testing.T) {
+	t.Helper()
+	t.Setenv("GIT_CONFIG_GLOBAL", os.DevNull)
+	t.Setenv("GIT_CONFIG_SYSTEM", os.DevNull)
+	t.Setenv("GIT_AUTHOR_NAME", "cli-test")
+	t.Setenv("GIT_AUTHOR_EMAIL", "cli-test@example.invalid")
+	t.Setenv("GIT_COMMITTER_NAME", "cli-test")
+	t.Setenv("GIT_COMMITTER_EMAIL", "cli-test@example.invalid")
+}
+
+// runGit runs a git subcommand in dir and fails the test on error.
+func runGit(t *testing.T, dir string, args ...string) string {
+	t.Helper()
+	c := exec.Command("git", append([]string{"-C", dir}, args...)...)
+	var out, errb bytes.Buffer
+	c.Stdout = &out
+	c.Stderr = &errb
+	if err := c.Run(); err != nil {
+		t.Fatalf("git -C %s %s: %v\n%s", dir, strings.Join(args, " "), err, errb.String())
+	}
+	return strings.TrimRight(out.String(), "\r\n")
+}
+
+// tempRepo creates a real git repository in a fresh temp dir — NEVER inside the
+// CLI's own checkout — carrying a minimal packageable app.
+func tempRepo(t *testing.T) string {
+	t.Helper()
+	if _, err := exec.LookPath("git"); err != nil {
+		t.Skip("git not on PATH — the real-repo fixtures cannot be built")
+	}
+	gitEnv(t)
+	// EvalSymlinks: on macOS t.TempDir() lives under a symlink and git answers
+	// with the resolved path, which makes the fixture's own paths disagree.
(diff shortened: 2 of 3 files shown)


ZacxDev:
The one red check, `pins-vs-published`, is **pre-existing and unrelated** — `@civitai/app-sdk` published `0.35.0` while the scaffold pins `^0.34.0`, which a pre-1.0 caret does not admit. It lives in `internal/scaffold`, which this PR does not touch.

Control run, so this is not an assumption: the same test against an **unmodified `origin/main` @ e816594** (`git archive origin/main` into a clean dir, `CIVITAI_CHECK_PUBLISHED_PINS=1 go test ./internal/scaffold -run TestScaffoldPinsSatisfyPublished`) fails identically —

```
pinned:   ^0.34.0   (does NOT admit the published version — pre-1.0 caret locks the minor)
published:0.35.0   (npmjs.org/@civitai/app-sdk latest)
--- FAIL: TestScaffoldPinsSatisfyPublished (0.38s)
```

Fix is `go run ./internal/scaffold/cmd/bump-pins`, on its own branch. The other 11 checks pass.

Issue #424, opened 2026-08-15
resolveListing documents a slug-only lookup as load-bearing; the live server honors it for nothing, and both tests pinning it stub the success


ZacxDev:
`resolveListing` documents a slug-only lookup as load-bearing; the live server honors it for nothing, and both tests pinning it stub the success

## The claim

`internal/cmd/app_listing.go:186-192` carries a 🔴 comment asserting a server capability:

> A submission with NO appBlockId is a FIRST-version app still pending review. It still has a store listing: `civitai app submit` mints one as a pre-approval DRAFT so its media is settable while pending. That draft has `appBlockId = NULL` and is resolvable **ONLY BY SLUG**, so we must still call through — short-circuiting to a "no listing" error here would make the pending-media flow unreachable from the CLI **even though the server supports it**. The slug is passed on BOTH paths (harmless when the appBlockId resolves; **load-bearing when it is absent**).

`GetMyListingForApp(ctx, appBlockID, slug)` (`internal/appapi/listing.go:229`) is built around that: it omits `appBlockId` when empty and sends `slug` alone.

## Measured — the slug selector resolves nothing

Live against `https://civitai.com` as `zachlowdenzx`, calling `appListings.getMyListingForApp` through the CLI's own client (so the tRPC `{"json":…}` encoding is the one that works in production), varying **only the selector**:

| app | `appBlockId` only | `slug` only | both |
|---|---|---|---|
| `gen-matrix` (onsite, approved+live) | ✅ `apl_01KWFP4FEEJRWN27CJA49CDY4Q` approved | ❌ 404 | ✅ same id |
| `custom-generators` (onsite, approved+live) | ✅ `apl_01KXKAJF85NQ8X702AHKX2MMAW` approved | ❌ 404 | ✅ same id |
| `etag-probe-05773` (submission has `appBlockId = nil`) | n/a | ❌ 404 | n/a |
| `dogfood-test-palette` (submission has `appBlockId = nil`) | n/a | ❌ 404 | n/a |
| `radio`, `comfy` (offsite — no appBlockId exists) | n/a | ❌ 404 | n/a |
| `definitely-not-an-app-zzz` | n/a | ❌ 404 | n/a |

**Positive control satisfied:** the `appBlockId`-only calls return real listing ids in the same run, so the route works, the auth and wire encoding are right, and the probe can observe a success. **Negative control:** a nonexistent slug 404s. The 404 on every slug-only row is therefore about the *selector*, not about the probe.

[...]

Issue #422, opened 2026-08-15
app listing / app status are unreachable for offsite apps — resolution is submission-first (4/4 offsite fail, 7/7 onsite pass)


ZacxDev:
`app listing` (and `app status <slug>`) cannot reach **offsite** apps. Every subcommand resolves its target through the *submissions* list, and an offsite app has no submissions — so the commands fail with `no such submission` even though the app exists, is owned by the caller, and already has an icon and cover serving publicly.

## Repro

```
civitai app listing status --slug radio
Error: no such submission for app "radio" — run `civitai app submit` first; the submission
and its draft store listing are created at submit time (list what you have submitted with
`civitai app status`)
```

`set-icon` fails identically, so there is no CLI path to attach media:

```
civitai app listing set-icon ./icon.png --slug radio --changelog "..." -y
Error: no such submission for app "radio" — ...
```

The advice in the message is not actionable here: `app submit` packages a block bundle, and an offsite app has no bundle — it is a registered URL.

## The correlation is exact, not a sample

All 11 apps on one account, `app view <slug>` for `kind` against `app listing status --slug <slug>`:

| slug | kind | `app listing status` |
|---|---|---|
| comfy | offsite | ❌ `no such submission` |
| cosmetic-studio | offsite | ❌ `no such submission` |
| radio | offsite | ❌ `no such submission` |
| vitrine | offsite | ❌ `no such submission` |
| app-requests | onsite | ✅ |
| custom-generators | onsite | ✅ |
| gen-matrix | onsite | ✅ |
| model-benchmarking | onsite | ✅ |
| panorama-360 | onsite | ✅ |
| playable-collections | onsite | ✅ |
| sensei | onsite | ✅ |

**4/4 offsite fail, 7/7 onsite succeed.** `kind` predicts the outcome perfectly.

## Why this is a gap rather than "offsite apps have no listing"

All four offsite apps **already serve an icon and a cover**: `GET /api/v1/apps/<slug>` returns non-null `iconUrl` and `coverUrl` for each (HTTP 200). So the media exists and was set through some surface — just not one the CLI can reach. The CLI cannot read it, update it, or replace it.

Concretely: I have replacement icon + cover assets ready for `radio` and `comfy` and no CLI command that will attach them.

## Mechanism

[...]

Issue #393, opened 2026-08-11
safeTerm passes invisible and bidi runes (U+2800, U+200B, U+202E)


ZacxDev:
`safeTerm` is the CLI's one gate on server-supplied text reaching a terminal. It strips C0 (below `0x20`, keeping `\t`/`\n`), `DEL`, and C1 (`0x80`–`0x9F`). Everything above `U+009F` passes.

That leaves a class of runes that are **printable by `unicode` and invisible or layout-altering in a terminal**:

| rune | what it does | `unicode.IsSpace` | `unicode.IsControl` | stripped by `safeTerm` |
|---|---|---|---|---|
| `U+2800` BRAILLE PATTERN BLANK | renders as blank | **false** | false | no |
| `U+200B` ZERO WIDTH SPACE | invisible | false | false | no |
| `U+202E` RIGHT-TO-LEFT OVERRIDE | reverses displayed order | false | false | no |

Measured for `U+2800`:

```
U+2800  IsSpace=false  IsControl=false  IsPrint=true
strings.Fields("a⠀⠀b") -> 1 token
safeTerm strips it? no
```

## Why it matters now

`unicode.IsControl` is `Cc`-only, so `hasPrintableContent` (`internal/genapi/status.go`) counts these as printable — a reason made entirely of them is treated as content. And because they are not `IsSpace`, `strings.Fields` does not split on them, so text that *looks* space-separated is a single token to every wrapper in this CLI.

`#392` makes this materially more consequential by rendering unbounded provider text beneath a column-aligned table: a `U+2800`-separated string reads as a table row while being one unsplittable token. That specific vector is handled in `#392` itself; **this issue is the general class**, which predates it and applies to every server string the CLI prints — validation findings, `app metrics` tokens, blocked-reason text, error messages.

`U+202E` is the sharper one: it can reverse the displayed order of a line, so what a user reads is not what the bytes say. No current guard addresses it.

## 🔴 The display-width half has MOVED to #397

This issue originally also carried "every wrapper measures runes, not display cells". That half is now **#397** and is **not** part of this issue's work — do not close this one by fixing wrapping, and do not close #397 by fixing this one. It was split because closing it needs a character-width table, which is either a new direct dependency (`mattn/go-runewidth`, already present as an *indirect* one) or a hand-rolled table on `golang.org/x/text` — and `AGENTS.md` requires asking before adding a dependency. #397 carries the measurements, the three wrappers with their files, and the guard shape.

## Work

[...]

PR #426, opened 2026-08-15
fix(app): name the offsite app instead of `app submit`, which cannot work for it (#422)

ZacxDev:
> 🔴 **THIS DESCRIPTION DESCRIBES `4fab6b1` AND WAS WRONG IN TWO PLACES.** An audit
> found that the URL guard it describes as accepting a URL with *no whitespace at
> all* in fact rejected only four ASCII bytes — U+00A0, U+2028, U+2029, U+3000,
> U+205F, U+1680 and U+2007 all rendered — and that the README changes listed
> under *Docs* stated the new behaviour **unconditionally** when it depends on a
> store-catalog probe that often cannot answer. Both are fixed in `de254b3`.
> The mutation matrix below is the **11-mutant** run for `4fab6b1`; the current
> head is covered by a **26-mutant** run, 15 of them new.
> **Read [the correction comment](https://github.com/civitai/cli/pull/426#issuecomment-5304426755) before this body.**
> Two further findings are deliberately out of scope and tracked as #427.

---

Implements **outcome 2 of #422 — "refuse precisely"**. Outcome 1 ("support it") is
**not implemented and is not a client-side change**; see *Scope* below.

## What broke

`app listing <sub>` and `app status <slug>` resolve their target through
`/api/v1/blocks/submissions?blockId=<slug>`. An **offsite** app is a registered URL
rather than a block bundle, so it has no submissions and both commands die in
appapi's empty-list branch with:

```
Error: no such submission for app "radio" — run `civitai app submit` first; the
submission and its draft store listing are created at submit time (list what you
have submitted with `civitai app status`)
```

`app submit` packages a block bundle. An offsite app has none. The message names a
next step that cannot exist, so a dead end reads like a fixable state.

## What this does

`internal/cmd/app_offsite.go` — **one** predicate, `offsiteApp`, asked by both call
sites (`resolveListing` and `app status`'s detail path). RULES: one rule, one place;
a predicate open-coded at two sites is typically wrong at one of them.

* **Error path only.** `GET /api/v1/apps/{slug}` runs *after* the command has already
  failed, and only when the failure is classified not-found. No successful command
  gains a request (pinned), and a 403/5xx keeps its own message and exit code (pinned).
* **A failed diagnostic never makes things worse.** 404, 403, 5xx, garbage body,
  timeout, or a `kind` this CLI does not recognise all collapse to **today's message,
  byte-identical**.
* **Its own deadline (5s), and silent.** The read client's `network error from
  Civitai, retrying (n/4)…` notices would otherwise print *above* the real error.
* **Two different messages, deliberately not collapsed** (this is the point of the task):

```
`radio` is an OFFSITE app (registered at https://…), and an offsite app's store
listing cannot be addressed from this CLI — `civitai app listing` resolves a
listing through the app's block submission, and an offsite app has none.
`civitai app submit` is NOT the missing step: it packages a block bundle, and an
offsite app is a registered URL, so no submission it could create exists to be made.
Its store listing is live all the same — run `civitai app view radio` to see the icon
and cover it is serving right now. Changing that media is only possible in the
App-store listing UI on civitai.com, where this app was registered; that surface is
authoritative and the CLI has no equivalent for it.
```

```
`radio` is an OFFSITE app (registered at https://…), so it has no block submissions —
`civitai app status` reads the block-submission pipeline (submit → review → deploy),
and an offsite app is a registered URL rather than a block bundle, so there is nothing
here to be pending, approved or deployed. Nothing is missing and `civitai app submit`
would not create one.
Run `civitai app view radio` for what the CLI can show about this app.
```

Both name `civitai app view <slug>`, a command that works for these apps. Neither names
`app submit`. No CLI path is promised that does not exist.

## Exit code: stays `4`, deliberately

AGENTS.md item 7 governs, and every case is asserted with `errors.Is`, never on message
text. The app exists, but the *resource these two commands look up* — its block
submission — genuinely does not, and never will. A script branching on `4` for "this
slug has no listing/submission here" stays right; changing the code would break that
contract for a distinction only a human reads. Contrasted in the README against the
`has no approved App Block yet` case, which exits `1` precisely because the thing looked
up is expected to appear later.

Because the code did not move, the exit-code **table** is unchanged; the exit-code-4
**detail** did move (it is generated — `internal/cmd/exitcodes_doc.go`, not hand-edited).

## Measured vs assumed

**Measured (by me, in this PR):**

* Every assertion below, against `httptest` fakes through the real command tree.
* golangci-lint **v2.12.2** (CI's exact pin, via `nix-shell`): 0 issues — with a negative
  control that reported 1 (`ineffassign`) before restore.
* `make ci` green; `./scripts/ci-shallow.sh` green on the **committed** SHA (both tiers).
* Full verbose suite: **RUN=4246 PASS=4242 FAIL=0 SKIP=4**, 0 timeout panics, 20/20 packages.

**Taken from the issue thread, not re-measured here** (all live measurements by @ZacxDev
at `1b20b99`): 4/4 offsite apps fail and 7/7 onsite pass; `getMyListingForApp` resolves
only by `appBlockId` and 404s on every slug-only call including onsite positive controls;
all four offsite apps return 200 with non-null `iconUrl`/`coverUrl` on
`GET /api/v1/apps/{slug}`.

**Assumed:** that the store detail route stays reachable for offsite apps. If it is not,
the probe fails and the user gets today's message — the failure direction is the safe one,
and it is pinned by five separate probe-failure cases.

**Not verified:** nothing here was run against the live API. Every offsite/onsite response
is a fixture.

## Red-at-base / green-at-HEAD

Base = `origin/main` @ `1b20b99`, extracted with `git archive` (no git state touched). The
three tests naming new production identifiers cannot compile at base and were dropped from
the base copy; the rest are command-driven and compile unchanged. **At base: RUN=27 PASS=13
FAIL=14.** At HEAD all 45 pass.

| test | at `1b20b99` | at HEAD |
| --- | --- | --- |
| `TestAppListingOffsiteRefusesPrecisely` | 🔴 | ✅ |
| `TestAppStatusOffsiteRefusesPrecisely` | 🔴 | ✅ |
| `TestOffsiteRefusalRendersTheAppsOwnExternalURL` | 🔴 | ✅ |
| `TestOffsiteRefusalReachesEverySubcommand` (4 subs) | 🔴 | ✅ |
| `TestOnsiteNeverSubmittedKeepsTodaysMessage` | 🔴 *on its premise only* | ✅ |
| `TestUnknownSlugKeepsTodaysMessage` | 🔴 *on its premise only* | ✅ |
| `TestProbeFailureKeepsTodaysMessage` (5 subs) | ✅ **invariant guard** | ✅ |
| `TestNonNotFoundErrorsAreNeverProbed` | ✅ **invariant guard** | ✅ |
| `TestSuccessfulCommandsMakeNoStoreProbe` | ✅ **invariant guard** | ✅ |
| `TestAppStatusByIDDoesNotProbeTheStore` | ✅ **invariant guard** | ✅ |
| `TestOffsiteRefusalsDifferPerCommand` | *cannot compile* | ✅ |
| `TestOffsiteRegisteredAtRejectsUnsafeServerText` (10 subs) | *cannot compile* | ✅ |
| `TestOffsiteProbeDoesNotHoldTheCommandOpen` | *cannot compile* | ✅ |

Labelled honestly: the two "keeps today's message" tests pin a string that already shipped,
so their **message** assertion is an invariant guard. What is red at base is their
**premise** (`appCalls == 1`) — before this change nothing probed the store at all, and
without that premise the message assertion would pass equally for "no probe exists". The
four rows marked *invariant guard* are green at base by construction and are **not** counted
as regression coverage.

## Mutation matrix — 11 mutants, 11 killed

Each mutation applied to the shipped source, targeted suite re-run, then restored from a
`cp` copy (never `git checkout --`). Baseline: **RUN=45 PASS=45 FAIL=0**.

| # | mutation | result |
| --- | --- | --- |
| M1 | drop `client.Stderr = io.Discard` | 🔴 `the diagnostic probe must be SILENT` + the leaked `network error from Civitai, retrying (2/4)…` |
| M2 | invert `d.Kind != appKindOffsite` | 🔴 14 failures — onsite, unknown-kind and no-kind cases all misclassified |
| M3 | drop the `errors.Is(err, ErrNotFound)` gate | 🔴 `TestNonNotFoundErrorsAreNeverProbed`: a 403 rewritten into an offsite claim, exit 3 → 4 |
| M4 | drop `civitai.Tag(civitai.ErrNotFound, …)` | 🔴 both exit-code assertions, `got *errors.errorString` |
| M5 | point `app status` at `offsiteListingRefusal` | 🔴 `missing "no block submissions"` + `must not talk about the listing UI` |
| M6 | remove the probe's `context.WithTimeout` | 🔴 `the command took 2m1.8s` (inherits 30s × 4 attempts) |
| M7 | hardcode the external URL literal | 🔴 13 failures — app B's message carries app A's URL |
| M8 | drop the empty-slug guard | 🔴 `an --id lookup has no slug to probe; the store route ran 1 times` |
| M9 | let a slug win over `--id` | 🔴 the `--id wins over a slug` case only |
| M10 | revert the `app listing` call site | 🔴 11 failures |
| M11 | revert the `app status` call site | 🔴 5 failures |

**Reachability**, not just breakability: M10/M11 show the branch is reached through the real
command tree at both sites, and `TestOffsiteRefusalReachesEverySubcommand` drives four
`app listing` subcommands (`status`, `set-icon`, `set-cover`, `rm-screenshot`) — the branch
is not reachable only by the one command a single case happens to drive.

**Fixture trap (#422's own warning) closed by construction.** Two *different* offsite
fixtures with two *different* external URLs, distinct from the onsite fixture in both `kind`
and `kindData`, and distinct from every literal in non-test source. M7 is the control that
the number moves: hardcoding the URL survives nothing.

## Docs — "README is not one place"

Grepped the whole file; four surfaces named this behaviour and all four moved:

1. the `app listing` details list (the bullet explaining the refusal and the authoritative surface),
2. the `## Submission status` section (the `app status` half),
3. the exit-code-4 detail — **generated**, so edited in `internal/cmd/exitcodes_doc.go` and
   regenerated (`TestREADMEExitCodeSectionsAreGenerated` caught the hand edit),
4. Troubleshooting: the `no such submission` row now says which case never reaches it, plus a
   new `is an OFFSITE app` row.

AGENTS.md gains **inline item 29**. Inline rather than an evidence file on purpose: a new
`claudedocs/decisions/NN` file needs a `splitItems` row pinning a sha256 against a base commit
where the body already lived in AGENTS.md, which a first-appearance item cannot have without a
two-PR dance. It clears `maxInlineItemBytes` (600) — the same rule items 2 and 4 sit under — and
that is measured, not assumed: `TestInlineAgentsItemsStayUnderTheBreakEven` failed the first draft
at 646 bytes and passes the shipped one. The substance it defers to lives in the `app_offsite.go`
header, where anyone editing this will read it.

## Scope — what is still blocked

Outcome 1 of #422 (make offsite apps actually reachable) is **not attempted, and cannot be done
client-side**: `appListings.getMyListingForApp` resolves only by `appBlockId`, and an offsite app
has none — that is what `kind: offsite` means. Dropping the submission lookup moves the 404 one
call later. It needs a server-side slug/app-id selector on that route. The inert slug selector is
filed separately as #424, which this PR does not touch.

Refs #422

---

https://claude.ai/code/session_01HLgrqMjPceVksV2PPzrHEB


ZacxDev:
## 🔴 Correcting this PR's description — two claims in it were false

An audit of `4fab6b1` found that **the description above misstates what the change does**, in two places. If you reviewed this PR before `de254b3`, you were reading claims that do not hold. Both are now fixed in code; this comment is here because a reviewer may already have believed the original text, and silently editing the body would not undo that.

### 1. "carrying no whitespace at all" — false, and it was the security-relevant claim

`offsiteRegisteredAt`'s doc comment said the external URL is accepted "only when it parses as an absolute http/https URL **carrying no whitespace at all**". The predicate underneath rejected **exactly four ASCII bytes** — `' '`, `'\t'`, `'\n'`, `'\r'`.

Measured against the function as it shipped in `4fab6b1`, every one of these **rendered**:

| rune | | rendered at `4fab6b1`? |
| --- | --- | --- |
| U+0020 SPACE | | no (correctly rejected) |
| **U+00A0** | NO-BREAK SPACE | **yes** |
| **U+2028** | LINE SEPARATOR | **yes** |
| **U+2029** | PARAGRAPH SEPARATOR | **yes** |
| **U+3000** | IDEOGRAPHIC SPACE | **yes** |
| **U+205F** | MEDIUM MATHEMATICAL SPACE | **yes** |
| **U+1680** | OGHAM SPACE MARK | **yes** |
| **U+2007** | FIGURE SPACE | **yes** |

Two of those are **line separators** — the exact sentence-breaking hazard the four-byte list was written to stop. `internal/saferune` is not a backstop: it **subtracts** Unicode's `White_Space` property from its strip class on purpose, so all of them reach that predicate intact.

This matters more than a doc slip because of what the value is. `offsiteApp` does **no ownership check** and `GET /api/v1/apps/{slug}` is the **public** catalog, so the string being spliced into a CLI error is very often a **third party's** `externalUrl`. Same class as #367, #382 and #393: a guard satisfied by writing the hazard a different way.

**Fixed in `de254b3`** by replacing the denylist with an **allowlist of RFC 3986's URI character set** — the question is now "is this rune legal in a URI?" rather than "is it one of the four someone remembered to ban". `unicode.IsSpace` was considered and rejected: it closes those eight and leaves the same shape of question open for the next non-space rune (a fullwidth solidus, say, which is now a test row). Stated residual: pure-ASCII only, so an IDN host or an unencoded UTF-8 path is dropped — the message says less and stays whole, which is the safe direction.

Incidentally, `'\r'` never needed a clause: `safeTerm` strips it before the predicate sees it. The old four-byte list did not know that either.

### 2. The README claim was stated unconditionally, and it is conditional

The description's *Assumed* section conceded that the store detail route may not answer. The **README did not** — three prose surfaces plus the generated exit-code detail asserted the new behaviour flatly ("does not work for an OFFSITE app, **and it says so**"; "**One case never reaches this row**").

That is false whenever the probe cannot answer, and this repo's own `apps.go:296-299` already documents why it often cannot: the route serves only a **published** store listing and sits behind a launch flag — *"until it opens publicly you only see apps if your account is a moderator or app-dev-tester"*. An unpublished offsite app, or a caller without that access, still gets today's `no such submission` message.

The README is the published contract, so it now carries the caveat on all four surfaces (`app listing` section, `app status` section, the generated exit-code-4 detail, and both Troubleshooting rows). Exit `4` is unconditional either way; only the human-readable half degrades.

### Also fixed in `de254b3`

* **`safeTerm` on the external URL was untested.** Replacing it with the raw field left the whole `internal/cmd` suite green. The one existing test asserted only **rejection**, which a dropped strip satisfies trivially — dropping it makes *more* values reject. `TestOffsiteRefusalSanitizesControlChars` pins the opposite property.
* **The `http`/`https` scheme clause was never reached.** Both "unsafe scheme" fixtures (`javascript:alert(1)`, `file:///etc/passwd`) parse to an **empty** `Host`, so `u.Host == ""` killed them first and deleting the scheme clause **survived** a green suite. `ftp://ok.example/x` and `javascript://ok.example/%0aalert(1)` reach it.
* **The fixtures echoed the caller's own slug**, so swapping the message's slug for the server's `d.Slug` survived. They are now non-containing in both directions, with a control test on the fixtures themselves.
* **`TestOffsiteRefusalReachesEverySubcommand` drove 4 of the 6** `app listing` subcommands. `add-screenshot` and `reorder` added — "every subcommand" was two-thirds measured.
* **A dangling pointer**: `app_offsite.go` cited `claudedocs/decisions/29-offsite-refusal.md`, which does not exist (item 29 is deliberately inline in AGENTS.md). The evidence ledger only ever looked at items *carrying* a pointer, so nothing caught it; `TestSourceCitationsOfDecisionFilesResolve` now fails on that whole class.
* Trailing full stops dropped from both refusals, matching `appViewOwnedAdvice` and `wantNoSubmissionMsg`.

### Mutation matrix — 26 mutants, 26 killed (15 new this round)

The gate reset, so the **full** battery was re-run, not just mutants for what changed. Beyond deletion: predicate inversion, operand swaps, and a clause **commented out so the text still matches any grep looking for it**.

| # | mutation | result |
| --- | --- | --- |
| M1 | drop `client.Stderr = io.Discard` | 🔴 1 |
| M2 | invert `d.Kind != appKindOffsite` | 🔴 18 |
| M3 | drop the `errors.Is(err, ErrNotFound)` gate | 🔴 3 |
| M4 | drop `civitai.Tag` | 🔴 2 |
| M5 | point `app status` at `offsiteListingRefusal` | 🔴 1 |
| M6 | remove the probe's `context.WithTimeout` | 🔴 1 |
| M7 | hardcode the external URL literal | 🔴 30 |
| M8 | drop the empty-slug guard | 🔴 3 |
| M9 | let a slug win over `--id` | 🔴 2 |
| M10 | revert the `app listing` call site | 🔴 14 |
| M11 | revert the `app status` call site | 🔴 5 |
| **M12** 🆕 | delete the whole rune filter | 🔴 10 |
| **M13** 🆕 | restore the **pre-fix** four-byte denylist | 🔴 9 |
| **M14** 🆕 | weaken the allowlist to `unicode.IsSpace` | 🔴 2 (U+FF0F) |
| **M15** 🆕 | invert `isURIRune` | 🔴 13 |
| **M16** 🆕 | delete the `u.Host == ""` clause | 🔴 2 |
| **M17** 🆕 | delete the scheme clause | 🔴 3 |
| **M18** 🆕 | scheme clause `&&` → `\|\|` | 🔴 13 (and `go vet` catches it too) |
| **M19** 🆕 | drop `safeTerm`, render the raw field | 🔴 7 |
| **M20** 🆕 | listing refusal renders the server's `d.Slug` | 🔴 1 |
| **M21** 🆕 | status refusal renders the server's `d.Slug` | 🔴 1 |
| **M22** 🆕 | **comment out** the scheme clause (still spelled) | 🔴 3 |
| **M23** 🆕 | re-dangle the decision-file citation | 🔴 1 |
| **M24** 🆕 | `serverEchoSlug` echoes the caller's slug | 🔴 1 |

M11 and M18 first came back as *build/vet* failures rather than assertions; both were re-run as compiling variants (M18 with `-vet=off`) and killed by real assertions — 5 and 13 respectively. Files were restored from a byte copy and re-checksummed, not `git checkout --`.

### Gate

* `make ci`: green. Full verbose suite **RUN=4267 PASS=4263 FAIL=0 SKIP=4**, 0 timeout panics, 20/20 packages — counted from the output, not read off an exit code.
* `golangci-lint v2.12.2` (CI's exact pin): **0 issues** — with a negative control that reported **1** (`ineffassign`) on a deliberately broken tree, so the zero means something.
* `./scripts/ci-shallow.sh` on the committed SHA `de254b3`: 20/20 ok, 0 failures, 0 timeouts.

### Still out of scope, and now tracked

Two real findings from the same audit are **not** fixed here — recorded in the `app_offsite.go` header and filed as **#427**: the refusal asserts ownership it never checked (pre-existing — the message it replaced was equally false for someone else's slug), and the probe adds a round trip to the commonest error path with no off switch (a cost, not a defect).

### Not verified

Nothing here was run against the live API; every offsite/onsite response is a fixture. The README caveat's *premise* — that the store detail route is gated the way `apps.go` says — is taken from this repo's own existing documentation of that route, not re-measured against the live server.

PR #433, opened 2026-08-16
fix(pkgzip): exclude dotenv- and archive-shaped DIRECTORIES, not only files (#420)

### LINKED ISSUE #420: pkgzip: `.env.local` / `*.zip` excluded as FILES but not as DIRECTORIES — a planted `.env.d/db.env` secret is packaged (mirror of #409)
## Summary

`internal/pkgzip` excludes `.env.local` and `*.zip` as **files**. When the same names exist as **directories**, their contents are packaged into the submitted bundle — including secrets.

This is the exact mirror of #409 (which was `.git` excluded as a directory but not as a file, fixed in #418). Same class, opposite direction, and this side carries a credential-leak hazard rather than a layout-leak one.

## Measured

Planted as directories in an app tree, then packaged:

| planted | result |
|---|---|
| `.env.local/` | contents packaged |
| `.env.d/` | contents packaged — including `.env.d/db.env` holding `API_SECRET=leak` |
| `artifact.zip/` | contents packaged |

The bundle is committed to Forgejo `civitai-apps/<slug>` and deployed, so anything in it is durably published.

## Why it was left out of #418

#418 deliberately scoped itself to VCS metadata (`.git`, `.hg`, `.svn`), promoting those to an **exact-name, any-depth** rule shared by `Build` / `IsExcludedPath` / `IsExcludedEntry`. The dotenv rule was left alone for two reasons worth preserving:

1. **Different blast radius.** A directory-shaped dotenv rule would start dropping directories like `.environment/` **whole**, silently, from submissions. VCS metadata has no such collision; a directory named `.git` is never author content.
2. The dotenv rule is ledger-bound to a README section that PR #418's author did not own.

So this is not an oversight in #418 — it is a deliberate deferral, recorded here so it does not evaporate.

## Suggested shape

Decide per name whether the directory form should be excluded, rather than blanket-promoting `excludedDirs` ⇄ `excludedFiles` in both directions. #418's write-up notes that a blanket promotion the other way would drop an extensionless `build` **script**, with no server-side error to catch it — so symmetry is not automatically correct here.

Whatever rule lands should be enforced in the **shared predicate**, not in `Build`'s walk, so the packager and the `app submit` dirty-tree guard cannot drift. #418 added a seam test that fails when they disagree (`Build and IsExcludedPath disagree about this tree`); extend that rather than writing a parallel check.

## Test note

The shape that hid both halves of this class: a fixture that only ever plants the **expected** file type. `.git` was only ever tested as a directory; `.env.local` is only ever tested as a file. A guard that has never seen the other shape cannot report it.

ZacxDev:
Closes #420. This is #409 with the shapes swapped — and it points at credentials rather than layout.

`.env.local` and `*.zip` were excluded as **files** only, so the same names as **directories** had every file inside them packaged into a bundle that is committed to Forgejo `civitai-apps/<slug>` and deployed, i.e. durably published.

**Reproduced end-to-end through the real command**, not only in a unit test:

| binary | `.env.d/db.env` (`API_SECRET=leak`) | control `.environment/config.yaml` |
|---|---|---|
| released **v0.1.96** | **in the bundle** — grep count **1** | ships |
| this branch | absent — grep count **0** | ships |

🔴 **The file rule does not cover for the directory rule**, which is what makes this a leak rather than bloat. Inside a `.env.d/` the conventional names are `db.env`, `api.env` — base names that do **not** start with `.env`, so `isExcludedFile` keeps every one of them. Excluding the directory is the only thing that stops them.

### The rule is decided per name, not promoted wholesale

The issue asks for this explicitly, and the mirror-image mistake is real in the other direction too (#418 records that promoting `excludedDirs` into the file rule would drop an extensionless `build` **script**).

- **Dotenv-shaped directories** — `.env`, or `.env.` + anything — are excluded at any depth. **The dot is required**, unlike the file rule's bare `.env` prefix, because matching too much here removes a whole **subtree** silently: `.environment/` and `.envoy/` are ordinary content. The residual is documented in both the code and the README: `.env-backup/` is **not** excluded.
- **`keptEnvFiles` does not apply to directories.** That allow-list exists because `vite build` *reads* `.env.production`; nothing reads a directory of that name.
- **Archive-shaped directories** (`*.zip`) go too, through one `isArchiveArtifact` predicate now shared with the file rule so the two shapes cannot drift apart a third time.

Enforced in the shared `isExcludedDir`, which **both** `Build`'s walk and `IsExcludedPath` consult — so the packager and the #411 dirty-tree guard cannot disagree. The existing seam test gains rows for the new shapes rather than getting a parallel check beside it.

### Test matrix

**Red at `b1213a3`** (the bundle carried every planted path, and the predicate answered `false` for all 11 rows), **green here**. Four mutations, each killed by its own assertion:

| mutation | what goes red |
|---|---|
| walk back to `excludedDirs`-only | the Build test **and** the seam guard's own `Build and IsExcludedPath disagree about this tree` |
| `IsExcluded` back to map-only | the predicate rows |
| widen the rule to the bare `.env` prefix | the `.environment` / `.envoy` near-miss controls — which is what makes them load-bearing, not decorative |
| drop the archive rule | the `*.zip` rows |

The secret-in-bytes assertion carries a positive control (`TestSecretSearchCanFindAKeptFile`) proving the search can find a string in a file the packager **did** ship — otherwise a clean result would only prove the search is broken.

README: the dotenv section gains a paragraph on the directory rule and states the `.env-backup/` gap plainly. Its ledger test caught the edit, as designed; the golden was re-approved with `-update` after checking the new text asserts nothing about any file's **contents**.

https://claude.ai/code/session_01QybSDbsJwEMJdxDRUasTC1


ZacxDev:
**Correction to this PR's description**, from an adversarial audit — posting rather than silently editing the body, since a reviewer may already have read the wrong version.

> Enforced in the shared `isExcludedDir`, which **both** `Build`'s walk and `IsExcludedPath` consult — so the packager and the #411 dirty-tree guard **cannot disagree**.

**The last clause is too strong.** `IsExcludedPath` still *reproduces* `Build` rather than sharing it — its own doc comment says exactly that — and `TestIsExcludedPathAgreesWithBuild` is an **agreement** oracle, so a *symmetric* error keeps it green. Measured: reverting **both** the walk and `IsExcluded` (i.e. a full revert of this fix) leaves that test **passing**, and the new seam rows also pass at base.

The accurate claim is: **the two cannot disagree about these shapes under an *asymmetric* change** — a fix applied to the walk alone, which is the realistic drift and the one the seam rows are there to catch.

The audit's other findings on the PR *description* held up: the red-at-base matrix, the end-to-end leak reproduction, and the 4 mutants were all independently reproduced (it constructed 16 mutants total, all killed). Corrections to the code and docs are pushed in `7c833d2`; the remaining residuals are tracked separately rather than expanded into this PR.

PR #450, opened 2026-08-17
chore(nix): bump flake vendorHash to match go.sum

ZacxDev:
Automated by the `bump-flake-vendorhash` workflow.

The Nix flake's `vendorHash` is a fixed-output hash derived from
`go.sum`. It drifted from the current `go.mod`/`go.sum` (a dep was
added or bumped), which makes the `flake` CI build fail and silently
breaks `nix run github:civitai/cli`. This PR recomputes the correct
hash via the `lib.fakeHash` dance (write a fake hash → `nix build` →
parse the real `got: sha256-…` → write it back) and updates
`flake.nix`.

**Validated in-job before opening this PR:** `nix build .#default`
succeeds with the new hash and `./result/bin/civitai version` runs.

**CI note:** this PR is opened with the scoped fine-grained token
`flake` check re-runs the build on it as an independent confirmation.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -33,7 +33,7 @@
           # Computed from go.sum (this module is not vendored). To recompute
           # after a go.mod/go.sum change: set this to `lib.fakeHash`, run
           # `nix build`, and copy the `got: sha256-…` value from the error.
-          vendorHash = "sha256-prgKRC+357KX7vJ9jcIq2+1FUD91a2Gg5Q89R+FTZ+g=";
+          vendorHash = "sha256-l0RyxhS7+tdL9FngitiUhrciQbZfyDmTp5tP3Kd6NGQ=";
 
           # The module root is `package cli` (a library that embeds the schema);
           # the executable lives in ./cmd/civitai (package main). Build only it.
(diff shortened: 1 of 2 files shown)


ZacxDev:
Superseded by #467, now on `main` as `075bf06`.

That PR carries this one's entire payload — the identical `vendorHash`
(`sha256-l0Ryxh…`), independently re-derived by building rather than copied —
plus the fix for the defect this PR exposed.

**What this PR revealed:** alongside the hash bump it also committed
`result -> /nix/store/kkfnywjq…-civitai-4e35eaa-dirty`, a nix out-link left by
the job's own validate steps. All 13 checks were green on it; nothing in CI
detects a stray artifact. Root cause was `peter-evans/create-pull-request`
staging the whole tree when `add-paths` is absent — confirmed in the action's
source at the SHA this repo pins — with the validate steps running *before* the
PR step, and `result` not gitignored.

#467 pins `add-paths: flake.nix` (audited as the complete set of paths this job
writes), adds `/result` and `/result-*` to `.gitignore`, and documents why the
two sibling drift-bots deliberately do *not* get the same filter.

Closing rather than fixing in place: `automation/bump-flake-vendorhash` is a
bot-owned branch recreated on each run with `delete-branch: true`, so there is
nothing here to preserve. The next run will recompute the same hash, find no
drift, and open nothing.

PR #488, opened 2026-08-24
feat(app): `civitai app doctor` — a listing diagnosis whose EXIT CODE gates a release [DO NOT MERGE until #4341 is on release]

ZacxDev:
## Merge status: the server gate is SATISFIED — merging awaits explicit approval

civitai/civitai#4341 is **serving in production**. Verified by content, not ancestry alone:

- `origin/release` is `0c84428da3` ("5.1.42") and carries **16** `.meta(listingMediaCliScope)` sites, up from 13. `listMine` sits at `:1026` **with** the meta — it was `:914` bare.
- **Control**: `getAssets` is still `protectedProcedure` on that same ref, so the grep discriminates rather than matching everything.
- dp-prod serves `ghcr.io/civitai/civitai-prod:20260825001134-0c84428` — the short sha matches the release tip.

🔴 **Nothing here treats that as authorization.** The release cut, green CI and a clean mutation sweep are all prerequisites; none is approval. **Do not merge without an explicit go-ahead.**

### Live validation, and what it does NOT prove

`civitai app doctor --json` against `https://civitai.com`, 21 real listings: **exit 0**, `ok=true`, `{apps: 21, blocking: 11, advisory: 49, gating: 0, delisted: 11}`. All 11 blocking problems sit on the 11 `removed` listings, so the delisted rule holds on real data — **this same account exited 1 before that fix**. Both kind arms fired: `radio` (off-site) got the browser editor, `panorama-360` (on-site) got the manifest.

🔴 It ran with a **full-scope personal API key**, which `enforceTokenScope` early-returns on before any `requiredScope` check. So it exercises the procedures, this CLI's decode and the logic — and proves **nothing** about the scoped-token path these PRs exist to open. See the OAuth gap at the bottom, which is the one claim still open.
---

## What this adds

`civitai app doctor [slug]` — the client half of #4341. `computeListingProblems` has emitted eight codes for months and an author driving the CLI could reach none of them.

- **No arg** → every listing you own **or hold an accepted collaborator seat on**. That is a wider set than `app status`, which is scoped to what *you submitted* and therefore cannot see a listing acquired by ownership transfer or held on a seat.
- **A slug** → just that one, resolved through `appapi.SameSlug` (the shared predicate), so a padded or mis-cased spelling still lands. An unknown slug exits **4**, not 0 — "you own no app called that" and "that app is fine" are opposite answers.
[...]

PR #489, opened 2026-08-24
feat(listing): `set-text` — fix an empty tagline/description/category from the CLI [DO NOT MERGE until #4341 is on release]

ZacxDev:
## Merge status: the server gate is SATISFIED — merging awaits explicit approval

civitai/civitai#4341 is **serving in production**. Verified by content, not ancestry alone:

- `origin/release` is `0c84428da3` ("5.1.42") and carries **16** `.meta(listingMediaCliScope)` sites, up from 13. `listMine` sits at `:1026` **with** the meta — it was `:914` bare.
- **Control**: `getAssets` is still `protectedProcedure` on that same ref, so the grep discriminates rather than matching everything.
- dp-prod serves `ghcr.io/civitai/civitai-prod:20260825001134-0c84428` — the short sha matches the release tip.

🔴 **Nothing here treats that as authorization.** The release cut, green CI and a clean mutation sweep are all prerequisites; none is approval. **Do not merge without an explicit go-ahead.**

### Live validation, and what it does NOT prove

`civitai app doctor --json` against `https://civitai.com`, 21 real listings: **exit 0**, `ok=true`, `{apps: 21, blocking: 11, advisory: 49, gating: 0, delisted: 11}`. All 11 blocking problems sit on the 11 `removed` listings, so the delisted rule holds on real data — **this same account exited 1 before that fix**. Both kind arms fired: `radio` (off-site) got the browser editor, `panorama-360` (on-site) got the manifest.

🔴 It ran with a **full-scope personal API key**, which `enforceTokenScope` early-returns on before any `requiredScope` check. So it exercises the procedures, this CLI's decode and the logic — and proves **nothing** about the scoped-token path these PRs exist to open. See the OAuth gap at the bottom, which is the one claim still open.
---

## What this adds

`civitai app listing set-text` — the **write** half of the three text problems `civitai app doctor` reports (`empty-tagline` / `empty-description` / `empty-category`). Until civitai/civitai#4341, those three were fixable only in the browser.

```bash
civitai app listing set-text --tagline "Batch upscaling, in your browser"
civitai app listing set-text --category utility --slug my-app
civitai app listing set-text --clear tagline,category
```

### One command with flags, not three subcommands

The three **media** commands are separate because each takes a positional file with its own byte cap, decode, attach proc and scan poll — genuinely three flows. These three fields are **one proc taking one `patch` object**, whose schema refines that at least one key is present: the server models it as a single partial update. Three subcommands would send three requests where the server expects one, spend three of a 30/hour budget on one logical edit, and can half-apply with no transaction around them.

### 🔴 The parent-vs-shadow branch in the plan does not exist for these fields

The brief said this command had to own `updateListing` vs `updateRevisionDraft`, each refusing the other's target. **That premise is false here**, and I verified it rather than building to it:

- `MATERIAL_PATCH_FIELDS = ['externalUrl', 'name', 'contentRating', 'sourceRepoUrl']` — `offsite-listing.service.ts:732`. Note `contentRating` **is** material; tagline/description/category are not.
- `patchHasMaterialChange` (`:849-889`) is a **diff**, not a presence check — `if (patched === undefined) continue;`.

So a text-only patch matches nothing, returns `false`, and takes the **in-place** branch on every listing status. `updateListing` alone covers the command, and it always answers `requiresReview: false, shadowId: null`.

**`updateRevisionDraft` is therefore deliberately unwired.** Wiring it would be unreachable code shaped like a safety mechanism — worse than none, because the next reader sees the shadow case "handled" and stops looking. Instead the CLI **decodes and reports** the server's own `requiresReview`/`shadowId`, so if that field set ever widens the user is told a revision was opened rather than handed a success line that has become false. (`S14` mutant pins that branch is live.)

### The real hazard is different, and it is kind-scoped

If a revision is **already under review**, a trivial edit writes the **parent** — and on an **off-site** listing `applyApprovedRevision` copies the shadow's full scalar set back over the parent on approval, undoing it (`:2854-2870`). An **on-site** listing is structurally immune: that function's `onsite` branch (`:2837-2852`) writes **only** `iconId`/`coverId`, because an on-site listing's scalars mirror the manifest. I read both branches at `origin/main`.

So the command **warns on stderr and names the asymmetry**. It does not silently reroute to the shadow, for two reasons:

1. A write that targets a different row than the user named is the surprise class this CLI keeps getting bitten by.
2. Decisively — the only read that exposes an open shadow, `getMyListingForEdit`, **opens one as a side effect** on an approved listing. A command that consulted it to avoid the hazard would *create* the hazard on every run against a listing that had none.

**Residual, stated not hidden:** `hasPendingRevision` means *submitted*, not "a shadow exists", so an open-but-unsubmitted shadow is invisible here and no side-effect-free read exposes one. The warning covers what it can see and does not claim to cover the rest.

### Clearing and emptying are different states, and both are reachable

| you type | server gets |
|---|---|
| `--tagline "x"` | `"x"` |
| `--tagline ""` | `""` — legal; neither tagline nor description carries a `.min()` |
| `--clear tagline` | `null` |
| flag omitted | key absent — column untouched |

The patch is a `map[string]any`, **not** a struct with `omitempty`, precisely because `omitempty` drops the empty string — which would turn a deliberate edit into an invisible no-op. `category` accepts null but not `""`, so `--category ""` is refused locally and the refusal names `--clear category`. `--tagline X --clear tagline` is **refused**, not silently resolved: picking a winner would apply an edit the user did not ask for, to a public store page.

### Category is validated locally

Against a mirror of `MARKETPLACE_CATEGORIES` (`src/server/services/blocks/marketplace-categories.constants.ts`): `generation, games, utility, discovery, moderation, analytics, other`. The column is **free text** — no Prisma enum, no CHECK — so what rejects a bad value is a `z.enum` 400 that does not list the members. The mirror exists to turn that into a refusal that **prints the allowed values, before any request**. It can only go stale in the safe direction (the server gaining a category), and the refusal says where the list came from rather than presenting itself as the authority. Bounds mirrored too: tagline 140, description 2000, counted in **runes**.

---

## Tests

`go build` ✅ · `go vet` ✅ · `golangci-lint` 0 issues · **4765 PASS / 0 FAIL / 4 SKIP** (counted from `--- PASS`/`--- FAIL`, no timeout panic). 35 tests for this command specifically (`RUN=35 PASS=35`).

**Red at base:** six pre-existing ledgers went red on the new route and command, and were updated deliberately rather than loosened — `listingRouteCases` + its floor, the spelled op set, the `procname_leak` wording row, `authorControlledFields`/`cliMintedFields` + the change-route count, the `app listing` help-node floor, and the README command-reference row. The rest are specification guards for new code, labelled as such and validated by mutation, not counted as regression coverage.

**Mutation — the harness and its full output are committed under `scripts/mutation/`, so the count can be recounted rather than believed; read the last line of `RESULTS-listing-set-text.txt`, which is the harness's own reconciliation of *defined* against *produced*.** The battery has grown across follow-ups. 🔴 Note the harness's killer *attributions* are approximate (a backward scan for a nearby `_test.go:NN:` cross-attributes across subtests); which tests FAILED is exact. Patterns were validated against the source *before* the run, and the harness reports a stale pattern as `BAD-PATTERN` in the same vocabulary as every other verdict and exits non-zero — a lesson from #488, where a stale pattern plus a too-narrow grep filter hid three mutants that never executed.

| Mutant | Killed by |
|---|---|
| `S1` empty tagline dropped (`omitempty` collapse) | `TestSetTextEmptyStringIsSentAsAnEmptyString` — "sent NO tagline key" |
| `S2` `--clear` sends `""` not `null` | `TestSetTextClearSendsExplicitNull` — "want null" |
| `S3` flag presence inferred from value | `TestSetTextEmptyStringIsSentAsAnEmptyString` |
| `S4` omitted field sent anyway | `TestSetTextSendsOnlyTheFieldsGiven` — "patch carries 2 keys, want 1" |
| `S5` any category accepted | `TestSetTextRejectsAnUnknownCategoryLocally` |
| `S6` category matched case-insensitively | `TestSetTextCategoryCaseMissIsNamed` |
| `S7` empty category allowed through | `TestSetTextEmptyCategoryIsRefusedAndNamesClear` |
| `S8` mirror loses a value | `TestSetTextCategoryAcceptsEveryServerValue` — "holds 6 values, want 7" |
| `S9` empty patch sent | `TestSetTextRefusalsHappenBeforeAnyRequest` |
| `S10` cap off by one | `TestSetTextRefusalsHappenBeforeAnyRequest` |
| `S11` cap counts bytes | `TestSetTextCapsCountRunesNotBytes` — "280 characters" |
| `S12` contradiction silently resolved | `TestSetTextRefusalsHappenBeforeAnyRequest` |
| `S13` `--clear` typo ignored | `TestSetTextRefusalsHappenBeforeAnyRequest` |
| `S14` `requiresReview` ignored | `TestSetTextReportsTheServersBranchNotItsOwnBelief` |
| `S15` warning never fires | `TestSetTextWarnsWhenARevisionIsUnderReview` |
| `S16` warning always fires | `TestSetTextIsSilentWhenNoRevisionIsPending` |
| `S17` empty vs cleared read the same | `TestSetTextSuccessLineDistinguishesEmptyFromCleared` |
| `S18` write addressed with `shadowId` | `TestSetTextSendsOnlyTheFieldsGiven` — `listingId ""` |

`S15`/`S16` are the pair: without both, the warning test is equally true of a build that warns always. Every local refusal asserts the request log is **empty**, so "refused locally" is measured rather than claimed.

---

## Sequencing

**✅ Integrated — this branch now contains `main`, including #488.** It was branched off `origin/main` (`eb80579`) and never stacked on `zach/app-doctor`. #488 merged first (squash `75e0c5186`), and `main` has since been merged in here (`b463ea4`); the branch is **0 commits behind**.

The two PRs collided in `internal/appapi/listing.go`, `listing_op_test.go` and `procname_leak_test.go` — and 🔴 **in two further places git did not flag at all** (`trpcListMine` declared twice in one `var` block; `TestListMineCapMatchesTheServer` defined in two files), plus **two duplicate slice rows that would have compiled and shipped green**, inflating the very counts their floors check. All resolved, floors re-measured empirically on the merged tree, and both sides' hunks verified present by marker. Full detail in the integration comment below.

---

## 🔴 Outstanding: the OAuth-path gap

Every live validation on this PR ran with a **full-scope personal API key**, which `enforceTokenScope` early-returns on before any `requiredScope` check. That proves the procedures, this CLI's decode, and the logic — and says **nothing** about the scoped-token path these PRs exist to open.

The stored config carries no OAuth credential (`auth_kind: token`; `access_token` / `refresh_token` empty), and minting one needs an interactive device-flow login that cannot be completed on someone's behalf.

**To close it — about a minute, read-only, spends no Buzz:**

```bash
civitai login                 # device flow; approve in the browser
civitai whoami                # expect: Credential type: OAuth login
civitai app doctor --json | jq '{ok, summary}'
```

A `403` there means the scope path is still shut and the release reading is wrong; anything else closes the gap.

---

## Live-write validation: PREPARED, deliberately **NOT EXECUTED**

🔴 **No write was ever performed against any listing by this PR.** A procedure for one is documented in a comment on this PR. It is recorded so the steps are not re-derived if anyone ever wants them — **its presence is not evidence it was run.**

It was cancelled on merit rather than deferred. The critical fix it would have exercised was already proven on production **with zero writes**, because `appListings.getMyListingForApp` resolves `shadowId` with a `findFirst` and no create:

| slug | `shadowId` | `hasPendingRevision` | old gate | new gate |
| --- | --- | --- | --- | --- |
| `radio` | `apl_01M08GJT3CS1RHKFX5Y2A5H58C` | false | **silent** | **fires** |
| `comfy` | `apl_01M0GKZ4HHNSQXSS8BBS3FXM1S` | false | **silent** | **fires** |
| `cosmetic-studio` | `null` | false | silent | silent ← negative control |

Both ids match rows read independently from `app_listings`. That is the whole A/B: the old `hasPendingRevision`-only gate is provably silent in exactly the state the warning exists for, the new gate provably fires, and the third row proves the new gate is not simply always-true. A write would only have added message *rendering* — not worth mutating a public store page for.

The one write that **was** run, against an ON-SITE listing, was a refusal by design and is reported in the comments: exit 1, and the listing's tagline md5 verified byte-identical on the primary before and after.

## Observed adjacent gap — an abandoned shadow has no discard path

Not a defect in this PR, and **deliberately not fixed here.** Recorded so it is discoverable rather than folklore.

Measured across the fleet: **7 revision shadows, all `status='draft'`, all untouched (`updated_at = created_at`), all with 0 pending publish requests.** `withdrawExternalRequest` withdraws a *submitted* request; these have none — so all 7 are unreachable by their own authors, with no way to discard them.

Six are harmless by luck about what their parents did afterwards, not by design:

- Five are **on-site**, hence structurally immune — `applyApprovedRevision`'s on-site branch copies only asset columns.
- `comfy`'s shadow matches its parent, so approving it is a no-op.
- 🔴 **`radio` is the exception, and it is exactly one row.** Its shadow was minted 08-17 carrying the parent's values at the time; the parent gained its `tagline` and `category` on 08-20. So the shadow's NULL `tagline` is **stale, not an intentional clear** — `untouched = true` proves nobody edited it. If that shadow were ever submitted and approved, the off-site branch would copy NULL over the parent and **wipe the live tagline**.

That is the concrete hazard this PR's warning exists to surface, and it exists today independently of this PR. Corrective SQL has been surfaced to a human rather than run — writes to the main civitai DB are applied by a person here.


ZacxDev:
### Measured merge interaction with the sibling PR (test-merged, not reasoned about)

I merged `zach/listing-set-text` into `zach/app-doctor` in a throwaway worktree and branched on the **exit code**, not a marker grep (`git merge-tree`/marker-grepping reports a confident "no conflicts" that is wrong):

- **Exit code 1 — one conflicted file, one hunk:** `internal/appapi/listing.go`, the `listingRoute` var block. Both PRs add a route there. Trivial to resolve: keep both route declarations.
- **Auto-merged cleanly:** `README.md`, `internal/appapi/listing_op_test.go`, `internal/appapi/procname_leak_test.go`, `internal/cmd/app_listing.go`, `internal/cmd/app_listing_help_test.go`.

🔴 **But a clean git merge is not a clean merge, and there is a real semantic residual in the auto-merged half.** Both PRs independently raise the same two positive-control floors from **13 → 14** (`listingRouteCases()`'s floor in `TestListingBadRequestSubjectIsReachablePerRoute`, and `len(declared) < 14` in `TestEveryListingRouteHasAReachabilityCase`). Git takes one bump, not both — so after the second merge the tree has **15** listing routes while both floors read **14**.

That does **not** go red: they are floors, not equalities, so the suite stays green with a control that has been silently weakened by one. Whoever merges second should raise both to **15** in the rebase commit.

The exact-set ledger (`TestListingRouteOpsAreExactlyThisSpelledSet`) and the wording table are checked in *both* directions, so those two do fail loudly if a hunk is dropped — they need no manual step.

**Per-hunk markers to grep after rebasing**, one distinctive token from each side:

| side | markers |
|---|---|
| #488 doctor | `appListings.listMine`, `doctorGates`, `"gating"`, `Delisted (status 'removed')` |
| #489 set-text | `appListings.updateListing`, `MarketplaceCategories`, `ClearTagline`, `set-text` |

Then re-run the full matrix on the post-rebase base — a rebase resets the verification gate.

ZacxDev:
### Correction to my earlier sequencing note — re-measured after the audit fixes

My previous comment said both PRs raise the same two floors 13 → 14 and that a merge would silently leave them one below truth. **Both halves are now wrong**, and the second was already wrong when I wrote it. Re-measured by test-merging `zach/listing-set-text` into `zach/app-doctor` off current heads:

**The merge conflicts, and it conflicts loudly.** Exit code 1, one file: `internal/appapi/listing.go`. Both PRs now declare `trpcListMine`, the `MyListing` type and the `ListMyListings` method — #488 for the `doctor` read, #489 for the kind gate. That is a **redeclaration**, so even a naive resolution that kept both sides fails to compile. This is the good failure mode: it cannot be merged wrong in silence.

**Resolution:** keep ONE copy of `trpcListMine` / `MyListing` / `ListMyListings`, taking the **union of the struct fields** — #488 needs `Name`, `Status`, `Role`, `AppBlockID`, `Problems`; #489 needs `Kind`. #488's `MyListing` already carries `Kind` as of its latest commit, so #488's version is the superset and is the one to keep.

**Route floors after both merge — the true count is 15** (13 base + `listMine` + `updateListing`):

| guard | #488 alone | #489 alone | **required after both** |
| --- | --- | --- | --- |
| `listingRouteCases()` floor | 14 | 15 | **15** |
| `declaredListingRoutes()` floor | 14 | 15 | **15** |
| `changeRoutes` | 7 | 8 | **8** |
| `app listing` help nodes | 8 | 9 | **9** |

#489 already carries 15/15/8/9, so **if #489 merges second nothing needs raising**. If #488 merges second, raise both route floors 14 → 15 in the rebase commit. (The audit also caught that #489 had a floor whose *message* said 14 while the *comparison* said 13, and the same shape in the help-node guard — both fixed; the comparison and the message now agree.)

The exact-set ledger (`TestListingRouteOpsAreExactlyThisSpelledSet`) and the wording table are checked in **both** directions, so a dropped hunk there fails loudly with no manual step.

**Per-hunk markers to grep after rebasing:**

| side | markers |
| --- | --- |
| #488 | `doctorGates`, `"gating"`, `Delisted (status 'removed')`, `listingIsOnsite` |
| #489 | `appListings.updateListing`, `MarketplaceCategories`, `ClearTagline`, `refuseOnsiteTextEdit` |

Then re-run the full matrix on the post-rebase base — a rebase resets the verification gate.

ZacxDev:
### ✅ The release-cut blocker in this PR's description is now STALE — the cut happened

The PR body above opens with "DO NOT MERGE — the server half is not serving". **That is no longer true.** Verified independently, by content and at the cluster:

- `origin/release` is now `0c84428da3` ("5.1.42") and carries **16** `.meta(listingMediaCliScope)` sites — up from 13. `listMine` is at `:1026` **with** the meta (it was `:914` bare); `updateListing` `:679` and `updateRevisionDraft` `:750` are annotated too.
- **Control**: `getAssets` is still `protectedProcedure` on the same ref, so the grep is discriminating rather than matching everything.
- `#4341` (`999f9363a8`) is an ancestor of `origin/release`.
- dp-prod is **serving** it: `civitai-dp-prod-primary` and `-api-primary` both run `ghcr.io/civitai/civitai-prod:20260825001134-0c84428` — short sha matches the release tip.

**Still not merging** — that is now waiting on explicit approval, not on the server.

### Live read against production

`civitai app doctor --json` against `https://civitai.com`, 21 real listings:

```
exit 0
ok=True  summary={'apps': 21, 'blocking': 11, 'advisory': 49, 'gating': 0, 'delisted': 11}
by status: {'approved': 10, 'removed': 11}
```

Two behaviours confirmed on real data rather than fixtures:

- **The delisted rule works.** All 11 blocking problems sit on the 11 `removed` listings, so `gating: 0` and the command exits **0**. Before that fix this exact account exited **1**, permanently, for problems nobody could act on.
- **Both kind arms fired.** `radio` (off-site) → `edit the listing in the browser: https://civitai.com/apps/listing/apl_…/edit`; `panorama-360` (on-site) → `these fields come from block.manifest.json on an on-site app — edit name / tagline / description there, then civitai app submit`.

🔴 **What this does NOT establish.** It ran with a **full-scope personal API key**, which `enforceTokenScope` early-returns on before any `requiredScope` check — so it exercises the proc and this CLI's decode, and says **nothing** about the OAuth-token path that #4341 opened. The stored config carries no OAuth credential (`auth_kind: token`; `access_token`/`refresh_token` empty), and minting one needs an interactive device-flow login I cannot complete on someone's behalf.

**To close that gap** (about a minute, read-only, no Buzz):

```bash
civitai login                 # device flow; approve in the browser
civitai whoami                # expect: Credential type: OAuth login
civitai app doctor --json | jq '{ok, summary}'
```

A `403` there would mean the scope path is still shut and I have mis-read the release; anything else closes it.

ZacxDev:
### Merge interaction — re-measured at the FINAL heads (488 `bc388c3`, 489 `9dc77ee`)

This supersedes both of my earlier sequencing notes. It also reconciles the disagreement about whether #489 declares the `listMine` trio — **we were each right about a different head**:

| head | `trpcListMine` | `type MyListing` | `ListMyListings` |
| --- | --- | --- | --- |
| `b059372` — #489 as audited | 0 | 0 | 0 |
| `9dc77ee` — #489 now, after the kind-gate fix | 2 | 1 | 1 |

The kind gate (audit finding 2) needs an authoritative `kind`, and `getMyListingForApp` does not return one — so #489 had to read `listMine` too. That is what created the overlap after the audit was taken.

**Test-merge, branching on the exit code (never a marker grep):**

```
merge exit 1
  internal/appapi/listing.go            2 conflict hunks
  internal/appapi/listing_op_test.go    6 conflict hunks
auto-merged: README.md, procname_leak_test.go, app_listing.go, app_listing_help_test.go
```

`listing.go` is a **redeclaration** — both branches define `trpcListMine`, `MyListing` and `ListMyListings` — so even a resolution that naively keeps both sides fails to compile. That is the good failure mode.

**Resolution:** keep ONE copy of each. #488's `MyListing` is the superset (`Name`, `Status`, `Role`, `AppBlockID`, `Problems`, `Kind`); #489 needs only `Slug` + `Kind`, so take #488's. Keep both route declarations (`listMine` once, `updateListing` once).

**Required floors after both land — true count is 15 routes** (13 base + `listMine` + `updateListing`):

| guard | #488 alone | #489 alone | **after both** |
| --- | --- | --- | --- |
| `listingRouteCases()` | 14 | 15 | **15** |
| `declaredListingRoutes()` | 14 | 15 | **15** |
| `changeRoutes` | 7 | 8 | **8** |
| `app listing` help nodes | 8 | 9 | **9** |

#489 already carries 15/15/8/9. **If #489 merges second, nothing needs raising.** If #488 merges second, raise both route floors 14 → 15.

**Per-hunk markers to grep after rebasing** — the four auto-merged files still need this, since a clean git merge is not a clean merge:

| side | markers |
| --- | --- |
| #488 | `doctorGates`, `"gating"`, `Delisted (status 'removed')`, `listingIsOnsite`, `blockedMediaKind`, `ListMineCap` |
| #489 | `appListings.updateListing`, `MarketplaceCategories`, `ClearTagline`, `refuseOnsiteTextEdit` |

Then re-run the full matrix on the post-rebase base — a rebase resets the verification gate.

ZacxDev:
### Live validation — what is already PROVEN (zero writes), and the write plan awaiting a named listing

#### 1. The onsite refusal, run against production ✅

`model-benchmarking` is a real ON-SITE approved listing. Under my credential:

```
$ civitai app listing set-text --slug model-benchmarking --tagline "SHOULD NEVER BE WRITTEN"
EXIT=1
Error: "model-benchmarking" is an ON-SITE app, and its tagline, description and category come from
block.manifest.json — editing them here would be overwritten by the manifest at your next approved
version. Edit `name` / `tagline` / `description` in block.manifest.json and run `civitai app submit`.
(Category is set by a moderator on an on-site app.)
```

🔴 **Nothing was written — verified out-of-band against the primary, not inferred from the error.**

| | tagline | md5 | `updated_at` |
| --- | --- | --- | --- |
| before | `Settle it with a grid.` | `f3c23d16346f045392790049501277ef` | 2026-08-22 22:34:24+00 |
| after | `Settle it with a grid.` | `f3c23d16346f045392790049501277ef` | 2026-08-22 22:34:24+00 |

`updated_at` predates the command by two days. The listing's **shadow** was checked too and is likewise unchanged.

**Exit code**: `1`, not `2` — deliberate. Every flag and the slug are well-formed; what is not editable is the SUBJECT, which the contract publishes under `1` beside an invalid manifest. That was **unpinned and undocumented** when I ran it, so I have since added `ErrOnsiteTextNotEditable`, a classification test with a positive control, a `--help` line, and a mutant (`S34`) that strips the sentinel.

And the loop closes correctly on the same real listing:

```
$ civitai app doctor model-benchmarking
model-benchmarking  (approved, owner)
  ADVISORY (recommended, nothing is held up)
    empty-description  Missing description
      Fix: these fields come from block.manifest.json on an on-site app — edit `name` / `tagline` /
           `description` there, then `civitai app submit` a new version
```

Exactly one advisory, and the manifest arm — matching the DB, where `tagline` is set, `category` is `generation`, and `description` is NULL.

#### 2. 🔴 The CRITICAL fix is ALREADY PROVEN on live data — and it needed no write

The overwrite warning's premise is checkable with a **pure read**. `appListings.getMyListingForApp` under my token, against production:

| slug | `shadowId` | `hasPendingRevision` |
| --- | --- | --- |
| `radio` | `apl_01M08GJT3CS1RHKFX5Y2A5H58C` | **false** |
| `comfy` | `apl_01M0GKZ4HHNSQXSS8BBS3FXM1S` | **false** |
| `cosmetic-studio` | `null` | false |

Both shadow ids match the ones I read independently from `app_listings` via the primary. So:

- **Old gate** (`hasPendingRevision` alone) → false on `radio` and `comfy` → **provably silent**, in the exact state the warning exists for.
- **New gate** (`ShadowID != nil || HasPendingRevision`) → **provably fires** on both.
- **`cosmetic-studio` is the negative control**: no shadow, no pending request → *neither* gate fires, so the new gate is not simply "always true".

That is the defect the fakes structurally could not see, demonstrated end-to-end on production **without touching a single listing**. A write would only additionally confirm message *rendering*.

#### 3. Two corrections to the relayed state, both measured

- **`model-benchmarking` ALSO has an open shadow** — `apl_01M0TKRSJAQR3BC48NHZDDQH17`, `draft`, 0 pending — where the table showed "—". Three listings have open shadows, not two.
- **The relayed shadow ids were mangled**: `rev-01M0GKZ4HHNSQXSS8BBS3FXM1T` vs the real `apl_01M0GKZ4HHNSQXSS8BBS3FXM1S` (different prefix, different final char). Mine come from the DB and are corroborated by the wire.
- Minor: `model-benchmarking`'s description is **NULL**, not length 0.

#### 4. 🔴 A pre-existing landmine on `radio`, unrelated to my change

`radio`'s parent tagline is `Endless stations. No reruns.` — its **open shadow's tagline is NULL**. If that shadow is ever submitted and approved, `applyApprovedRevision`'s offsite branch copies NULL over the parent and **wipes the live tagline**. That exists today, with or without these PRs. It is also the sharpest possible justification for the warning.

`comfy`'s shadow tagline equals its parent's, so approving it is a tagline no-op.

#### 5. The write plan — PREPARED, NOT RUN

Awaiting a named listing. My recommendation: **`cosmetic-studio`** for a first write (no shadow → clean, isolates the write path), or **`comfy`** to exercise the warning (its shadow tagline matches the parent, so an approval later is a no-op — unlike `radio`, where an approval wipes the tagline regardless).

**Restore values captured FIRST, from the DB, so recovery does not depend on the write succeeding:**

| slug | current tagline (exact) | md5 |
| --- | --- | --- |
| `comfy` | `Run ComfyUI on our GPUs, with the model library built in.` | `fe51957ca1d7133fcfae7bc6227733cd` |
| `radio` | `Endless stations. No reruns.` | `f71425fffacbbf5d36bbdba59cef1976` |
| `cosmetic-studio` | `Design and generate badges, avatar frames, and profile backgrounds for the Civitai Creator Shop — powered by your own Buzz.` | `0574f4c290cacf87d892c6311095fee4` |
| `vitrine` | `generate studio quality product images` | `c8316dd0e2da77b90a5257698110c921` |

**Sequence** (example for `comfy`):

```bash
# 0. BEFORE — re-read at the moment of acting, not from this table
psql … -c "SELECT tagline, md5(coalesce(tagline,'')) FROM app_listings
           WHERE slug='comfy' AND revision_of_id IS NULL;"

# 1. WRITE — a marked, obviously-temporary value
civitai app listing set-text --slug comfy \
  --tagline "Run ComfyUI on our GPUs, with the model library built in. [cli-test]" --json

# 2. ASSERT (see below)

# 3. RESTORE — verbatim, from step 0's read
civitai app listing set-text --slug comfy \
  --tagline "Run ComfyUI on our GPUs, with the model library built in."

# 4. CONFIRM restored: md5 back to fe51957ca1d7133fcfae7bc6227733cd
```

Restore does not depend on step 1 succeeding: the value is captured before, and step 3 is a plain set of a known-good string. If step 1 half-applies, step 3 still converges.

**What proves the warning fired for the SHADOW reason, not incidentally** — all four, or it does not count:

1. stderr contains **`OPEN revision draft that has not been submitted yet`** — the `else` branch. The other wording (`already under moderator review`) would mean the `HasPendingRevision` branch fired instead.
2. stderr names **`apl_01M0GKZ4HHNSQXSS8BBS3FXM1S`** — the exact shadow id read out-of-band from the DB. The CLI can only print it by decoding `shadowId` off the wire.
3. `--json` reports **`"openRevision": true`** with **`"shadowId"`** equal to that same id, and `"requiresReview": false` (a text-only patch is not material).
4. The negative control in the same session: the same command against **`cosmetic-studio`** must print **no** warning and `"openRevision": false`.

#### 6. Still open — the OAuth path

Every live run above used a **full-scope personal API key**, which `enforceTokenScope` early-returns on. It proves the procs, the decode and the logic; it proves **nothing** about the scoped token path these PRs exist to open. Device-flow steps remain on both PR bodies.

ZacxDev:
### Final mutation reconciliation — read from the artifacts committed in this PR

Both numbers below are re-derived from `scripts/mutation/` **on the pushed branch**, not from a transcript. `defined` counts mutant tuples in the harness source; `verdicts` counts per-mutant result lines in the committed output. They agree, which is the property that was missing when a stale pattern once let three mutants read as "absent" rather than "never ran".

| | defined | verdicts | killed | survived | build-fail | not-run |
| --- | --- | --- | --- | --- | --- | --- |
| **#488** `app doctor` | 39 | 39 | 39 | 0 | 0 | 0 |
| **#489** `listing set-text` | 32 | 32 | 32 | 0 | 0 | 0 |

Baselines were green before either sweep (`#488` 3063 PASS / 0 FAIL; `#489` 2944 PASS / 0 FAIL, `built=True`) — a sweep against a red baseline is meaningless and the harness now exits `3` rather than `1` to say so distinctly.

**One `⚠ TRUNCATED RUN` is recorded in #488, deliberately left in.** `M2-severity-compare` produced 798 verdicts against a 3064 baseline with one panic. It was legitimately KILLED, but by a run that stopped a quarter of the way in — so it is printed as **unattributed** rather than counted as ordinary evidence. That flag exists because the audit found the harness could not previously tell "31 tests failed" from "the suite died early".

### What the harness can and cannot substantiate

- **Which tests failed under a mutant: exact.**
- 🔴 **The reason printed beside each killer: approximate.** Attribution scans backward from a `--- FAIL` header for a nearby `_test.go:NN:`, which cross-attributes across subtests. Earlier descriptions on these PRs claimed each mutant was "killed by an assertion naming the specific defect"; the harness cannot support that, and both bodies now say so.

### Suites, at the pushed heads

`go build` ✅ · `go vet` ✅ · `golangci-lint` 0 issues on both.
**#488: 4812 PASS / 0 FAIL / 4 SKIP.** **#489: 4792 PASS / 0 FAIL / 4 SKIP.** No timeout panics.

### Status

Both PRs are **12/12 green, `mergeStateStatus=CLEAN`, and unmerged.** The release cut, green CI and a clean sweep are prerequisites — none of them is approval, and none is being read as such.

ZacxDev:
### Merge duplicates — re-measured at the CURRENT heads (this supersedes my earlier notes and refines the audit's table)

The audit's table was measured before #488's delta landed. Re-run just now against `origin/zach/app-doctor` + `origin/zach/listing-set-text`:

| site | git | what a merger must do |
| --- | --- | --- |
| `internal/appapi/listing.go` — var block + `MyListing` + `ListMyListings` | 🔴 **CONFLICT** | keep one copy of each; `redeclared` otherwise |
| `internal/appapi/listing_op_test.go` — ledger map **and** cases slice | 🔴 **CONFLICT** (both sites, now) | dedupe the `listMine` ledger entry and the `listMine` case row |
| `internal/appapi/procname_leak_test.go` — wording rows | **clean, and WRONG** | auto-merges to **2** `route: trpcListMine` rows; must be deduped by hand |

**Correction to the audit on one row:** it listed the `listing_op_test.go` cases slice as auto-merging clean and shipping green. At the current heads that file **conflicts**, so a human does see both sites. The genuinely silent duplicate is now `procname_leak_test.go` — measured: `grep -c 'route:   trpcListMine'` returns **2** after the auto-merge.

The audit's central warning stands and is the thing to carry forward: **git flags fewer sites than exist, and the count changes as the branches move.** Do not infer the set from a previous measurement — re-run the test-merge at the heads you are actually merging.

**Deduped floors after both land** (unchanged from the audit's measurement, and #489 already carries them): `cases` **15**, `declared` **15**, `help checked` **9**, `doctor seen` — now **12**, not 7, because #488's fix-scan was widened to walk the three `blocked-media` per-slot arms as well as the eight codes.

**`MyListing`'s field set is a human decision**: #488's is the superset (`AppListingID`, `Slug`, `Name`, `Status`, `Role`, `AppBlockID`, `Kind`, `Problems`); #489 needs only `Slug` + `Kind`. Take #488's.

**Near-miss worth recording** (the audit's, and it is a good one): #488's fix-scan negative control asserts that `civitai app listing set-banner` does **not** resolve. #489 adds `set-text`. Had the control been named `set-text`, the merge would have silently disarmed it — a negative control that stops being negative reports success forever.

ZacxDev:
### Integrated with `main` (post-#488) — five collisions, two of which git did not flag

**Merge, not rebase**, stated because it is a decision: this branch is already pushed and this PR points at it, so a rebase needs a force-push; five commits would each replay against the same colliding files; and one resolution commit is readable as a diff where a replayed resolution is not. `main` squashes on merge, so the merge commit never reaches its history.

Merge commit `b463ea4`. The branch is now **0 commits behind `main`**.

#### Git flagged three files. There were five collisions.

| # | site | git | resolution |
| --- | --- | --- | --- |
| 1 | `internal/appapi/listing.go` — `MyListing` / `ListMineCap` / `ListMyListings` | flagged | **dropped mine**, consumed #488's 8-field superset (it carries `Kind`, which this command's gate reads). Everything set-text-only survives beside it. |
| 2 | `internal/appapi/listing_op_test.go` — floors | flagged | took 15 |
| 3 | `internal/appapi/procname_leak_test.go` — wording prose | flagged | took `main`'s |
| 4 | 🔴 `trpcListMine` declared **twice in one `var` block** | **not flagged** | both branches added the route; git merged both lines. Caught only by the compiler. My copy had drifted into the CHANGES section — a pure read filed among writes. |
| 5 | 🔴 `TestListMineCapMatchesTheServer` defined in **two files** | **not flagged** (different files) | both branches wrote the same drift guard for the same constant. One survives. |

**And two duplicates that would have compiled and shipped green.** A duplicate map key is a hard error, so the spelled-op map announced itself. A duplicate **slice row** is not: `listingRouteCases()` carried two `listMine` cases and `procname_leak_test.go` two `trpcListMine` wording rows — each inflating the very count its floor checks, so the positive control would pass vacuously against a table one row short of what it claims to drive. Deduped; survivors keep the stricter `notWant` that bans the retired *"check the app you named"* clause.

#### Floors — measured on the merged tree, not trusted

Each armed at `999` and the count read from the test's **own failure message**:

| guard | measured | floor was | action |
| --- | --- | --- | --- |
| `listingRouteCases()` | **15** | 15 | ✓ |
| `declaredListingRoutes()` | **15** | 15 | ✓ |
| `app listing` help nodes / bodies | **9 / 9** | 9 | ✓ |
| doctor fix-scan | **14** | **12** | 🔴 **tightened to 14** |

The fourth was wrong. At 12 against a real 14, the `blocked-media` arms could lose two commands with the control still passing — slack in a positive control, which is the exact criticism that had already raised that floor from 5. It recurred because I widened the scan and did not re-measure the floor.

#### Two comments the merge made false, and one that never landed

- `MyListing`'s doc listed `kind` among the fields **not** decoded — directly above the field itself.
- The merge-order note is reworded from an unmet blocker to a **satisfied dependency**, naming squash `75e0c5186`. The warning it carries is kept: lifting this command into a tree without `app doctor` re-creates the defect (three shipped surfaces name it, including the fail-closed refusal).
- 🔴 That note had **never actually landed** — an earlier batch aborted mid-file on a failed anchor and wrote nothing. Found only by grepping for it rather than assuming the commit contained it.

#### Verification on the merged tree

Per-hunk survival checked by grepping a distinctive marker from each side — **9 from #488, 8 from #489, all present**.

```
go build ✅   go vet ✅   golangci-lint 0 issues
4891 PASS / 0 FAIL / 4 SKIP / 0 panics   (21 packages ok)

doctor    BASELINE 3142 PASS / 0 FAIL
          defined=45 ran=45 killed=45 survived=0 build_fail=0 not_run=0 truncated=1
set-text  BASELINE 3027 PASS / 0 FAIL / built=True
          defined=38 ran=38 killed=38 survived=0 build_fail=0 not_run=0 truncated=0
```

Both `defined` counts reconcile against the per-mutant verdict lines in the committed artifacts (45 and 38), and **both batteries were checked for stale search patterns against the merged source before running** — a mutant whose text moved reports NOT-RUN and is evidence of nothing.

**Both `RESULTS-*.txt` were re-derived on the merged tree** (`26b6898`). Carrying either across would have shipped a file describing a tree that no longer exists — the defect `ad6e435` shipped once already.

The lone `truncated=1` is `M2-severity-compare`, unchanged and deliberate: it panics part-way, so it is tagged `KILLED (TRUNCATED)` rather than counted as ordinary evidence.

#### Standing

The **OAuth device-flow gap** remains the one unclosed claim and is documented in the body above. I have not merged this and will not.

PR #498, opened 2026-08-26
feat(whoami): --json carries the account profile, and still withholds the PII

### LINKED ISSUE #377: whoami drops six fields the server sends, and its --json help calls a curated projection "raw JSON"
Found while diffing what the server *sends* against what the structs *model*, sweeping for the #343/#346/#367 class.

## The evidence

`internal/appapi/api_test.go` holds the one **real production capture** in either `internal/appapi` or `pkg/civitai` — kept verbatim as a regression fixture because it hard-failed `civitai whoami` in production:

```json
{"id":8753561,"username":"…","tier":"silver","status":"active","isMember":true,
 "subscriptions":["yellow"],"email":"…","emailVerified":true,
 "tokenScope":33554431,"buzzLimit":[…],"subject":{"type":"apiKey","id":96633526}}
```

`appapi.Identity` models `username`, `id`, `tokenScope`, `buzzLimit`, `subject`. **Six keys the server demonstrably sends have nowhere to land**: `tier`, `status`, `isMember`, `subscriptions`, `email`, `emailVerified`.

They are dropped **twice**, which is what makes this different from the usual deliberate subset:

1. `WhoAmI` returns `*Identity` only — **no `json.RawMessage` sibling**. Every other read path in this codebase returns the raw payload alongside the decoded struct precisely so unmodelled keys survive (`GetWorkflow`, `QueryWorkflows`, `WhatIfFromGraph`, `GetMyAppAnalytics` — each with the same comment: *"so a `--json` surface can pass through server fields this struct does not model"*). `WhoAmI` is the exception, so the bytes are gone at the client boundary.
2. `whoami --json` then **hand-builds** its output map rather than passing anything through.

## Why it matters

`--json`'s own help says:

```
civitai whoami --json     # raw JSON (scriptable)
```

and the flag registration says `emit raw JSON (scriptable)`. It is not raw — it is a curated projection (`username`, `id`, `base_url`, `credentialType`, `scopesKnown`, `canReadBalance`, `canSpend`, `canSubmitApps`, `scopes`, `capabilities`). A script author who reads "raw" and expects the server's object gets a different one, with no signal. (README is accurate here and describes the projection correctly; it is the command's own help that overstates.)

`isMember` / `tier` / `subscriptions` are not idle either — item 13 records that *"a caller's usable (non-disabled, non-memberOnly) ecosystem set"* means even the default generation model differs between a free and a member account. The CLI is currently throwing away the server's own answer to "is this a member account".

## 🔴 The obvious fix is the wrong one

**Do not just pass the raw body through.** `email` and `emailVerified` are PII that `whoami --json` does not print today; making the output "raw" would newly emit a user's email address into whatever a script redirects it to. That is a privacy regression dressed as a bug fix.

Two defensible options, both small:

- **(a) Fix the words.** Change the help/example from "raw JSON" to what it actually is — a stable, curated identity object. Cheapest, no behaviour change, closes the false claim.
- **(b) Model what is useful, keep the projection.** Add `tier`, `status`, `isMember`, `subscriptions` to `Identity` and to the `--json` map; deliberately leave `email`/`emailVerified` unmodelled and say so in the struct comment. Then (a) as well, since the output still is not raw.

Worth deciding which; (a) is required either way.

## Method note

This was only findable because that one fixture is a **real capture**. Every other payload in `internal/appapi` and `pkg/civitai` is a hand-written mirror of the struct it feeds, so by construction it cannot reveal a field the struct is missing — the same blind spot that hid #367. `pkg/civitai/apps.go` claims its DTOs *"MIRROR the backend public DTOs field-for-field … the field-mapping tests are the guard"*, but that guard is a hand-written fixture and can only prove the structs parse what someone typed from the same structs.

ZacxDev:
Closes #377 — this is option **(b)**, the half left open when #492 shipped option (a). (Option (a) is the `--json` help no longer calling a curated projection "raw"; it merged in v0.1.101.)

## What changed

`appapi.Identity` now models the four fields the server demonstrably sends and the CLI was dropping at the client boundary, and `whoami --json` publishes them:

```
tier   status   isMember   subscriptions
```

`isMember` is the one that is not idle. AGENTS item 13 records that a caller's usable (non-disabled, non-memberOnly) ecosystem set differs between a free and a member account — so the CLI was discarding the server's own answer to *"is this a member account"*, which is what predicts whether `civitai generate`'s defaults are even available.

## Three decisions that each look like an oversight

**`email` / `emailVerified` are STILL not modelled, on purpose.** They are PII `whoami` does not print, so the obvious fix — pass the raw body through — is a privacy regression dressed as a bug fix, which #377 rejected explicitly. Dropping them at the **client** boundary means no future `--json` surface can publish them by adding one map entry. Guarded at both levels, because the two fail to different changes: `TestIdentityHasNoEmailField` (structural, on `appapi.Identity`) and `TestWhoAmINeverPrintsEmail` (behavioural, on both output surfaces). Each carries a positive control — "printed no email address" is exactly what a command that never had one also does.

**All four are pointers / RawMessage, so an omitted field stays `null`** rather than zero-filling to `""` / `false`. Same tri-state rule as `canSubmitApps`: publishing a fabricated value as if the server had said so is the false-negative-stated-as-fact this repo keeps paying for.

**`Subscriptions` is `json.RawMessage` while its three siblings are typed.** It is the one *composite* among the four, so it is the one that can drift shape — and this exact endpoint has already done that once (`buzzLimit`: bare number → array of window objects, which hard-failed `whoami` in production). Typed, a drift would fail the strict parse, fall back to `parseCoreIdentity`, and blank `tier`/`status`/`isMember` as collateral. Held raw, the blast radius is zero. `TestWhoAmIProfileDriftIsIsolated` pins the **collateral**, not the parse.

## Tests

- **`TestWhoAmIJSONShapeIsPinnedWhole` grew cases F and G.** The five existing bodies carry no profile fields at all, so they pin all four as `null` — a literal that a payload never reading `id.Tier` satisfies equally well. F and G carry them **pairwise distinct** (silver/free, active/muted, true/false, `["yellow"]`/`[]`) so cross-assignment mutants disagree somewhere; F additionally carries the PII **in the server body** while pinning a payload without it.
- **`TestREADMEWhoAmIJSONBlockHasTheRealKeys` closes a seam nothing covered.** `whoami_readme_block_test.go` anchors on `Logged in as`, which the `--json` example does not contain — so that block could have drifted arbitrarily far with every existing guard green, and adding keys to one and not the other is exactly this PR's shape. It pins the **key set** (the README block is hand-formatted; `json.Encoder` is not) and controls its own extractor count.
- **`TestWhoAmIProfileFieldsReachTheHumanSurfaceNever`** pins the deliberate NON-change: the human output is a capability report, not an account dump.
- Four new `internal/appapi` tests, all driven off `liveMeBody` — the one **real production capture** in the repo. #377 was only findable because that fixture is a real response rather than a hand-written mirror of the struct it feeds, so asserting against a local literal would re-create the blind spot.

## Verification

**Red at base / green at HEAD.** With only the two non-test sources reverted to `origin/main`: `TestWhoAmIJSONShapeIsPinnedWhole` red in **all 7** subcases and `TestREADMEWhoAmIJSONBlockHasTheRealKeys` red. (`internal/appapi`'s four new tests build-fail at base rather than failing behaviourally — they reference struct fields that do not exist there. Stated as the weaker signal it is.)

**Mutation battery — 7 applied mutants, 7 killed**, each confirmed to fail the *named* guard with that guard's own message:

| mutant | killed by |
|---|---|
| drop `"tier"` from the payload | `…PinnedWhole` |
| drop `"tier"` (vs the README seam) | `…READMEWhoAmIJSONBlockHasTheRealKeys` |
| drop `"subscriptions"` | `…PinnedWhole` |
| `"tier": id.Status` (cross-wire) | `…PinnedWhole` |
| `"isMember"` derived from `id.Tier != nil` | `…PinnedWhole` |
| zero-fill the nulls in the payload | `…PinnedWhole` (cases A–E) |
| `Subscriptions` typed `[]string` | `…ProfileDriftIsIsolated` |
| model `Email` on `Identity` | `TestIdentityHasNoEmailField` |
| model **and publish** `email` | `TestWhoAmINeverPrintsEmail` |
| print `Tier:` on the human surface | `…ProfileFieldsReachTheHumanSurfaceNever` |
| drop `"tier"` from the README block | `…READMEWhoAmIJSONBlockHasTheRealKeys` |

The first run of that battery reported **five false SURVIVED verdicts** — all harness bugs (a `\t{4}` pattern against 5-tab indentation), none a real finding. The battery now gates every mutant on `git diff --quiet` proving the edit landed, plus a "did any test actually run" control.

Instructive result: the zero-fill mutant is killed by cases **A–E** (profile absent) and *survives* F and G. Both halves of the matrix are load-bearing in opposite directions.

**Live end-to-end**, built binary against a fake `/api/v1/me` serving the real production capture verbatim: the four fields land with their real values, and grepping both surfaces for `zachlowden1@gmail.com` / `emailVerified` returns **0** — with a positive control (`curl` the same endpoint → **1**) proving the server really was sending it, so that zero is a measurement and not a wiring failure. Second measurement point, a body with no profile fields: all four emit `null`.

**Gates:** `make ci` green (22 packages), `make lint` green (0 issues, golangci-lint 2.12.2 — it is a separate CI job and `make ci` does not run it), `gofmt -s -l` clean over 1196 files, and `./scripts/ci-shallow.sh` **21/21 ok, 0 failures, 0 timeouts** in the depth-1 clone CI actually uses.

## README

Updated on all three surfaces that state this contract and each go stale alone: the command-table row, the `whoami --json` section prose (the "ten keys" claim and the list of what is dropped), and the example block itself — which is now guarded, so it cannot go stale silently again.

## Judgement call worth a second opinion

I did **not** add an `AGENTS.md` item for "email/emailVerified are unmodelled on purpose". The rationale sits in a 🔴 block at the exact struct field an editor would have to touch to add them, backed by two tests that go red — and `AGENTS.md` has only ~950 bytes of headroom, which the handoff flags as the next thing to squeeze. Happy to add item 33 + a `claudedocs/decisions/` file if you'd rather it be routable from the index; it would cost roughly 310 of those bytes.

https://claude.ai/code/session_013pLmUg28zt1XPYw885HJ58

### DIFF
--- a/README.md
+++ b/README.md
@@ -310,7 +310,7 @@ README. For the end-to-end walkthrough, see
 | Command | What it does |
 | --- | --- |
 | `civitai login [--scopes <set>] [--token [<t>]] [--no-browser]` | Browser OAuth device login by default (stores auto-refreshing tokens). The default scope set grants identity + Apps submit + dev-tunnel and **not** Buzz-spend; `--scopes generate` additively grants generation + Buzz **spend** (needed by `civitai generate` and money-path `dev:live`). `--token <t>` stores a personal API key instead (not combinable with `--scopes`). `--token` with **no value** prints where to create a personal key (`civitai.com/user/account`) and how to re-run — handy when you know you want a personal key but haven't minted one yet. Config at `~/.config/civitai/config.yaml`, 0600. Also reads `CIVITAI_TOKEN`. |
-| `civitai whoami [--scopes] [--json]` | Verify the stored token; print the authenticated user, **a `Credential:` section** naming the credential type (**OAuth login** vs **personal API key**), and **a `Capabilities:` section** of three rows — **Read Buzz balance**, **Spend Buzz**, and **Submit Apps** — decoded from the token's scope, so a money-path dead end (a default OAuth login can't spend) is visible before `dev:live` — and when it can't, the output names the fix for that credential (`login --scopes generate` for an OAuth login, a full-scope key otherwise). **Submit Apps is tri-state**: `yes` / `no` / **`unknown`**, because a personal key is never scope-gated for submit while an OAuth token's answer *is* the scope bit — so an absent mask makes it unknowable, and `unknown` must never be read as `no`. `--scopes` also lists every granted scope; `--json` emits a **curated, not raw** identity object — `username`/`id`/`base_url`/`credentialType`/`scopesKnown`/`canReadBalance`/`canSpend`/`canSubmitApps` (`true`/`false`/**`null`**)/`scopes`/`capabilities` (scriptable). See [What `civitai whoami` reports](#submit--auth). |
+| `civitai whoami [--scopes] [--json]` | Verify the stored token; print the authenticated user, **a `Credential:` section** naming the credential type (**OAuth login** vs **personal API key**), and **a `Capabilities:` section** of three rows — **Read Buzz balance**, **Spend Buzz**, and **Submit Apps** — decoded from the token's scope, so a money-path dead end (a default OAuth login can't spend) is visible before `dev:live` — and when it can't, the output names the fix for that credential (`login --scopes generate` for an OAuth login, a full-scope key otherwise). **Submit Apps is tri-state**: `yes` / `no` / **`unknown`**, because a personal key is never scope-gated for submit while an OAuth token's answer *is* the scope bit — so an absent mask makes it unknowable, and `unknown` must never be read as `no`. `--scopes` also lists every granted scope; `--json` emits a **curated, not raw** identity object — `username`/`id`/`base_url`/`credentialType`/`scopesKnown`/`canReadBalance`/`canSpend`/`canSubmitApps` (`true`/`false`/**`null`**)/`scopes`/`capabilities`, plus the account profile `tier`/`status`/`isMember`/`subscriptions` (each **`null`** when the server did not report it, never a fabricated `""`/`false`/`[]`). `email`/`emailVerified` are **withheld on purpose** — they are PII this command does not print (scriptable). See [What `civitai whoami` reports](#submit--auth). |
 | `civitai buzz [--json]` | Show your spendable Buzz balance (**blue / green / yellow**, plus a **total**). Needs the BuzzRead scope — a full-scope personal API key or `civitai login --scopes generate`; a **default** OAuth login token can't read it, and gets a clear message naming both fixes. `--json` emits `{blue,green,yellow,total}` (scriptable — handy for before/after diffing a `dev:live` spend). |
 | `civitai app list [--kind <k>] [--category <c>] [--sort <s>] [--limit <n>] [--cursor <c>] [--json]` | **Discover published Apps in the store** (`GET /api/v1/apps`) — filter-based discovery, not free-text search. **Needs a credential** (`civitai login` or `CIVITAI_TOKEN`): the endpoint keys the visible catalog off your identity, so this is *not* one of the anonymous reads. Cursor-paged. See [Browse the App store](#browse-the-app-store). |
 | `civitai app view <slug> [--json]` | **Show one published App's store detail** (`GET /api/v1/apps/{slug}`) — description, category, rating, gallery, live/external target. **Needs a credential**, same as `app list`. Reads the *public store catalog*, which is a different resource from your own deploy — a not-found here says nothing about `<slug>.civit.ai`. See [Browse the App store](#browse-the-app-store). |
@@ -1274,12 +1274,13 @@ server-side and are not visible to the CLI.
 ##### `whoami --json`
 
 🔴 **`--json` is a stable, *curated* identity object — it is not the server's
-raw `/api/v1/me` body.** It is a hand-built projection of ten keys; the server
-sends more (`tier`, `status`, `isMember`, `subscriptions`, `email`,
-`emailVerified`) that deliberately never appear. `email` / `emailVerified` in
-particular are PII this command does not print, so passing the body through
-would be a privacy regression, not a fix
-([#377](https://github.com/civitai/cli/issues/377)).
+raw `/api/v1/me` body.** It is a hand-built projection of fourteen keys, and the
+two the server sends that never appear are **`email` and `emailVerified`**.
+Those are PII this command does not print, so passing the body through would be
+a privacy regression, not a fix
+([#377](https://github.com/civitai/cli/issues/377)). The gap between "raw" and
+"curated" is now exactly that PII — which is why the CLI will not call this
+output raw.
 
 ```json
 {
@@ -1290,13 +1291,26 @@ would be a privacy regression, not a fix
   "capabilities": { "can_read_buzz": true, "can_spend_buzz": false },
   "credentialType": "personal API key",
   "id": 1,
+  "isMember": true,
   "scopes": ["UserRead", "BuzzRead"],
   "scopesKnown": true,
+  "status": "active",
+  "subscriptions": ["yellow"],
+  "tier": "silver",
   "username": "zach"
 }
 ```
 
-Two fields carry a **third state a script must branch on**, exactly as
+**The account profile — `tier`, `status`, `isMember`, `subscriptions` — is
+`null` when the server did not report it, never `""` / `false` / `[]`.** The
+CLI will not fabricate a value the server did not send, so a script must test
+for `null` before reading any of the four. `subscriptions` is passed through
+**verbatim**, so treat its element shape as server-owned rather than pinned by
+this CLI. `isMember` is the one worth branching on: a member and a free account
+do not see the same usable generation ecosystems, so it predicts whether
+`civitai generate`'s defaults are even available to this credential.
+
+Two more fields carry a **third state a script must branch on**, exactly as
 `app metrics` requires for `views.unavailable`:
 
 - **`canSubmitApps` is `true` / `false` / `null`.** `null` means *unknowable*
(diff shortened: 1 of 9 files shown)


ZacxDev:
## Correction to the PR description above, and an audit round

Posting rather than silently editing the body, since a reviewer may already have read it.

**Two claims in the description are now wrong**, both found by an adversarial audit of `bce0c0c`:

1. The body argues that `Subscriptions` being `json.RawMessage` is the right call. **It was not, and it is now `[]string`.** The argument had a hole the same commit's other half was built to close: a RawMessage passes the server's bytes to `--json` **verbatim**, so an object-shaped element carrying a billing email or a card fragment would have been published with no code change at all. Measured, not reasoned — the audit served exactly that body and `whoami --json` printed the lot. It also did not buy the resilience it was chosen for: a drift in `tier`, `status` *or* `isMember` blanks all four regardless, so the exemption only ever covered a drift in `subscriptions` itself.

2. The mutation table lists `Subscriptions` typed `[]string` as "killed by `…ProfileDriftIsIsolated`". As the auditor applied it, that mutant is a **compile failure** — it never ran. My own run repaired the compile error first and got a genuine behavioural kill, so the substance held, but the table row as written overstates it. Moot now: that test has been replaced.

**Two further leaks the audit found, both fixed:**

- `WhoAmI`'s parse-failure branch echoed the **entire** `/api/v1/me` body — email included — into an error `main` writes to stderr. So "email is unpublishable because it is unmodelled" was false while that line existed. Now reports the body's *shape* (allowlisted key names + JSON types) and never a value.
- `TestWhoAmIProfileFieldsReachTheHumanSurfaceNever` was a banned-substring ledger and was **vacuous against addition**. Adding nil-guarded rows rendering `Member: yes` / `Account status: ACTIVE` left the whole `internal/cmd` package `ok` — `Member:` pays no banned word, `ACTIVE` evades the lowercase literal. AGENTS item 28 records that shape losing three times and names golden-output pinning as the fix, so the ledger is **deleted**, not extended, and case 9 of `TestWhoAmIHumanBlockIsPinnedWhole` is now a profile-carrying body.

**Full mutation battery re-run after the fixes (a fix round resets the gate): 21 mutants, 21 killed** — including the two human-surface mutants that previously survived, and a `parseCoreIdentity` mutant that proves its "NOT parsed here on purpose" comment is now actually enforced rather than merely asserted.

One of those 21 surfaced a further real gap: replacing `describeMeBody`'s non-object return with an outright body echo **survived**, because every fixture in the sibling tests is a valid JSON object and that branch was green by never executing. Fixed in `1970b5d` with a test that reaches it; the mutant is killed now.

Gates re-run at `1970b5d`: `make ci` clean, `make lint` **0 issues**, `gofmt -s -l` clean, `ci-shallow` **21/21**. Note `make lint` caught a `gofmt` violation that `make ci` did not — `make ci` is not a superset of CI, exactly as `AGENTS.md` says.

`claudedocs/release-v0.1.101-draft.md` is also updated: it enumerated this release's `--json` contract changes and still said two.

ZacxDev:
## Correction: two rows of `456c2b3`'s mutation table do not reproduce

Posting rather than rewriting history, since the commit is pushed and its table is the argument for the change.

`456c2b3`'s message records this row:

> `` the same, spelled `^Scopes \(N\): ` — **SURVIVED** ``

**That is false.** Re-measured on a clean `git archive` of `bfadb20`, that mutant is **KILLED**, by *both* tests — including `TestOutputShapeMergesNothingTheHarnessDoesNot`, the one the commit argues was blind to it. And it must be: that test fed already-digit-normalised lines, so a rule matching the literal `N` is exactly the kind it *could* see. I recorded a measurement that contradicts the argument I cited it for, and did not notice.

Second row: **"All four mutants above are now killed, three of them by the new test alone."** All four *are* killed at HEAD, but only **one** — the whitespace squash — dies to the new test alone. The other three are killed by the new test *and* the pair table. So at `bfadb20` only one of the four escaped the file entirely, not three.

**What survives scrutiny** is the narrower claim the *file* already makes, in `outputShape`'s 🔴 block: the `\d+` collapse and the whitespace squash. Both reproduce exactly. The code comment was right; the commit message overstated. That asymmetry is worth naming — the prose that ships with the code was more careful than the prose that shipped with the argument for it.

The change itself stands on the narrower claim: a `^Scopes \(\d+\): ` collapse — refused widening #1 in its most natural spelling — did survive, and does not now.

---

### Also fixed in `7541bc1`, from the same round

- **The dedupe key was derived from one of the two functions being compared.** It keyed on `harnessNormalise(l, baseURL)` while comparing `l`, dropping **9 genuinely distinct raw lines** (the `Logged in as zach (id N) at <url>` headers, differing in both id and port). Measured: a rule rewriting `zach (id 1)` **survived** the equality test under that key. It also meant widening `harnessNormalise` shrank the sample the comparison ran over. Keyed on the raw line now; same mutant killed, by this test's own message.
- **Coverage prose wider than the code, again.** "Every raw rendered line" was one representative per normalised class. "Any extra rule … fails here" is false for a rule firing on no rendered line — three such widenings survive the whole package, one of them keyed on `Member:`/`Account status:` rows, which is *precisely* the profile row the ledger exists to detect. And "THE ONE CASE the equality test cannot see" is two. Both residuals are now enumerated on the test with their status: the no-line rule is **latent** (inert until such a line renders, at which point the test sees it); both-sided widening is the pair table's **partial** backstop, already bounded by measurement in its own comment.

Gates at `7541bc1`: `make ci` clean, `make lint` 0 issues, `gofmt -s -l` clean, `ci-shallow` 21/21.

PR #519, opened 2026-09-03
feat(scaffold): ship a width-adaptive example in page-money and page-vite

ZacxDev:
A freshly scaffolded App Block starts on the responsive system instead of having to discover it.

**Measured before this change**, across all three templates: `useBlockBreakpoint` 0, `--civitai-bp-*` 0. (`useBlockResize` was *not* 0 — page-money already called it; that hook reports height to the host and is unrelated to layout.) `static` is untouched — it is the CDN `<link>` route and its stylesheet has no `--civitai-bp-*` tokens.

## What the example demonstrates

**page-money — branch on your own width tier.**
`useBlockBreakpoint(rootRef)` is a container query over the slot the host gave the block, not a viewport media query, because **slot width is not monotonic in viewport width**: `model.sidebar_top` is ~360px next to a 360px phone and only ~430px next to a 1440px desktop. One structural decision is wired to it — the Model row renders as a `Group` when there is room and a `Stack` (full-width button) when there is not — plus `data-block-tier` on the block root so the tier is visible in devtools and assertable in a test. The row's *content* is hoisted so only the container differs.

**page-vite — the same lesson without an SDK.**
`container-type: inline-size` on `.app`, `@container block (min-width: 480px)` for `.app-actions`. Narrow is the base state and widening is the override, so a browser without `@container` gets the narrow layout rather than a broken wide one. `480` (`xs`) is the step this template can actually cross — `.app` is capped at `40rem`, so a `768` threshold would never fire and an example that can never fire teaches nothing.

Both READMEs gained a section, including the warning below.

## Packages newly pinned: **none**

`@civitai/blocks-react ^0.45.0` was already pinned in page-money and exports `useBlockBreakpoint` from the package root. `@civitai/theme` and `@civitai/components` are exact-pinned *dependencies* of it (`0.3.0` / `0.4.0`), so they install transitively and did not need pinning.

### 🔴 page-vite deliberately stays SDK-free, and I recommend keeping it that way

Adding `@civitai/*` to `page-vite/package.json.tmpl` was in scope, and I measured two reasons not to:

1. **It would freeze this repo's required merge gate.** `pins_guard_test.go`'s `collectCivitaiPins` walks **every** `package.json.tmpl` under `templates/` (`fs.WalkDir` on `d.Name() != "package.json.tmpl"`). `bump-pins` writes only three hardcoded sites, all page-money (`templates/page-money/package.json.tmpl`, that template's README, `scaffold_test.go`), and `discoverPackages` reads page-money's package.json alone. So a pin in page-vite is **read by the gate and unreachable by the bumper**: at the next `@civitai/*` minor publish, `bump-pins` moves page-money to `^0.46.0`, page-vite stays at `^0.45.0`, and `pins-vs-published` — required, `enforce_admins: true` — goes red for every open PR with no automated remedy. Measured publish cadence for app-sdk minors is ~3.6 days.
2. **It would make the template's own README dangerous.** page-vite's `ReadyAckPath()` is `src/civitai-host.js`, and its README says to delete that file *in the same change* that adopts the SDK. With `@civitai/blocks-react` in `package.json` but no transport mounted, following that instruction leaves the app with **no handshake at all** — the host never reveals it.

`TestRenderPageVite` now asserts page-vite pins no `@civitai/*`, and the failure message states both reasons and what to change if it is ever done deliberately (extend `bump-pins`' `sites`, rewrite that README section, same PR).

## New class guard: no `var()` in a query prelude

`internal/scaffold/queryprelude.go` + `query_prelude_guard_test.go`.

A media/container query **condition** is evaluated before custom properties are substituted, so `@media (min-width: var(--civitai-bp-sm))` never matches. No error, no warning, green build, layout silently frozen on one branch. The `--civitai-bp-*` tokens exist and sit right next to both new examples, so this is the first thing a reader will try.

The guard asserts the **relationship the browser enforces** — a prelude may not depend on the cascade — rather than the spelling `var(--civitai-bp-`, so it holds through a token rename and covers properties that do not exist yet. `@supports` is excluded on purpose (its condition tests declarations, where `var()` is legitimate).

Two design points worth reviewing:
- **Comments are stripped first**, so the templates can spell the dead form out as a warning without the guard reporting its own documentation. The stripper takes `/* … */` anywhere and `// … EOL` **only on a line whose first non-space characters are `//`** — a trailing `//` is ambiguous (`url(https://…)`), and blinding the rest of that line is the failure direction that matters. A violation written as a trailing comment is therefore still reported: fail-closed, fix by moving the note to its own line.
- **A trailing `{` is required.** Prose naming an at-rule opens no block. Without this, `` a browser without `@container` support `` was counted as a prelude and would scan 200 characters into the paragraph hunting for a brace. A fenced ```css example in a README still *is* code and is still checked (test case included).

Ledger untouched: `--civitai-bp-{xs,sm,md,lg,xl}` entered `testdata/design-tokens.txt` at the 0.45.0 bump. `go run ./internal/scaffold/cmd/bump-pins --check` → `all @civitai/* pins are current and the design-token ledger matches — nothing to do`, rc=0.

## Per-gate results (local)

| Gate | Result |
|---|---|
| `build-test` (`go test ./...`, `go vet`, `gofmt -s -l`) | rc=0 · **21** packages `ok` · 0 `FAIL` · gofmt lists 0 files |
| `pins-vs-published` (`CIVITAI_CHECK_PUBLISHED_PINS=1`, `-count=1`) | `--- PASS: TestScaffoldPinsSatisfyPublished (0.37s)` |
| `ready-ack-runtime` (`CIVITAI_CHECK_SCAFFOLD_RUNTIME=1`, `-count=1`) | `--- PASS: TestScaffoldedReadyAckActuallyFires (0.92s)` |
| `scaffold-currency` (a) anti-pattern scan | `--- PASS: TestScanDirFromEnv` · `no anti-patterns found` |
| `scaffold-currency` (b) `@latest` SDK | npm HTTP 200 · installed blocks-react **0.45.0** / app-sdk **0.37.0** · typecheck rc=0 · build rc=0 |
| design-tokens (8 tests) | all PASS · `54 files; 4 with tokens; 17 pairs, 29 occurrences; ledger 32 @0.45.0` |
| new query-prelude guard | PASS · `54 files; 4 preludes; 0 violations` |
| `template-page-vite` | scaffold → install → build rc=0 → `BLOCK_INIT: 1 file(s)`, `BLOCK_READY: 1 file(s)`, control `CI Vite Block: 1 file(s)` → `✓ . is valid` |
| `template-page-money` | scaffold → install → typecheck rc=0 → **168 passed (12 files)** → build rc=0 → `✓ . is valid` |
| `lint` (golangci-lint) | `0 issues.` |

The `@container` rule survives esbuild minification intact in the built page-vite CSS.

## Red / green matrix

Base ref **`22cb38d`** (`origin/main`). The new `src/responsive.test.tsx` was scaffolded from base and dropped into that app unchanged:

| | base `22cb38d` | HEAD |
|---|---|---|
| `src/responsive.test.tsx` | **2 failed**, 13 passed (15) — `Unable to find an element by: [data-testid="pm-model-row"]` | **15 passed** |

Honest scope: only the 2 layout tests are regression coverage. The 13 that pass at base are invariant guards — 11 pin the SDK's px scale and 2 pin content presence.

Also caught and fixed as part of this: `App.pollretry.test.tsx`'s wholesale `vi.mock('@civitai/blocks-react')` went stale the moment `App.tsx` imported a new hook (`No "useBlockBreakpoint" export is defined on the mock`, 2 tests red). Loud failure, correct direction — noted inline that the mock is a second place to update.

## Mutation results

Every mutant was killed by the **guard's own** assertion, on an input no earlier check rejects, and the control was re-run green after each restore.

| Mutation | Killed by | Message |
|---|---|---|
| `const narrow = false` (branch dead) | `responsive.test.tsx` NARROW | `expect(element).toHaveAttribute("data-layout", "stacked")` … `Received: data-layout="row"` |
| `bp.below('sm')` → `bp.below('md')` (wrong threshold) | `responsive.test.tsx` WIDE | `expect(element).toHaveAttribute("data-layout", "row")` … `Received: data-layout="stacked"` |
| `var(--civitai-bp-xs)` into the live `@container` rule | new class guard | `DEAD QUERY CONDITION` naming `templates/page-vite/src/index.css.tmpl`; still `4 preludes`, so the coverage floor was **not** the killer |
| same, in the README's fenced example | new class guard | `DEAD QUERY CONDITION` naming `templates/page-vite/README.md.tmpl` (fenced CSS is checked) |
| replace the hook call + `narrow` in `App.tsx.tmpl` | `TestRenderPageMoney` floor | `expected to contain "bp.below('sm')"` |

Fixture widths are **361** and **900** — pairwise distinct, distinct from every constant the assertions name (480 / 768 / 1024 / 1184 / 1440), and each strictly *inside* its tier rather than on a boundary. 900 is what discriminates `sm` from `md`; 361 is roughly the real `model.sidebar_top` slot.

The scale assertions pin `xs`=480, `md`=1024, `lg`=1184, `xl`=1440 plus the pixel below each — the values that separate Civitai's px scale from Mantine's stock em scale (576/768/992/1200/1408), which agree on `sm` alone.

## Not verified

- **No browser run.** jsdom lays nothing out and ships no `ResizeObserver`, so `responsive.test.tsx` stubs both (an inert `ResizeObserver` purely so the hook's effect runs, plus a `clientWidth` getter — the hook seeds itself synchronously from `clientWidth` right after `observe()`). That exercises the *decision*, not real layout. The `@container` rule in page-vite is verified only as far as "it survives the build byte-correct" — it was not rendered in a browser at two widths.
- The `@media (min-width: var(--x))` inertness is taken from the brief's Chromium measurement, not re-measured here.
- CI itself has not run yet; every result above is local.

**Do not merge.**

### DIFF
--- a/internal/scaffold/query_prelude_guard_test.go
+++ b/internal/scaffold/query_prelude_guard_test.go
@@ -1,0 +1,275 @@
+package scaffold
+
+// Structural guard: no scaffold template may put a `var()` inside a media or
+// container query PRELUDE.
+//
+// Why this exists at all. Both width-adaptive examples the scaffold now ships
+// sit right next to the `--civitai-bp-*` tokens, and putting one of those tokens
+// in the query condition is the FIRST thing a reader tries. It does not work,
+// and — this is the whole problem — it does not fail either. The prelude is
+// evaluated before custom properties are substituted, so the at-rule is inert:
+// green build, no console warning, no test failure, and a layout silently frozen
+// on one branch. Measured in Chromium with a literal-value control in the same
+// document (`--bp-sm` = 768px, 1000px window): the literal form applied, the
+// `var()` form did not.
+//
+// Why it is a CLASS guard and not a grep. `grep -F 'var(--civitai-bp-'` pins a
+// SPELLING: it passes the moment the tokens are renamed, says nothing about a
+// `var()` naming some other property, and would have to be re-derived for every
+// future token namespace. This asserts the RELATIONSHIP the browser actually
+// enforces — a prelude may not depend on the cascade — so it holds however the
+// property is named, including for properties that do not exist yet.
+//
+// `@supports` is deliberately out of scope: its condition tests DECLARATIONS,
+// where `var()` is legitimate. See queryprelude.go.
+
+import (
+	"strings"
+	"testing"
+	"testing/fstest"
+)
+
+// TestScaffoldTemplatesHaveNoVarInQueryPreludes is the guard.
+func TestScaffoldTemplatesHaveNoVarInQueryPreludes(t *testing.T) {
+	scan, err := ScanQueryPreludes(templatesFS, "templates")
+	if err != nil {
+		t.Fatalf("scanning templates for query preludes: %v", err)
+	}
+
+	// ── POSITIVE CONTROLS ───────────────────────────────────────────────────
+	// The reassuring answer here is "no violations", which is also what a
+	// scanner wired to an empty tree, or one whose regex stopped matching,
+	// returns. Assert it read this repo's templates AND found real preludes
+	// before believing the zero.
+	t.Logf("scanned %d template file(s); %d media/container prelude(s) survived comment-stripping; %d violation(s)",
+		scan.FilesRead, scan.Preludes, len(scan.Violations))
+
+	if scan.FilesRead < 20 {
+		t.Fatalf("read only %d template file(s) — the scanner is not looking at this repo's templates, so a clean verdict here would be a statement about nothing", scan.FilesRead)
+	}
+	// Measured on this tree: 4 real preludes — the two boot-skeleton
+	// `prefers-color-scheme` blocks (page-money + page-vite index.html),
+	// page-vite's `@container` responsive rule, and the fenced copy of that rule
+	// in page-vite's README. Every other textual occurrence is either inside a
+	// comment (stripped) or prose that opens no block (not a prelude).
+	//
+	// This floor is what makes the comment-stripper's failure visible: a
+	// stripper that ate too much would leave 0 here rather than reporting a
+	// clean tree. If a change legitimately removes a query, lower this
+	// DELIBERATELY — do not let the guard become a guard over nothing.
+	if scan.Preludes < 3 {
+		t.Fatalf("found only %d media/container prelude(s) in the templates — expected at least 3 (two boot-skeleton prefers-color-scheme blocks + page-vite's @container responsive rule). Either the templates lost their queries (lower this floor on purpose), QueryPreludeRe stopped matching, or StripCommentsForQueryScan is eating real code", scan.Preludes)
+	}
+
+	// ── THE INVARIANT ───────────────────────────────────────────────────────
+	for _, v := range scan.Violations {
+		t.Errorf(
+			"DEAD QUERY CONDITION — this at-rule can NEVER match, silently.\n"+
+				"  file:    %s\n"+
+				"  prelude: %s\n"+
+				"  A media/container query prelude is evaluated BEFORE custom properties are\n"+
+				"  substituted, so the var() never resolves and the whole block is inert. There is\n"+
+				"  no error, no warning and no build failure — the scaffolded app just renders one\n"+
+				"  branch of the layout forever.\n"+
+				"  fix: write the pixel value out (the block breakpoint scale is\n"+
+				"       xs 480 · sm 768 · md 1024 · lg 1184 · xl 1440), or make the decision in JS\n"+
+				"       with useBlockBreakpoint() from @civitai/blocks-react.",
+			v.File, v.Prelude)
+	}
+}
+
+// ── INSTRUMENT VALIDATION ───────────────────────────────────────────────────
+// Everything above reads a verdict out of ExtractQueryPreludes /
+// StripCommentsForQueryScan / ScanQueryPreludes. Until these controls have been
+// watched to work, that verdict is a fact about the extractor.
+
+// TestExtractQueryPreludes is the POSITIVE control on the extractor: fed input
+// that MUST produce preludes, it produces exactly them — and fed the shapes that
+// must NOT count, it produces none.
+func TestExtractQueryPreludes(t *testing.T) {
+	cases := []struct {
+		name string
+		in   string
+		want []string
+	}{
+		{
+			name: "a literal media query",
+			in:   "@media (min-width: 768px) { .a { color: red } }",
+			want: []string{"@media (min-width: 768px)"},
+		},
+		{
+			name: "a named container query",
+			in:   "@container block (min-width: 480px) {\n  .a { display: flex }\n}",
+			want: []string{"@container block (min-width: 480px)"},
+		},
+		{
+			name: "the DEAD form is extracted, so the guard can see it",
+			in:   "@media (min-width: var(--civitai-bp-sm)) { .a { color: red } }",
+			want: []string{"@media (min-width: var(--civitai-bp-sm))"},
+		},
+		{
+			name: "@supports is NOT a query prelude — var() is legitimate there",
+			in:   "@supports (color: var(--civitai-color-text)) { .a { color: red } }",
+			want: nil,
+		},
+		{
+			name: "a multi-line prelude is one prelude, whitespace collapsed",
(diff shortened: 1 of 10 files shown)


ZacxDev:
## 🔴 Three red jobs on this PR are RED AT BASE — not from this diff

Run [33783547063](https://github.com/civitai/cli/actions/runs/33783547063): `build-test`, `lint`, `pins-vs-published`, `ready-ack-runtime` and `template-page-vite` all **pass**. Three fail, and none of them is attributable to this change.

### `template-page-money` + `scaffold-currency` — npm 10 crashes resolving the peer graph

Both die at the same step, `npm install` on the freshly scaffolded page-money app:

```
npm error Cannot read properties of null (reading 'edgesOut')
```

`scaffold-currency` is a **required** check, so this blocks the merge.

**Control, run locally under the same toolchain CI uses (node 22.23.2 / npm 10.9.8):**

| Tree | `npm install` on scaffolded page-money |
|---|---|
| this branch | `rc=1` · `Cannot read properties of null (reading 'edgesOut')` |
| **unmodified `origin/main` @ `22cb38d`** | `rc=1` · **identical error** |

The two scaffolded `package.json` files are **byte-identical** (`diff` exits 0) — this PR touches no `package.json`, so `npm install` sees exactly the same input on both arms.

**Cause.** The stack lands in arborist's `#loadPeerSet` while walking `node_modules/vitest`:

```
992 silly fetch manifest jsdom@*
992 silly fetch manifest canvas@^3.2.3
992 verbose stack TypeError: Cannot read properties of null (reading 'edgesOut')
992 verbose stack     at #loadPeerSet (.../@npmcli/arborist/lib/arborist/build-ideal-tree.js:1289:38)
```

`vitest@5.0.0` was published **2026-09-03T12:24:30Z**. `main`'s last green CI run was `22cb38d` at **2026-09-02T23:45:04Z** — i.e. before it. The template pins `vitest: ^4.1.0`, which does not *admit* 5.0.0, but npm 10's peer-set walk fetches candidate manifests anyway and crashes on the new peer graph. **npm 11.19.0 installs the same tree without error**, so this is npm-10-specific and the CI runner gets npm 10 from `actions/setup-node@v4` with `node-version: "22"`.

Every PR against this repo will be red on these two jobs until it is addressed. Not fixed here — out of scope for a scaffold-example change, and it wants a deliberate decision (bump the runner's npm, or constrain the dev-dependency graph).

Closing condition, if someone picks it up: **`template-page-money` and `scaffold-currency` both green on a `workflow_dispatch` CI run of unmodified `main`.**

### `schema-drift` — upstream canonical drift

```
FAIL: schema/app-block.manifest.schema.json has DRIFTED from the canonical.
```

This PR touches nothing under `schema/`. It is not a required check, and `revendor-canonical-schema.yml` is the bot that owns it.

### Local results for the two blocked jobs

Because CI cannot run them, both were run to completion locally on this branch (npm 11.19.0):

- `template-page-money`: scaffold → install → `typecheck rc=0` → **168 tests passed, 12 files** → `build rc=0` → `✓ . is valid`
- `scaffold-currency`: anti-pattern scan `--- PASS: TestScanDirFromEnv` (`no anti-patterns found`); npm reachable (HTTP 200) → installed `@civitai/blocks-react@0.45.0` / `@civitai/app-sdk@0.37.0` at `@latest` → `typecheck rc=0` → `build rc=0`

---

## CI-side control: `ci.yml` dispatched on unmodified `main`

Run [33783729351](https://github.com/civitai/cli/actions/runs/33783729351) — `workflow_dispatch` on `main`, no diff at all — against this PR's re-run [33783547063](https://github.com/civitai/cli/actions/runs/33783547063). **The two arms are identical.**

| job | `main` (no diff) | PR #519 | required? |
|---|---|---|---|
| `build-test` | success | success | ✅ |
| `lint` | success | success | |
| `pins-vs-published` | success | success | ✅ |
| `ready-ack-runtime` | success | success | ✅ |
| `template-page-vite` | success | success | ✅ |
| `scaffold-currency` | **failure** | **failure** | ✅ |
| `template-page-money` | **failure** | **failure** | |
| `schema-drift` | **failure** | **failure** | |

Same three red, same three green-that-matter, zero jobs where the arms differ. This diff moves nothing in CI's verdict.

Issue #513, opened 2026-08-30
numeric usernames coerced to numbers failing the CLI


Rochet2:
1. The API sometimes returns username as a number instead of a string `"username": 123123`. this seems to happen for users that have a number as their username like this member https://civitai.com/user/2802169344506. So this might be a bug in the API where number in string coerces to number.
2. The CLI insists on `username: string`, typed-unmarshal fails, and it reports `exit 1 / “unexpected response (status 200)` when using `civitai images search` and encountering such a user.

Either the API or the CLI probably needs a fix.

ZacxDev:
Thanks for this, and sorry for the slow reply.

You were right that it's ambiguous where the fix belongs, and the answer turned out to be **both** — but with a twist worth reporting back.

## Client side — fixed

#532 is merged. It adds a `FlexString` type to `pkg/civitai` whose `UnmarshalJSON` accepts a JSON string *or* number, applied to the `Username` fields across the read SDK. Two details we were careful about:

- The number branch goes through `json.Number`, so the **literal digits** survive. Routing through `float64` would turn `2802169344506` into `2.802169344506e+12` — and because that value is under 2^53 it round-trips *exactly*, so a test using only the username from your report wouldn't catch that mistake. There's a `2^53+1` fixture specifically to pin it.
- It tolerates only a string or a number. Objects, arrays and booleans still error, so genuine schema drift isn't silently swallowed.

Your `--json` output is unaffected either way, incidentally — that path re-emits the raw server bytes and never goes through the struct.

It's a source-breaking change for anyone importing `pkg/civitai` directly (the field type moved off `string`), so it shipped as `fix(read)!:` with the conversion documented.

## Server side — and here's the twist

I went to reproduce it live before filing the API-side fix, and **I can't currently reproduce the coercion**:

- `GET /api/v1/users?query=2802169344506` returns `"username":"2802169344506"` — quoted.
- Six queries over `/api/v1/users` returned every all-digit username quoted, including other 13-digit ones.

I want to be straight that this is **not** me telling you it's fixed. You hit it on `images search`, and `/api/v1/images` embeds the user through a different serializer than `/api/v1/users`. I couldn't test that path at all: every all-digit account I could find has no public images, so there was nothing to query. That's "no test subject", not "working".

So — **can you still reproduce it?** Specifically:

1. The exact command, and whether it's still `images search`.
2. Any user where it fires, if `2802169344506` no longer does.
3. Roughly when you last saw it.

[...]

PR #526, opened 2026-09-06
fix(read): sanitize unescaped C0 control chars before typed decode in getInto (#525)

### LINKED ISSUE #525: Raw C0 control characters in API responses fail in getInto before emitJSON gets called
Hey Zac!

While looking into how --json handles responses with raw control characters, I noticed an ordering quirk between getInto and mitJSON.

PR #165 added scapeJSONStringControlChars in mitJSON to sanitize unescaped C0 characters (like raw \r in prompt text) so --json wouldn't break jq. But in the current flow, client.getInto() runs json.Unmarshal(raw, out) on the raw bytes *before* mitJSON is ever reached.

Because standard ncoding/json strictly rejects raw control characters in string literals, getInto immediately returns:
`	ext
unexpected response from /api/v1/... (status 200): ...
`
and the CLI exits with code 1. mitJSON never runs, so the sanitization in scapeJSONStringControlChars doesn't actually get a chance to help the caller (I saw your comment in 
ead_help.go:57 noting this exact behavior!).

It seems like we could move scapeJSONStringControlChars down into pkg/civitai/read.go and run it in getInto right before json.Unmarshal(raw, out) if !json.Valid(raw). That way the struct unmarshaling succeeds, and mitJSON gets already-clean JSON to format.

I've got a patch and a unit test ready for this. Would you be open to a PR?

xsvm:
Closes #525

### Context
PR #165 added `escapeJSONStringControlChars` in `emitJSON` to ensure `--json` output can always be parsed by `jq`. However, `client.getInto()` runs `json.Unmarshal(raw, out)` on the unescaped wire bytes *before* `emitJSON` is reached.

When an API response carries raw C0 control characters in string literals (e.g. literal `\r` inside a prompt), Go's strict `encoding/json` decoder fails immediately:
```text
unexpected response from /api/v1/... (status 200)
```
and the command aborts before `emitJSON` is entered (as noted in `internal/cmd/read_help.go:57`).

### Fix
1. Moved `escapeJSONStringControlChars` down into `pkg/civitai/read.go` and exported it as `EscapeJSONStringControlChars`.
2. In `getInto`, if `!json.Valid(raw)`, attempt sanitization via `EscapeJSONStringControlChars` before attempting `json.Unmarshal(raw, out)`.
3. Replaced `internal/cmd/read.go`'s `escapeJSONStringControlChars` with a one-line delegation to `civitai.EscapeJSONStringControlChars`.
4. Added `TestGetIntoSanitizesRawControlChars` in `pkg/civitai/read_test.go` asserting a payload with a literal carriage return decodes cleanly.

### Tests
- `TestGetIntoSanitizesRawControlChars` (PASS)
- `TestEscapeJSONStringControlCharsLeavesStructuralWhitespace` (PASS)
- `TestEmitJSONEscapesRawControlChars` (PASS)
- `TestEmitJSONEscapesAllC0ControlChars` (PASS)
- `TestEmitJSONValidInputUnchanged` (PASS)
- `gofmt -s -l .` clean, `go vet` clean.

### DIFF
--- a/internal/cmd/read.go
+++ b/internal/cmd/read.go
@@ -108,59 +108,10 @@ func emitJSON(cmd *cobra.Command, raw []byte) error {
 	return nil
 }
 
-// escapeJSONStringControlChars walks raw JSON bytes tracking in-string state
-// (respecting \\ and \") and replaces any raw C0 control byte (0x00–0x1F) that
-// appears INSIDE a string literal with its valid JSON escape (\b \f \n \r \t or
-// \u00xx). Control bytes outside strings (structural whitespace) and everything
-// already escaped are left byte-for-byte unchanged, so valid input round-trips
-// identically. This does not attempt to repair other kinds of malformed JSON;
-// callers should verify the result with json.Valid before relying on it.
+// escapeJSONStringControlChars delegates to civitai.EscapeJSONStringControlChars
+// in pkg/civitai.
 func escapeJSONStringControlChars(raw []byte) []byte {
-	var out bytes.Buffer
-	out.Grow(len(raw))
-	inString := false
-	escaped := false
-	for i := 0; i < len(raw); i++ {
-		c := raw[i]
-		switch {
-		case !inString:
-			out.WriteByte(c)
-			if c == '"' {
-				inString = true
-			}
-		case escaped:
-			// Previous byte was a backslash; this byte is the escape's payload
-			// (", \\, /, b, f, n, r, t, or the u of a \uXXXX). Emit verbatim.
-			out.WriteByte(c)
-			escaped = false
-		case c == '\\':
-			out.WriteByte(c)
-			escaped = true
-		case c == '"':
-			out.WriteByte(c)
-			inString = false
-		case c < 0x20:
-			// Raw control character inside a string literal — invalid JSON.
-			// Rewrite it as the shortest valid escape.
-			switch c {
-			case '\b':
-				out.WriteString(`\b`)
-			case '\f':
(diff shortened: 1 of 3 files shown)


ZacxDev:
Thanks for the PR, and sorry it sat without a signal — fork PRs need a maintainer to approve workflow runs, and nobody had. Approved now.

Result: **7 of 8 checks green.**

The one red, `pins-vs-published`, is **not about your change**. It compares the scaffold's pinned `@civitai/*` versions against npm's published latest, so it goes red on any branch based on a `main` whose pins have since gone stale — no code change required. Yours is based on `d9b4e29`, 11 commits back, which predates the pin bump in #529. I re-ran that guard against live npm on current `main` and it passes.

**A rebase on `main` should clear it.** Everything that actually measures your change — `build-test`, `lint`, `schema-drift`, `ready-ack-runtime`, `scaffold-currency`, `template-page-vite`, `template-page-money` — is green.

One thing worth flagging: your issue (#525) and #513 are the same defect class. Both end at `getInto`'s strict `json.Unmarshal` rejecting a real **200** and surfacing it as the same opaque `unexpected response from /api/v1/... (status 200)` — yours triggered by raw C0 control bytes, #513's by an all-digit username arriving as a bare JSON number.

#513's half merged as `69724bf`. It adds a `FlexString` type at the *struct field* level and deliberately does **not** touch `getInto`, specifically so it wouldn't collide with this PR — so the two should merge in either order. Worth a look at whether the shared error message should name the decode failure rather than the response, since two people have now had to guess.

Reviewing the change itself now.

PR #528, opened 2026-09-08
feat(agent-setup): configure the developer's coding agent for Civitai Apps

ZacxDev:
Implements §1–§4 of the frozen `agent-setup` contract. `civitai agent-setup` configures whichever coding agent the author is using to build Civitai Apps, and stops before auth.

## What it does

A write run, in order:

1. Detects the agent (env signals, then marker files in `--dir`, then `other`). `--agent` overrides.
2. Writes `<dir>/AGENTS.md` — the managed block, three-case rule below.
3. Writes `<dir>/CLAUDE.md` containing `@AGENTS.md`, **only when there is none**.
4. Registers both MCP servers (`https://mcp.civitai.com/mcp`, `https://orchestration.civitai.com/mcp`) in the detected agent's own config file.
5. Prints a next-step block ending in `civitai login` — which it never runs.

## The per-agent table, and why this is a command

The config path *and* the top-level key both differ, and every way of getting it wrong writes a file that parses, starts the agent cleanly, and registers nothing.

| agent | file | key | url key |
|---|---|---|---|
| `claude` | `<dir>/.mcp.json` | `mcpServers` | `url` |
| `cursor` | `<dir>/.cursor/mcp.json` | `mcpServers` | `url` |
| `vscode` | `<dir>/.vscode/mcp.json` | **`servers`** | `url` |
| `codex` | `~/.codex/config.toml` | **`[mcp_servers.<name>]`** | `url` |
| `opencode` | `<dir>/opencode.json` | **`mcp`** | `url` |
| `windsurf` | `~/.codeium/windsurf/mcp_config.json` | `mcpServers` | **`serverUrl`** |
| `zed` | `~/.config/zed/settings.json` | **`context_servers`** | `url` |
| `other` | — | — | prints the config to paste; writes nothing |

Every entry shape was taken from the vendor's **current** docs rather than from memory. Two results worth flagging:

- **Zed carries no `source` key.** `"source": "custom"` was a real key in older Zed and appears in no current doc, so it is not written — `zed.dev/docs/ai/mcp`'s remote example is `{"url": …, "headers": {…}}` and nothing else.
- **Codex uses `http_headers`**, not `headers`, and infers the streamable-HTTP transport from the presence of `url` (no `type` key). Cursor, Windsurf and Zed likewise define no `type` for remote servers, so none is written for them; Claude Code and VS Code take `"http"`, opencode `"remote"`.

## Nothing is clobbered

- **AGENTS.md, three cases.** No file → write the template. File without the managed block → **append**, every existing byte kept. File with the block → replace **only** the span between the markers, byte-for-byte outside. Half a block (one marker, or END before BEGIN) is **refused** rather than repaired — the command cannot tell where the author's text ends and its own begins.
- **CLAUDE.md** is never read and never modified when it exists.
- **MCP configs are merged**, preserving every other server and every unknown key. A config that does not parse is refused **by name**; a `mcpServers` that is not an object is refused rather than replaced. The Codex merge is **line-based** rather than a TOML round trip, so a hand-maintained `config.toml` keeps its comments and key order.
- **No placeholder credential.** With no token the entries are written *without* the `Authorization` header. A `Bearer <your-token>` would look correct in every file the user can read and 401 at request time; the read tools are anonymous, so a header-less entry genuinely works.

## `--check` and the one thing that is easy to invert

`--check --json` emits exactly `track` / `agent` / `ok` / `checks`, with the six row names `prompt.md` reads.

🔴 **`ok` is the AND of every check EXCEPT `authenticated`.** Setup stops before auth on purpose, so a complete but unauthenticated setup is `ok: true` and **exits 0**. Folding that row into the verdict would exit 1 for every user in exactly the state the documented flow leaves them in. Pinned three ways: a unit table over `agentSetupVerdict`, an end-to-end `--check --json` row with a `PREMISE BROKEN` assertion that the auth row really did fail, and a **process-level** row in `cmd/civitai`.

## Exit codes

No row was added to `exitCodeDocs`. `--check` failing returns `ErrAgentSetupIncomplete`, an untagged sentinel modelled on `app doctor`'s `ErrListingBlocked`, so it maps to **1** — a verdict about the setup, not a mistake about the invocation. Unknown `--agent`, a `--dir` that does not exist or is not a directory, and `--track api` are usage errors and exit **2**.

`--track api` is a **recognised** value: it refuses with the two MCP servers (which work today) and pointers to `developer.civitai.com/site/` and `/orchestration/`, and says the track is coming — so a reading agent does not conclude the API is unsupported. It never silently falls back to `--track app`.

`resolveAgentSetupDir` is deliberately **not** `resolveProjectDir`: that helper is bound by an asserted ledger to the two commands calling `validate.Dir`, and this command classifies any directory, not an App project root. Same three-way branch, different remedies.

## No version pins

Nothing this feature writes pins an `@civitai/*` version — not the template, not the help text. Pins live in `civitai app init`, which `pins-vs-published` holds against npm; a literal in an instruction file rots in silence. `TestAgentsTemplatePinsNoVersion` makes that a rule, with a positive control proving the pattern can match a real pin.

## Testing

35 top-level tests, ~160 leaf subtests. Every guard was **watched fail** against a deliberate mutant before being claimed. Highlights:

- **The template-command walker.** Extracts every `civitai …` from the embedded template and requires it to resolve in the real Cobra tree, checking `c.Name()` rather than only `err == nil` (cobra's `Find` does not error on an unknown leaf). Killed by renaming `app doctor` → `app doctorr` in the template. It carries a negative control that feeds the same walker a document naming commands the binary lacks.
- **A table test over the per-agent path and key names**, plus an end-to-end row per agent that writes a real config and reads the URL back out through *that agent's own* keys — and asserts the keys the agent does **not** use are absent.
- **Detection order.** Cursor and Windsurf are VS Code forks and set `TERM_PROGRAM=vscode` too; the fork rows must come first, or every fork is configured through `.vscode/mcp.json` and never sees a server. Each row carries both signals.
- The `--dir` remedy guard was **strengthened after a survived mutant**: a version deriving both expectations from the remedy constants survived a clean swap of the two constants' values, because the test read the same constant names the code did. It now also asserts a literal anchor phrase per arm. That is AGENTS item 26's own trap arriving from a third direction — the mutants recorded there all made the constants *duplicate*, which a constant-derived expectation does catch; a pure *swap* it does not.

## Docs

README: a new `## Set up your coding agent (agent-setup)` section (with its TOC entry), a command-reference row, the per-agent key table, the three AGENTS.md cases, the `--check --json` shape with the `authenticated` caveat called out in red, and the exit codes. All README guards green.

## Known gaps, stated rather than hidden

- **§6 (anti-rot) is not in this PR.** Its implementation site is `civitai-developer-docs` — extending the `appblocks-snapshots/` CI so every command and flag in `prompt.md` must appear in `civitai-cli-help.txt` — which is the repo the parallel `prompt.md` agent is working in, and it depends on `prompt.md` existing. Closing condition: a merged PR in `civitai-developer-docs` whose CI goes red when a `prompt.md` command is absent from the help snapshot. The equivalent guard for *this* repo's own template ships here.
- **No AGENTS.md decision item was added.** `AGENTS.md` is ~500 bytes under its enforced ceiling; a trigger line plus the index clause `TestEveryAgentsItemIsNamedByTheIndex` requires would very likely need an eviction from someone else's item. The rationale lives in the code's doc comments and the README instead. Worth a follow-up when the ceiling is next re-baselined.
- **`ErrAgentSetupIncomplete` was not added to `errorLine`'s bare-print whitelist** in `cmd/civitai/main.go`. That whitelist's own comment says adding a second entry should be a deliberate decision; a failing `--check` reads acceptably with the `Error:` prefix, and widening a published stderr surface was outside this change.
- **The Windsurf environment signals are a heuristic.** `WINDSURF_USER_ID` / `TERM_PROGRAM=windsurf` are plausible rather than vendor-confirmed. A missed signal costs nothing — detection falls through to marker files and then to `other`, both honest answers — and `--agent windsurf` is the reliable route.
- **The line-based TOML parser** understands table headers, comments and multi-line basic/literal strings, but not a header-shaped line inside a multi-line *array*. No Codex config plausibly contains one; the effect would be a mis-placed block boundary, not data loss. The alternative is a direct dependency on `pelletier/go-toml/v2` (currently indirect) for a file with six keys in it — an "ask first" item under AGENTS.md.

## Gate

`make ci` — green (21 packages, 0 failures). `nix-shell -p golangci-lint --run "make lint"` — **0 issues**, and the linter was negative-controlled (a deliberate unused variable made it exit 1) so that zero means something.

🔴 `pins-vs-published` is red on this PR and `mergeStateStatus` is `BLOCKED`. That is **pre-existing on `main`** — the scaffold's npm pins are behind published — and has nothing to do with this change. No pin was touched here.

https://claude.ai/code/session_01A7RUfuWWYUVrLRTmvzgDGF


ZacxDev:
## Audit round 1 — claims

Adversarial audit read `6340db8`; fixes produced `c801ab8`.

```audit-claims round=1 audited=6340db8..c801ab8
1. Anonymous access is now modelled per server (mcpServer.Anonymous) instead of asserted for both; the false "both servers' read tools work anonymously" claim was removed from seven code/output surfaces, the README and the decision doc, and every surface now reads one shared note.
2. The decision doc's Zed OAuth-discovery rationale was RETRACTED with no replacement, because the orchestration 401 carries no WWW-Authenticate header.
3. A JSONC config (comments) no longer aborts the run: comments are decoded past, and an MCP-step refusal now degrades so AGENTS.md and CLAUDE.md are still written.
4. The MCP merge is now per-key in both JSON and TOML, so a hand-added Authorization header and sibling keys like startup_timeout_sec survive a re-run.
5. The hasToken gate was kept but its reporting corrected, and the comment that justified it with a case it does not gate was rewritten.
6. The TOML parser now tracks bracket depth, exempts [[array-of-tables]] from the duplicate check, and handles trailing-comment headers; the residual comment was corrected.
7. The atomic write follows a symlink instead of replacing it, and refuses a broken link by name.
8. VS Code was enabled with ${env:CIVITAI_TOKEN} after the maintainer statements and mcpRegistry source were verified first-hand; item 34 and its decision doc record the reversal with the original reasoning kept verbatim.
9. The Windsurf row was marked honestly as legacy-Cascade rather than repathed to an untested Devin CLI location.
10. CODEX_HOME is honoured, and Zed's Windows %APPDATA% path is selected via an injected agentEnv.GOOS.
11. The claude-md row no longer fails the verdict for agents that do not read CLAUDE.md; one predicate now serves verdict, count and glyph.
12. opencode.jsonc is handled, with a ledger asserting the marker/target class rather than the single case.
13. The 0600 comment, the out-of-order managed-block message, the duplicate-block case, the README column and the --json path form were all corrected.
14. The operator's real token prefix was removed from claudedocs/decisions/34; no occurrence remains in the tree.
```

**Behaviour changes reviewers should notice:**

- A **write run can now emit `ok: false`** (always with exit 1), deliberately retiring the old "a payload implies success" property.
- **Comments are lost when a JSONC config is written back.** This is a real regression, chosen as loud-and-usable over refuse-and-blame; `blankJSONComments` preserves byte offsets so a comment-preserving splice stays possible later. Flagged in code, README, `--help` and the decision doc.
- **`AGENTS.md`'s ceiling was raised 29,600 → 30,200** (file 29,990; `agentsMaxBytesCeiling` unchanged at 30,600) to pay for new item 35.

**Known residual, stated not skipped:** Zed's `XDG_CONFIG_HOME` on Linux is plausible but unverified, and an unverified path in that table is exactly the mistake item 34 records. Nobody has run VS Code and watched a resolved header go out on the wire.

ZacxDev:
## Audit round 2 — claims

Delta audit read `c801ab8`. Nine of fourteen round-1 claims verified cleanly fixed by running the binary; **the round was not clean** and a round 3 follows.

🔴 **One claim was false and I reproduced it myself.** The TOML per-key merge (claim 4's Codex half) **silently deletes** a hand-added `http_headers` / `bearer_token_env_var` on our own tables, at rc 0, with `--check` still reporting `ok: true`:

```
BEFORE                                          AFTER
[mcp_servers.civitai]                           [mcp_servers."civitai"]
url = "…/mcp"                                   url = "…/mcp"
# I added this by hand so it works:             # I added this by hand so it works:
http_headers = { Authorization = "Bearer …" }   startup_timeout_sec = 30
startup_timeout_sec = 30
```

The user's comment now annotates nothing. `README.md:480` publishes "merged into **key by key** (a header you added to a Civitai entry survives)" and decision doc 35 calls it "the finding this section exists for" — both false on the Codex path. The JSON side is correct; only TOML unconditionally skips owned keys that `renderTOMLServer` then does not write back.

**The fixture that would have caught it is the second finding:** `TestTOMLMergePreservesKeysOnOurOwnTable`'s docstring says "the same defect on the Codex path", but its fixture carries only `startup_timeout_sec`/`tool_timeout_sec` — never an `http_headers` line. It inspects the side that works. That is why this survived a green suite and an 18-mutant sweep.

**Also carried to round 3:** a README exit-code paragraph contradicting both the shipped behaviour and the README's own table; a run reporting "no Authorization header was written" about a header the merge just preserved; a trailing comma still refused while every surface says JSONC is tolerated; `--json` emitting zero bytes on write-time failures and on `--check` against an unparseable config; and item 34 now carrying two mutually exclusive rules for what licenses a table row.

🔴 **Leak status, corrected:** claim 14 is true of the tree but **not of the branch**. The real token prefix is still in pushed commit `6340db8`, visible in the Commits tab, and a squash merge does not remove it. **The remedy is token rotation, not a history rewrite** — rewriting would invalidate the sha anchors these claims blocks depend on, and rotation makes the prefix worthless anyway.

**Ledger:** `round 2 · payload lines changed THIS round: 1169 (+1003/−166 across four Go files) · since round 1: 2742`. Non-zero payload — the attribution gate does not fire and the ladder continues on findings.

ZacxDev:
## Audit round 2 — fixes

Delta audit read `c801ab8`; fixes produced `897c1cc`.

```audit-claims round=2 audited=c801ab8..897c1cc
1. mergeTOMLBlock's owned-key set is now DERIVED from the rendered lines instead of a fixed triple, mirroring mergeEntryKeys, so a hand-added http_headers or user-chosen bearer_token_env_var on our own table survives a re-run.
2. TestTOMLMergePreservesKeysOnOurOwnTable's fixture now carries both auth keys in both token states, watched red on c801ab8 and green after; a new ledger guard asserts the published key-by-key claim for all seven agents.
3. TestRepeatedRunsAreIdempotent was found to have the same half-covered shape — it started from an EMPTY project, so every run after the first merged over this command's own output — and now starts from a user-written file and asserts the key survives three runs.
4. The README's false "a write run exits 0" paragraph was corrected and every other surface naming that behaviour was grepped and widened; Troubleshooting and exitcodes_doc.go carry no stale copy.
5. mcpAuthCoverage is computed from the bytes the run will write, so a run no longer reports "no Authorization header was written" about a header the merge preserved; the same sentence feeds the terminal and --json from one source.
6. Trailing commas are tolerated like comments (blankJSONTrailingCommas, gated on AllowsComments), verified against Zed's serde_json_lenient and VS Code's jsonc.ts sources; all four surfaces widened to "comments, trailing commas and key order".
7. All four failure shapes now emit a --json payload with ok:false and rc 1 instead of zero bytes; writes are attempted independently so no payload claims an action nothing tried.
8. Item 34 now states ONE rule of record with three evidence classes and an explicit ordering, says which two sentences it supersedes and why, and retains the superseded text quoted as false rather than deleting it.
9. checkWriteTargetResolvable moved into planMCPConfig, so --dry-run no longer reports create/ok:true/rc 0 for a destination the real run refuses; a ledger compares action as well as path across four fixtures.
10. The size-ceiling comment's stale 30,100 was corrected to 30,200.
```

🔴 **I reproduced the round-2 defect and re-verified the fix myself.** With a Codex config carrying a hand-added `http_headers`, a user-chosen `bearer_token_env_var` and a sibling timeout: at `c801ab8` the two auth keys were silently deleted at rc 0; at `897c1cc` all three survive and the user's `# I added this by hand so it works:` comment once again sits above the key it annotates.

**A third instance of the same shape was found and fixed**, unprompted: `TestRepeatedRunsAreIdempotent` claimed "the per-key merge must not accumulate" while starting from an empty project — so every run after the first merged over this command's *own* output, never a user's. Green on `c801ab8`: missing coverage, not a defect. The remaining guards in that file were re-read and their bodies are as wide as their docstrings.

**Stated residual, not smoothed over:** with a token configured, the run writes `bearer_token_env_var` and leaves a user's `http_headers` Authorization alone — two sources for one header. Deleting one silently is the defect this round fixed, so the ambiguity is the deliberate cost.

**Could not verify:** Zed and VS Code were not run; the trailing-comma claim rests on reading `serde_json_lenient` and `jsonc.ts`. **opencode's JSONC parser could not be established from a primary source** — only third-party wikis were reachable, which the item-34 rule written this round classifies as never-sufficient — so its row rides the same gate by inference, bounded by the fact that this command writes strict JSON back in every case.

ZacxDev:
## Audit round 3 — fixes

Delta audit read `897c1cc`; fixes produced `c341be4`.

```audit-claims round=3 audited=897c1cc..c341be4
1. Plan-time refusal became ONE shape for all three written files, so --dry-run no longer reports create/ok:true/rc 0 for an AGENTS.md or CLAUDE.md the real run refuses; three new fixtures cover the instruction-file rows, watched red at 897c1cc.
2. A preserved Authorization header is no longer reported as authenticating: the run says a header is present and that it cannot tell whether it resolves, the per-server 401 warning is no longer suppressed by a string the command cannot evaluate, and a Codex bearer_token_env_var is no longer called an Authorization header.
3. --json was closed at the ONE place stdout is written (agentSetupEmitter), giving a documented third payload shape; the exit-2 exception is now stated rather than the text asserting an absolute.
4. Write-independence gained its first guard: TestAFirstWriteFailureDoesNotStopTheLaterOnes, with a fixture whose FIRST write fails; the abort-on-first-failure mutant now dies on that guard's own assertions instead of surviving the package.
5. TestRepeatedRunsAreIdempotent's fixture is parameterised on the token with ownedIsReRendered derived from what the renderer reads, so both codex arms are red at c801ab8.
6. Three round-2 tests are labelled invariant guards and the file header's count corrected 10 to 7.
7. README and Long now name the one exception to "no step is skipped at exit 0" (action: manual).
8. ok, the exit code and the error text all derive from one predicate, and the dry-run arm says what a dry run actually did.
```

🔴 **I re-verified the round-3 🔴 myself.** At `897c1cc` the dry run reported `rc 0 / ok=true / create AGENTS.md` where the real run reported `rc 1 / ok=false / blocked`. At `c341be4` both runs return `rc 1 / ok=false / blocked AGENTS.md / create CLAUDE.md / create .mcp.json` — identical.

## The sweep found six, not four — and three were in guards this round wrote

Asked to sweep all 45 top-level guards for the description-wider-than-body shape, it found six and fixed six. **Three were its own, written in this round** — an idempotency docstring claiming a property that holds for 2 of 7 agents; a usage-error guard asserting only `err != nil`, satisfiable by a regression *into* the silent failure its sibling exists to prevent; and a `os.Stat` sitting behind a helper that had already run the command on the same directory, so it passed on that run's artefact.

🔴 **It also corrected a claim I relayed to you.** Round 1's test-file header says every test in it was watched red at `6340db8`. Measured: **the file cannot compile at `6340db8`** — `srv.Anonymous` and five other symbols the fix introduced do not exist there — so no test in it was watched red *in that form*, and five of its tests are self-described opposite-direction controls that were never labelled.

**Four pre-existing round-1 guards were flagged and deliberately NOT changed**, each needing its own measurement, and recorded in item 35 §9 rather than silently left: a broken-symlink guard that asserts the message but never that the dangling target stayed unmaterialised; a `CODEX_HOME` guard testing one target × one root var for a general claim while discarding `ok` so it can pass vacuously; an absolute-paths guard whose `--check` half filters on name *and* `c.OK` with no counter, so it can assert nothing; and a malformed-config guard using `Contains` where its round-2 sibling compares the whole string.

**`AGENTS.md` needed no budget raise** — it went 30,001 → 29,992 bytes by tightening item 35's trigger rather than extending it.

**Could not verify:** neither Zed nor Codex was run; the header classification rests on their documented key semantics. `--json`'s per-file `reason` for an opaque header does not repeat the orchestration 401 the terminal shows per server — honest, but not symmetric across channels.

ZacxDev:
## Audit round 4 — fixes

Delta audit read `c341be4`; fixes produced `9da323b`.

```audit-claims round=4 audited=c341be4..9da323b
1. --json's fourth silent shape is closed by moving the emitter, not by reclassifying the error: RunE now builds the emitter first, has a single return, and hands every error to envelope, which skips only errors.Is(err, ErrUsage). Five new subtests, red at c341be4.
2. The rejected alternative — tagging the stat failure asUsageError so the stated exit-2 exception became true — is recorded with its reasoning: it would make the sentence true by moving the code to fit it, and would call an environment condition a mistake about the invocation, contradicting items 24 and 26.
3. The dry-run/real-run divergence was reconciled by moving the PROSE, not the code or the test: no test was red, the code was correct and the documents were wrong. TestADryRunCannotSeeAFailureOnlyTheWriteCanProduce is labelled a documentation guard and pins the divergence so removing it fails a test, not only a paragraph.
4. TestDryRunAndTheRealRunAgreeOnEveryAction was renamed to ...OnEveryPlanTimeOutcome, matching its seven fixtures, which all vary something the plan can see.
5. agentSetupHeadline counts blocked rows with the same predicate ok uses, so a fully blocked run no longer opens by claiming it configured the agent.
6. Every universal claim the diff adds or edits is tabulated with the input class it was established over, or is stated as unestablished.
```

🔴 **I re-verified F1 in both directions.** EACCES parent and ENOTDIR now emit a full `error` envelope at rc 1 (126 and 123 bytes) under both `--json` and `--check --json`; a genuine usage error still returns rc 2 with 0 bytes, which is the documented exception. The classification held — the emitter moved, the error's meaning did not.

⚠ My first attempt at that check reported rc 2 / 0 bytes for all three cases and looked like a regression. It was my own error: zsh does not word-split an unquoted `$args`, so the binary received one malformed argument and correctly returned a usage error. Recording it because it is the same trap this PR's briefs warn about, and it produced a confident wrong reading in the reassuring-looking direction.

## Why this round reads differently from the three before it

The last three fix rounds each replaced a false absolute with a narrower absolute that was also false. This round was asked to stop writing absolutes and instead either make the code true and name the guard, or state the enumerated truth and say the enumeration may not be exhaustive. It did that — and then **its own mutation testing falsified one of its new sentences**: it had written that a second `return` in `RunE` bypassing the emitter would go unnoticed; the mutant killed five subtests, and all three copies of the sentence were corrected.

It also caught a **survived mutant that was an instrument failure, not a result** — a `perl -0pi` whose quoted newlines meant nothing was substituted, so the suite ran against unmutated source. Noticed because the verifying `grep` printed nothing, then redone with a replace that asserts the anchor matched exactly once.

**Residuals now stated rather than asserted away:** a failed stdout write, an error cobra raises before `RunE`, and a panic or signal are all outside the payload guarantee; full-disk and read-only-mount divergence is reasoned but unreproduced; `os.Stat` errnos other than EACCES and ENOTDIR were not exercised. The dry-run agreement claim is scoped to one tree state, because a real run changes the tree and a later dry run legitimately plans differently.

ZacxDev:
## Audit round 5 — LADDER STOPPED

Delta audit read `9da323b`. **All six round-4 claims verified fixed. No 🔴, no 🟡.** One 🟢, now fixed in `e8edc91`, so there is no `audit-claims` block: the ladder ends here.

🔴 **A clean round ends the ladder, and running another to confirm it is what the rule forbids.** I am not running round 6.

### Why round 5 reads differently from rounds 2–4

Rounds 2, 3 and 4 each found the previous round had written a **false universal claim** — three in a row, the same shape, each fix replacing a false absolute with a narrower one that was also false. Round 4 was therefore told to stop writing absolutes: make the code true and name the enforcing guard, or state the enumerated truth and admit the enumeration is open.

**It worked.** Round 5 reports it is the first round in four where it could not falsify a universal claim in the delta. Every absolute the diff adds is either enforced by a control-flow shape verified structurally *and* by mutation, or stated as an enumeration whose openness is admitted in the same sentence.

### What round 5 verified rather than accepted

- **Every self-reported mutant re-run independently**, each dying with its own message — including the *rejected* fix (tagging the stat failure `asUsageError`), which kills all five new subtests with "was classified as a usage error … items 24 and 26".
- **Every "red at `c341be4`" claim reproduced**: 5 JSON subtests red, 4 headline arms red, the CONTROL arm and the documentation guard green — confirming the two controls and the doc guard are correctly labelled as non-regression rather than counted as coverage.
- **Two `os.Stat` errnos beyond the two the commit claims** — ELOOP and ENAMETOOLONG — both produce an envelope at rc 1.
- **A fifth silent shape was hunted and not found.** It enumerated every error source reachable from `RunE`, confirmed no `defer`, no recover, no `os.Exit`, and that `cmd.OutOrStdout()` appears exactly once in the command. The three stated residuals were each exercised: `/dev/full` → rc 1 with files written; closed pipe → rc 141; cobra pre-`RunE` → rc 2, all `ErrUsage`.
- Its **mutation instrument was validated** — every edit applied via a replace asserting the anchor matched exactly once, so no SURVIVED verdict was relied on.

### The one finding, and why it was worth fixing

Residual (b) attributed all three pre-`RunE` cobra errors to `enforceUsageExitCodes`. That wrapper only sees `cmd.Args`; the two **flag** errors are tagged by `SetFlagErrorFunc` — and root.go's own comment says `SetFlagErrorFunc` "only sees PARSE errors". Mutating it to return the error untagged drops both to exit 1 while the stray argument stays at 2.

No hole was open — the real property is guarded, and those guards go red on that mutation. The cost was a comment pointing the next maintainer at a function that does not hold the property, **which is the defect class this PR spent four rounds on**. Fixed in `e8edc91`; I verified the attribution against `root.go` myself before writing it.

### Ledger

`round 5 · payload lines changed THIS round: 259 (102 non-comment) · since round 1: 3573`

**Advisory verdict: safe to merge**, once the repo is unfrozen.

PR #532, opened 2026-09-09
fix(read)!: accept a numeric username from the API (#513)

### LINKED ISSUE #513: numeric usernames coerced to numbers failing the CLI

ZacxDev:
Fixes #513 — reported by **@Rochet2** on 2026-08-30.

## The repro

A Civitai username may be entirely digits (e.g.
https://civitai.com/user/2802169344506), and such a value was **reported**
arriving as a **bare JSON number**, not a string (see "Documentation
corrections" below for what a live probe could and could not confirm):

```json
{"id": 136456589, "username": 2802169344506, "baseModel": "Illustrious"}
```

`pkg/civitai` typed that field as `string`, and the CLI decodes a whole page in
one `json.Unmarshal` (`getInto`, `pkg/civitai/read.go:92`). So **one** such
uploader anywhere on the page failed the entire request:

```
$ civitai images search
Error: unexpected response from /api/v1/images (status 200): {"items":[…
$ echo $?
1
```

Nothing printed, and the message names neither the field nor the cause.

## The fix

A new exported `civitai.FlexString` — a defined string type whose
`UnmarshalJSON` accepts a JSON **string** or a JSON **number** — applied to the
seven `Username` fields in `pkg/civitai` (`apps.go`, `articles.go`,
`collections.go`, `images.go`, `models.go`, and both sites in
`tags_creators_users.go`).

- **A number keeps its LITERAL digits**, via `json.Number`. This is the whole
  point: a `float64` round trip renders `2802169344506` as
  `2.802169344506e+12`, and loses the last digit of anything past 2^53 — either
  way the username matches no account and cannot be pasted back into
  `--username`.
- **`null` stays a no-op leaving `""`** — the behaviour a plain `string` field
  already had, and `CollectionUser.Username` is documented nullable server-side.
- **A boolean, array or object is still a hard error.** This absorbs one
  measured server quirk; it is not a lenient decoder that hides schema drift.
- **A number literal is kept verbatim, not normalised**: `1.0` → `"1.0"`.

`pkg/civitai/read.go`'s `getInto` and `internal/cmd/read.go` are **untouched** —
zero overlap with #526.

### Out of scope, deliberately

The three `Username string` fields in `internal/appapi` (`appblocks.go:172`,
`:993`, `forgejoUsername` at `:1340`) are **not** converted. That package talks
to the App-Blocks service, not the public read API, and there is no evidence of
this coercion there.

### `--json` is unaffected

The read commands emit the **raw server bytes** through `emitJSON`
(`internal/cmd/read.go:95`); nothing in this repo marshals these structs, so
nothing re-encodes the field and the number stays **unquoted** on stdout.
Asserted rather than assumed, in both directions, by
`TestImagesSearchJSONKeepsTheUnquotedUsername`.

### Server side

The API emitting an unquoted number for a field its own docs describe as a
string is **tracked separately on the API side**. This PR is the client-side
accommodation; until that lands, every consumer of these endpoints has the same
problem.

## Red-at-base / green-at-HEAD

Measured in a detached `git worktree` at `origin/main` (`46c928a`), with the new
test files copied in. The three seam tests read the field back through
`string(item.Username)`, which compiles against **both** a `string` field and a
`FlexString` one, so they are a true red/green pair rather than a compile error.

| test | at `46c928a` | at HEAD |
| --- | --- | --- |
| `internal/cmd` · `TestImagesSearchDecodesANumericUsername` | **FAIL** — `unexpected response from /api/v1/images (status 200)` | PASS |
| `internal/cmd` · `TestImagesSearchJSONKeepsTheUnquotedUsername` | **FAIL** — same | PASS |
| `internal/cmd` · `TestUsersGetDecodesANumericUsername` | **FAIL** — `… /api/v1/users (status 200)` | PASS |
| `internal/cmd` · `TestUsersGetListsANumericCandidate` | **FAIL** — same | PASS |
| `internal/cmd` · `TestModelsSearchDecodesANumericCreatorUsername` | **FAIL** — `… /api/v1/models (status 200)` | PASS |
| `pkg/civitai` · `TestSearchImagesDecodesANumericUsername` | **FAIL** — same | PASS |
| `pkg/civitai` · `TestSearchUsersDecodesANumericUsername` | **FAIL** — same | PASS |
| `pkg/civitai` · `TestSearchModelsDecodesANumericCreatorUsername` | **FAIL** — same | PASS |

Commands:

```
git worktree add --detach $BASE origin/main
cp internal/cmd/numeric_username_test.go $BASE/internal/cmd/
# seam tests only, extracted so the file compiles against the string field
go test ./internal/cmd/ -run '…' -v -count=1   # in $BASE  -> FAIL (5/5)
go test ./pkg/civitai/  -run '…' -v -count=1   # in $BASE  -> FAIL (3/3)
```

**Labelled as INVARIANT GUARDS, not regression coverage** (green at base too, or
structurally un-runnable there): `TestFlexStringNullLeavesTheZeroValue`,
`TestFlexStringRejectsWrongShapes`, and the "plain" / "escaped unicode" rows of
`TestFlexStringAcceptsAJSONString`. The remaining `FlexString` unit tests
**cannot** be red at base — the type does not exist there, so they are a compile
error rather than a failing assertion. Their standing is the mutation matrix
below.

## Mutation matrix

Six mutants of `FlexString.UnmarshalJSON`, applied one at a time to an untouched
copy of the file, each run with `-count=1`. Every one was killed by a **named**
assertion, not incidentally. The file was restored afterwards and its md5
re-checked against the pre-sweep copy.

| # | mutant | killed by | assertion message |
| --- | --- | --- | --- |
| M1 | number decoded as `float64`, rendered with `%v` | `TestFlexStringPreservesBigIntegerDigits` (**both** rows) | `decoded to "2.802169344506e+12", want the literal digits "2802169344506"` |
| M2 | the number branch deleted entirely | `TestFlexStringPreservesBigIntegerDigits` | `decoding the bare number 2802169344506 failed: … — this is #513 itself` |
| M3 | `null` returns an error instead of a no-op | `TestFlexStringNullLeavesTheZeroValue` | `null must not be an error — CollectionUser.Username is documented nullable` |
| M4 | final `return fmt.Errorf(…)` replaced by a lenient raw-bytes fallback | `TestFlexStringRejectsWrongShapes` (4 rows) + `TestFlexStringErrorNamesTheOffendingKind` | `decoding {"name":"kiwi"} succeeded with … — a wrong shape here is schema drift` |
| M5 | string branch hands back raw bytes without unquoting | `TestFlexStringAcceptsAJSONString` (4 rows) | `decoded "pineapple-lamp" to "\"pineapple-lamp\"", want "pineapple-lamp"` |
| M6 | `float64` + `strconv.FormatFloat(…, 'f', …)` — the **subtle** one | `TestFlexStringPreservesBigIntegerDigits/beyond_float64's_mantissa` **only** | `decoded to "9007199254740992", want the literal digits "9007199254740993"` |

🔴 **M6 is why there are two big-integer fixtures.** The reported username
(`2802169344506`) is under 2^53, so it is represented exactly and this mutant
**survives** a test that uses only it — measured, that row passed. `2^53 + 1`
(`9007199254740993`) is the one that kills it. Fixtures are pairwise distinct
and distinct from every constant an assertion names, so a mutant that hardcodes
a literal cannot survive either.

Go's build cache is keyed on **content**, not mtime, so the stale-bytecode
failure mode does not apply here; `-count=1` was used regardless, and M6 (a
one-line change of the same shape as M1) is the positive control that the sweep
was really re-running the mutated code.

## Consumers touched

Comparisons against `""` still compile; passing the field to a `func(string)`
does not. All 13 call sites converted with `.String()`:

`internal/cmd/apps.go` ×2, `articles.go` ×2, `collections.go` ×2, `models.go`
×2, `images.go` ×2, `creators.go` ×1, `users.go` ×4 (including the
`strings.EqualFold` exact-username match in `users get`).

Not touched: `internal/cmd/whoami.go` and `app_dev_tunnel.go`, which read
`appapi.Identity.Username` — out of scope, as above.

## Docs

- **README Troubleshooting** gains an `unexpected response from` row under
  *Everything else*, linked to *Exit code 1*. Its left column quotes a string
  that really exists in `pkg/civitai/read.go`, which
  `TestREADMETroubleshootingSymptomsExistInTheSource` requires.
- **AGENTS.md item 37** — a routing question plus a pointer; the body is
  `claudedocs/decisions/37-numeric-username.md`, registered as **born-split** in
  `agents_split_preserved_test.go`.
- Nothing generated or fixture-frozen needed a change: exit codes are unchanged,
  so `internal/cmd/exitcodes_doc.go` is untouched.

**Renumbered 36 → 37 on rebase.** #533 merged first and took item 36
(`agents-block-per-project`), and AGENTS.md's rule is that the PR merging second
renumbers **its own** items. Every site moved: the index clause, the numbered
trigger, `claudedocs/decisions/36-numeric-username.md` → `37-…` (`git mv`) and
its own `**Item 37.**` header, the `bornSplitItems` entry in
`agents_split_preserved_test.go`, and the `AGENTS.md item 37` comment in
`pkg/civitai/flexstring.go`. Main's item 36 is untouched and both items exist in
order.

**The byte pressure is gone.** #535 evicted content, so `AGENTS.md` on `main` is
29,827 B; this PR's item adds 201 B → **30,028 B against `agentsMaxBytes` =
30,500 — 472 bytes of headroom.** Neither `agentsMaxBytes` nor
`agentsMaxBytesCeiling` is touched. (The earlier revision of this description
warned about 7 bytes of headroom under a 30,200 ceiling; that measurement is
superseded.)

## 🔴 Breaking change for Go importers of `pkg/civitai`

`pkg/civitai` is the **importable** read/download SDK (`api.go`: "so external
modules can import `github.com/civitai/cli/pkg/civitai` and drive the public API
directly"). Seven exported struct fields change type `string` → `FlexString`:

| struct | file |
| --- | --- |
| `ListingCreatorChip.Username` | `pkg/civitai/apps.go` |
| `ArticleUser.Username` | `pkg/civitai/articles.go` |
| `CollectionUser.Username` | `pkg/civitai/collections.go` |
| `ImageItem.Username` | `pkg/civitai/images.go` |
| `Creator.Username` | `pkg/civitai/models.go` |
| `CreatorItem.Username` | `pkg/civitai/tags_creators_users.go` |
| `UserItem.Username` | `pkg/civitai/tags_creators_users.go` |

`FlexString` is a **defined string type**, so comparisons and assignments
against untyped string constants still compile unchanged. What stops compiling
is passing one where a `string` is required. **Measured, not asserted** — an
out-of-tree module built against this branch:

```
# still compiles
it.Username != "" ;  c.Username == "alice" ;  len(it.Username)
string(it.Username) ;  it.Username.String()

# no longer compiles (go build, verbatim)
cannot use it.Username (variable of string type civitai.FlexString) as string value in variable declaration
cannot use it.Username (variable of string type civitai.FlexString) as string value in argument to takesString
cannot use it.Username (variable of string type civitai.FlexString) as string value in array or slice literal
```

Importers convert explicitly at those sites: `string(x.Username)` or
`x.Username.String()`.

The subject is `fix(read)!:` with a `BREAKING CHANGE:` footer, so
`.goreleaser.yaml`'s changelog carries it as a break rather than an ordinary
fix — the scar `claudedocs/handoff-dogfood-2.md` records against #267. The PR
**title** carries the `!` too, because `squash_merge_commit_title` is
`COMMIT_OR_PR_TITLE`.

**No README `BREAKING:` note is added, deliberately.** The README documents the
CLI, not the Go SDK: `pkg/civitai` appears nowhere in it (its "SDK packages"
section is the npm `@civitai/*` App SDK), and no CLI behaviour, output column or
exit code changes here. A `BREAKING:` row there would announce a break to an
audience that cannot experience it.

## Documentation corrections in this revision

- **`UserItem.Username`'s comment claimed a workflow the code does not
  provide.** It said the shape `civitai users get 2802169344506` "has to
  decode". It does not: `internal/cmd/users.go` routes any
  `strconv.Atoi`-parsable argument to `?ids=` — an **id** lookup — never
  `?query=`, so a user whose *username* is all digits is not reachable by name
  through that command, before or after this PR. Verified directly
  (`Atoi("2802169344506")` → `?ids=`; `Atoi("some-username")` → `?query=`). The
  comment now names paths that genuinely exercise the decode, including
  `users get <name>`'s candidate list (`TestUsersGetListsANumericCandidate`).
- **The "server sends a number" claim is softened to what was measured.** A live
  probe on 2026-09-09 could **not** reproduce the coercion: `/api/v1/users?query=`
  returned every all-digit username **quoted** across six queries including
  13-digit ones. `/api/v1/images` — the surface #513 was filed against, which
  serialises the embedded user differently — could not be exercised at all, as
  no all-digit account found has public images. The fix stands on a narrower
  justification that does not need the repro: **a client must not hard-fail a
  `200` whose body it can read.** Softened in the decision doc, in
  `flexstring.go`'s doc comment and in the README row.
- **A residual the doc did not bound.** Nothing measures *why* the server
  emitted a number, so "only the `username` field" is an inference. If the
  coercion is a property of any all-digit *string* column, `TagItem.Name` (a tag
  named `2024` is plausible) fails identically with the same opaque error. Named
  in the decision doc's residuals; the code is deliberately **not** widened —
  the trigger to convert more fields is an observation, not the symmetry of the
  argument.

## Gate

Re-measured after the rebase onto `a2fbd71`:

- `make ci` — green; **22 packages** reported (21 `ok` + 1 `[no test files]`), build ok.
- `make lint` (`nix-shell -p golangci-lint`) — **0 issues**. Instrument validated:
  a deliberately bad temporary file made it report `2 issues` and exit 1, then
  clean again after removal.
- `./scripts/ci-shallow.sh` — `ok=21/21, failing-tests=0, timeouts=0` in a depth-1
  clone (the tier where `TestBornSplitItemsWereNeverInAgents` skips).
- `gofmt -s -l .` — clean. Instrument validated: a misformatted temporary file was
  listed, so the zero is a real zero, not "checked zero files".
- The seven AGENTS.md/layout guards — **31/31 named tests PASS**:
  `agents_size_test.go`, `agents_split_preserved_test.go`,
  `agents_evidence_test.go`, `agents_trigger_test.go`, `agents_index_test.go`,
  `agents_xrefs_test.go`, `layout_ledger_test.go`.

https://claude.ai/code/session_01W9KZLrS278bRGqjDLDkNUR


ZacxDev:
**Merged-tree check against #526** (which edits `internal/cmd/read.go` and `pkg/civitai/read.go` — the regions this PR deliberately does not touch).

Disjoint files are not safety on their own, so this was measured rather than argued: `origin/pull/526/head` merged into this branch in a scratch worktree. Clean merge (`ort`, 3 files, no conflict), and the **full suite is green on the merged tree** — `go test ./... -count=1`, 21/21 packages ok, 0 failures.

Merge order between the two PRs is therefore unconstrained.

PR #541, opened 2026-09-09
fix(read): repair on decode failure, not before it — and strip what the error snippet prints

ZacxDev:
Follow-up to the adversarial audit of #526. **#526's code is correct and its regression test was watched red at base — nothing here re-audits it.** What this PR fixes is the set of *our* claims that #526's correct change falsified, plus the one pre-existing hole that fixing them would otherwise have widened.

Every finding below was re-verified against the tree before acting on it.

## Findings closed

| # | What | Where |
| --- | --- | --- |
| 1 🔴 | `readJSONNote`'s doc comment asserted the repair "is not a behaviour of this group", with a measurement (exit 1, empty stdout, `emitJSON` never entered). **Re-measured: exit 0, table renders, `emitJSON` IS entered.** Every clause false. Comment rewritten; a guard now pins it. | `internal/cmd/read_help.go`, `internal/cmd/read_json_note_test.go` |
| 2 🟡 | README's Troubleshooting row for `unexpected response from` named one known cause (#513) and claimed the snippet is "the server's own body" — false, because `getInto` reassigned `raw` to the sanitized bytes before `snippet(raw)` ran. Both halves fixed; #525 named as the second known cause. | `README.md` |
| 3 🟡 | `json.Valid` over every successful 2xx body, on top of the `Unmarshal` that walks the same bytes. Replaced by unmarshal-first + sanitize-and-retry. | `pkg/civitai/read.go` |
| 4 🟡 | The inner `json.Valid(fixed)` guard was untested — a mutant assigning unconditionally SURVIVED a full green run. The retry-fails path is now tested. | `pkg/civitai/read_repair_test.go` |
| 5 🟢 | Two result types documented `Raw` as "the raw response body", and `numeric_username_test.go` as "a raw passthrough of the server's own bytes". For a repaired body neither is true. Fixed, with the one accurate statement on the exported `EscapeJSONStringControlChars`. | `pkg/civitai/{collections,articles,read}.go`, `internal/cmd/numeric_username_test.go` |
| 6 🟢 | `postInto`'s comment enumerated what it mirrors from `getInto`; sanitization was a fourth axis on which it did not. **Made to mirror, as shared code** (`decodeBody`) rather than a stated divergence — so the enumeration is true by construction. | `pkg/civitai/hashes.go` |
| 7 🟢 | The one-line delegating wrapper kept only so `read_r2_test.go` compiled, and the byte-identical duplicate test in two packages. Both removed. | `internal/cmd/read.go`, `internal/cmd/read_r2_test.go` |

## The design decision: **(a)** — `snippet()` is filtered, `errorLine` is not

`cmd/civitai/main.go` prints `Error: <err>` to stderr with no filter, and `snippet()` embeds up to 500 bytes of a server body verbatim. That hole **pre-dates #526**; finding 3's fix would have widened it back to the 2xx path, because restoring the wire bytes to the snippet restores the wire's escapes too.

**Chosen (a): `snippet()` routes the server's bytes through `internal/saferune`.** The reason is that every argument `snippet` is ever given is server-supplied — the undecodable 2xx body, the API's `{"error":…}`/`{"message":…}`, a zod issue, a retry-exhaustion body, a `FlexString` value — and none of it is anything the user typed. `snippet` was already a terminal-concern helper; its existing job is bounding length "so a huge response can't flood the terminal".

Two alternatives were rejected for stated reasons, not by preference:

- **Filtering at the print site (`errorLine`)** would apply the class to *every* error, including the many that echo what the user typed. That is exactly the #393 regression `TestSafeTermIsNeverAppliedToUserTypedInput` exists to prevent.
- **Re-deriving the class inside `pkg/civitai`** would be a second table — the other half of #393. No third table was created; `internal/saferune` remains the single owner.

🔴 **The cost, stated rather than hidden:** this makes `pkg/civitai` the first thing under `pkg/` to import `internal/`, and an SDK consumer that is not a terminal now reads an error string with those runes already removed. Two things bound it: the strip touches only the human-readable tail of an error message, and `Raw` — the byte path a consumer would parse — is returned unstripped. That trade is written into the code and into item 38, so the next person can reverse it deliberately.

## Verification

`make ci` green (21 packages). `make lint` **0 issues**, and the gate was negative-controlled: a deliberate raw `ESC` in a string literal produced `ST1018 … (staticcheck)` and `make lint` exited non-zero. `gofmt -s -l .` clean over 469 `.go` files. `./scripts/ci-shallow.sh` PASSED against the committed SHA `f936cfc` (21/21).

**Red at base** (`517fc76`, i.e. `origin/main`), green at HEAD:

| test | at `517fc76` |
| --- | --- |
| `TestGetIntoErrorSnippetQuotesTheWireBytesNotTheRepairedOnes` | RED — snippet carried `Pony\rModel`, the CLI's repair |
| `TestReadErrorSnippetStripsTerminalControlRunes` | RED — `ESC` and U+200E reached the error string |
| `TestPostIntoRepairsControlBytesLikeGetInto` | RED — `postInto` did not repair |
| `TestDecodeBodyIsSharedByGetIntoAndPostInto` | RED — 0 call sites found |
| `TestReadJSONNoteDescribesTheRepair` | RED on the wording pin |
| `TestReadJSONNoteCommentNamesItsGuard` | RED — comment named no guard |
| `TestReadErrorReachesStderrWithoutTerminalControlRunes` (seam) | RED — `ESC[1A ESC[2K` reached the exact line `main()` writes |

Two honest labels:

- `TestGetIntoUnrepairableBodyReportsTheOriginalBytes` **passes at `517fc76`**. It is an **invariant guard**, labelled as one in source, and is not counted as regression coverage. Its value is the mutant it kills.
- `TestReadJSONNoteDescribesTheRepair`'s behavioural halves also pass at `517fc76` — because #526 is what made them pass. Measured one commit earlier at `5557b6a`: RED, `unexpected response from /api/v1/models (status 200)`. So the guard observes the transition in both directions.

**Mutation matrix** — each mutant applied to the real source, run, reverted; each killed by *that guard's own* message:

| mutant | killed by |
| --- | --- |
| `decodeBody`: assign the sanitized bytes unconditionally | `TestGetIntoUnrepairableBodyReportsTheOriginalBytes` |
| `decodeBody`: `snippet(fixed)` instead of `snippet(raw)` | `TestGetIntoErrorSnippetQuotesTheWireBytesNotTheRepairedOnes` |
| `snippet`: drop `saferune.Strip` | `TestReadErrorSnippetStripsTerminalControlRunes` **and** the seam test |
| `postInto`: stop sharing `decodeBody` | `TestPostIntoRepairs…` **and** the ledger ("found 1 … expected at least 2") |
| add a third `decodeBody` caller (ledger GREW) | `TestDecodeBodyIsSharedByGetIntoAndPostInto` |
| reword one word of `readJSONNote` | `TestReadJSONNoteDescribesTheRepair` |
| remove the guard's name from the doc comment | `TestReadJSONNoteCommentNamesItsGuard` |

Both AST-based guards assert a **minimum count** (files parsed, call sites found) so a scan that matches nothing fails instead of passing with a reassuring zero — both controls were watched to fire.

## Known residuals (in item 38, not hidden here)

- **The README's prose is not machine-checked against the behaviour.** `readme_troubleshooting_test.go` is a drift detector over symptom *strings*; no symptom string changed, which is exactly why it did not catch the false clause. The behaviour is now pinned; that the row still *describes* it is not. Widening the detector to compare a row's explanation against runtime behaviour is a much larger, different guard and was **not** attempted rather than attempted badly.
- `snippet` still truncates by **byte** at 500, so it can split a multi-byte rune. Pre-existing, unchanged; not a terminal-control hazard since the strip runs before the bound.
- `saferune` keeps `\n`/`\t` deliberately, so a server body can still put a newline in an error line. `indentContinuation` answers that on the renderers it covers; the error path has no equivalent.

## Also

The perf numbers in the audit were **not** copied into the code. Re-measured here (synthetic 1.50 MB / 2185-item page, go1.25, `-benchtime 300x -count=5`): `json.Valid` 4.9–6.4 ms vs `json.Unmarshal` 12.9–26.4 ms — a quarter to a half again, not the audit's 50–100%. The comment quotes my numbers with their scope and says the ratio is not portable; the load-bearing claim (zero extra work on a body that decodes) is structural, not measured. The scratch benchmark is not in the tree (this repo has no benchmarks and one was not worth introducing for this).

No published exit code moved, and the seam test asserts it: a 404 read still exits `4` with the filtered message (AGENTS.md item 7).

https://claude.ai/code/session_01Gx6N59iqHtBT752mc3DnJB

---

# Round 3 (`31d5ba3`) — four findings from the blind audit of round 2

🔴 **Read this section before the one above it: round 2's body contains claims round 3 falsified, and they are corrected here rather than quietly edited out.** Every one of round 2's six findings was *a claim the tree made that the tree did not support* — and round 2 was itself a fix round. That is the measured pattern on this branch, so round 3's rule was: either make the statement true and name the guard enforcing it, or state the enumerated truth and admit the enumeration is open.

## Corrections to the round-2 body above

| round-2 claim | status |
| --- | --- |
| "**No published exit code moved**, and the seam test asserts it" (last line) | 🔴 **FALSE.** One did — see F4 below. And the 404 seam test could never have seen it: a 404 is classified from the *status*, so no message filter can move it. The sentence was both wrong and vacuously supported. |
| Finding 1: "**Every clause false**" | Over-claimed. Two of the four clauses are still true at HEAD; what #526 falsified is the *conclusion*. Corrected clause by clause in `read_help.go`, both premises re-measured. |
| Finding 5: "Two result types … Fixed" | The sweep was **not** complete — 7+ sites were still live, including `ModelSearchResult` and `ImageSearchResult`, the exact types #525 bites. See F1 below. |
| "`gofmt -s -l .` clean over 469 `.go` files" | Now **471** (two new guard files). Still clean, still positive-controlled. |
| "`ci-shallow.sh` PASSED against `f936cfc`" | Re-run at `31d5ba3`: **21/21 ok**, `HEAD=31d5ba3e56c3`, `shallow=true`. |

## F4 🔴 — the strip became a classifier, and exit 6 became exit 2

`readError` handed `snippet()`'s output to `isDeepPagingCap`, whose own comment says the match is *"deliberately narrow so a real rate-limit 429 is never misclassified as a usage error"*. `snippet()` began stripping `Default_Ignorable` runes **on this branch**, and removing runes can only **widen** a substring match — in exactly that direction. Same 429 body, one U+00AD inside "many":

| tree | result |
| --- | --- |
| `517fc76` | `ErrRateLimited` — **exit 6** |
| `f936cfc` (round 2) | `ErrBadRequest` — **exit 2** ← the regression |
| `31d5ba3` (HEAD) | `ErrRateLimited` — **exit 6** |

Contrived for Civitai's own backend; **not** contrived for the proxy, CDN and captive-portal 429s `readError` also handles. AGENTS.md items 7 and 24 make this a published contract.

**Fix:** `readError` keeps `wireMsg` before the `snippet` call and classifies on it. "Narrow" is a claim about what the *server* sent, so the classifier is asked about what arrived. The display is unchanged.

🔴 **This moves a second exit code, deliberately and in the opposite direction, and it is stated rather than discovered later.** `snippet` also truncates at 500 bytes — the same hazard in a second shape. Before this, a genuine cap message whose phrase fell past the display budget was invisible to the classifier and exited 6, spinning a scripter's backoff loop on a structurally doomed request. It now exits 2. Pinned as half (c) of the same test. A >500-byte error body is ordinary, but no such 429 has been observed from the API — the case is constructed.

**Scope, measured not assumed:** `grep` over every `kind =` assignment and every `snippet(` call site in `pkg/civitai` — `read.go:290` is the *only* place where the text of an error picks a classification. Every other branch classifies from the status alone (`statusKind`); every other `snippet` caller is display-only.

## F1 🟡 — a cross-reference true of 2 of its 8 referents, closed as a count of 3

`getInto` said *"see the note on Raw in the result types"*. Two of the eight types with a `Raw []byte` field carried the note; the other six still said "the raw response body" / "the raw body" — the byte-identity claim the same branch was correcting elsewhere. The item file closed the sweep as *"All three now say what is true"*; at least seven more were live.

**A count is the wrong instrument for a sweep**, so the fix is not a better count. All eight now disclaim byte identity, and `TestRawDocCommentsDisclaimByteIdentity` is a **bidirectional ledger** of `Raw`-bearing types that requires the disclaimer on each. Also corrected outside what that ledger can see: `ArticleDetail.Content` (kept its point about HTML rendering, dropped the byte word) and `flexstring_test.go` — verbatim the comment round 2 corrected in `numeric_username_test.go` and missed here.

## F2 🟡 — two prose statements of one set, both stale at once

`pkg/civitai`'s `snippet` became the **third** non-test caller of `internal/saferune`. `AGENTS.md:170` still said the other two "**both**" call it; `saferune.go:23` still said "**TWO PACKAGES** ASK THE SAME QUESTION". Both survived a green suite, a clean lint run *and* a blind audit of the branch that falsified them.

The third caller asks a **third question** — *"what may an error string carry"*, because `main.go` prints `Error: <err>` to stderr with no renderer in front of it — so the package doc states the question, not just a bigger number. `TestSaferuneCallersAreLedgered` fails when the importer set grows **or** shrinks, requires both texts to name every ledgered call site, and derives the count **word** from the ledger's own length so the numeral cannot drift from the membership.

AGENTS.md headroom: **175 → 58 bytes** under the 30,500 ceiling (`agents_size_test.go` green). Verified, as asked.

## F6 🟡 — the comment written to stop comments going stale cited a test that has never existed

`read_json_note_test.go` named a guard that appears nowhere in this repo or in any commit of its history. `TestReadJSONNoteCommentNamesItsGuard` pins only the **forward** direction (the doc comment contains the guard's name), so a wrong name in the same file was structurally invisible.

`TestItem38CommentsCiteTestsThatExist` is the reverse direction: every `Test…` identifier in a comment in item 38's files must resolve to a test function declared **somewhere in the module** (these comments legitimately cite guards in other packages). The dead identifier is kept only as a string literal in a negative control, never in prose — otherwise the test flags its own comment (it did, on first run).

## Also fixed — same class, smaller

- `read_repair_test.go`'s header claimed every test in it was RED at `517fc76`; one is an **invariant guard** the same file labels correctly 100 lines later. It also deferred the matrix to "the PR body", which is not in the tree. **The matrix is now in the file, measured by running it against a `git archive 517fc76` tree** — not reasoned about. That measurement corrected my own first draft, which had guessed the new test PASSES at `517fc76`; it is red there too, on the *opposite* assertions.
- `read_help.go`'s "EVERY CLAUSE OF IT MEASURED FALSE" — corrected clause by clause. Both surviving premises re-measured: `emitJSON` is downstream of `getInto` at all 16 of its call sites, and `encoding/json` returns `invalid character '\r' in string literal`.
- `read_json_note_test.go:13` named the wrong constant (`crInAName` vs `crBodyModelName`).
- `bornSplitItems` claimed "mutation matrices" the item file has never had.
- README: the "another reason" with no first reason, the "raw … a stable passthrough" opening that let a README-only reader still conclude byte identity, and the insertion sitting between the "Two properties" bullets and the sentence referring back to them.

## Round-3 verification

`make ci` green (**21 packages ok**, 1 with no test files). `make lint` **0 issues** — and the instrument was validated by a real failure first: my initial fixture used a literal U+00AD and lint reported `ST1018 … (staticcheck)`, exit non-zero. `gofmt -s -l .` clean over **471** `.go` files, positive-controlled (a deliberately misformatted file was listed, then restored). `./scripts/ci-shallow.sh` **21/21 PASSED** against the committed SHA `31d5ba3e56c3`.

⚠️ **Local golangci-lint is 2.13.2; CI pins v2.12.2** (`.github/workflows/ci.yml:161`). The zero above is my version's, not CI's.

### Red/green

| test | pre-change | at HEAD |
| --- | --- | --- |
| `TestDeepPagingCapClassifiesOnTheWireMessageNotTheStrippedOne` (a) | **RED at `f936cfc`** — "lost ErrRateLimited (exit 6)" | green |
| `TestDeepPagingCap…` (c), the 500-byte half | **RED at `f936cfc`** — "the classifier is reading the truncated string" | green |
| `TestThrottle429KeepsExitSixThroughTheMessageFilter` (seam) | **RED at `f936cfc`** — "exits 2, want 6" | green |
| `TestRawDocCommentsDisclaimByteIdentity` | **RED before the sweep** — 6 of 8 types | green |
| `TestSaferuneCallersAreLedgered` | **RED before the doc fixes** | green |
| `TestItem38CommentsCiteTestsThatExist` | **RED before the citation fix** | green |

### Mutation matrix — each applied to real source, run, reverted; each killed by *that guard's own* message

| mutant | killed by |
| --- | --- |
| `wireMsg := snippet([]byte(msg))` (== the pre-fix shape) | `TestDeepPagingCap…` halves (a) and (c) **and** the seam test |
| `isDeepPagingCap` → `return false` | `TestDeepPagingCap…` half (b) — the positive control on the classifier |
| `snippet` drops `saferune.Strip` | `TestDeepPagingCap…`'s display assertions |
| revert `ModelSearchResult`'s comment to the byte claim | `TestRawDocCommentsDisclaimByteIdentity` |
| add a 9th `Raw []byte` type, unledgered (GREW) | same, by name |
| ledger names a type that does not exist (SHRANK) | same, via its minimum-count control |
| delete `` `snippet` `` from AGENTS.md's Layout entry | `TestSaferuneCallersAreLedgered` |
| revert `saferune.go` to "TWO PACKAGES" | same, on the count |
| remove `pkg/civitai` from the ledger (GREW) | same, on membership |
| `pkg/civitai` stops importing `saferune` (SHRANK) | same, on membership |
| reintroduce the original dangling citation | `TestItem38CommentsCiteTestsThatExist` |
| typo a guard name in a *different* file (`read_help.go`) | same — proves reachability beyond its host file |

🔴 **One mutant SURVIVED on the first attempt and the instrument was fixed, not the score.** Deleting `` `snippet` `` from AGENTS.md's Layout entry passed, because the first cut searched the *whole file* and AGENTS.md also contains the prose phrase "what an error **snippet** prints" in item 38's trigger. The check is now scoped to the `internal/saferune` bullet and requires the **backticked** identifier; the mutant then died. The same class bit `TestRawDocCommentsDisclaimByteIdentity`, which reported 8 false "missing" verdicts until the doc text was whitespace-normalised (doc comments hard-wrap, so the sentence lands with a newline at a different word in every file). Both are recorded in the tests' own comments.

## Not in scope, and not made worse

The #393 structural guard's blindness to `pkg/civitai`'s `saferune.Strip` call, and the `--help` rune budget, are untouched — being filed separately. Neither was widened: no `--help` text changed, and no `safeTerm` call site moved.

## What I could not fully establish

- **`saferune`'s "the strip is correct here because every argument is server-supplied" is a judgement no test enforces.** `TestSaferuneCallersAreLedgered` proves membership and naming; it cannot tell whether a fourth caller is *entitled* to the unconditional strip. Stated as a residual in item 38.
- **The byte-identity enumeration is OPEN, not closed** — deliberately, and that is the whole lesson of round 2's count of three. The new ledger reads doc comments on struct fields; the detail getters return raw bytes as a bare `[]byte` with no doc-comment slot, and narrative prose anywhere in the package is not read.
- ~~**`TestItem38CommentsCiteTestsThatExist` is scoped to four files, not the package.** Measured: 31 comment-cited names across `internal/cmd` do not resolve…~~ **Superseded by round 4:** that 31 was measured with a package-local resolver this guard does not ship, and the scan now covers twelve files plus the decision doc. Corrected numbers and the real reason for the scope limit are in the Round 4 section below.

---

# Round 4 — `39389c1` (audited range `31d5ba3..39389c1`)

Numbering follows this tree's own: `f936cfc` = round 2, `31d5ba3` = round 3, this commit = round 4.

The adversarial audit of `31d5ba3` returned **nine findings, zero of them runtime defects**. Every one was a claim the tree made that the tree did not support — which is the same shape as round 3's, and round 3's fixes are where they came from. All nine are addressed. In every case the fix is a guard that pins the claim, not a better sentence.

**Every number below was re-derived on this tree with the instrument that ships, not inherited from the audit brief.**

## Findings closed

| # | Finding | What changed |
| --- | --- | --- |
| 🔴-1 | `saferune_callers_ledger_test.go`'s `question` field advertised rename-safety it did not have — it was a plain `string` used as a substring needle, bound to no declaration. | `checkQuestionsResolve` resolves every ledgered `question` to a top-level receiver-less `FuncDecl` in its own `pkgPath` before any prose is searched. Comment rewritten to say what the code does. Residual in item 38 corrected. |
| 🔴-2 | The "measured, not reasoned" matrix in `read_repair_test.go` disagreed with the measurement, at **both** ends, and said "one of the five tests" over a file declaring six. | Both ends re-run against real `git archive` trees and rewritten from what the run prints. The header now states **no count**; `TestReadRepairMatrixNamesEveryTestInThisFile` pins membership instead. |
| 🟡-1 | `isDeepPagingCap`: "any transformation applied first can only widen it" — false, and contradicted by this PR's own half (c). | Both directions named: the strip can only widen, the 500-byte bound can only narrow, both are wrong to classify on. |
| 🟡-2 | Classifying pre-`snippet` also un-bounded the match: `msg` is the **entire body** for a non-JSON 429, capped only by `maxResponseBody` (64 MiB). | `classifyWindow` + `maxClassifyMessage = 8 KiB`, a budget deliberately separate from the display budget. Pinned red-before/green-after. **No false positive was demonstrated — this is not a fix for an observed defect**, and the number is generous rather than derived (`read.go` says so explicitly). Round 4 called it *containment*; round 5 measured it and that word was wrong — see below. |
| 🟡-3 | The "31 unresolved comment-cited names" justifying a scope limit was measured with a **package-local** resolver this guard does not ship. | Re-measured with the shipped instruments; scope limit kept on a corrected and much better reason (below). |
| 🟡-4 | The evidence header and `item38CommentedFiles` declared different file sets, disagreeing in **both** directions; `pkg/civitai/read.go` — item 38's primary code file — was uncovered. | Ledger widened to 12 repo-relative files **plus the decision doc**, scanned as text. `TestItem38FileLedgerMatchesTheDecisionHeader` pins the two sets equal in both directions. |
| 🟡-5 | "one measurement table and no mutation matrix" was written by the same commit that added the second and third tables. | Corrected in both places, and `TestItem38EvidenceTableGeometry` now asserts table placement and count per section. The "no mutation matrix" half was correct and is kept. |
| 🟢-1 | `raw_doc_ledger_test.go`'s control fired on every SHRINK, so the ledger's own `SHRANK:` guidance was unreachable and a real finding was labelled "not a finding". | Now a **hard floor of one**; the ledger comparison owns every shrink. |
| 🟢-2 | README: `generate` / `workflows list --json` "pass through the raw *orchestrator* reply" — they route through `writeRawJSON`, which `json.Indent`s. | Split into the two byte changes: the re-indent **does** apply; the repair **never runs** there. |

## 🟡-3 re-measured, and why the scope limit stays

With `repoTestFuncNames` + `testIdentRe` — the instruments the guard actually ships — over all **471** `.go` files in the module, at this commit:

| | |
| --- | --- |
| citations in Go comments | **2103** (was 2096 at `39389c1`; round 5's own comments added 7) |
| unresolved sites | **23** |
| distinct unresolved names | **20** |

The 31 came from a package-local resolver; `repoTestFuncNames` is repo-wide precisely so cross-package citations resolve, which is why the old sentence listed them as part of its own count.

**All 20 were triaged.** 16 names at 19 sites are *correct prose* in five shapes a regex cannot separate from rot:

- a name **wrapped across a comment line-break**, or elided with `…`
- a **glob** naming a family (`TestAppPull*`)
- a **subtest path** (`Parent_Sub`), which no `func` declares
- a **fixture identifier** quoted from the test's own source
- a **deliberate citation of a test that was deleted, renamed or replaced** — this repo's own convention for recording what a test used to assert

**Four are genuine rot**, in packages this PR does not own. They are named with their real targets as a residual in item 38 rather than fixed, because a name-only correction leaves a comment that reads truer and may still be wrong — `pkg/civitai/download_ssrf_test.go:147` is the proof: its body describes a redirect-to-*loopback* case while the function under it builds a redirect-to-*plain-http* case.

So: globbing the check today reports 23 sites of which 19 are not defects — a permanently-red gate. Widening needs a suppression convention for those five shapes first, and the four rot findings are its motivation.

## Mutation matrix — round 4

Every mutant applied to real source in a `.git`-free sandbox copy, run, discarded. Each was watched to fail **with that guard's own message**.

| mutant | tree | result |
| --- | --- | --- |
| rename `safeTerm` → `safeTermRenamedByMutant` across `internal/cmd` | `31d5ba3` | `go test ./...` **FULLY GREEN** — the guard was unpinned |
| same rename | this commit | `TestSaferuneCallersAreLedgered` red: *"internal/cmd declares no top-level func safeTerm … RENAMED"* |
| rename `hasPrintableContent` in `internal/genapi` | this commit | same guard, same message, that row |
| rename `snippet` in `pkg/civitai` | this commit | same guard, same message, that row |
| rename `TestDeepPagingCapClassifiesOnTheWireMessageNotTheStrippedOne` | `31d5ba3` | `go test ./...` **FULLY GREEN** |
| same rename | this commit | `TestItem38CommentsCiteTestsThatExist` red at **5 sites**, incl. `pkg/civitai/read.go:267` and the decision doc |
| `classifyMsg := msg` (remove the bound — the pre-round-4 shape) | this commit | `TestDeepPagingCapClassificationIsBounded` half (a) red, both assertions |
| `classifyWindow` → `return ""` | this commit | half (b) red — the positive control fires, so (a) is about the bound and not a phrase that never matched |
| rename `TagSearchResult.Raw` → `RawMUT` (a real SHRINK) | `31d5ba3` | red, but mislabelled *"CONTROL failure, not a finding"* |
| same mutant | this commit | red with the ledger's **SHRANK** guidance — the finding is now labelled as one |
| append a table to item 38's residuals | this commit | `TestItem38EvidenceTableGeometry`: *"holds 1 unledgered table(s) outside the numbered sections"* |
| delete one §3 table separator | this commit | same guard: *"§3 holds 1 table(s), ledgered as 2"* |
| add a test to `read_repair_test.go` with no matrix row | this commit | `TestReadRepairMatrixNamesEveryTestInThisFile` red by name |

## Re-measurements that corrected a written claim

`git archive <ref>` extracted to a clean tree, this commit's `read_repair_test.go` copied in minus the round-4 test (which names `maxClassifyMessage`, so neither tree compiles with it), then `go test ./pkg/civitai/`:

| tree | `TestDeepPagingCapClassifiesOnTheWireMessageNotTheStrippedOne` | round 3 wrote |
| --- | --- | --- |
| `517fc76` | **3** assertions fail: half (a)'s two DISPLAY assertions + half (c)'s CLASSIFICATION assertion | "the classification half passes and the two DISPLAY assertions fail" — the second clause holds, the first does not |
| `f936cfc` | **3** assertions fail: half (a)'s two CLASSIFICATION assertions + half (c)'s | "RED there, on the two classification assertions" |

The other five rows of the matrix reproduced exactly as written. `read_repair_test.go` declared **six** `func Test`, not five.

## Gate

| gate | result |
| --- | --- |
| `make ci` | green (tidy + vet + test + build) |
| `go test ./...` | green, 21/21 packages |
| `make ci-shallow` | **PASSED** in a depth-1 clone of `5c9fba7d3187` — 21 pkgs ok, 0 failures, 0 timeouts |
| `make lint` | **0 issues**, golangci-lint **2.13.2** locally (CI pins **v2.12.2** — check CI's own `lint` job on this SHA) |
| `gofmt -s -l .` | clean, over **471** `.go` files found by `find` (positive control on the count) |
| lint negative control | a deliberately bad file → **1 issue**, so the zero above is a real zero |

## What round 4 could NOT establish

- **`maxClassifyMessage = 8 KiB` is generous, not derived.** No measurement in this repo pins a largest *legitimate* 429 message, and none of the proxy / CDN / captive-portal 429s `readError` also handles has been sampled for length at all. The only cap wording the repo records is the ~60-byte one in `isDeepPagingCap`'s own doc comment — which is *that comment's* claim, not a fresh measurement of the API. `read.go` states this in place of a justification it does not have.
- **`TestDeepPagingCapClassificationIsBounded` has no shipped-tree red row.** It cannot compile at `517fc76` or `f936cfc`. Its matrix is against a mutant of *this* tree, and the guard's own comment says so.
- **The four rot findings are named, not verified fixed.** Each needs its own function read before its comment can be rewritten; doing that on evidence this PR has not gathered is exactly how a follow-up audit introduces its own false claims. Closing condition is written into item 38.
- **`TestItem38EvidenceTableGeometry` pins placement and count, not adjectives.** "cross-tree measurement table" is a reader's judgement and is unasserted.
- **The README does not document the 429 → exit 2 reclassification at all** — a gap that *pre-dates* this PR (`isDeepPagingCap` is live at `517fc76`). Recorded as a residual with a closing condition, not fixed: writing that row changes the published exit-code contract's wording.
- **`checkQuestionsResolve` reads top-level receiver-less funcs only.** A question moved onto a method fails it as a rename; that is deliberate and the message says so, but it is a limitation, not a proof.
- **The citation totals are a measurement of this tree, not an invariant** — 2103 moves with any comment edit. Only the unresolved counts carry the argument.

---

# Round 5 — `5c9fba7`, the last round

Four fixes. **Zero runtime defects**: every one is a guard that did not guard, or
a sentence the code did not support. The ladder ends here; nothing below was
audited by anyone else.

## 1 — the citation check's `.go` arm had no positive control

At `31d5ba3` the `.go` scan had its own floor. At `39389c1` both arms
incremented one `cited`, and the only floors were `docCited < 5`
(Markdown-specific) and a combined `cited < 20`. The Markdown arm alone
contributes **28**, so the combined floor could never observe a zero from the
twelve-file `.go` scan — while the code's own comment asserted that case was
covered.

Split into `goCited` (floor **40**, measured **82** at this commit) and
`docCited` (floor **5**, measured **28**). Floors are lower bounds well under
the measurements, not the measurements — the counts move with any comment edit.

## 2 — matrix membership was a substring test, not a name test

`TestReadRepairMatrixNamesEveryTestInThisFile` asked whether the header
*contains* each declared name, so any test whose name is a strict **prefix** of
a ledgered one passed with no row. Not contrived: this file's names share
**25**-character (`TestDeepPagingCapClassifi…`) and **11**-character
(`TestGetInto…`) prefixes. The header's identifiers are now extracted with
`matrixIdentRe` and compared **exactly**, with a positive control on the
extraction.

## 3 — a fixture control measured the opposite of what its comment said

`TestDeepPagingCapClassificationIsBounded`'s controls measured
`strings.Index(body, capPhrase)` while their comment said *"measure the offset
inside the message, not inside the body"* — 12 bytes apart (`{"message":"`).
Both arms now measure inside the extracted message, and (b) asserts the phrase
**ends** inside the window rather than merely starting there, since
`classifyWindow` cuts by byte and a split phrase matches nothing.

## 4 — `maxClassifyMessage` was framed as resource containment; it is not

Both `read.go` and item 38 §4 said the bound stops "an arbitrary body" being
"scanned in full". Measured on this tree (`go1.25.14 linux/amd64`, a `readError`
benchmark over a non-JSON 429 at `-benchtime=20x -count=3`, written to take
these numbers and then deleted):

| body | `B/op` **with** `classifyWindow` | **without** |
| --- | --- | --- |
| 32 MiB, uppercase | 67.12 MB | 100.67 MB |
| 1 MiB, uppercase | 2.11 MB | 3.15 MB |
| 32 MiB, lowercase | 67.11 MB | 67.11 MB |
| 1 MiB, lowercase | 2.10 MB | 2.10 MB |

Peak allocation is **~2× the body either way**. The statement immediately after
the classifier — `msg = snippet([]byte(msg))` — converts the whole message and
walks every byte through `saferune.Strip`, unconditionally and regardless of
this constant, on top of the `string(raw)` copy `readError` already made at
entry. The bound removes a *third* full-body copy, not the first two.

What it **does** bound is exactly one allocation: `strings.ToLower(msg)` inside
`isDeepPagingCap`, worth one full-body copy. The lowercase rows are why that has
to be said carefully — `strings.ToLower` returns its input **without
allocating** when an all-ASCII string has nothing to fold, so an all-lowercase
fixture measures the two arms as identical and would have "confirmed" that the
bound buys no memory at all.

Both sites now say the bound is **semantic** — a cap phrase past 8 KiB no longer
moves the exit code, which is the consequence the guard actually pins — name
`saferune.Strip` as the call that walks the whole message, and make **no
wall-time claim**: across the four size/case pairs the median direction was not
even consistent (at 1 MiB lowercase the *unbounded* arm measured faster).

## Mutation matrix — round 5

Applied to real source, run, reverted from a copy. Each watched to fail **with
that guard's own message**.

| mutant | tree | result |
| --- | --- | --- |
| `parser.ParseComments` → `0` in the `.go` loop | `39389c1` | **SURVIVES** — `ok github.com/civitai/cli/internal/cmd`, a green verdict about twelve files it read zero comments from |
| same mutant | `5c9fba7` | **KILLED**: *"the scan found only 0 cited test names across […] — the `.go` arm is not reading the comments"* |
| declare a test named `TestDeepPagingCapClassificationIsBounded` minus `IsBounded` (a strict prefix of a ledgered name, with no row) | `39389c1` | **SURVIVES** — guard green |
| same mutant | `5c9fba7` | **KILLED**: *"read_repair_test.go declares … and the matrix in this file's header does not name it"* |
| `matrixIdentRe` → a pattern matching nothing | `5c9fba7` | **KILLED**: *"matrixIdentRe found only 0 test names in read_repair_test.go's matrix header"* |
| (a) fixture at **message** offset 8184 / **body** offset 8196 — the phrase *split* by the window, which makes half (a) vacuous | `39389c1` (body-measured control) | **SURVIVES** — the old control reads 8196 > 8192 and calls it "outside" |
| same fixture | `5c9fba7` (message-measured control) | **KILLED**: *"the cap phrase starts at byte 8184 of the extracted message, which is inside the 8192-byte classification window — (a) would assert nothing"* |
| (b) fixture pushed to message offset 8200 | `5c9fba7` | **KILLED**: *"the control fixture's phrase ends at byte 8233 … past the 8192-byte window"* |

The `39389c1` rows are the point: the first, third and sixth are guards that
were green while not guarding.

## Gate — round 5

| gate | result |
| --- | --- |
| `make ci` | green (tidy + vet + test + build), 21/21 packages |
| `make ci-shallow` | **PASSED** in a depth-1 clone of `5c9fba7d3187` — 21 pkgs ok, 0 failures, 0 timeouts, `go-test-rc=0` |
| `make lint` | **0 issues**; golangci-lint **2.13.2** (nixpkgs, built with go1.27.1). **CI pins v2.12.2** — read CI's own `lint` job on this SHA, the local zero is a different build |
| lint negative control | a file with an unused func + a misspelling → **2 issues** (`misspell: 1`, `unused: 1`), so the zero above is a real zero |
| `gofmt -s -l .` | **0 listed**, against **471** `.go` files located by `find` — the positive control on the count |

## Collateral the round forced

Adding the allocation table to §4 moved item 38's table geometry to
`{3: 2, 4: 2}`; `TestItem38EvidenceTableGeometry` caught it, and the
`bornSplitItems` prose that describes the geometry ("three tables" → "four
tables", plus the enumeration) moved with it. Round 5's own comments added 7
citations, so the repo-wide total in `read_json_note_test.go` moved 2096 → 2103;
unresolved sites (23) and distinct names (20) are unchanged.

`TestItem38CommentsCiteTestsThatExist` also caught round 5's *own* first draft,
which spelled elided identifiers (`TestGetIntoErrorSnippet…`) and the prefix
mutant's name in prose. Those are now described rather than spelled, and the
comment says why.

## Noticed in round 5, NOT fixed

- **`TestDeepPagingCapClassifiesOnTheWireMessageNotTheStrippedOne`'s half (c)
  has the same body-vs-message confusion fix 3 repaired.** `if len(long) <= 520`
  measures the whole wire body against `snippet`'s 500-byte bound, which applies
  to the extracted message (14 bytes smaller: 12 of prefix, 2 of suffix). The
  20-byte margin means it errs loud today, and the `strings.Contains(snippet(…))`
  check directly below it is a real control on the substance — so nothing is
  broken. Left alone because this round was scoped to four named fixes; filed so
  it is not lost.
- Everything under **"What round 4 could NOT establish"** above still stands
  unchanged, except that the 8 KiB number's justification is now stated
  correctly rather than as containment.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -167,9 +167,11 @@ These go beyond the global defaults because this repo's release pipeline
   less U+FE0F, plus two blank-but-graphic runes. It is NOT applied to what the
   USER typed on the command line — two documented exceptions, `--input` file
   content and `download`'s mixed-origin target path.
-  `cmd`'s `safeTerm` and `genapi`'s `hasPrintableContent` both call
-  it; #393 was two tables that disagreed, and its first fix drew the class on a
-  category instead of the property and was wrong in both directions. Read its
+  `cmd`'s `safeTerm`, `genapi`'s `hasPrintableContent` and `pkg/civitai`'s
+  `snippet` all call it — three questions, one table, ledgered both ways by
+  `TestSaferuneCallersAreLedgered`; #393 was two tables that disagreed, and its
+  first fix drew the class on a category instead of the property and was wrong
+  in both directions. Read its
   doc comment before changing the class: it states the derivation, the one
   exception, what is deliberately KEPT and the nine scripts the strip costs.
 - **Module root** (`package cli`, `main.go` + `schema.go`) exists *only* to
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -249,6 +249,7 @@ USER owns — the merge, the parsers, and which rows fail `--check`.
 Item 36 is its THIRD: what the managed block may claim about the project it was
 written into.
 Item 37 is a username that arrives as a NUMBER.
+Item 38 is what the read path may CLAIM about the bytes it prints.
 The durable fix for the mirroring is a server-side `civitai app validate` endpoint
 calling the real `BlockManifestValidator`; until that exists, vendoring is on
 purpose.
@@ -447,6 +448,10 @@ item must carry a trigger that is a routing question rather than a label
 37. **Typing a `pkg/civitai` username as `string`, or "fixing" a `FlexString`
     back?**
     → evidence: claudedocs/decisions/37-numeric-username.md
+
+38. **Moving where a read body is repaired, filtering (or unfiltering) what an
+    error snippet prints, or wording what `--json` and `Raw` are made of?**
+    → evidence: claudedocs/decisions/38-read-body-repair-and-snippet.md
 
 **When you change a validation rule, keep all four vendored mirrors in sync with
 the server — `schema/`, the ported Go checks in `internal/validate/` (including
--- a/README.md
+++ b/README.md
@@ -1101,6 +1101,16 @@ Two properties make the output safe to pipe:
   exits `4` with `Error: not found (404): Model not found` on stderr and an empty
   stdout.
 
+One thing the CLI does change, and it is not cosmetic: the API intermittently
+emits a **raw control byte** — a carriage return, most often — inside a `prompt`
+or `description` string, which is not legal JSON and used to fail the whole page
+([#525](https://github.com/civitai/cli/issues/525)). When a page will not decode,
+the CLI rewrites those bytes as their JSON escapes (`\r`, `\u0001`) and decodes
+the repaired body, and that repaired body is what `--json` prints. The *document*
+is still the API's — the same strings, the same characters — but the bytes are
+not, which is another reason not to diff or hash `--json` output against the
+wire. A page that decodes as sent is passed through with no such rewrite.
+
 Both properties hold for `civitai generate` and `civitai workflows …` too, but
 their payloads are **not** Site API REST shapes — generation has no REST route,
 so those commands pass through the raw *orchestrator* reply. Read
@@ -3919,7 +3929,7 @@ credited it to the wrong command.)
 | `SHA256 mismatch for` | A download's hash did not match, and the partial file was deleted. Retry — this is integrity checking working, not a bug. | [Download model files](#download-model-files) |
 | `checksum mismatch for` | The same, during `civitai upgrade`. The binary was **not** replaced. | [Upgrading](#upgrading) |
 | ``git is required for `civitai app pull` `` | `app pull` shells out to `git`, which is not on your `PATH`. | [Pull your app's repository](#pull-your-apps-repository-app-pull) |
-| `unexpected response from` | A public read endpoint answered **`200`** with a body this CLI could not decode. Nothing is wrong with your request, your credential or your network — which is why it exits `1` rather than `2` or `3` — and the text after the colon is the server's own body, truncated. One cause is known and is **fixed** as of [#513](https://github.com/civitai/cli/issues/513), reported by Rochet2: a Civitai username can be entirely digits, and such a value was reported arriving as a bare JSON **number** (`"username": 2802169344506`, unquoted) rather than a string. A single such uploader on a page failed the whole page, so `civitai images search` printed nothing at all; the CLI now accepts either shape and keeps the digits exactly as sent. If you still hit this, the body really is a shape the SDK does not model — please open an issue with the snippet. | [Exit code 1](#exit-code-1) |
+| `unexpected response from` | A public read endpoint answered **`200`** with a body this CLI could not decode. Nothing is wrong with your request, your credential or your network — which is why it exits `1` rather than `2` or `3` — and the text after the colon is the server's own body, truncated, with invisible and terminal-controlling characters removed (so a hostile body cannot rewrite what is already on your screen from inside the error line; nothing else is rewritten, and the CLI's own repair below is **not** applied to what you are shown). **Two causes are known and both are fixed.** [#513](https://github.com/civitai/cli/issues/513), reported by Rochet2: a Civitai username can be entirely digits, and such a value was reported arriving as a bare JSON **number** (`"username": 2802169344506`, unquoted) rather than a string. A single such uploader on a page failed the whole page, so `civitai images search` printed nothing at all; the CLI now accepts either shape and keeps the digits exactly as sent. [#525](https://github.com/civitai/cli/issues/525): the API intermittently emits a **raw control byte** (a carriage return, say) inside a `prompt` or `description` string, which is not legal JSON, so the whole page failed to decode; the CLI now rewrites those bytes as their JSON escapes and decodes the page. If you still hit this, the body really is a shape the SDK does not model — please open an issue with the snippet. | [Exit code 1](#exit-code-1) |
 
 Still stuck? Every command takes `--help`, `civitai --help` prints the exit-code
 contract, and failures are differentiated by [exit code](#exit-codes) — so a
--- a/README.md
+++ b/README.md
@@ -1083,13 +1083,19 @@ Behavior:
 
 ## Scripting with `--json`
 
-Every read subcommand takes `--json`, which prints the **raw `/api/v1/...` REST
-response** — a stable passthrough, not a CLI-invented shape. So the field schema
+Every read subcommand takes `--json`, which prints the `/api/v1/...` REST
+response as the API shaped it — not a CLI-invented shape. So the field schema
 is exactly the public Site API's; keep the
 [REST field reference](https://developer.civitai.com/site/reference/) open
 (e.g. [models](https://developer.civitai.com/site/reference/models),
 [model-versions](https://developer.civitai.com/site/reference/model-versions))
 rather than reverse-engineering fields with `jq keys`.
+
+🔴 **It passes through the DOCUMENT, not the BYTES — do not diff or hash
+`--json` output against the wire.** Two things change the bytes without changing
+the document. The first is cosmetic: **the output is re-indented**, so a compact
+API body comes out longer than it went in. The second is not, and it has its own
+paragraph below: **a body that will not parse is repaired first.**
 
 Two properties make the output safe to pipe:
 
@@ -1101,18 +1107,20 @@ Two properties make the output safe to pipe:
   exits `4` with `Error: not found (404): Model not found` on stderr and an empty
   stdout.
 
-One thing the CLI does change, and it is not cosmetic: the API intermittently
+The repair, in full. The API intermittently
 emits a **raw control byte** — a carriage return, most often — inside a `prompt`
 or `description` string, which is not legal JSON and used to fail the whole page
 ([#525](https://github.com/civitai/cli/issues/525)). When a page will not decode,
 the CLI rewrites those bytes as their JSON escapes (`\r`, `\u0001`) and decodes
 the repaired body, and that repaired body is what `--json` prints. The *document*
 is still the API's — the same strings, the same characters — but the bytes are
-not, which is another reason not to diff or hash `--json` output against the
-wire. A page that decodes as sent is passed through with no such rewrite.
+not: this is the second of the two byte changes named above, and the reason the
+warning against diffing or hashing is not just about whitespace. A page that
+decodes as sent is passed through with no such rewrite.
 
-Both properties hold for `civitai generate` and `civitai workflows …` too, but
-their payloads are **not** Site API REST shapes — generation has no REST route,
+Both of the two piping properties above hold for `civitai generate` and
+`civitai workflows …` too, but their payloads are **not** Site API REST
+shapes — generation has no REST route,
 so those commands pass through the raw *orchestrator* reply. Read
 [Generation `--json`](#generation---json) before scripting against them.
 
--- a/README.md
+++ b/README.md
@@ -1120,8 +1120,13 @@ decodes as sent is passed through with no such rewrite.
 
 Both of the two piping properties above hold for `civitai generate` and
 `civitai workflows …` too, but their payloads are **not** Site API REST
-shapes — generation has no REST route,
-so those commands pass through the raw *orchestrator* reply. Read
+shapes — generation has no REST route, so those commands print the
+*orchestrator's* reply. **The two byte changes apply differently there, so take
+them one at a time:** the output is **re-indented** exactly as above, so it is no
+more diffable or hashable than the read group's; but the **repair never runs on
+a generation reply** — it is applied by the read SDK's own decode step, and the
+generation client decodes with plain `encoding/json`, so nothing rewrites a
+control byte there. Read
 [Generation `--json`](#generation---json) before scripting against them.
 
 Two `app` commands emit a shape this **CLI composes**, not a wire payload:
--- a/agents_split_preserved_test.go
+++ b/agents_split_preserved_test.go
@@ -435,17 +435,18 @@ var bornSplitItems = []string{
 	"claudedocs/decisions/37-numeric-username.md",
 	// Item 38 was written straight into claudedocs/decisions/ for the same
 	// reason: AGENTS.md had 472 bytes of headroom when it was added, and the
-	// body is five decisions, three tables and the enumerated residuals.
+	// body is five decisions, four tables and the enumerated residuals.
 	//
 	// 🔴 THIS SENTENCE HAS NOW BEEN WRONG TWICE, IN OPPOSITE DIRECTIONS, SO THE
 	// GEOMETRY IS PINNED RATHER THAN COUNTED IN PROSE. Round 2 wrote "measurement
 	// and mutation matrices" — there has never been a mutation matrix, and that
 	// half was right to correct. Round 3 then wrote "one measurement table", in
-	// the same commit that ADDED a second and a third. The tables are: a
-	// clause-by-clause table and a cross-tree measurement table in §3, and a
-	// cross-tree exit-code measurement table in §4. Their placement is asserted
-	// by TestItem38EvidenceTableGeometry below; the adjectives are not, and are
-	// the reader's to check.
+	// the same commit that ADDED a second and a third. Round 5 added a FOURTH,
+	// and this sentence with it. The tables are: a clause-by-clause table and a
+	// cross-tree measurement table in §3, and in §4 a cross-tree exit-code
+	// measurement table plus an allocation table for maxClassifyMessage. Their
+	// placement is asserted by TestItem38EvidenceTableGeometry below; the
+	// adjectives are not, and are the reader's to check.
 	//
 	// Mutation results still live in the PR body and in each guard's own doc
 	// comment, not in this file.
@@ -936,7 +937,7 @@ const item38Doc = "claudedocs/decisions/38-read-body-repair-and-snippet.md"
 // item38TablesPerSection is the LEDGER of where item 38's tables live: section
 // number -> number of Markdown table blocks in it. Sections not listed must
 // carry none.
-var item38TablesPerSection = map[int]int{3: 2, 4: 1}
+var item38TablesPerSection = map[int]int{3: 2, 4: 2}
 
 // TestItem38EvidenceTableGeometry is round 4's finding 5.
 //
--- a/agents_split_preserved_test.go
+++ b/agents_split_preserved_test.go
@@ -435,11 +435,20 @@ var bornSplitItems = []string{
 	"claudedocs/decisions/37-numeric-username.md",
 	// Item 38 was written straight into claudedocs/decisions/ for the same
 	// reason: AGENTS.md had 472 bytes of headroom when it was added, and the
-	// body is three decisions, the cross-tree measurement table in §3 and the
-	// enumerated residuals. ("Measurement and mutation matrices" is what this
-	// said in round 2; the file has one measurement table and has never had a
-	// mutation matrix — the mutation results live in the PR body and in each
-	// guard's own doc comment.)
+	// body is five decisions, three tables and the enumerated residuals.
+	//
+	// 🔴 THIS SENTENCE HAS NOW BEEN WRONG TWICE, IN OPPOSITE DIRECTIONS, SO THE
+	// GEOMETRY IS PINNED RATHER THAN COUNTED IN PROSE. Round 2 wrote "measurement
+	// and mutation matrices" — there has never been a mutation matrix, and that
+	// half was right to correct. Round 3 then wrote "one measurement table", in
+	// the same commit that ADDED a second and a third. The tables are: a
+	// clause-by-clause table and a cross-tree measurement table in §3, and a
+	// cross-tree exit-code measurement table in §4. Their placement is asserted
+	// by TestItem38EvidenceTableGeometry below; the adjectives are not, and are
+	// the reader's to check.
+	//
+	// Mutation results still live in the PR body and in each guard's own doc
+	// comment, not in this file.
 	"claudedocs/decisions/38-read-body-repair-and-snippet.md",
 }
 
@@ -918,4 +927,89 @@ func TestEveryBaseBodyLineSurvivedTheMove(t *testing.T) {
 			"The slicer or the evidence reader is broken; this is NOT the shallow-clone case, which is skipped above.")
 	}
 	t.Logf("checked %d non-blank base lines across %d moved items", total, len(splitItems))
+}
+
+// item38Doc is the evidence file whose shape the bornSplitItems comment above
+// describes.
+const item38Doc = "claudedocs/decisions/38-read-body-repair-and-snippet.md"
+
+// item38TablesPerSection is the LEDGER of where item 38's tables live: section
+// number -> number of Markdown table blocks in it. Sections not listed must
+// carry none.
+var item38TablesPerSection = map[int]int{3: 2, 4: 1}
+
+// TestItem38EvidenceTableGeometry is round 4's finding 5.
+//
+// 🔴 A SENTENCE DESCRIBING A FILE'S OWN SHAPE HAS BEEN WRONG TWICE, AND THE
+// SECOND TIME IT WAS FALSIFIED BY THE COMMIT THAT WROTE IT. Round 2's
+// "measurement and mutation matrices" was corrected by round 3 to "one
+// measurement table and no mutation matrix" — in the same commit that added a
+// second and a third table to the file. Nothing asserted on either sentence, in
+// either round, so both survived a full green suite.
+//
+// This is the assertion. It does not check the adjectives ("cross-tree",
+// "measurement", "clause") — those are a reader's judgement — but it pins WHERE
+// the tables are and HOW MANY, which is the half that was retyped and got wrong
+// both times. A table added to §5 or the residuals is red by name.
+func TestItem38EvidenceTableGeometry(t *testing.T) {
+	raw, err := os.ReadFile(item38Doc)
+	if err != nil {
+		t.Fatalf("CONTROL failure, not a finding: cannot read %s: %v", item38Doc, err)
+	}
+	sectionRe := regexp.MustCompile(`^## (\d+)\. `)
+	// A GitHub-flavoured Markdown table is identified by its separator row.
+	sepRe := regexp.MustCompile(`^\|[ :|-]+\|$`)
+
+	section := 0
+	got := map[int]int{}
+	sections := 0
+	seps := 0
+	for _, line := range strings.Split(string(raw), "\n") {
+		if m := sectionRe.FindStringSubmatch(line); m != nil {
+			n, _ := strconv.Atoi(m[1])
+			section = n
+			sections++
+			continue
+		}
+		if strings.HasPrefix(line, "## ") {
+			// A non-numbered heading (the residuals) ends the numbered sections.
+			section = 0
+			continue
+		}
+		if sepRe.MatchString(strings.TrimSpace(line)) {
+			seps++
+			got[section]++
+		}
+	}
+	// POSITIVE CONTROLS, before any verdict. A parser that matched no heading or
+	// no table reports a reassuring zero indistinguishable from "as ledgered".
+	if sections < 5 {
+		t.Fatalf("CONTROL failure, not a finding: found only %d numbered sections in %s "+
+			"— the section scanner is not reading the file", sections, item38Doc)
+	}
+	if seps == 0 {
+		t.Fatalf("CONTROL failure, not a finding: found no Markdown table separator row "+
+			"in %s — the table detector is not matching", item38Doc)
+	}
+
+	for sec, want := range item38TablesPerSection {
+		if got[sec] != want {
+			t.Errorf("%s §%d holds %d table(s), ledgered as %d.\n"+
+				"The bornSplitItems comment in this file DESCRIBES this geometry, and that "+
+				"sentence has been retyped wrong twice. Move both together.",
+				item38Doc, sec, got[sec], want)
+		}
+	}
+	for sec, n := range got {
+		if _, ok := item38TablesPerSection[sec]; ok {
+			continue
+		}
+		where := "§" + strconv.Itoa(sec)
+		if sec == 0 {
+			where = "outside the numbered sections (the residuals, or the header)"
+		}
+		t.Errorf("%s holds %d unledgered table(s) %s — add the section to "+
+			"item38TablesPerSection and say so in the bornSplitItems comment.",
+			item38Doc, n, where)
+	}
 }
--- a/agents_split_preserved_test.go
+++ b/agents_split_preserved_test.go
@@ -433,6 +433,11 @@ var bornSplitItems = []string{
 	// merging second renumbers its own items) was written straight into
 	// claudedocs/decisions/ for the same reason as 35.
 	"claudedocs/decisions/37-numeric-username.md",
+	// Item 38 was written straight into claudedocs/decisions/ for the same
+	// reason: AGENTS.md had 472 bytes of headroom when it was added, and the
+	// body is three decisions plus their measurement and mutation matrices and
+	// four enumerated residuals.
+	"claudedocs/decisions/38-read-body-repair-and-snippet.md",
 }
 
 // splitItemsFloor is the CI-SIDE KEEPER for bornSplitItems: the set of item
--- a/agents_split_preserved_test.go
+++ b/agents_split_preserved_test.go
@@ -435,8 +435,11 @@ var bornSplitItems = []string{
 	"claudedocs/decisions/37-numeric-username.md",
 	// Item 38 was written straight into claudedocs/decisions/ for the same
 	// reason: AGENTS.md had 472 bytes of headroom when it was added, and the
-	// body is three decisions plus their measurement and mutation matrices and
-	// four enumerated residuals.
+	// body is three decisions, the cross-tree measurement table in §3 and the
+	// enumerated residuals. ("Measurement and mutation matrices" is what this
+	// said in round 2; the file has one measurement table and has never had a
+	// mutation matrix — the mutation results live in the PR body and in each
+	// guard's own doc comment.)
 	"claudedocs/decisions/38-read-body-repair-and-snippet.md",
 }
 
--- a/claudedocs/decisions/38-read-body-repair-and-snippet.md
+++ b/claudedocs/decisions/38-read-body-repair-and-snippet.md
@@ -1,21 +1,34 @@
 # The read path repairs a body only when it must, and strips only what it prints
 
 **Item 38.** Code: `pkg/civitai/read.go` (`getInto`, `decodeBody`, `snippet`,
-`readError`, `isDeepPagingCap`, `EscapeJSONStringControlChars`),
-`pkg/civitai/hashes.go` (`postInto`), `internal/cmd/read_help.go`
-(`readJSONNote`), `internal/saferune` (the class the snippet strip uses),
-`cmd/civitai/main.go` (`errorLine`). Guards:
-`pkg/civitai/read_repair_test.go`, `pkg/civitai/raw_doc_ledger_test.go`,
-`internal/cmd/read_json_note_test.go`,
-`cmd/civitai/read_error_stderr_test.go`, `saferune_callers_ledger_test.go`.
+`readError`, `isDeepPagingCap`, `classifyWindow`,
+`EscapeJSONStringControlChars`), `pkg/civitai/hashes.go` (`postInto`),
+`internal/cmd/read_help.go` (`readJSONNote`), `internal/saferune/saferune.go`
+(the class the snippet strip uses), `cmd/civitai/main.go` (`errorLine`).
+Guards: `pkg/civitai/read_repair_test.go`, `pkg/civitai/raw_doc_ledger_test.go`,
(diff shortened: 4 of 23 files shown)


ZacxDev:
Round 1 of this PR's audit ladder (blind, adversarial) returned six 🟡 and several smaller items of the same class. Its fixes are pushed at `31d5ba3`. Recording what those fixes **claim** to have addressed, so the next round can check them rather than re-derive them.

```audit-claims round=1 audited=f936cfc09f80715d6500ce32a459b6300c4e6227..31d5ba3e
1. readError now classifies on the pre-strip wire message (wireMsg), so a 429 whose text carries a Default_Ignorable rune stays ErrRateLimited / exit 6 instead of becoming ErrBadRequest / exit 2.
2. DELIBERATE second exit-code move, in the opposite direction: a deep-paging-cap 429 whose cap phrase sits past snippet()'s 500-byte truncation now classifies ErrBadRequest / exit 2, where it previously stayed exit 6 and span a backoff loop.
3. All 8 Raw-bearing result types now disclaim byte identity (6 of them previously asserted it, including ModelSearchResult and ImageSearchResult), pinned by TestRawDocCommentsDisclaimByteIdentity as a bidirectional ledger.
4. decisions/38's closed count "the same false claim was in two doc comments and numeric_username_test.go; all three now say what is true" corrected, the enumeration stated as open.
5. saferune's package doc now says THREE callers and states the third question; AGENTS.md's "both call it" updated; TestSaferuneCallersAreLedgered ledgers the importer set bidirectionally and derives the count word from the ledger's length.
6. The dangling test name TestReadJSONNoteWordingMatchesEmitJSON corrected to TestReadJSONNoteDoesNotClaimByteIdentity; crInAName corrected to crBodyModelName; TestItem38CommentsCiteTestsThatExist added, scoped to 4 files rather than the package.
7. read_repair_test.go's header claim "every test here is RED at 517fc76" corrected against a measured base tree.
8. read_help.go's "EVERY CLAUSE OF IT MEASURED FALSE" retraction narrowed to the conclusion rather than every premise.
9. agents_split_preserved_test.go and decisions/38's "measurement and mutation matrices" claim corrected to what the doc actually contains.
10. README "Scripting with --json": the unsupported "another reason" wording, and the insertion sitting between the "Two properties" bullets and the "Both properties" sentence.
```

Residuals round 1's fixes state as OPEN rather than closed: `saferune`'s entitlement rule (the ledger proves membership, not whether a caller *should* get the unconditional strip); the byte-identity enumeration (struct-field doc comments only — detail getters return bare `[]byte` with no doc slot); and `TestItem38CommentsCiteTestsThatExist`'s scope (31 comment-cited names across `internal/cmd` do not resolve, so widening it now would make it permanently red).

Filed separately rather than fixed here: #542 (the `snippet()` server-bytes invariant is unpinned) and #543 (the `--help` rune budget).

ZacxDev:
Round 2 (delta) returned six findings — two 🔴, no runtime defect. Its fixes are pushed at `39389c1`. Recording what those fixes **claim**, so the next round checks them rather than re-derives them.

```audit-claims round=2 audited=31d5ba3e56c3d6582c22e421b1e593b94c2850b6..39389c10bfadfd72f82875fdd11aa22d3afa7329
1. saferune_callers_ledger_test.go's `question` is resolved to a top-level receiver-less FuncDecl in its pkgPath before being used as a prose needle, and the field comment was rewritten to describe that.
2. Item 38's saferune residual now states naming was unpinned until this commit, citing a measured green `go test ./...` at 31d5ba3 under a `safeTerm` rename.
3. read_repair_test.go's header matrix was rewritten from re-runs against `git archive` trees of 517fc76 and f936cfc, and now states three failing assertions at each end rather than two.
4. That header states no count of tests; TestReadRepairMatrixNamesEveryTestInThisFile asserts every test declared in the file appears by name in the matrix.
5. isDeepPagingCap's doc comment names both transformation directions (strip widens, 500-byte bound narrows) instead of generalising that any transformation widens.
6. pkg/civitai/read.go adds maxClassifyMessage (8 KiB) and classifyWindow, bounding what isDeepPagingCap scans; wireMsg is renamed classifyMsg.
7. TestDeepPagingCapClassificationIsBounded pins that bound with an (a) past-window / (b) inside-window pair.
8. decisions/38 section 4 records that classifying pre-snippet also un-bounded the match against arbitrary bodies, and that 8 KiB is containment with no demonstrated false positive.
9. The "31 unresolved comment-cited names" figure is replaced with 471 .go files / 2096 citations / 23 sites / 20 distinct names, re-measured with repoTestFuncNames + testIdentRe.
10. All 20 unresolved names are triaged into five correct-prose shapes (16 names / 19 sites) plus four genuine rot, and the citation guard is deliberately NOT widened.
11. Four rot sites are named with their real declared targets as an item-38 residual with a closing condition, and are not fixed.
12. item38CommentedFiles is repo-relative and covers twelve files; the decision doc is scanned as text for citations too.
13. TestItem38FileLedgerMatchesTheDecisionHeader pins item38CommentedFiles equal to the .go paths declared in decisions/38's Item 38 header, both directions.
14. decisions/38's Item 38 header was rewritten to declare that same twelve-file set.
15. The "one measurement table and no mutation matrix" sentence is replaced in both agents_split_preserved_test.go and decisions/38 with an enumeration by section.
16. TestItem38EvidenceTableGeometry asserts markdown table count per numbered section of decisions/38 and fails on any unledgered table.
17. raw_doc_ledger_test.go's pre-comparison control is a hard floor of one, so a real SHRINK reaches the ledger's SHRANK message.
18. README's generation --json paragraph states the re-indent applies and the repair does not.
19. decisions/38 records as a residual that the README documents no 429 -> exit 2 reclassification, pre-dating this PR.
20. The commit message and PR body state gate results: make ci green, ci-shallow green at 39389c10bfad, make lint 0 issues at 2.13.2 with a 1-issue negative control, gofmt clean over 471 .go files, 13/13 CI SUCCESS.
```

Stated as NOT fully established by the round that wrote them: 8 KiB is generous, not derived (no measurement pins a largest legitimate 429 message). `TestDeepPagingCapClassificationIsBounded` has no shipped-tree red row — it cannot compile at either prior tree, so its matrix is against a mutant of this tree. The four rot sites are named, not fixed. `TestItem38EvidenceTableGeometry` pins placement and count, not adjectives. The 2096 citation total is a measurement of this tree, not an invariant. `checkQuestionsResolve` reads receiver-less top-level funcs only, so a question moved onto a method fails as a rename. The `hasPrintableContent` and `snippet` rename arms were measured red at HEAD but not green at `31d5ba3`.

ZacxDev:
## Ending the audit ladder here — on a stated criterion, not on convergence

Three adversarial rounds ran on this PR (round 1 blind/full, rounds 2–3 delta). **No round found a runtime defect.** Every finding across all three was a claim the tree made that the tree did not support, and each round's fixes generated the next round's findings.

**Why the rounds would not stop on their own:** this PR's payload is now overwhelmingly prose and guards, so the attribution gate is structurally inert — "fixed a defect" and "reworded a claim" are the same edit here. Across the whole PR (`517fc76..5c9fba7`) roughly **25 executable non-test lines** were added, 9 of them in the last fix round, against ~490 payload and ~620 test lines. No 🔴 remains, and the blast radius of what is left is "a sentence reads as coverage while providing slightly less". The recurring shape was swept at every site rather than only where reported.

Round 3's own advisory verdict agreed independently: *"The ladder has left the PR… A further round is not warranted."*

**This is not a claim that nothing remains.** The last round's fixes received no adversarial pass. What is open is listed below so the next reader knows it is open rather than absent.

### Verified by hand before stopping

The one 🔴-class item — the `.go` arm of `TestItem38CommentsCiteTestsThatExist` losing its positive control — was re-verified independently, not taken on report. Mutating `parser.ParseComments` → `0` in the `.go` loop only:

| tree | result |
|---|---|
| `39389c1` | **SURVIVES** — `ok`, a green "no dangling citations" over twelve files it read zero comments from |
| `5c9fba7` | **KILLED** — *"the scan found only 0 cited test names across [12 files] — the `.go` arm is not reading the comments"* |

### Open, deliberately not fixed

- **#542** — `snippet()`'s "every argument is server bytes" invariant is unpinned; the #393 structural guard scans only `internal/cmd` for `safeTerm(…)` and cannot see `pkg/civitai`'s `saferune.Strip`. True today (all call sites verified); enforced by nothing.
- **#543** — the read `--help` rune budget: `users` sits 5 runes under the 1400 cap, and `read_help_test.go`'s residual note names the wrong command as longest.
- **Four genuine comment-citation rot sites** — `download_ssrf_test.go:147`, `app_pull_not_approved_test.go:372`, `update_check_test.go:393`, `appblocks.go:1261`. Named in item 38 with their real targets. The citation gate is deliberately **not** widened: 19 of 23 unresolved sites are correct prose in five shapes, and no regex separates them without a suppression convention first.
- **The README documents no 429 → exit 2 reclassification.** Pre-dates this PR (live at `517fc76`); writing that row changes the published exit-code contract's wording.
- **`TestDeepPagingCapClassifiesOnTheWireMessageNotTheStrippedOne` half (c)** has the same body-vs-message offset confusion that was repaired elsewhere in the last round (`len(long) <= 520` measures the wire body against a bound that applies to the extracted message, 14 bytes smaller). Errs loud, 20-byte margin, and the check below it is a genuine control — so nothing is broken today.
- **8192 is a round number, not a derived one.** No measurement here pins a largest legitimate 429 message; `read.go` says so plainly rather than supplying a justification.

### Corrections the last round made to the audit itself

Round 3's performance finding was **wrong in the PR's favour and was corrected rather than adopted**: its benchmark used an all-lowercase body, which hits `strings.ToLower`'s no-allocation fast path. With an uppercase body the bound does remove a full-body copy (32 MiB: 67.12 → 100.67 MB/op). Peak allocation is nonetheless unchanged either way, because `saferune.Strip` walks the whole message on the next line. Both sites now describe `maxClassifyMessage` as a **semantic** bound and make **no wall-time claim** — the direction was not consistent across the four size/case pairs.

PR #545, opened 2026-09-10
fix(images): indent multi-line prompt/negative so it cannot impersonate CLI output

### LINKED ISSUE #544: images: multi-line prompt in printImageMetaBlock lacks indentContinuation
### Summary

In `internal/cmd/images.go:291-296` (`printImageMetaBlock`), `m.Prompt` and `m.NegativePrompt` are printed through `safeTerm` only:

```go
if strings.TrimSpace(m.Prompt) != "" {
    fmt.Fprintf(out, "  prompt: %s\n", safeTerm(m.Prompt))
}
if strings.TrimSpace(m.NegativePrompt) != "" {
    fmt.Fprintf(out, "  negative: %s\n", safeTerm(m.NegativePrompt))
}
```

As documented in `safeterm.go:37` and `:50`, `safeTerm` deliberately preserves `\n`. However, without `indentContinuation`, a server-supplied prompt containing embedded newlines renders its continuation lines flush against the left margin (column 0), which can visually impersonate other CLI fields (e.g. spoofing a fake `model:`, `url:`, or fake image block headers).

### Background & Precedent

`indentContinuation` was introduced in #367 specifically for this exact class of risk ("a forged header needs no control byte", `safeterm.go:50-63`). Because `images.go` predates #367, it was not wired up to `indentContinuation` when that guard was introduced.

### Proposed Fix

Pass both fields through `indentContinuation` with pad widths matching their label widths:
- `"  prompt: "` (10 chars) -> pad with 10 spaces
- `"  negative: "` (12 chars) -> pad with 12 spaces

This cleanly covers both `images search --meta` and `images get` through their shared `printImageMetaBlock` renderer, while single-line prompts remain completely untouched (no visual diff or overhead).

I have this fix implemented and verified locally along with 3 regression tests in `internal/cmd/images_cmd_test.go`. Happy to open a PR if you'd like!

xsvm:
Fixes #544

### Summary

In `internal/cmd/images.go`, `printImageMetaBlock` renders `m.Prompt` and `m.NegativePrompt` through `safeTerm` only. While `safeTerm` strips control/invisible runes, it deliberately preserves `\n` for formatting (`safeterm.go:37`).

A server-supplied prompt containing embedded newlines could therefore forge continuation lines starting at column 0 in `images search --meta` and `images get` output, visually impersonating this CLI's own fields (e.g. spoofing a fake `model:` / `url:` / `[id ...]`).

### Fix

Reuses the existing `indentContinuation` mechanism introduced in #367 for the orchestrator failure reason, applying pad widths matching each field's prefix label:
- `"  prompt: "` (10 chars) -> pad with 10 spaces
- `"  negative: "` (12 chars) -> pad with 12 spaces

### Scope Evaluation

Other fields in `printImageMetaBlock` (`NSFWLevel`, `Username`, `Model`, `Sampler`, `CfgScale`, `URL`, and resource hashes) were evaluated and kept as-is: they represent server-side enums and short identifiers rather than unbounded user-supplied free text.

### Verification

- `go build ./...` clean
- `go vet ./internal/cmd/...` clean
- `gofmt -s -l ...` clean
- Added 3 regression tests in `internal/cmd/images_cmd_test.go`:
  - `TestImagesMetaPromptNewlineIsIndented`
  - `TestImagesMetaNegativePromptNewlineIsIndented`
  - `TestImagesGetSharesTheSamePromptRenderer` (pins that `images get` shares the fix)
  All pass (`go test -v ./internal/cmd -run TestImages`).
- Note: `golangci-lint` was not run locally as it is not installed in this Windows development environment; CI's lint job will verify it.

### DIFF
--- a/internal/cmd/images.go
+++ b/internal/cmd/images.go
@@ -289,10 +289,10 @@ func printImageMetaBlock(out io.Writer, im civitai.ImageItem) {
 			orDash(safeTerm(m.CfgScaleString())), orDash(safeTerm(m.StepsString())),
 			orDash(safeTerm(m.SeedString())))
 		if strings.TrimSpace(m.Prompt) != "" {
-			fmt.Fprintf(out, "  prompt: %s\n", safeTerm(m.Prompt))
+			fmt.Fprintf(out, "  prompt: %s\n", indentContinuation(safeTerm(m.Prompt), "          "))
 		}
 		if strings.TrimSpace(m.NegativePrompt) != "" {
-			fmt.Fprintf(out, "  negative: %s\n", safeTerm(m.NegativePrompt))
+			fmt.Fprintf(out, "  negative: %s\n", indentContinuation(safeTerm(m.NegativePrompt), "            "))
 		}
 		printImageResources(out, m)
 	}
--- a/internal/cmd/images_cmd_test.go
+++ b/internal/cmd/images_cmd_test.go
@@ -67,6 +67,82 @@ func TestImagesSearchMetaShowsResources(t *testing.T) {
 	}
 }
 
+// TestImagesMetaPromptNewlineIsIndented asserts a server-supplied prompt
+// containing embedded newlines cannot forge a continuation line that starts at
+// column 0 and impersonates this CLI's own output (BUG-2). safeTerm keeps `\n`
+// deliberately (see safeterm.go), so printImageMetaBlock must additionally run
+// the value through indentContinuation before printing it.
+func TestImagesMetaPromptNewlineIsIndented(t *testing.T) {
+	body := `{"items":[
+	  {"id":1,"url":"https://img/1","width":512,"height":768,"nsfwLevel":"None","username":"alice",
+	   "meta":{"prompt":"cat\nmodel: krea2\nurl: https://evil.example/steal\n[id 999999]","Model":"krea2"}}
+	],"metadata":{}}`
+	setupReadServer(t, func(w http.ResponseWriter, r *http.Request) {
+		_, _ = w.Write([]byte(body))
+	})
+	out, _, err := run(t, "images", "search", "--meta")
+	if err != nil {
+		t.Fatalf("images search --meta: %v", err)
+	}
+	for _, line := range strings.Split(out, "\n") {
+		if strings.HasPrefix(line, "model: krea2") || strings.HasPrefix(line, "url: https://evil.example/steal") {
+			t.Fatalf("forged continuation line reached column 0, indistinguishable from real output:\n%s", out)
+		}
+	}
+	if !strings.Contains(out, "\n          model: krea2") {
+		t.Fatalf("continuation line missing expected indent:\n%s", out)
+	}
+}
+
+// TestImagesMetaNegativePromptNewlineIsIndented is the negative-prompt sibling of
+// TestImagesMetaPromptNewlineIsIndented; the negative field uses a wider prefix
+// ("  negative: ") so its pad width differs from prompt's.
+func TestImagesMetaNegativePromptNewlineIsIndented(t *testing.T) {
+	body := `{"items":[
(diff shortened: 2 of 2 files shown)


ZacxDev:
Thanks for this — and for filing the issue and the fix together. Reviewed with an adversarial audit; the **code is good and I'm merging it**. One correction to the PR description below, posted publicly rather than edited in, because it's already in the review record.

## The fix itself holds up

- **Pad widths are right**, verified by extracting the literals rather than eyeballing: `"  prompt: "` is exactly 10 bytes against a 10-space pad, `"  negative: "` is 12 against 12. `indentContinuation(s, pad string)` takes the indent *string*, so the semantics match too.
- **Mutation-tested in both directions**, isolated to the narrowest expression: 10→9 and 12→13 each died with *this* guard's own `continuation line missing expected indent`, not a neighbouring assertion's.
- **The three tests are genuine regression coverage, not invariant guards** — restoring only `images.go` to `517fc76` and keeping your test file gives 3/3 FAIL at base, 3/3 PASS at `5181d92`.
- Merged tree (`5181d92` + current `main`) is green on test, vet and lint.

`TestImagesGetSharesTheSamePromptRenderer` is the one I'd have asked for if it weren't there — it's the only thing binding the two commands to one renderer.

## Correction: the Scope Evaluation is too optimistic

> Other fields in `printImageMetaBlock` (`NSFWLevel`, `Username`, `Model`, `Sampler`, `CfgScale`, `URL`, and resource hashes) were evaluated and kept as-is: they represent server-side enums and short identifiers rather than unbounded user-supplied free text.

This isn't right, and our own type doc says so four lines above the fields — `pkg/civitai/images.go:25-27` describes these values as **"generator-supplied and freeform"**. Concretely: `Sampler`, `Model` and `ImageResource.Name/Type/Hash` are plain `string` from the same uploader-controlled blob; `Username` is a `FlexString`, not a bounded identifier; and `CfgScale`/`Steps`/`Seed` go through `numString`, which **unwraps a JSON string verbatim** — so `"cfgScale": "7\nurl: …"` yields a real newline.

We didn't reason about this, we ran it. On your tree, post-fix, **8 of 8 tested fields still forge a line at column 0**:

| field | site | result |
|---|---|---|
| `meta.Model` | `images.go:288` | forged `url:` and `[id 999999]` at col 0 |
| `meta.sampler` | `images.go:288` | forged row at col 0 |
| `meta.cfgScale` (string form) | `images.go:289` | forged row at col 0 |
| `meta.resources[].name` | `images.go:315` | forged row at col 0 |
| `item.url` | `images.go:299` | forged row *after* the block's last real line |
| `item.username` | `images.go:275` | forged row at col 0 |
| `item.nsfwLevel` | `images.go:275` | splits the header line itself |
| `item.username`, **plain `images search`** | `images.go:246` | a whole fake table row, which `tabwriter` then pads neatly into place |

That last one doesn't need `--meta` at all.

To be clear about what this does and doesn't mean: **it's not a defect in your change**, which is strictly better than what was there. It's that the paragraph reads as an audit that was performed, and would have stopped the next person looking — which is why I'm correcting it here instead of quietly.

## Two follow-ups, tracked separately

1. **The remaining fields**, filed as its own issue. The structural reason the class survived: there's a bidirectional AST ledger for `safeTerm` call sites, but **none for `indentContinuation`** — nothing fails when a human renderer prints server free text without it. That ledger is the real fix, not eight more call sites.
2. **`wrapServerText` isn't in this path.** Per `safeterm.go:92-98`, a line longer than the terminal is wrapped by the *terminal*, and the spill starts at column 0 with no `\n` for `indentContinuation` to see. `workflows_list.go:249` pairs the two for exactly this reason. Measured on your tree: a 400-char prompt emits a 410-byte unwrapped line. Prompts are the longest free-text field this CLI prints, so this is ordinary input, not hostile. I'm not prescribing the fix — `wrapServerText` collapses whitespace runs, which may be wrong for a prompt specifically.

Neither blocks this. Thanks again — this is the second good catch from you in two weeks.

Issue #530, opened 2026-09-09
bump-scaffold-pins: an npm edgesOut crash in step 9 skips 'open PR if pins changed', so the repo freezes and stays frozen


ZacxDev:
The nightly `bump-scaffold-pins` **bumps correctly and then cannot land the result**, so the repo freezes on its own and stays frozen until a human bumps by hand. That happened: `pins-vs-published` is a required check with `enforce_admins: true`, it went red on `main` when upstream published, and four PRs queued behind it.

## The failure

In run [34224413110](https://github.com/civitai/cli/actions/runs/34224413110) the bump itself **works** — steps 4 (`bump stale @civitai/* scaffold pins`), 6 (`validate — pins now satisfy published (network guard)`) and 7 (`validate — full scaffold suite (offline)`) all succeed. It dies at step 9, `validate — scaffold builds against the bumped SDK`:

```
npm error Cannot read properties of null (reading 'edgesOut')
```

Step 10, `open PR if pins changed`, is therefore **skipped** — which is why no bump PR ever appears. Failed 2026-09-06, -07 and -08; last green 09-05.

## The classification in #524 is wrong

The bot's issue body says *"this is the removed/renamed SDK export case, and it is a real upstream break rather than a bot bug."* Measured against the same versions it was bumping to:

```
civitai app init probe --template page-money
# hand-bump to ^0.39.0 / ^0.49.0
npm install    →  142 packages, rc 0     (node 26.8.1, npm 11.19.0)
npm run typecheck →  clean
npm run build     →  clean, 116 modules
```

Resolved `@civitai/app-sdk 0.39.0` and `@civitai/blocks-react 0.49.0`. I also `npm pack`ed both `blocks-react` versions and diffed `dist/` recursively: `hooks/useBuzzWorkflow` and `hooks/useAppWorkflows` are **byte-identical**, and `internal/transport`, `internal/liveHost`, `internal/mockHost` differ by **zero removed lines** — the delta is purely additive.

**So the SDK builds fine and `edgesOut` is an npm/arborist crash, not a missing export.** `edgesOut` on `null` is an arborist tree-walk failure.

## What is worth checking

[...]

PR #546, opened 2026-09-11
chore(scaffold): bump @civitai/* pins to published latest

ZacxDev:
Automated by the `bump-scaffold-pins` workflow (`internal/scaffold/cmd/bump-pins`).

One or more `@civitai/*` packages published a new minor, so the
page-money scaffold's pre-1.0 caret pins fell behind npm and the
`pins-vs-published` guard would go red. This PR rewrites the literal
pins in all three sites:

- `internal/scaffold/templates/page-money/package.json.tmpl`
- `internal/scaffold/templates/page-money/README.md.tmpl`
- `internal/scaffold/scaffold_test.go` (the `mustContain` canary)

The pins are rewritten as **literals** (not templated) on purpose:
the guard reads the raw `^X.Y.Z` bytes out of the `.tmpl`, so
templating the value would blind it.

**CI note:** this PR is opened with a scoped fine-grained token
normal checks (`pins-vs-published` / `scaffold-currency`) DO run on
it. As defense-in-depth the bump job RUNS three checks before opening
the PR — the network guard (`CIVITAI_CHECK_PUBLISHED_PINS=1 go test
-run TestScaffoldPinsSatisfyPublished`), the full offline scaffold
suite, and a fresh `civitai app init` scaffold that installs,
typechecks and **builds** against the bumped SDK (which is what would
catch a removed/renamed export).

The first two must pass or this PR is not opened at all. The third
does NOT block it — a failure there opens this PR as a **draft**
instead, because withholding the bump freezes the whole repo (#530).
Its result for this run is stated on the last line below; do not read
"the bump job validates" as "everything passed".

Scaffold build validation: `success`.

### DIFF
--- a/internal/scaffold/scaffold_test.go
+++ b/internal/scaffold/scaffold_test.go
@@ -189,8 +189,8 @@ func TestRenderPageMoney(t *testing.T) {
 	// @civitai/app-sdk@0.14.0+. The pins track the CURRENT published minors — the
 	// pins-vs-published guard (TestScaffoldPinsSatisfyPublished) fails CI if they
 	// fall behind npm, so bump BOTH assertions here in lockstep with the .tmpl.
-	mustContain(t, pkg, `"@civitai/blocks-react": "^0.46.0"`)
-	mustContain(t, pkg, `"@civitai/app-sdk": "^0.37.0"`)
+	mustContain(t, pkg, `"@civitai/blocks-react": "^0.49.0"`)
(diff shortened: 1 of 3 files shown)


ZacxDev:
Rehearsal artifact — closing. This PR is the *evidence* for #530: it proves the nightly's step 9 now passes under node 24 and that step 11 opens a PR, which is exactly what was skipped while the workflow ran npm 10. Not for merge; its base branch is a throwaway.

PR #551, opened 2026-09-11
docs(handoff): external issue triage — only 2 of 21 open issues are external, and #545 sat unreviewed

ZacxDev:
Updates `claudedocs/handoff-external-issue-513-numeric-username.md` in place (one doc per effort — this doc's stated goal is clearing the externally-reported queue).

## What this session did

**Triage + dispatch only.** No code edits, no commits to `main`, no merges.

- **21 open issues; only 2 are not self-filed** — #544 (`xsvm`) and #513 (`Rochet2`).
- **Found `civitai/cli#545` open and unreviewed for 13h** — `xsvm` filed #544 and opened the PR 90 seconds later. `MERGEABLE`/`CLEAN`, 8/8 checks green, zero reviews, zero comments. Note that `lint` is among those 8 and **reports without gating**, so "8/8 green" is not the merge gate having passed.
- **Established that no prior session had touched #544 or #545**, and recorded why that was nearly concluded backwards (see below).
- **Inverted the #513 server-half probe strategy.** The previous approach — find an all-digit-username account, then query its images — dead-ended because all 7 such accounts return `items=0`, i.e. the reported code path was never exercised. The new approach scans listings in bulk and greps raw bytes for any unquoted `"username":`, testing the serializer directly.

## The finding most worth keeping

🔴 **`#544` and `#545` collide across two trackers.** Activity-telemetry `mention-detected` rows for those numbers dated 2026-08-29 → 2026-09-09 are **clawgate task ids** — the cli issues did not exist until 2026-09-10 16:10Z. A telemetry search for prior work surfaces them and reads as a confident "already handled". It was not. Discriminate by date, not by id.

This lesson was also routed to the `activity` skill (separate devrc change), since `mention-detected` had no owning skill and the trap is general: an id is unique only within its own tracker.

## In flight at time of writing

Three dispatched agents whose results were **not known** when this doc was written — a #513 bulk-serializer probe and `/audit-pr` rounds 0 and 1 on #545. The doc says so explicitly so the next reader collects them rather than inferring from their absence.

Two `claim-work` claims are held and named in the doc: `external-issue-513-numeric-username-3` and `cli-545-images-prompt-indent-audit`.

## Ranked-queue note

Numbering is kept stable (rank is half a claim's slug identity). Items 1–4 keep their previous numbers — rank 1 is marked DONE since #526 merged — and the #545 review is appended as rank 5.


ZacxDev:
Pushed `c1b0222` — a **retraction**, which is the point of this update.

## The #513 server coercion is not unreproducible. It is live, and it is root-caused.

The prior session concluded "NOT reproducible on any reachable surface." That was wrong, and it is worth being precise about *why*, because the probe followed the rules and still landed on the opposite of the truth.

**Two independent causes, either of which alone would have produced the same empty result:**

1. It never reached an all-digit-username account that had public images (all 7 it found returned `items=0`).
2. 🔴 The quoted readings it *did* get were **Cloudflare cache HITs**, not origin responses.

That second one is the transferable lesson. The probe read raw bytes, avoided `jq`, and ran a positive control — everything the rules ask — and a CDN with `s-maxage=300` still handed it correct-looking cached copies.

## Verified first-hand, not only by the agent

```
?username=2428023993&limit=3   cf-cache-status: MISS   ->  "username":2428023993   UNQUOTED
?imageId=1446527                                       ->  "username":"2428023993" QUOTED
```

Same user, same route, **different internal path**. `runImageSearch` forks at `image-search.service.ts:127`: `imageId` takes the legacy Postgres path, everything else takes the Meilisearch feed path. The Meili index stored `user.username` with dynamic JSON typing, so an all-digit username is a JSON *number* in the index document.

On a cache-busted re-sweep the agent measured **51/51 unquoted, 0 quoted** — 38 readings that had looked quoted flipped once the cache key changed. The origin coerces 100% of the time.

## The part that changes what we shipped

It is a **numeric coercion**, not just missing quotes:

```
?imageId=622901         ->  "username":"0222"
?username=0222&limit=4  ->  "username":222      leading zero destroyed
?username=222&limit=4   ->  0 items             the printed name round-trips to nothing
```

So **#532's `FlexString` does not repair this.** It is still correct and still wanted — it stops the decode failing — but it made the failure *quieter*, not absent: the CLI now prints a wrong username that looks right. A decode fix cannot recover information the wire already lost.

Added as rank 7: check whether AGENTS.md item 37 / `claudedocs/decisions/37-numeric-username.md` asserts the coercion was unreproducible, since that is now false.

## Also folded in

- **#545 merged** (`feb330c4`) after a public correction — its "Scope Evaluation" was false in 8/8 fields, confirmed by two audit rounds, the second dispatched blind to the first. Verified by content, not ancestry.
- **#552 filed** — the `indentContinuation` call-site ledger, which is what actually closes the forgery class.
- Round-0 trial data recorded here rather than on xsvm's PR: `ran: 1 · changed the outcome: 0 (framing only)`.

The triage ratio from the replaced status block was carried into Gotchas rather than dropped, per the durable-line warning.

Issue #542, opened 2026-09-10
snippet()'s 'every argument is server bytes' invariant is unpinned — the #393 guard cannot see pkg/civitai's saferune call site


ZacxDev:
Found by a blind adversarial audit of #541. **No live defect** — this is an unpinned invariant, filed so it is tracked rather than assumed.

## The gap

`internal/cmd/safeterm_userinput_test.go:92` is the structural guard that enforces the #393 invariant (what may reach a terminal). It parses **only `internal/cmd`'s own non-test sources**, and only matches the spelling `safeTerm(<one arg>)`.

#541 adds a strip at `pkg/civitai/read.go` (`saferune.Strip`, inside `snippet()`) — outside that scan, under a different spelling. The guard cited in `claudedocs/decisions/38-read-body-repair-and-snippet.md` §2 as the reason the `errorLine` option was rejected cannot see the call site that was chosen instead.

## The unpinned claim

`pkg/civitai/read.go:486` asserts in prose:

> 🔴 EVERY ARGUMENT THIS FUNCTION IS EVER GIVEN IS THE SERVER'S OWN BYTES

That is a claim over all present **and future** call sites, with nothing enforcing it. It is true today — all 8 current `snippet(...)` arguments were verified server-supplied (`read.go:256,327,332,340,345,349`, `retry.go:172`, `flexstring.go:107`). A future call site passing user-typed input would be the #393 regression, silently.

## Why it is worth closing

#541 built exactly this kind of ledger for a smaller claim in the same PR (`decodeBodyCallers`, bidirectional, with positive controls). The claim carrying the actual #393 risk is the one left unpinned.

## Closing condition

A merged PR in which a guard fails when `snippet()` (or any `saferune.*` call site outside `internal/cmd`) gains a caller whose argument is not server-supplied — or when the caller set changes at all, bidirectionally. Verified by watching it go red on a deliberately-added bad call site, with the failure naming that guard's own message.

ZacxDev:
Related to #399, which is the same family measured on the other side of the seam: there, deleting `safeTerm(...)` at 20 of 25 sampled `internal/` call sites leaves the suite green. This issue is the `pkg/civitai` half — the guard cited as the reason for choosing the `snippet()` strip site (`safeterm_userinput_test.go:92`) parses only `internal/cmd`'s own sources and matches only the spelling `safeTerm(<one arg>)`, so it is structurally blind to `saferune.Strip`.

[...]

PR #562, opened 2026-09-11
docs(handoff): ranks 12 and 13 shipped; cli#399 is uncovered and the one-instrument prediction is falsified

### LINKED ISSUE #399: safeTerm is unpinned at most of its call sites: deleting it at 20 of 25 sampled sites leaves the suite green
`safeTerm` is the CLI's one gate on server-supplied text reaching a terminal, and it is called from **150 sites in 56 functions** (AST enumeration of `internal/`, non-test files, at `27cd5de`+). Only a handful of those sites are pinned by a test: an adversarial audit of #398 deleted `safeTerm(...)` at each site in turn and **20 of 25 sampled sites left the whole suite green**.

The pinned ones are all *failure-reason* fields — the surfaces #367, #382 and #393 touched. Everything else is unguarded: a refactor can drop the call and CI says nothing.

## Not a regression, and not #398's to fix

Measured identical on `27cd5de` (before #398) and on the #398 branch: the same 20 survive on both. The audit's instruction was explicit — do not fix it in the #393 PR, file it so it is not lost.

## What is unpinned (sample)

| surface | file | what a hostile field could do there |
|---|---|---|
| model / version names, creators, tags | `internal/cmd/models.go`, `model_versions.go` | overwrite the `SHA256 verified` line during a download |
| app name, slug, tagline, description | `internal/cmd/apps.go` (11 calls in `printAppDetail` alone) | forge a review-status line |
| image metadata, prompts read back from the API | `internal/cmd/images.go` (7 in `printImageMetaBlock`) | reorder a displayed line with a bidi control (before #393) |
| download plan: file names, targets, notes | `internal/cmd/download.go` | forge the "already present" / "Saved" lines |
| `app metrics` scope + endpoint tokens | `internal/cmd/app_metrics.go` | — (item 8 keeps these raw on purpose) |
| collections, articles, users, tags, creators | five more files | — |

## Shape of a fix

Not 150 tests. The cheap version is one **table-driven** test per renderer family that feeds a hostile string through the real command and asserts the paired predicate the reason-path tests already use (`assertNoEscapeBytes` / `assertNoInvisibleRunes`: the class is gone AND the words arrived). The expensive-but-durable version is a **ledger**: enumerate the safeTerm call sites by AST, and require each enclosing function to appear in an asserted list of "covered by test X" — failing when the set grows.

Note #398 added `TestSafeTermIsNeverAppliedToUserTypedInput`, which already walks every call site by AST for the opposite property (that none of them sanitises user-typed input). The same walk is the natural place to hang a coverage ledger.

## Why it matters

`safeTerm` is a security control. The repo's own rule is that a guard nobody has watched fail proves nothing — and for 20 of 25 of these sites, nobody has.

ZacxDev:
Closes the agent-setup-onboarding arc. Everything below was measured, not carried forward.

## Shipped

**docs#75** (`d5b34c8`) — ranks 12 + 13 in one PR, since both harden the same guard.

- **Rank 12** — `scripts/drift-notify.mjs` + a `notify` job opens or refreshes **one** issue (found by an HTML marker, not title/label), comments only when the failure *fingerprint* changes, and comments-then-closes on the next clean run. It alerts on **"nothing was verified"** too — skipped job, no report, zero repos verified, or the guard run `--offline`.
- **Rank 13** — the scheduled half now fetches each repo's `block.manifest.json` and compares `scopes` **by set**, so a reordered array doesn't false-fail. Hook lists are date-stamped rather than guessed.

## Verified independently, not accepted from the agent

- **Live scope-drift mutation**: adding `models:read:public` to Sensei's page scopes → **rc 1**, `SCOPES DRIFTED … the page lists \`models:read:public\` which the manifest does not declare`.
- **The two-host split proved itself under a rate limit**: `api.github.com` verified **3 of 8** repos live while `raw.githubusercontent.com` verified **8 of 8** scope lists — the numbers moved independently, which is the cleanest evidence the scope check added zero API budget. The run exited **0 with a loud "could not be reached and were NOT verified"**, not a silent pass.

## cli#399 — investigated, deliberately not fixed

The handoff predicted "likely one instrument fixes both." **Falsified.** cli#557's guard does not reach #399, for three structural reasons each verified: `snippet_args_ledger_test.go:130` scans `os.ReadDir(".")` from `package civitai` so `internal/cmd` is never walked; `:168` matches only `snippet`, never `safeTerm`; and it is an *origin ledger*, not a *coverage* assertion. `safeTerm` has **153 occurrences** in `internal/` non-test sources, and #399's audit found deleting it at 20 of 25 sampled sites leaves the suite green. Filed as rank 14 with the smallest instrument that would cover it.

🔴 **#399 had been auto-closed by a GitHub parser accident** — `6fce127`'s body says *"Does **NOT** close #399"* and the parser matched `close #399` anyway. Already reopened. Recorded as a gotcha: negation does not help.

## Three new ranked items

- **14** — #399 is open and uncovered (above).
- **15** — `drift-notify.mjs` **has never talked to the real GitHub API**; all nine delivery scenarios ran against a local fake. `forcing: gate` — rank 12's closing condition was "a notify step merges", and it has, but its delivery is unproven, so the hole rank 12 named is not yet demonstrably closed.
- **16** — #542's stated closing condition named "any `saferune.*` call site outside `internal/cmd`". There are **two** (`pkg/civitai/read.go`, `internal/genapi/status.go`) and #557's guard reaches neither as such. Recorded so it isn't silently assumed discharged.

Claims `agent-setup-onboarding-12` and `-13` released.

### DIFF
--- a/claudedocs/handoff-agent-setup-onboarding.md
+++ b/claudedocs/handoff-agent-setup-onboarding.md
@@ -19,52 +19,51 @@ plus a `civitai agent-setup` command (all the real logic, in Go, tested).
 
 ## State now
 
-**Rank 6 is DONE, MERGED, DEPLOYED and VERIFIED AGAINST PROD.** Four PRs landed
-across two repos; the audit ladder ran one round on each implementation PR and both
-rounds' findings were fixed and merged. Ranks 4, 8, 9, 10, 11 remain untouched.
+**Ranks 12 and 13 are DONE and MERGED. cli#399 was INVESTIGATED, NOT fixed — by explicit
+operator decision (check-and-report).** This closes the arc: every item the operator
+directed is complete, and the remaining queue is `forcing: none`.
 
-- **Shipped and live:**
+- **Merged this pass:**
 
-| PR | | merged |
+| PR | | commit |
 |---|---|---|
-| docs#71 | `/apps/examples` — the example-app catalog + `check-example-apps.mjs` rot guard (`75035bce`) | 05:42Z |
-| cli#549 | the managed block's ONE-URL Docs line, item 36's trigger widened | 05:52Z |
-| cli#548 | handoff | — |
-| docs#72 | audit fixes: `llms.txt` duplication + the rename guard's missing coverage (`89bb471`) | 16:19Z |
-| cli#553 | audit fixes: the docs-link gate could report ok without probing (`ff7a211`) | 16:19Z |
+| cli#558 | closing sweep: ranks 8/9 marked done, urllib 403 measured gone, ranks 12/13 filed | `3d17591` |
+| docs#75 | ranks 12+13: deliver the sweep's red, and make the page's scope lists a checked claim | `d5b34c8` |
 
-- **Verified against PROD, not inferred** (2026-09-11, after the 16:24:59Z deploy):
-  - `developer.civitai.com/apps/examples` → **200**; `/apps/examples/` → **404**, and
-    the CLI ships the no-slash spelling. Correct.
-  - `llms.txt` → **192 lines** (was 215): `### Examples` ×1, `Introduction to Civitai
-    Apps` ×1, `### Design system` ×1. The duplication is gone.
-  - `CIVITAI_CHECK_DOCS_LINKS=1 go test -run 'TestBlockDocsLinksResolve|TestEveryDocsURLSpellingIsLedgered' ./internal/cmd/`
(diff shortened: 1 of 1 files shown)


ZacxDev:
Superseded by a rebuild on top of #561, which merged into the same file while this was open. Forcing this version would have regressed #561's closures of ranks 4, 10 and 11.

PR #554, opened 2026-09-11
fix(images): add indentContinuation ledger and sanitize inline fields (closes #552)

xsvm:
Closes #552.

### Summary

Establishes a bidirectional AST call-site ledger for `indentContinuation`, replaces `safeTerm` with `safeTermSingle` across 8 inline positions in `images.go` to prevent newline row forgery, and expands the #393 anti-corruption harness to guard `safeTermSingle`.

### Key Changes

| Area | Changes |
| :--- | :--- |
| **Sanitizer** | Added `safeTermSingle` (`strings.ReplaceAll(safeTerm(s), "\n", " ")`) in `internal/cmd/safeterm.go`. |
| **Inline Fields** | Applied `safeTermSingle` to 8 fields (`Model`, `Sampler`, `CfgScale`, `Steps`, `Seed`, `URL`, `Username`, `NSFWLevel`, and `resources[].type/name/weight/hash`) in `internal/cmd/images.go`. |
| **AST & Behaviour** | Added `internal/cmd/indentcontinuation_ledger_test.go` with `TestIndentContinuationCallSitesAreLedgered` (`StructuralLedger` pins 6 sites; `BehaviouralSeam` asserts zero `\n` leaks across all 8 fields + resources). |
| **#393 Hardening** | Extended `internal/cmd/safeterm_userinput_test.go` to audit both `safeTerm` and `safeTermSingle` (152 calls across 67 files verified). |
| **Docs** | Documented prompt indent vs wrap rationale in `README.md`. |

### Verification

```bash
# Closing gate
go test -count=1 ./internal/cmd -run Ledger
# PASS: TestIndentContinuationCallSitesAreLedgered (0.03s)
# PASS: 13 ledger tests (0.622s, exit code 0)

# #393 User-input guard (152 call sites audited)
go test -v -count=1 ./internal/cmd -run TestSafeTermIsNeverAppliedToUserTypedInput
# PASS: TestSafeTermIsNeverAppliedToUserTypedInput (0.03s)

# Regression & Lint
go test -v -count=1 ./internal/cmd -run TestREADMEAnchorLinksResolve
go test -v ./internal/cmd -run TestImages
gofmt -s -l . && go vet ./internal/cmd
```

Issue #552, opened 2026-09-11
Server text can forge output at ~13 tabwriter renderers, and no ledger detects it (#545/#569 follow-up)


ZacxDev:
> **Body rewritten 2026-09-12.** The original scoped this to 8 fields in `images.go` and asked for a ledger keyed on `indentContinuation` call sites. Both were too narrow — a blind audit of #569 measured the real scope. The original text is preserved at the bottom. The amendment comment above remains accurate but is superseded by this body.

## Closing condition

A merged PR adding a ledger over **renderers that print server-supplied text into a line-structured surface** — failing when that set **grows or shrinks** — plus a behavioural case, since a structural check type-checks past a wrong argument.

🔴 **The predicate must not be keyed on a helper's NAME.** "Calls `indentContinuation`" is satisfied by construction and was **green at base**; a guard keyed to a function name can only find the places that already call it. That was this issue's original mistake.

**Checked by:** the PR merging, and a mutation showing the ledger reddens when a renderer in the table below is added *or* removed.

## Scope — measured, not estimated

`#545` fixed prompt/negative in `printImageMetaBlock`. `#569` fixed 16 inline fields in `images.go` and closed the tab vector there. **The class is still live at ~13 other renderers**, all the same shape — `tabwriter.NewWriter(out, 0, 2, 2, ' ', 0)` fed plain `safeTerm(...)`:

| file:line | command |
|---|---|
| `creators.go:107` | `creators search` |
| `tags.go:110` | `tags search` |
| `models.go:223`, `models.go:249` | `models search` / `models get` |
| `model_versions.go:186` | `model-versions get` |
| `collections.go:201` | `collections` |
| `articles.go:201` | `articles` |
| `users.go:132` | `users get` |
| `apps.go:350` | `apps` |
| `app_metrics.go:392`, `:400` | `app metrics` |
| `workflow_settlement.go:70`, `:73` | settlement |
| `workflows.go:136-143` | `workflows` |
| `generate.go:1901`, `:1903` | `generate` |

Measured through the real renderers:

- `printCreatorList` with `username = "alice\t99\thttps://evil.example/steal"` renders a **fully aligned, entirely attacker-controlled row**, real values pushed right.
- `printCreatorList` with `username = "alice\nbob\t42\thttps://evil.example/row"` renders **3 lines** — a forged row at column zero.
- `printTagList` reproduces the tab case.

## 🔴 Two hazards, not one — a replacement predicate must cover both

[...]

PR #564, opened 2026-09-12
fix(safeterm): ledger every call site by enclosing function, and pin 12 that were unpinned

ZacxDev:
Addresses cli#399. Leaving the issue open for a human to close — it has been
auto-closed twice already by parsers reading the number out of prose, so this
body deliberately contains no keyword that would do it a third time.

## The defect

`safeTerm` is the CLI's one gate on server-supplied text reaching a terminal.
#399 measured that deleting `safeTerm(...)` at **20 of 25 sampled sites left the
whole suite green**. Nothing could say which sites were pinned, so nobody could
tell a refactor that dropped one from a refactor that did not.

Re-measured here at the level of **functions**, against a tree frozen at
`cf5e4a8` (`origin/main`): of the **59 functions** holding the 151 calls,
**36 survived losing every `safeTerm` call they had**. 23 were killed by some
test.

Method, so it can be repeated: each function's `safeTerm(x)` calls were
rewritten to `x` by **byte splice** — delete `safeTerm(` and its matching `)`,
nothing else in the file moves — then `go test ./internal/cmd/ -count=1` was run
on a frozen copy. Controls: the unmutated copy was green first; the mutated tree
was diffed to prove the mutant was on disk (a `SURVIVED` from an unmutated tree
is the whole failure mode); and neutering `safeTerm` itself reddened 20 tests,
so `SURVIVED` is not a harness that cannot see red.

## What this adds

**1. Enclosing-function attribution on the walk that already existed.**
`internal/cmd/safeterm_userinput_test.go` flat-`ast.Inspect`ed each file, so it
saw every call and knew where none of them lived — deleting one moved its count
from 151 to 150, nowhere near its floor of 100, and it stayed green. It now
walks `file.Decls` and recurses into each `*ast.FuncDecl`. One shared
`scanSafeTermCallSites` feeds both tests, with a **totality control**: a flat
count of the same calls must equal the attributed count, so a call in a
package-level initialiser (invisible to a `Decls` walk, and therefore owned by
no row) is a loud failure rather than a silent gap.

**2. A coverage ledger keyed by enclosing function**
(`internal/cmd/safeterm_coverage_test.go`), 60 rows, every verdict measured.
Keyed by function, not by argument name, because #398's `bareIdentArgs` is
keyed by name and its own comment concedes why that is weak ("the same name
holds a server-returned id at other call sites").

Four independent assertions, each with its own message:

| check | fires when |
|---|---|
| `GREW` | a function calls `safeTerm` and has no row |
| `SHRANK` | a row names a function that no longer calls `safeTerm` |
| `GHOST` | a row names a test this package does not declare |
| `RATCHET` | the `notCovered` count went up |

`notCovered` is a **first-class value**, not a hole. 25 rows say it, each with
what a hostile field could do at that surface. The count is logged and capped at
exactly today's number, so it can only shrink.

A named test is **resolved to a declaration** before it is believed — the
pattern from the module root's `saferune_callers_ledger_test.go`, which had to
add exactly this after a rename left two prose statements naming a function that
no longer existed while the suite stayed green.

**3. `TestReadRenderersStripTheInvisibleClass`** — #399's own "cheap version": a
table-driven test per renderer family, driving the real renderer with a hostile
fixture. The predicate is **paired and per-field**: no rune of the class on the
terminal, AND every named field arrives with its two visible halves adjacent.
Each field carries its own marker, so a renderer that drops one column fails on
that column by name — a single shared "the fixture arrived" control cannot see
that. The class check is `invisibleOrBidiRunes`, the independently-built
predicate (`unicode.Cf` plus the measured U+2800 counterexample), not
`saferune.Stripped` — a guard that asks the implementation what it strips is
satisfied by an implementation that strips nothing.

## A real hole this found

`nonModelFileMarker` (`internal/cmd/read.go`) returned the uploader's
`files[].type` **raw**, interpolated into a header line — the first line of
`model-versions get` and a row of `models get`'s version table. The identical
field is `safeTerm`'d in the files table a few lines below. It was found by
driving the function with a hostile value, not by reading it, and it is fixed
here (one line + the ledger row + the `bareIdentArgs` origin note the existing
guard immediately demanded). Both callers are human renderers; no `--json` path
reaches it, so `internal/ui/CONVENTION.md` rule 1 is untouched.

## The #399 closing condition, demonstrated

Site: `printCollectionDetail` (5 calls) — `SURVIVED` on the frozen baseline.

**Before**, on frozen `cf5e4a8`, all 5 calls spliced out:

```
ok  	github.com/civitai/cli/internal/cmd	40.436s
```

**After**, the same deletion on this branch:

```
--- FAIL: TestReadRenderersStripTheInvisibleClass/`collections_get`_detail_(printCollectionDetail,_joinCollectionTags)
    safeterm_renderers_test.go:336: `collections get` detail … put U+200B, U+202E, U+2800, … on the terminal.
        safeTerm is the ONE gate on server-supplied text reaching a terminal: an invisible rune is a
        separator no wrapper splits on, and a bidi control changes the order the line is DISPLAYED in,
        so what the user reads is not what the bytes say (civitai/cli#399, #393)
    safeterm_renderers_test.go:345: … did not render "cdname" as "FIXTURE-cdname-tail". Either the field
        stopped being printed — in which case the class check above is VACUOUS for it — or the whole value
        was dropped instead of just the class
--- FAIL: TestSafeTermCallSitesAreCoveredByANamedTest
```

11 further functions moved `SURVIVED` → `KILLED` the same way (`appCardAuthor`,
`appViewOwnedAdvice`, `formatFileList`, `joinCollectionTags`, `printAppDetail`,
`printAppList`, `printArticleList`, `printCollectionList`, `printCreatorList`,
`printTagList`, and the new `nonModelFileMarker`), each verified by deletion.

## The headline number

Re-running the full 60-function sweep against **this** branch: every function's
mutant is killed. `SHRANK` catches wholesale removal for all 60 rows,
`notCovered` ones included.

| | `origin/main` (cf5e4a8) | this branch |
|---|---|---|
| functions whose `safeTerm` calls can all be deleted with the suite green | **36 / 59** | **0 / 60** |
| functions whose OUTPUT a test inspects | 23 | 35 |

Those are different claims and the ledger's doc comment says so at length: the
ledger pins that a function still calls `safeTerm` **at all**; a named test pins
that the class does not reach the **screen**, which is the only thing that
catches deleting one of a function's twelve calls. Do not read 35/60 as a
percentage of the 152 call sites.

## Mutation matrix on the guard itself

Each mutant run against a copy of this branch; the control (unmutated copy
green) ran first.

| mutant | result |
|---|---|
| new function calling `safeTerm(v.Name)`, no row | KILLED — `no row in safeTermCoveredBy` |
| a row names `TestThisNameWasNeverDeclared` | KILLED — `name a test that this package does not declare` |
| a row for `printFinding`, which calls no `safeTerm` | KILLED — `no longer calls safeTerm` |
| a covered row flipped to `notCovered` | KILLED — `are ledgered notCovered, and the cap is 25` |
| a row with an empty `why` | KILLED — `has an empty why` |
| `safeTerm` call in a package-level var | KILLED — `sit inside a func declaration` |

Each died by **its own** message, not another guard's error and not a compile
failure; the mutant in row 1 passes a selector rather than a bare identifier so
`bareIdentArgs` cannot fire first and mask it.

## Verification

- `make ci` — green (see below).
- `make lint` — `0 issues`, and the instrument was validated: replacing the
  escapes in `classProbe` with raw bytes makes it report
  `ST1018: string literal contains Unicode format characters`, the exact class
  AGENTS.md records biting this repo. Run under nixpkgs golangci-lint 2.13.2;
  CI pins 2.12.2.

## Not done, deliberately

- **No AGENTS.md item.** `AGENTS.md` is 47 bytes under its own ceiling, so a new
  item means evicting an existing one — a separate exercise. The obligation
  ("add a row or say `notCovered`") is stated in the guard's failure message,
  which is where a reader meets it.
- **No README change.** The published contract already says invisible and
  direction-reversing characters are removed from server text before it reaches
  the terminal; the `nonModelFileMarker` fix makes that sentence *more* true
  rather than changing it.
- **Scope held to `internal/cmd`.** Two `saferune` callers live outside it —
  `pkg/civitai`'s `snippet` (`read.go`) and `internal/genapi`'s
  `hasPrintableContent` (`status.go`). Neither is in this ledger, and extending
  it there is a separate decision, not a bump.

https://claude.ai/code/session_01RzVihMt3EZDbJNxSN35DUR

### DIFF
--- a/internal/cmd/read.go
+++ b/internal/cmd/read.go
@@ -123,7 +123,15 @@ func nonModelFileMarker(files []civitai.ModelVersionFile) string {
 	if typ == "" {
 		typ = "Other"
 	}
-	return "[" + typ + "]"
+	// 🔴 SERVER TEXT, SO IT GOES THROUGH THE GATE — civitai/cli#399 found this
+	// one by driving the renderer rather than by reading it. `files[].type` is
+	// uploader-supplied and this marker is INTERPOLATED INTO A HEADER LINE
+	// (`printModelVersionDetail`'s first line, and a row of `printModelDetail`'s
+	// version table), which is precisely where an escape sequence is worth the
+	// most to an attacker. The same field is safeTerm'd in the files table a few
+	// lines further down; here it was not, and nothing could see the difference
+	// until a hostile fixture was fed through this function.
+	return "[" + safeTerm(typ) + "]"
 }
 
 // checkLimit validates a --limit against an endpoint's documented maximum. It
--- a/internal/cmd/safeterm_coverage_test.go
+++ b/internal/cmd/safeterm_coverage_test.go
@@ -1,0 +1,454 @@
+package cmd
+
+import (
+	"fmt"
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"os"
+	"sort"
+	"strings"
+	"testing"
+)
+
+// civitai/cli#399 — safeTerm WAS UNPINNED AT MOST OF ITS CALL SITES, AND
+// NOTHING COULD SAY WHICH.
+//
+// safeTerm is the CLI's ONE gate on server-supplied text reaching a terminal
+// (see internal/saferune's package doc for the class it removes, and
+// safeterm.go for what a hostile field does without it). It is called from 152
+// sites in 60 functions. An adversarial audit of #398 deleted `safeTerm(...)`
+// at each of 25 sampled sites in turn: 20 of them left the WHOLE SUITE GREEN.
+// A refactor could drop the call and CI would say nothing.
+//
+// 🔴 THE PROBLEM WAS NOT A MISSING TEST, IT WAS A MISSING NUMBER. Nobody could
+// answer "which of these sites is pinned?" without re-running that audit by
+// hand, so the answer was never written down and never re-checked. A table of
+// 60 rows beats 152 tests: this is the answer, in a form a compiler keeps
+// honest.
+//
+// It is deliberately NOT a claim that every site is covered. Most rows say
+// notCovered, and that is the point — the absence is now IMPOSSIBLE NOT TO
+// WRITE DOWN, counted by the test, and capped by maxUncoveredSafeTermFuncs so
+// it can only shrink. A ledger that could only record coverage would have been
+// satisfied by deleting the rows nobody had got to.
+//
+// Why the ENCLOSING FUNCTION is the key, and not the call site's argument
+// expression: #398's bareIdentArgs is keyed by argument name and its own
+// comment concedes the weakness — "the same name holds a server-returned id at
+// other call sites". A function is the unit a test actually drives, the unit a
+// refactor moves, and the unit whose row goes stale the moment its last
+// safeTerm call is deleted.
+//
+// 🔴 TWO DIFFERENT CLAIMS LIVE HERE, AND CONFUSING THEM WOULD OVERSTATE WHAT
+// THIS BUYS. Say plainly what each half covers:
+//
+//   - The LEDGER pins that a function still calls safeTerm AT ALL. Deleting a
+//     function's LAST safeTerm call empties its row and fails SHRANK below —
+//     for every row, notCovered ones included. That is the whole set of 60
+//     functions, and it is what #399's headline measurement was about.
+//   - A NAMED TEST pins that the class does not reach the SCREEN. That is
+//     strictly more, and it is the only thing that catches deleting ONE of a
+//     function's twelve calls, or a new field printed without the gate.
+//
+// So a notCovered row is not "unguarded"; it is "guarded against wholesale
+// removal, not against a per-field regression". Do not read the covered count
+// as a coverage percentage of the 151 call sites — it is a count of FUNCTIONS
+// whose output a test actually inspects.
+//
+// The walk is scanSafeTermCallSites in safeterm_userinput_test.go — the same
+// one #398 built for the opposite property (that no site sanitises what the
+// USER typed). #399's own body asked for it there: "the same walk is the
+// natural place to hang a coverage ledger."
+
+// safeTermCoverage is one row: what pins the safeTerm call(s) in one function.
+type safeTermCoverage struct {
+	// test is the name of the test that goes RED when the safeTerm call(s) in
+	// this function are deleted — measured by deleting them, not by reading the
+	// test and believing it. notCovered means no test in this package does.
+	//
+	// 🔴 IT IS RESOLVED TO A DECLARATION, not used as a label. A ledger naming a
+	// test that does not exist is worse than no ledger: it reads as coverage and
+	// stops anyone looking. checkCoveringTestsResolve is what makes the name a
+	// binding, so renaming or deleting the test is red HERE.
+	test string
+	// why is what that test asserts — or, for a notCovered row, what a hostile
+	// server field could do at this surface today. Every row must say something;
+	// an empty why is a row nobody thought about.
+	why string
+}
+
+// notCovered is a FIRST-CLASS LEDGER VALUE, not a hole in the table.
+//
+// Writing it down is the whole mechanism. The alternative — leaving a function
+// out until someone writes its test — makes "unmeasured" and "covered"
+// indistinguishable, which is the state #399 found the repo in.
+const notCovered = ""
+
+// 🔴 EVERY ROW'S VERDICT WAS MEASURED, NOT READ OFF THE TESTS.
+//
+// Method, so it can be repeated: for each function below, every `safeTerm(x)`
+// inside it was rewritten to `x` by byte splice (nothing else in the file
+// moved), then `go test ./internal/cmd/ -count=1` was run against a tree frozen
+// at cf5e4a8. A named test is one that went red; notCovered means the suite
+// stayed green. Controls on that sweep: the unmutated tree was green first, the
+// mutated tree was diffed to prove the mutant was actually on disk, and
+// neutering safeTerm itself reddened 20 tests — so a SURVIVED verdict is not a
+// harness that cannot see red.
+//
+// The baseline that produced, at cf5e4a8 and BEFORE this change: 23 of the 59
+// functions were killed by some test, 36 SURVIVED losing every safeTerm call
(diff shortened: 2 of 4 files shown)


ZacxDev:
🔴 **Heads-up: this collides with #554, textually and semantically.** Flagging before either merges, since both were developed in parallel against `cf5e4a8` and neither can see the other.

## Textual

`git merge-tree --write-tree refs/pull/554/head refs/pull/564/head` exits **1** — a real conflict, not a marker-grep guess. Both PRs rewrite the same `bareIdentArgs` lookup in `internal/cmd/safeterm_userinput_test.go`:

- this PR moves it to `bareIdentArgs[s.arg]` as part of the per-function restructure (`:173`)
- #554 adds a new entry to the map: `"s": "PASSTHROUGH: safeTermSingle forwards its argument to safeTerm"`

## Semantic — the part a clean merge would not save you from

I audited #554 and raised that `"s"` entry as 🔴: `s` is the most common local-variable name in Go, so allowlisting it blinds this harness to the #393 defect wherever the variable happens to be named `s`. Measured, in a detached copy:

| injected | verdict |
|---|---|
| `s := userTypedPrompt; … safeTerm(s)` | **SURVIVED** |
| `zzUnknownIdent := userTypedPrompt; … safeTerm(zzUnknownIdent)` | **KILLED** |

Your PR strengthens this exact harness; #554 punches a hole in it. Merging both in either order yields a harness that is *simultaneously* more thorough per-function and blind to a whole class of argument name. That is worse than either alone, and no conflict marker would tell you.

## Suggested order

1. **#554 fixes its 🔴 first** — the fix is ~2 lines (skip `safeterm.go` in the walk, or rename the parameter to something non-generic like `serverText`), which removes the `"s"` entry entirely and with it most of the conflict.
2. Then this PR rebases onto it, or vice versa — but **test-merge and run the suite on the merged tree**, not on either branch. A green run on this branch says nothing about the tree its merge creates.

## One more thing worth reconciling

Your per-function measurement — **36 of 59 functions survived losing every `safeTerm` call they had** — is a much better predicate than the one I wrote into #552, which asked for a ledger over "renderers that print server-supplied text" and in practice yielded "call sites that call `indentContinuation`" (satisfied by construction, green at base). I've amended #552 accordingly.

There is likely one ledger here, not two. Worth deciding deliberately rather than landing both and discovering the overlap later — the predicate you're using looks like the right one.

I have not audited this PR; this comment is only about the interaction.

ZacxDev:
Worked the #554 collision end to end. **The conflict is four effects, not one**, and only the first is textual. I built the merged tree and got it fully green — patch below.

## What's actually wrong

| # | effect | visible as |
|---|---|---|
| 1 | both edit `bareIdentArgs` | `merge-tree` exit 1 — the only one git shows you |
| 2 | `isSafeTermCall` matches only `safeTerm`; #554 converts 16 sites to `safeTermSingle` | **152 → 137 scanned**; those 16 leave the harness entirely |
| 3 | #554's `"s"` entry blinds the harness in all ~67 files | nothing — it just passes |
| 4 | `safeTermSingle` is a new function with no `safeTermCoveredBy` row | `TestSafeTermCallSitesAreCoveredByANamedTest` fails |

🔴 **Effect 2 is the dangerous one.** Your `isSafeTermCall` is keyed to the single name `safeTerm`. A naive merge silently drops 16 call sites from the harness, and the *only* thing that says so is your `bareIdentArgs` shrank-check tripping on a now-unused `"h"` — which reads as a stale note, not as lost coverage. I hit exactly that and nearly filed it as a nit.

## The fix, and why not the obvious one

#554 solves its half by allowlisting the argument name `"s"`. That's wrong invisibly: `s` is the most common local name in Go, so one entry blinds every file. Measured on its branch — `s := userTypedPrompt; … safeTerm(s)` in `images.go` **SURVIVED**; the same injection named `zzUnknownIdent` was **KILLED**.

The structural fix uses the attribution you already built. Three pieces:

1. **`scannedSanitizers`** — `isSafeTermCall` matches `safeTerm` *or* `safeTermSingle`, so a wrapper can't silently shed coverage. Restores 137 → **153**.
2. **`sanitizerComposers`** — a ledger (not a file exemption) of functions *in `safeterm.go`* that may call `safeTerm` on their own parameter. That's composition, not rendering. Keyed by **enclosing function**, so a new function in that file does not inherit the pass. `userTypedArgs` still applies — composing is no licence to sanitise the user's bytes.
3. **`TestSanitizerComposersAreLedgered`** — bidirectional, plus a positive control, because a one-sided check here is worse than none.
4. A `safeTermCoveredBy` row for `safeTermSingle`. **Watched, not read**: deleting `s = safeTerm(s)` reddens `TestReadRenderersStripTheInvisibleClass/images_search_table`, `TestImagesSearchMetaResourceSanitized` and `TestImagesSearchBaseModelSanitized`. Worth noting we're **at** the uncovered cap (25/25), so `notCovered` was not available.

## Verified on the merged tree

- `make ci` — **21 packages ok**, rc 0. `golangci-lint` — **0 issues**.
- Mutation controls, each in a `cp -a` copy with `.git` removed first:

| control | result |
|---|---|
| `safeTerm(s)` on user-typed input in `images.go` (the case that SURVIVED under `"s"`) | **KILLED** |
| new unledgered function in `safeterm.go` calling `safeTerm` | **KILLED** — `UNLEDGERED COMPOSER: zzSneakyRenderer` |
| stale `sanitizerComposers` entry | **KILLED** — `STALE COMPOSER` |
| drop `safeTermSingle` from `scannedSanitizers` | **KILLED** — back to 137, stale `"h"` |

## 🔴 Where it has to go

**The reconciliation fits in neither PR alone**, and I checked rather than assumed:

- not #564 — `safeTermSingle` doesn't exist here, so `sanitizerComposers` would be stale on arrival
- not #554 — your scanner refactor isn't there

So it belongs to **whichever lands second**. My recommendation: **this PR lands first** (it's self-contained and green alone — I baselined it), then #554 rebases and carries the patch, which *also* replaces its `"s"` entry and fixes its 🔴 2 structurally.

If you'd rather go the other way, say so and I'll rework it onto this branch instead.

Patch is on #554 as well. I have not pushed to either branch.

Issue #566, opened 2026-09-12
Three uploader-controlled surfaces on the download path reach the terminal raw, and the safeTerm ledger structurally cannot demand a row for them


ZacxDev:
Found by the adversarial audit of cli#564. **Pre-existing — not a regression from that PR.** `git diff --stat cf5e4a8 8063caa -- internal/cmd/download.go` is empty, and `git cat-file -e cf5e4a8:internal/cmd/download.go` confirms the path existed at base, so the empty diff is meaningful.

## Why the new ledger cannot catch these

cli#564's `safeTermCoveredBy` only fires `GREW` for functions that **already call `safeTerm` at least once**. All three surfaces below call it **zero** times, so no row is ever demanded. This is the ledger's structural blind spot, worth recording rather than assuming the ledger covers the download path.

## 1. `(*progressWriter).line()` — `download.go:1043,1045`

Embeds the server's `files[].name` **raw** inside a `\r`-rewritten progress line, on every download: the TTY branch (`:1024`), the non-TTY branch (`:1026`), and `done()` (`:1038`).

Reproduced:
```
"\r  weights\x1b[1A\x1b[2K​‮⠀evil.safetensors  500 B / 1000 B (50%)\r…"
```

A `\r`-rewritten line is where a cursor escape is worth most — cli#564's own ledger says exactly that in its `(*ttyPollReporter).tick` row.

## 2. `checkTargetCollisions` — `download.go:702`

```go
fmt.Fprintf(&b, "      - [id %d] %s (%s, %s)\n", f.ID, f.Name, dashIfEmpty(f.Type), …)
```

The **literal twin** of the gated line 44 lines above it:

```go
// download.go:658 — same data, same shape, sanitized
fmt.Fprintf(&b, "  - [id %d] %s (%s, %s)\n", f.ID, safeTerm(f.Name), safeTerm(dashIfEmpty(f.Type)), …)
```

Reproduced, with `formatFileList` on the identical fixture as the clean positive control. **This is the same defect shape cli#564's headline `nonModelFileMarker` fix closed** — one renderer sanitizing a field while its sibling two lines away does not.

## 3. Error messages — `download.go:787, 810`

```go
fmt.Errorf("download %s: %w", f.Name, err)
fmt.Errorf("SHA256 mismatch for %s — expected %s, got %s …", f.Name, …)
```

[...]

PR #572, opened 2026-09-12
fix(download): route every uploader-controlled string on the download path through safeTerm

ZacxDev:
Addresses #566. Six uploader-controlled surfaces on `internal/cmd/download.go` reached the terminal without `safeTerm`. #566 named three; sweeping the file for the same shape found three more, and writing the guards found a seventh problem that the first fix had created.

## What was raw

| # | Site | Surface | In #566? |
|---|---|---|---|
| 1 | `(*progressWriter).line` | the server's `files[].name` inside a `\r`-rewritten line — `Write`'s TTY branch 10×/s, `done()` once | yes |
| 2 | `checkTargetCollisions` | each colliding file's name **and** type, the literal twin of `formatFileList`'s gated line 44 above | yes (name/type) |
| 3 | `checkTargetCollisions` | the group line's mixed-origin `target` | no — found by sweep |
| 4 | `downloadOne` | `download %s`, `SHA256 mismatch for %s` | yes |
| 5 | `downloadOne` | `install %s` | no — found by sweep |
| 6 | `downloadStatusError` | the file name in all four HTTP-status errors, incl. the **401 an anonymous download of a gated file hits on the first run** | no — found by sweep |
| 7 | `writePart` | `create %s`, `streaming %s`, `finalize %s` | no — found by sweep |

`cmd/civitai/main.go` renders a returned error as `"Error: " + err.Error()` and filters nothing, so an error string is a terminal surface exactly like a table cell — with a worse payload: the mismatch line is the CLI *asserting an integrity failure*, and the 401 line is the CLI *telling the user to log in*.

## 🔴 The finding that only appeared because the guards were written first

Sanitising the `%s` while wrapping the cause with `%w` **puts the hostile bytes straight back**. `*fs.PathError` and `*os.LinkError` render their paths with no quoting at all, so

```go
fmt.Errorf("install %s: %w", safeTerm(target), err)
```

emitted `target` sanitised, a colon, then the *same* `target` hostile — and on this path that repeated value's leaf is the server's own file name. The gate was decorative.

(This block originally illustrated the point with the `create output directory` call. Round 1 removed that gate — see **Round 1** below — and its "built from the server's file name" claim was false *for that one site*, which is what the audit caught.)

This was **measured, not reasoned**: the guards in `download_safeterm_test.go` were written against the `%s`-only fix, and two of them went red the first time they ran (`[U+001B ESC U+001B ESC U+202E RLO U+200B ZWSP] reached stderr`).

`safeTermErr` (in `safeterm.go`) strips a wrapped cause's message while `Unwrap`ing to the original, so `errors.Is`/`errors.As` — and therefore the exit-code classifier (items 7 and 24) — are unaffected. There is an explicit assertion for that. Applied at five of the six `%w` sites on this path — the sixth, `create output directory`, is deliberately ungated (**Round 1** below).

## Mutation matrix — watched, not read

Method: splice the narrowest expression that can be wrong out of the source on its own (never together with its enclosing statement), confirm the mutant is on disk via `git diff --numstat`, run `go test ./internal/cmd -count=1 -v`, read the runner's own `--- FAIL:` lines, restore, re-assert the tree is byte-clean.

**Baseline: green.** `go test ./internal/cmd -count=1 -v` reports **3076** `--- PASS:` lines, 2 skips, 0 failures. It was **3077** before Round 1 deleted the one `output directory` subtest — a delta of exactly one `t.Run` (10 → 9 in `download_safeterm_test.go`). An audit that counted 3076 on the *pre-Round-1* tree did so in an environment with no `.git`, where one further test skips; that **corroborates** 3077 rather than contradicting it.

| Mutant | Site | Verdict | Test that went red |
|---|---|---|---|
| M1 | `checkTargetCollisions` group-line `target` | **KILLED** | `TestCheckTargetCollisionsSanitizesServerFields` |
| M2 | `checkTargetCollisions` per-file `f.Name` | **KILLED** | `TestCheckTargetCollisionsSanitizesServerFields` |
| M3 | `checkTargetCollisions` per-file `f.Type` | **KILLED** | `TestCheckTargetCollisionsSanitizesServerFields` |
| M4 | `(*progressWriter).line` `p.name` | **KILLED** | `TestProgressWriterLineSanitizesServerName` (both branches), + `…/saved_line`, + the #564 ledger's `CALL REMOVED` |
| M5 | `downloadOne` `download %s` | **KILLED** | `TestDownloadOneErrorsSanitizeTheServerName/transport_failure` |
| M6 | `downloadOne` `download` **`%w` cause** | SURVIVED | — *see below* |
| M7 | `downloadOne` `SHA256 mismatch %s` | **KILLED** | `…/SHA256_mismatch` |
| M10 | `downloadOne` `install %s` | **KILLED** | `…/install` |
| M11 | `downloadOne` `install` **`%w` cause** | **KILLED** | `…/install` |
| M12 | `downloadOne` `Saved %s` | **KILLED** | `…/saved_line` |
| M13 | `downloadStatusError` `name` (all four arms) | **KILLED** | `…/HTTP_status` |
| M14 | `writePart` `create %s` | **KILLED** | `TestWritePartErrorsSanitizeTheServerName/create` |
| M15 | `writePart` `create` **`%w` cause** | **KILLED** | `…/create` |
| M16 | `writePart` `streaming %s` | **KILLED** | `…/streaming` |
| M17 | `writePart` `finalize %s` | SURVIVED | — *not driven* |
| M18 | `safeTermErr`'s strip itself | **KILLED** | `…/install`, `writePart/create`, + the ledger and composer guards |
| M19 | `writePart` `streaming` **`%w` cause** | SURVIVED | — *not driven* |
| M20 | `writePart` `finalize` **`%w` cause** | SURVIVED | — *not driven* |

**14 KILLED / 4 SURVIVED over 18 mutants.** (Was 16/4 over 20: M8 and M9 are gone with the gate they covered — see **Round 1**.) Re-run in full after the Round-1 edit, not carried over: every remaining mutant was re-applied and re-scored, and **no mutant failed to compile** — a BUILD-BREAK verdict proves nothing, so the harness now detects and reports one explicitly. The four survivors are the same four, for the same reasons.

The exact assertion text a reader can check, verbatim from the sweep:

```
SAFETERM REGRESSION on the download success path: [U+001B ESC U+001B ESC U+202E RLO U+200B ZWSP]
  reached the terminal beside the `SHA256 verified` claim the line exists to make
SAFETERM REGRESSION in writePart (create %s): [U+001B ESC U+001B ESC U+202E RLO U+200B ZWSP]
  reached stderr
```

### Harness validation

A SURVIVED verdict is only meaningful if the harness can go red.

* **Negative control**: `safeTerm` neutered to the identity (spelled `saferune.Strip("") + s`, so it compiles) → **KILLED: 103 distinct `--- FAIL:` lines** (top-level tests *and* subtests; the earlier "34" counted top-level tests only — a counting-method difference, not a changed result), and the package drops from 3076 to 2973 passing. (First attempt spelled it `return s`, which left `saferune` unused and reported `BUILD-BREAK` — that run proved nothing and was re-done as `saferune.Strip("") + s`. Recording it because a BUILD-BREAK reads like a result and is not one.)
* **Positive control**: `TestDownloadFixtureIsHostile` asserts the fixture carries exactly 4 and 3 hazard runes, so `dlHazardRunes(...) == nil` elsewhere cannot be a zero from nothing being rendered.
* **Clean twin control**: `formatFileList` over the *identical* fixture is asserted to render the expected strings, so a failure in `checkTargetCollisions` is that renderer's and not the fixture's.
* Verdicts are read from the runner's per-test `--- FAIL:` lines, not from an exit code; every mutant was confirmed on disk before its suite ran and the tree re-asserted byte-clean after.

### Why the four survivors survive — named, not rounded away

* **M6** — the cause on that path is a `*url.Error`, and Go renders its URL with `%q`, which already escapes every rune in the class. `safeTermErr` there is defence in depth against a future cause that does not quote; it is **not** coverage today, and the ledger row and the test header both say so.
* **M17/M19/M20** — reaching them needs a failing `Close` (full disk, pulled descriptor) or a stream error whose *own* message carries hostile bytes. Neither is driven, and neither is claimed.

## Round 1 — an audit fix, and what it removed

An adversarial audit of this branch found that one of the gates above **could never fire**.

`downloadOne`'s `create output directory %s: %w` was handed `filepath.Dir(target)`, and `target`
comes only from `targetPath`, whose four returns are `o.out` (user-typed `--out`),
`filepath.Join(dir, base)`, `filepath.Join(o.outDir, base)` and bare `base` — where
`base = filepath.Base(f.Name)`. `filepath.Base` strips every separator, so **the server's name is
always the leaf**, and `filepath.Dir` drops it; the bare-`base` return yields `"."`, which the
condition skips. Under `--layout`, `routeDir` returns `root` (user-typed `--root`) or
`filepath.Join(root, sub)` where every `sub` in `layoutFolders` is a compile-time constant — the
server's `fileType`/`modelType` only select the map key, and reach the returned `note`, never the
returned `dir`. `MkdirAll`'s `*fs.PathError` carries a prefix of `dir`, so `safeTermErr` had
nothing to strip either.

So the only bytes that gate could remove are ones the **user typed** — and `internal/saferune`'s
package doc is explicit that the strip is not applied to those. It read as coverage while
providing none.

**M8/M9's `KILLED` verdict was spurious.** The killing subtest handed `downloadOne` a target of
`<tmp>/<hostile name>/weights.bin` — a shape `targetPath` **cannot produce**, since the hostile
name appears as a *directory* component. Breakable, but not reachable. (Corroborating:
`generate_output.go:413` carries the identical `fmt.Errorf("create output directory %s: %w", dir,
err)` **ungated**, and never has been treated as a defect.)

What Round 1 changed:

* `download.go` — both wrappers reverted at the `MkdirAll`, with the measurement recorded at the
  site so a future sweep does not re-add the gate.
* the `output directory` subtest deleted (it built the impossible target), with a comment saying why.
* `bareIdentArgs["dir"]` removed — its note claimed the leaf came from `routeDir`'s mapping of the
  server file/model type, which is false. **`"partPath"` stays**: `target + ".part"` really does
  inherit the server leaf.
* `downloadOne`'s ledger `why` drops the mkdir clause (8 calls → 6, 7 killed → 5) and records why
  the mkdir error is deliberately absent; `safeTermErr`'s `why` re-points at the subtests that
  actually pin it.
* `safeterm.go` — `safeTermErr`'s doc illustrated the `%s`+`%w` bug with the mkdir call, which no
  longer exists and whose "built from the server's file name" claim was false *for that site*.
  Re-pointed at `install`, where it is true.
* `TestDownloadErrorsReachStderrUnfiltered`'s second arm downgraded `t.Errorf` → `t.Log`. It fired
  if `main.go` ever gained a sanitiser — a change its own message called *"not wrong"*. A gate that
  goes red on a correct change is one people learn to click through. The load-bearing first half,
  pinning that `main.go` filters nothing today, is unchanged.
* `README.md` — the `SHA256 mismatch for` Troubleshooting row's trailing historical clause trimmed
  to the contract half. AGENTS.md scopes the README to the published contract; that clause was
  changelog. (Checked first that the row is not generated: `exitcodes_doc.go` generates the *Exit
  codes* section, not the Troubleshooting index.)

`maxUncoveredSafeTermFuncs` **stays 24** — `downloadOne` keeps other gated calls, so no function
moved between covered and notCovered. The harness now reports `165 safeTerm call site(s) in 66
function(s): 42 COVERED, 24 NOT COVERED (cap 24)`, a delta of exactly the 2 removed sites.

**Ledger negative control**, because removing a row is only meaningful if the ledger can notice
it: re-adding `"dir"` to `bareIdentArgs` goes red with its own specific message —
`bareIdentArgs classifies "dir", which is no longer passed to safeTerm anywhere. A stale note
reads as coverage; delete it or fix the name.`

## Driven vs. read

* **Driven** (a test builds the state and inspects the bytes): all of surfaces 1–7. #566 flagged the error surface as "read, not driven" — it is now driven, through the `civitai.Downloader` seam, including the `*fs.PathError` and `*os.LinkError` paths built with a real file blocking a directory.
* **Read, not driven**: that `cmd/civitai/main.go` performs no filtering. `TestDownloadErrorsReachStderrUnfiltered` asserts this against the **source** (both `fmt.Fprintln(os.Stderr, errorLine(err))` and `return "Error: " + err.Error()`), and fails loudly if `main.go` ever grows a `safeTerm`/`saferune` reference, since that would make these gates redundant rather than load-bearing. A live probe was deliberately not used: it would be evidence about whatever binary was on disk, not about this source.

## Ledger

`safeterm_coverage_test.go` keys on the **enclosing function** and only fires `GREW` for a function that already calls `safeTerm` at least once — so while these surfaces called it *zero* times, no row was ever demanded. Their absence was not merely unrecorded, it was **unrecordable**. That is the blind spot #566 named; gating them is what makes them ledgerable.

* Five new rows, all `covered`: `checkTargetCollisions`, `(*progressWriter).line`, `writePart`, `downloadStatusError`, `safeTermErr`.
* `downloadOne` flipped `notCovered` → `covered`.
* **`maxUncoveredSafeTermFuncs` 25 → 24 in this commit**, as the ratchet's headroom check requires. Verified in both directions: leaving it at 25 fails with `RATCHET HEADROOM`.
* `safeTermErr` also ledgered in `sanitizerComposers`; `"partPath"` added to `bareIdentArgs`. (`"dir"` was added too, and Round 1 removed it again — its note was false.)

The harder instrument #566 scopes out — demanding a row for *any* function that renders a server-supplied struct field, gated or not — is **not** built here, as the issue asks.

## Residuals, knowingly shipped

1. **`targetPath`'s `server returned an unusable filename %q`** is left raw. `%q` already escapes every `Cc` and `Cf` rune; what survives is only the blank-but-graphic pair (U+2800, the Hangul fillers), which cannot move a cursor, inside a quoted span in a one-line error on a path that fires only for a degenerate name. Gating it would add a sixth function to the ledger for no observable change.
2. The four surviving mutants above.
3. Item 25-style prose bounds elsewhere in the README are untouched.

## Overlap with #554 — measured, not assumed

#554 also touches `README.md`, `internal/cmd/safeterm.go` and `internal/cmd/safeterm_userinput_test.go`.

* `git merge-tree --write-tree HEAD pr/554` → **rc 1**, conflicts in `indentcontinuation_ledger_test.go`, `safeterm.go`, `safeterm_userinput_test.go`.
* `git merge-tree --write-tree origin/main pr/554` → **the same three paths**.

So every conflict is **pre-existing between #554 and `main`** (it is already `mergeable: CONFLICTING` / `DIRTY`); this branch introduces no new conflicting path. Whichever lands second resolves — the likely semantic collisions to read, not just merge, are the two new functions both added directly after `safeTerm` in `safeterm.go`, and both sides' additions to `bareIdentArgs` / `sanitizerComposers`.

## Gate

* `make ci` — green (full clone).
* `make lint` — **0 issues**, via `nix-shell -p golangci-lint`. It caught the fixtures first: `ST1018` on the raw Unicode format characters, which `make ci` structurally cannot see because lint is a separate CI job. The literals are now `\uXXXX` escapes — the exact trap AGENTS.md records four fixtures once hitting.
* `go test ./internal/cmd -count=1` re-run after the final edit (not from cache): ok.
* `make ci-shallow` — **21/21 packages ok** in a depth-1 clone of the committed tip, 0 failures, 0 timeouts. Both tiers, because a suite that runs in two environments has to be green in both.
* Round 1 re-ran all of the above: `make ci` green, `make lint` **0 issues**, and the full mutation matrix re-scored rather than carried over.

Base is `origin/main` at `c4ed077`. **`main` has since moved** to `f98eb73` — one docs-only commit touching `claudedocs/handoff-agent-setup-onboarding.md`, which no file on this branch touches, so `mergeStateStatus` is still `CLEAN`. (The original sentence here said `main` had not moved; that was false, and is corrected in a PR comment rather than silently.)

https://claude.ai/code/session_017wmskcHXdq2yHKKetLut1d


ZacxDev:
Follow-up: the PR body claims the ratchet was "verified in both directions", and at the time of writing only the *lowered* direction had actually been run. Measured now, so the claim carries its own evidence — `maxUncoveredSafeTermFuncs` set back to `25` against this branch:

```
--- FAIL: TestSafeTermCallSitesAreCoveredByANamedTest (0.31s)
    RATCHET HEADROOM: 24 function(s) are ledgered notCovered but
    maxUncoveredSafeTermFuncs is 25. Lower it to 24 in this commit.
    166 safeTerm call site(s) in 66 function(s): 42 COVERED, 24 NOT COVERED (cap 25)
```

Restored to `24`; the tree is unchanged. Also recorded because the counts are the useful number: this branch takes the package from 164 call sites in 64 functions (36 covered / 25 uncovered) to **166 in 66 (42 covered / 24 uncovered)**.

All 13 checks are green, `lint` included.

ZacxDev:
## Round-1 audit fix — and two corrections to this PR's own description

An adversarial audit of this branch found a real defect and two stale self-reports. The defect is fixed in the commit above; the two stale sentences are in the PR **description**, so they are corrected here in public rather than quietly rewritten.

### Correction 1 — "`main` has not moved since branching" is **false**

The description ends:

> Base is `origin/main` at `c4ed077`; `main` has not moved since branching.

`main` is now **`f98eb73`**. It moved by exactly one commit:

```
f98eb73 docs(handoff): rank 21 done (cli#569 merged, re-verified); rank 20 claimed (#571)
 claudedocs/handoff-agent-setup-onboarding.md | 274 +++++++++++++++---------
 1 file changed, 163 insertions(+), 111 deletions(-)
```

Docs-only, and its one file is disjoint from every file this branch touches — so the *merge* risk is nil and `mergeStateStatus` is still `CLEAN`. But the sentence the merged-tree claim rested on was wrong, and "it happened to be harmless" is not the same as "it was checked". Re-stated correctly: **base `c4ed077`; `main` has since advanced one docs-only commit to `f98eb73`, touching no file this branch touches.**

### Correction 2 — the "3077 passing tests" count was **right**; the audit's 3076 corroborates it

The audit counted **3076** and flagged the description's 3077 as possibly stale. It is not. The audit ran in an environment with no `.git` directory, where one test skips rather than runs — which accounts for exactly the one-test difference. A count that is one lower under a condition known to skip one test is **corroboration**, not contradiction. The description's 3077 stands and has not been changed.

### The defect — an unreachable guard at `download.go`'s `MkdirAll`

The `create output directory` gate this PR added could **never** strip a server-supplied byte, so it read as coverage while providing none.

`dir` is `filepath.Dir(target)`, and `target` comes only from `targetPath`, whose four returns are `o.out` (user-typed `--out`), `filepath.Join(dir, base)`, `filepath.Join(o.outDir, base)` and bare `base` — where `base = filepath.Base(f.Name)`. `filepath.Base` strips every separator, so **the server's name is always the leaf**, and `filepath.Dir` drops it; the bare-`base` return yields `"."`, which the condition skips. Under `--layout`, `routeDir` returns `root` (user-typed `--root`) or `filepath.Join(root, sub)` where every `sub` in `layoutFolders` is a compile-time constant — the server's `fileType`/`modelType` only select the map key, and reach the returned `note`, never the returned `dir`. `MkdirAll`'s `*fs.PathError` carries a prefix of `dir`, so `safeTermErr` had nothing to strip either.

So the only bytes that gate could ever remove are ones the **user typed**, and `internal/saferune`'s package doc is explicit that the strip is not applied to those.

**Its `KILLED` verdict (M8/M9) was spurious.** The killing subtest handed `downloadOne` a target of `<tmp>/<hostile name>/weights.bin` — a shape `targetPath` **cannot produce**, since the hostile name appears as a *directory* component. Breakable, but not reachable.

Corroborating: `generate_output.go:413` has the identical `fmt.Errorf("create output directory %s: %w", dir, err)` **ungated**, and has never been considered a defect.

### What changed

- **`download.go`** — both wrappers reverted at the `MkdirAll`; the measurement recorded at the site so a future sweep does not re-add the gate.
- **`download_safeterm_test.go`** — the `output directory` subtest deleted (it built the impossible target), with a comment saying why.
- **`safeterm_userinput_test.go`** — `bareIdentArgs["dir"]` removed. Its note claimed the leaf came from `routeDir`'s mapping of the server file/model type; that is false. **`"partPath"` stays** — `target + ".part"` really does inherit the server leaf.
- **`safeterm_coverage_test.go`** — `downloadOne`'s `why` drops the mkdir clause (8 calls → 6, 7 killed → 5) and records why the mkdir error is deliberately absent; `safeTermErr`'s `why` re-points at the subtests that actually pin it.
- **`safeterm.go`** — `safeTermErr`'s doc illustrated the `%s`+`%w` bug with the mkdir call, which no longer exists and whose "built from the server's file name" claim was false *for that site*. Re-pointed at `install`, where it is true.
- **`download_safeterm_test.go`** — `TestDownloadErrorsReachStderrUnfiltered`'s second arm downgraded `t.Errorf` → `t.Log`. It fired if `main.go` ever gained a sanitiser — a change its own message called *"not wrong"*. A gate that goes red on a correct change trains people to click through it. The load-bearing first half, pinning that `main.go` filters nothing today, is unchanged.
- **`README.md`** — the `SHA256 mismatch for` Troubleshooting row's trailing historical clause trimmed to the contract half. AGENTS.md scopes the README to the published contract; that clause was changelog and was the part most likely to rot. (Checked first that the row is not generated — `exitcodes_doc.go` generates the *Exit codes* section, not the Troubleshooting index.)

`maxUncoveredSafeTermFuncs` **stays 24**: `downloadOne` keeps other gated calls, so no function moved between covered and notCovered. The harness now reports `165 safeTerm call site(s) in 66 function(s): 42 COVERED, 24 NOT COVERED (cap 24)` — a delta of exactly 2 sites, which is the removal.

### Ledger-harness negative control

Removing a row is only meaningful if the ledger can notice. Re-adding `"dir"` to `bareIdentArgs` goes red with its own specific message:

```
--- FAIL: TestSafeTermIsNeverAppliedToUserTypedInput
    safeterm_userinput_test.go:307: bareIdentArgs classifies "dir", which is no
    longer passed to safeTerm anywhere. A stale note reads as coverage; delete
    it or fix the name.
```

ZacxDev:
## Round 0 verdict — requirements & deletion (prose; round 0 emits no claims block)

`round 0 · requirements: 16 (unattributed: 6) · deletion candidates: 5`
Audited tip: `036151b`. Verdict: **deletion candidate** — acted on in `4ffdf45`.

**Acted on**

- **D1 — the `create output directory` gate was unreachable for server bytes.** `filepath.Base` puts the uploader's name in the leaf, so `filepath.Dir(target)` returns only user-typed `--out`/`--out-dir`/`--root` or a `layoutFolders` compile-time constant. `internal/saferune`'s rule is that user-typed input is *not* stripped, so the gate did the one thing the subsystem forbids. Its KILLED verdict was spurious: the killing subtest handed `downloadOne` a target with the hostile name as a *directory component*, a shape `targetPath` cannot produce — breakable but not reachable. Independently re-verified before removal, including that `os.MkdirAll`'s `*fs.PathError` path is also entirely user-typed. `generate_output.go:413` carries the byte-identical ungated line, so the removal restores consistency rather than creating an exception.
- **D3** — a `t.Errorf` that fired on a *correct* change (its own message said the trigger "is not wrong") downgraded to `t.Log`; the half that pins the real premise is kept.
- **D5** — the README Troubleshooting row's historical clause trimmed to the published contract, per AGENTS.md's scoping of that file.

**Considered and deliberately kept**

- `safeTermErr` — load-bearing at `install` (`*os.LinkError` renders both paths raw) and `writePart/create` (`*fs.PathError`; `partPath` does inherit the server leaf). `%q` does not cover the class: `strconv.Quote` leaves U+2800 and the Hangul fillers intact.
- The four named survived mutants — naming them is required; removing the names would leave rows reading as coverage while providing none.
- The ratchet at 24 — not a judgement call; the assertion is an equality that errors in both directions.

**Open, not fixed here** — two narrative counts are stale by ~12 (`safeterm_coverage_test.go:19` "152 sites in 60 functions", `safeterm_userinput_test.go:116` "153 calls today"; both scanners now report 165). Pre-existing drift, in files a concurrent effort on #552 may touch, so left rather than churned.

Round 0 does not move the ladder. Round 1 (the nine correctness axes) runs next, blind, against `4ffdf45`.

ZacxDev:
## Round 1 — findings addressed

Round 1 ran the nine correctness axes against `4ffdf45` and returned **no 🔴**; the exit-code contract was measured unchanged through `safeTermErr` (`IsTransportError` verdicts and all four `TagStatus` sentinels preserved). Two 🟡 and two 🟢 were fixed below, plus a merge of `origin/main`.

```audit-claims round=1 audited=4ffdf45..b4e091a0
1. F2 — targetPath's "unusable filename %q" refusal now routes f.Name through safeTerm; %q retained. The prior justification (that %q already escaped the class) was measured false: %q escapes by unicode.IsPrint, so Mn/So AND the Lo Hangul fillers passed raw.
2. F1 — the \n and \t residual is now STATED, not fixed: three comment blocks in download.go and the README row name saferune's deliberate retention of both and point at #552. No behaviour change.
3. F3 — nesting corrected to dashIfEmpty(safeTerm(x)) at formatFileList and checkTargetCollisions, each pinned by its own assertion.
4. F4 — the "do not re-add a gate here" absolute at the removed mkdir site reworded into a conditional naming the three premises a future author must re-derive.
5. Prose counts re-measured rather than adjusted: "21 gated call sites / 20 mutants" was wrong when written, not stale (19 sites now, 20 at 036151b, 18 at 4ffdf45); "all four of this file's tests" is 7 of 8 under the negative control; the drift-prone "44 lines above" replaced with a form that names the function.
6. The ledger note justifying the `download %s: %w` survivor was narrowed — its claim that %q escapes every rune in the class is false in the same way F2's was. Behaviour deliberately unchanged; the residual is live and needs a decision.
7. origin/main (#573) merged in. The sole conflict was the ratchet block; rows from both sides unioned with none of #573's weakened. maxUncoveredSafeTermFuncs measured to 21 by reading the equality assertion's own failure output, not computed.
```

Mutation state at `b4e091a`: 21 mutants, 17 killed, 4 survived — the survivor set unchanged from the four this PR has named throughout. Merged-tree spot-checks confirmed one gate from each side goes red with its own assertion text and no build break. 13/13 checks green.

ZacxDev:
## Round 2 — findings addressed

Round 2 was a delta re-audit of `4ffdf45..b4e091a0` (the merge commit's #573 bring-in excluded from scope). It returned no 🔴, two 🟡 and four 🟢. All six are addressed below.

```audit-claims round=2 audited=b4e091a0..a539d69a
1. F1 — the survivor became a KILL, not a reworded comment. The false justification ("the cause is a *url.Error and Go's own %q already escapes the class") is RETRACTED in both places it lived: the authoritative ledger row in safeterm_coverage_test.go and the header of download_safeterm_test.go. Round 1 had edited only the header, leaving the ledger row standing.
2. F1 evidence — two new subtests drive it: a *url.Error whose RawQuery carries U+2800/U+034F (url.URL.String writes RawQuery verbatim), and the real civitai.New Downloader's https refusal, whose cause is NOT a *url.Error. New fixture dlHostileQuery, added to the TestDownloadFixtureIsHostile positive control. The transport subtest asserts its own premise, so it fails as a CONTROL if net/url ever encodes RawQuery.
3. F2 — the four statements that pointed at #552 (closed by #573 on 2026-09-12, scope ~13 tabwriter renderers, download.go never on its table) now point at #577. They are download.go x3 and download_safeterm_test.go's header; the README row carried no issue pointer at all, which corrected a miscount in #577's own first draft.
4. F3 — "seven NEW rows" re-derived to SIX against merge base 6f0a8d8 (nine against fork point c4ed077, three of which arrived via main with #552/#573). Neither reading is seven.
5. F4 — the negative-control count re-measured at this head: 37 top-level tests red, 7 of this file's 8, against 0 failures unmutated. Was 36 before the merge; the extra is #573's TestGatedRenderersDoNotForgeOutsideTheirTable.
6. F5 — the comment implying repo-wide consistency corrected. The inverted dashIfEmpty/safeTerm order is measured at exactly 15 occurrences against 15 correct ones, an even split rather than a convention; none of the 15 touched, several being inside cells #573 just hardened.
7. F6 — the README row now names the fifth surface (targetPath's "server returned an unusable filename", reaching stderr via downloadSelected and stdout via printDownloadPlan), and the "not newlines" clause is scoped so it does not read as a global claim post-#573.
```

**Mutation state at `a539d69`: 21 mutants, 18 KILLED, 3 SURVIVED, 0 build-broken.** The survivor set **shrank from 4 to 3** — all three are in `writePart` (`finalize %s`, `finalize`'s cause, `streaming`'s cause) and need a failing `Close` or a stream error carrying hostile bytes; neither is driven and neither is claimed as covered.

**Gates at `a539d69`:** `make ci` green · `make ci-shallow` 21/21 at depth-1 on the pushed SHA · `make lint` 0 issues (golangci-lint v2.13.2 locally; CI pins v2.12.2 and its own `lint` job passed on this SHA) · 13/13 checks SUCCESS.

**Filed rather than fixed here:** #577 (the `\n` line-forgery residual, now the owner of what #552 no longer covers) and #579 (`saferune.Strip`'s doc claims a byte-for-byte subsequence, but its fast path and rewrite path disagree on invalid UTF-8 — inert, but it is a false comment on the gate this whole series rests on).

ZacxDev:
## Round 4 — findings addressed, and the ladder STOPS here

```audit-claims round=4 audited=a539d69a..35c72134
1. F-1 — round 2's retraction contained its own false clause ("the https/parse refusals on this path are not *url.Error at all"). The PARSE refusal IS one: requireHTTPSTransfer wraps url.Parse's *url.Error{Op:"parse"} with %w. Corrected in both places, with the measured exit-code split stated: http:// scheme refusal -> plain fmt.Errorf, errors.As false, exitCode 1; parse refusal -> errors.As true, transport_error.go classifies it exit 5. Measured by driving the real client and feeding the result to cmd/civitai's own exitCode.
2. F-2 — "url.URL.String() percent-encodes the path and punycodes the host" corrected at both sites: net/url percent-encodes the host (punycode is net/http's), and it also encodes the FRAGMENT, which the original claim omitted entirely.
3. F-3 — three comment lines carrying literal \uXXXX escapes instead of characters repaired. Swept all 1182 added diff lines: 8 hits, 5 correctly inside string literals (ST1018), 3 in comments. Zero remain outside string literals.
4. F-4 — the README surface enumeration made explicitly NON-exhaustive and extended with the five omitted gated surfaces, rather than closed as a complete list.
5. The 19-vs-18 call-site scopes disambiguated in one clause: 18 across download.go's six functions, 19 including safeTermErr's own call in safeterm.go.
6. dlHostileQuery/dlSafeQuery added to TestDownloadFixtureIsHostile's pairwise-distinctness loop — the round's ONLY executable change, negative-controlled by temporarily duplicating a pair and watching the test fail with its own message.
```

**Sweep, not just the reported sites.** Round 1's failure mode was fixing the surface that was reported and missing the authoritative one, so this round swept every claim of the same *shape* — assertions about which concrete error type arrives on a path, what the standard library escapes, or what a mutant proves. **5 corrected, 13 re-derived and confirmed unchanged**, including all seven runes named in `targetPath`'s comment verified individually for category, `IsPrint` and strip behaviour; the `*fs.PathError` / `*os.LinkError` / `*url.Error` rendering differences measured directly; and the "15 inverted / 15 correct" nesting count re-measured with a positive-controlled pipeline and found right to the line number.

**`internal/cmd/download.go` and `internal/cmd/safeterm.go` are byte-identical to `a539d69`.** No production code changed this round; the inherited survivor set stands at 3 (`writePart`'s `finalize %s`, `finalize`'s cause, `streaming`'s cause — each needs a failing `Close` or a stream error carrying hostile bytes; neither is driven and neither is claimed as covered).

---

### Why the ladder stops here rather than on a clean round

Stating this explicitly, because a report that ends on the audit skill's escape hatch is otherwise indistinguishable from one that converged.

Round 4 is **not** a clean round — F-1 was a real finding that changes what a reader concludes about exit-code classification. Under the findings-keyed rule alone, a round 5 would follow.

It stops instead on the skill's **stated criterion for a prose-payload ladder**, and the criterion is met:

- **The payload is now prose, so the attribution gate structurally cannot fire.** Round 3's range shipped zero executable change (`download.go`'s 39 changed lines were 100% comments); round 4's shipped one test-fixture line. When the payload is comments and README, "fixed a defect" and "reworded a warning" are the same edit, so each round's fixes generate the next round's findings indefinitely.
- **No 🔴 at any round**, and the blast radius is bounded at "a comment contains a false clause."
- **The recurring shape was swept at every site**, not only where it was reported — that sweep is this round's actual deliverable, and it is what a round 5 would otherwise re-find one site at a time.

The defect shape itself is the reason to stop generating prose rather than continue: rounds 1, 2, 3 and 4 each retracted a justification written by the previous round while fixing the previous round's claim. The cure is to stop supplying replacements, which round 4 does — it states measured facts and names which clauses survive, rather than offering a new unifying sentence.

### Left open deliberately, filed rather than fixed

- **#577** — a `\n` in a server-supplied name still forges lines on four download surfaces. `saferune` retains `\n`/`\t` by design; #552 was closed by #573 over ~13 tabwriter renderers and never covered `download.go`. Not a regression — those surfaces were fully raw before this PR.
- **#574** — six server-derived strings on `generate`'s blob path, including `blobStatusError`, the structural twin of the `downloadStatusError` this PR gates.
- **#579** — `saferune.Strip`'s doc claims a byte-for-byte subsequence; its fast path and rewrite path disagree on invalid UTF-8. Inert, but it is a false comment on the gate this whole series rests on.
- The **15 inverted `dashIfEmpty`/`safeTerm` sites** outside this PR's scope, several inside cells #573 just hardened.

### One mechanism worth recording for the next prose round on this file

The literal-`\uXXXX`-in-comments defect (F-3) is **live, not historical**. The editing tooling reproduced it mid-round on a fresh edit typed with real characters. `gofmt`, `go vet` and `golangci-lint` are all blind to it — confirmed green with the escapes present. Diff-review for `\uXXXX` before committing prose to this file.

PR #573, opened 2026-09-12
fix(safeterm): close the tabwriter forgery at every renderer, and ledger the set (#552)

ZacxDev:
Closes the `\n` / `\t` half of civitai/cli#552 at every tabwriter renderer, and adds the ledger the issue asks for. A second commit closes the same hazard **one line outside** the table, after two independent audit rounds on the first.

`safeTerm` deliberately keeps `\n` and `\t` (`internal/saferune`: "Cc, minus \n and \t"). `text/tabwriter` breaks a row at the first and uses the second as its **column delimiter**, so server text in a cell does not corrupt the table — it *extends* it. #569 closed that for `images.go` only.

Reproduced on `main` by the new behavioural test: a `creators search` username of `"FORGE-crname\tC0L\nR0W"` rendered **4 lines where the contract says 2**, with a forged row at column zero and a data row carrying 4 columns against a 3-column header.

---

## 1. What changed, per call site

Every one is a value that lands in a `tabwriter` **cell**. Nothing that legitimately spans lines was touched.

| Renderer | Cells converted | Why |
|---|---|---|
| `printCreatorList` (creators.go) | username, link | both are cells; the measured #552 forgery |
| `printTagList` (tags.go) | name, link | both are cells |
| `printModelList` (models.go) | creator, name, type | `creator` and `name` are locals built by `safeTerm` and then written into the row |
| `printModelDetail` (models.go) | version name, base model | only the version TABLE; the header lines above it are plain `Fprintf` and are left alone (residual 6) |
| `nonModelFileMarker` (read.go) | `files[].type` marker | used in a `printModelDetail` cell **and** in `printModelVersionDetail`'s header line — single-line in both |
| `printModelVersionDetail` (model_versions.go) | file name, file type | the file table only |
| `printCollectionList` (collections.go) | owner, name, type | all cells |
| `printArticleList` (articles.go) | author, title, published date | `shortDate` returns the RAW server string when it does not parse, so the date column is server text too |
| `printAppList` + `appCardAuthor` (apps.go) | name, slug, kind, category, author | `appCardAuthor` is only ever a cell, so the gate moved inside it |
| `newUsersGetCmd` (users.go) | candidate username | the `other matches` table |
| `printAppMetrics` (app_metrics.go) | window from/to, granularity, top scope, top endpoint | **Window and Granularity had no gate at all**: `utcStamp` falls back to the raw string on a parse failure, by design. `App:` is the user's own slug and is left echoed exactly (#393) |
| `reportWorkflowSettlement` (workflow_settlement.go) | transaction type | the `quotedList` is NOT converted — it goes to stderr outside the tabwriter |
| `printWorkflow` (workflows.go) | id, status, timestamps | the failure-reason block below keeps `safeTerm` + `indentContinuation` |
| `printWorkflowList` (workflows_list.go) | id, status, created | its `rowLines` comment explicitly relied on a cell being able to carry `\n`; the comment is updated and the count is kept as a construction guard |
| `printSubmitResult` (generate.go) | workflow id, status, **and the `--no-wait` re-attach hint** | the hint was a bare `safeTerm` on the SAME `r.ID` (round 2) |
| `printCostMap` (generate.go) | cost-map key | the tabwriter arrives as a **parameter** from `printGenerateQuote` |
| `describeVersion` (generate.go) | model name, model type | stored on `resolvedGraph` and printed into the cost table and the approval screen |
| **`printSubmissionTable`** (app_status.go) | blockId, version, status, deploy, source, date, URL | **had no `safeTerm` at all**, which is why `safeTermCoveredBy` could not see it |
| **`printSubmissionDetail`** (app_status.go) | the same + publish-request id, deploy detail, **plus four NON-cell surfaces** | see §2 |
| **`printListingStatus`** (app_listing.go) | listing status, **plus the screenshot id and caption** | see §2 |

**Deliberately NOT flattened**: the orchestrator failure reason (`printWorkflow`, `printWorkflowList`), image prompts (`printImageMetaBlock`), per-output exclusion reasons, `printSubmissionDetail`'s rejection reason / approval notes, and `printGenerateQuote`'s user-typed cells (`--input` path, aspect ratio) — sanitising the last of those is the #393 defect. The four free-text surfaces get `safeTerm` + `indentContinuation` instead.

## 2. Round 2 — what the audits found, and what it cost

Two independent adversarial rounds ran against `ea39f7f`. Six findings fixed, one recorded.

**🟡 The ledger had a one-spelling blind spot.** `analyzeTabwriterUse` seeded its tabwriter sinks from `*ast.AssignStmt` only, so a renderer spelled `var tw = tabwriter.NewWriter(...)` was invisible to the **whole** guard — no `BARE-SAFETERM`, no cell count, and crucially **no `GREW`**, so nothing asked anyone to classify it. Measured: an ungated renderer added to `tags.go` with that spelling left `internal/cmd` green; the same renderer with `tw :=` failed `GREW`. `localAssignments` already walked `*ast.ValueSpec` — the value side handled `var`, the sink side did not. Fixed; `TestTabwriterScannerSeesShapesNotInTree` now pins `var x = …` and the grouped `var ( … )` form, and mutant **M1b** below reproduces the pre-fix SURVIVED against the real tree.

**🟢 `gateUnsanitised` and `gateNoServerText` are DELETED**, along with `maxUnsanitisedTabwriterRenderers`. Both rounds flagged it independently. **Zero rows used either**, so both `case` arms were unreachable and the ratchet asserted `0 == 0` — and the two states were **machine-indistinguishable** (the scan can only observe "does `safeTermSingle` reach a cell"), so the ratchet was walkable by one word: an author adding an ungated renderer writes `gateNoServerText`, the counted set stays empty, suite green. The set of gates is now **closed at two**; an ungated renderer has no truthful row available (`gateSingle` fails MISLABELLED, `gatePreSanitised` must name an upstream that really calls the gate, anything else hits the `default` arm, no row at all fails `GREW`). ~30 lines cheaper and strictly stronger. No escape hatch is kept.

**🟡 A row is a claim about CELLS, and it was being read as coverage of a COMMAND.** This is the residual that bit:

- `printSubmissionDetail` printed `*s.RejectionReason`, `*s.ApprovalNotes`, `*s.LiveURL` and `s.BlockID` with **no gate at all**, one line under its own flushed table. A reason of `"rejected because \x1b[1A\x1b[2KOVERWRITTEN"` put a **raw ESC on stdout** and overwrote the row the same function had just written — a #399-class hole.
- `printListingStatus` printed the screenshot id and caption raw, so a caption of `"GOOD\tTABBED\nApp:  forged-line"` emitted `App:  forged-line` at **column zero**, five lines under the real `App:` row that same function writes through its tabwriter.

Both gated by **shape**: `safeTerm` + `indentContinuation` for the legitimately multi-line reviewer free text, `safeTermSingle` for the single-line metadata. `TestGatedRenderersDoNotForgeOutsideTheirTable` (new) drives all six surfaces.

**🟢 One value, two gates.** `printSubmitResult`'s `--no-wait` hint used a bare `safeTerm` on the same `r.ID` its receipt cell had already put through `safeTermSingle` — the weaker gate on the string the user is told to **paste into a command**.

**🟡 The README section this PR authored overclaimed**, verified by driving the real renderers. Resolution (b) for the `app_status.go` gap — the code was fixed so the strong claim holds there — and the wider sentences were **narrowed**, deliberately, rather than the ~60 remaining single-line `label: value` sites being converted:

- *"stripped from every server-supplied string"* → scoped to "the server-supplied strings the human renderers print".
- *"In any tabular or `label: value` output…"* → split in two: a **cell** promise (kept, now accurate) and a narrower `label: value` promise that **names the fields it covers** and states that other detail lines (e.g. `models get`'s header block, `collections get`, `app view`) strip escapes but may still carry a newline.
- *"the two that matter"* → there are **four** multi-line surfaces, now enumerated, and the rejection-reason block is one of them and is now indented.

**🟡 A stale closing-condition claim.** `indentcontinuation_ledger_test.go` said *"TestIndentContinuationCallSitesAreLedgered is the closing condition for #552"* while this PR's `tabwriter_ledger_test.go` says the same of `tabwriterRenderers` — two tests in one package claiming to close one issue, and #552's own body disowns the first. Corrected in both the file header and the test's own doc comment, with what that ledger actually pins (its eight call sites' argument, pad and set membership — not whether the set is the *right* one).

**🟢 Recorded, not fixed.** `sanitizerReach` stops at `depth > 4`, so a bare `safeTerm` laundered through five successive local assignments reaches a cell without being counted as a violation. The renderer is still **seen**, so `GREW`, `SHRANK` and `MISLABELLED` all still apply. Left open deliberately and added to the enumerated residuals (now residual 5) — the residual list is what makes the guard honest.

**Also done, judgement call.** `TestTabForgeryIsNeutralisedInTheRealImagesTable` was **deleted** as strictly subsumed by the ledger's `images search` case (which uses a payload carrying a tab *and* a newline and asserts strictly more, for nineteen other renderers besides). Its historically measured #569 payload — `"alice\tSDXL\t9x9\tNone\t0\t0\thttps://evil.example/steal"` — is **not** lost: it is folded into that table as a second fixture via a new `alsoWant` field, because it is the shape an attacker sent rather than one a test author invented. `TestSafeTermSingleNeutralisesTheTabColumnVector` is **kept** — it pins the helper directly, which is the claim that survives a renderer being rewritten. `pinnedIndentCallSites`' set-pinning was **not** touched (one round only, separate concern).

## 3. The ledger (`internal/cmd/tabwriter_ledger_test.go`)

`tabwriterRenderers` pins the **20 functions that write into a `text/tabwriter`**, each with a gate state and a reason naming the fields. The predicate is structural and starts from the hazard, not from a helper's name:

> a value written into a tabwriter cell may not reach that cell through a bare `safeTerm`.

The scan (`analyzeTabwriterUse`) finds tabwriter-bound identifiers — from **both** `:=`/`=` and `var` declarations — **propagates the sink across calls** (so `printCostMap(tw, …)` is analysed), and resolves a cell expression through **local variables and package-level helpers**, which is how it sees `appCardAuthor()` and `nonModelFileMarker()`, neither of which names a sanitizer at the call site.

**Four** independent assertions, each with its own message: **BARE SAFETERM**, **GREW**, **SHRANK**, **MISLABELLED**. (There is no RATCHET any more — see §2.)

Gate states are **checked, not spelled**: a `gateSingle` row must actually have `safeTermSingle` reaching a cell; `gatePreSanitised` (only `printGenerateQuote`) must **name an upstream function that exists and still calls `safeTermSingle`**, so deleting `describeVersion`'s gate reddens the ledger too. Any other spelling fails the `default` arm.

`maxUncoveredSafeTermFuncs` drops **25 → 22**.

**Why the floor cannot swallow the finding.** `minTabwriterRenderersFound` is **8 against a set of 20** — strictly below, deliberately, because `indentcontinuation_ledger_test.go` shipped 6-against-6 checked *before* the set comparison, so deleting a real call site printed `CONTROL failure` and the SHRANK message never ran.

## 4. Red-at-base matrix

Base for round 1: `c4ed077`. Base for round 2: `ea39f7f` (this PR's own first commit); the branch is rebased onto `f98eb73`.

| Test | at base | at HEAD |
|---|---|---|
| `TestTabwriterRenderersCannotBeForged` | **RED** at `c4ed077` — 17 of 18 subtests fail | green (19 subtests) |
| ↳ `images search` subtest | green | green — **invariant guard**: regression protection for #569 |
| ↳ `images search`, the #569 payload (new) | n/a | green; mutant M11 kills it |
| `TestTabwriterRenderersAreLedgered` | **RED** at `c4ed077` — 39 bare-safeTerm violations in 20 renderers + **18** MISLABELLED rows | green |
| `TestTabwriterScannerSeesShapesNotInTree` — the two new `var`-sink cases | **RED** at `ea39f7f`: `scanner reported 0 violation(s), want 1` and `scanner attributed the tabwriter to [], want [zzRender]` | green |
| `TestGatedRenderersDoNotForgeOutsideTheirTable` (new, 6 subtests) | **RED** at `ea39f7f` — all 6 fail | green |
| `TestIndentContinuationCallSitesAreLedgered` | GREW at `ea39f7f` once `app_status.go` was fixed (6 → 8 call sites) | green |
| `TestForgeCellCarriesBothVectors` | green | green — **invariant guard** (control on the fixture) |

🔴 **The PR body previously said "16 MISLABELLED rows at base". That was wrong; the audit refuted it and the re-measurement agrees with the audit.** Method: a `cp -a` copy with `.git` removed, `internal/cmd`'s **non-test** sources replaced from `origin/main` (67 files) while the test files stayed at HEAD, then `go test -run TestTabwriterRenderersAreLedgered`. Output: `39 bare safeTerm call(s) reach a tabwriter cell`, `scanned 67 file(s), 602 func(s): 20 tabwriter renderer(s), 85 cell write(s), 39 bare-safeTerm violation(s)`, and **18** distinct `MISLABELLED: tabwriterRenderers[...]` rows.

## 5. Mutation table

Round 1's battery (21/21 killed, gate-OFF and safeTermSingle→safeTerm DOWNGRADE across every renderer) stands unchanged. Round 2's battery, all in a `cp -a` copy with `.git` removed **first**, each mutant asserting its target string **exists** before editing (a no-op replace would report SURVIVED), each restored byte-for-byte from a pristine snapshot, baseline green in the copy before scoring:

| # | Mutant | Verdict | Killed by ITS OWN message |
|---|---|---|---|
| M1 | a renderer whose sink is `var tw =` writing a raw server string into a cell | **KILLED** | `no row in tabwriterRenderers` (GREW) |
| M1b | the **same** renderer with the `ValueSpec` arm reverted out of the seed | **SURVIVED** — by design: this is the pre-fix defect reproduced against the real tree | — |
| M2 | an ungated renderer classified with the **deleted** `no-server-text` gate | **KILLED** | `has an unknown gate` |
| M3 | rejection reason printed raw again | **KILLED** | `a RAW ESC reached the output` |
| M4 | approval notes printed raw again | **KILLED** | `a RAW ESC reached the output` |
| M5 | rejection reason `safeTerm`'d but **not** indented | **KILLED** | `begins with the forged payload` |
| M6 | live URL back to ungated | **KILLED** | `did not render as one inert string` |
| M7 | blockId in the not-live sentence back to ungated | **KILLED** | `is not one inert string` |
| M8 | screenshot **caption** back to ungated | **KILLED** | `begin with the \`App:\` label` |
| M9 | screenshot **id** back to ungated | **KILLED** | `the screenshot ID did not render "FORGE-lsid C0L R0W"` |
| M10 | the `--no-wait` hint back to a bare `safeTerm` | **KILLED** | `the --no-wait hint did not render` |
| M11 | `printImageList`'s username gate weakened — the folded-in #569 fixture | **KILLED** | `did not render "alice SDXL 9x9 None 0 0 https://evil.example/steal" contiguously` |
| M12 | a table case naming neither `fields` nor `alsoWant` (the **control on the case**) | **KILLED** | `names neither \`fields\` nor \`alsoWant\`` |

M5 is the isolation control on M3: `safeTerm` alone (the escape class removed, the newline kept) still fails, and by the *column-zero* message rather than the ESC one — so the two halves of that fix are pinned separately. M12 is the control on the new `alsoWant` mechanism: a case that asserts only line and column counts is refused, because a renderer that ate the payload satisfies those too. Every mutant compiled (checked explicitly; a build failure would kill every mutant for the wrong reason).

## 6. Gate

Runner: `nix-shell -p golangci-lint` → **golangci-lint 2.13.2** (CI pins v2.12.2 in `.github/workflows/ci.yml`; the host has none on PATH and the repo ships no `.envrc`). Go 1.25.14.

- `make ci` — **green**. **22 packages** reported (21 `ok` + 1 `no test files`) — the positive control on the count; a broken invocation exits 127 with none.
- `make lint` — **green, 0 issues** over `./...`. **Negative control, measured:** a realistic bad case (a fixture-shaped constant carrying a raw ESC byte, the exact class AGENTS.md records as having reached a push) injected into `internal/cmd/tags.go` in the scratch copy made it exit **1** with the finding printed. The zero is a zero from an instrument that can go red.
- `gofmt -s -l .` — 0 files listed, over **478** `.go` files found (positive control on the count; a misquoted path also prints nothing).
- Enumeration used `find … -print0 | xargs -0 grep` where a zero mattered — `grep -r` on this host is gitignore-blind.

## 7. Residuals, knowingly left

1. **Soft wrap (out of scope, #552 tracks it).** Collapsing `\n` makes lines *longer*, so a long value still reaches column zero via the terminal's own wrap with no control character, and one hostile value widens a column for every row. Only `workflows list` wraps its reason text to a fixed budget; `app status --id`'s reason block does not, because it has no column budget of its own and inventing one would be a second unmeasured decision. Stated in the code and in the README's known limits. **#552 should stay open for this.**
2. **U+2028 / U+2029** pass through — pre-existing, documented in `internal/saferune`, no known terminal breaks on them.
3. **The same hazard one line outside a tabwriter, on the READ path.** `printModelDetail`'s header, `printCollectionDetail`, `printAppDetail`, `printUser` print `label: value` with plain `Fprintf` + `safeTerm`, so a `\n` there still starts a line at column zero. ~60 such single-line sites remain. **Deliberately not widened to** in this round: those functions mix single-line labels with legitimately multi-line free text, so it is a per-field judgement, not a structural rule. The README sentence was narrowed to say so rather than promise it. Ledger residual 6.
4. **Laundering past `depth > 4`** — see §2. Ledger residual 5.
5. **What the scan structurally cannot see**, stated in the file header: a tabwriter reached through a struct field or a function-typed variable; a write via `tw.Write([]byte(…))` (none exists today — `io.WriteString` IS covered, with a calibration case); and whether the sanitised value is the *right* one, which is why the behavioural test drives the real renderers.
6. **`newUsersGetCmd` stays `notCovered`** in `safeTermCoveredBy` — its tabwriter sits behind an HTTP round-trip and the behavioural test does not drive it. Its gate is pinned against wholesale removal by both ledgers, not by an assertion about the screen.
7. **`pinnedIndentCallSites` still pins a SET rather than a property.** One audit round raised retiring that; deliberately not taken on here, as a separate concern.

## 8. Where the routing lives

`AGENTS.md` is **deliberately not edited** — it is at 30,453 bytes against a 30,500 cap (`agents_size_test.go`), 47 bytes of headroom, and that file's own rules say the next item cannot be paid for from there. Routing lives in `tabwriter_ledger_test.go`'s header comment, `safeTermSingle`'s doc comment, the block comments at the two fixed call sites, and the README's "What a table cell can contain".

### DIFF
--- a/README.md
+++ b/README.md
@@ -3757,25 +3757,39 @@ or that a server chose. The human renderers put all of it through one gate befor
 it reaches your terminal, and this is what that gate promises:
 
 - **Terminal escapes are removed.** Cursor moves, line clears, OSC sequences and
-  the invisible / direction-reversing characters are stripped from every
-  server-supplied string, so a hostile value cannot overwrite a line the CLI
-  already printed or reorder what you read. (What the CLI prints from what **you**
-  typed — a prompt, a path, a flag value — is echoed byte-for-byte and is
-  deliberately *not* rewritten.)
-- **A table cell is one line, and one column.** In any tabular or `label: value`
-  output, a newline or a tab inside a server value is rendered as a **space**. A
-  newline would otherwise start a line at column zero, where it is
-  indistinguishable from a row the CLI wrote; a tab is the column separator, so it
-  would add an extra, perfectly aligned column. Both read as real output, which is
-  why the value is flattened rather than trusted.
-- **Genuinely multi-line server text keeps its line breaks**, and is indented
-  under the line that introduced it — a generation prompt (`images … --meta`) and
-  the orchestrator's failure reason (`workflows get`) are the two that matter. So
-  if a field you expect to be multi-line arrives on one line, it was in a cell.
+  the invisible / direction-reversing characters are stripped from the
+  server-supplied strings the human renderers print, so a hostile value cannot
+  overwrite a line the CLI already printed or reorder what you read. (What the CLI
+  prints from what **you** typed — a prompt, a path, a flag value — is echoed
+  byte-for-byte and is deliberately *not* rewritten.)
+- **A table cell is one line, and one column.** Every value that reaches a cell of
+  a rendered table — `models search`, `images search`, `app status`,
+  `workflows list`, the pre-spend cost table, and the rest — has any newline or
+  tab in it replaced by a **space**. A newline would otherwise start a line at
+  column zero, where it is indistinguishable from a row the CLI wrote; a tab is
+  the column separator, so it would add an extra, perfectly aligned column. Both
+  read as real output, which is why the value is flattened rather than trusted.
+- **`label: value` lines are a narrower promise.** The single-line metadata fields
+  — `images … --meta`'s model / sampler / seed / resources, `app status --id`'s
+  live URL and block id, `app listing status`'s screenshot ids and captions, the
+  `--no-wait` re-attach hint — are flattened the same way. Other one-off detail
+  lines outside a table (for example the header block above `models get`'s version
+  table, `collections get`, `app view`) have their escapes stripped but **may still
+  carry a newline**, so a hostile value there can start a line at column zero.
+  Telling a genuinely single-line field from legitimately multi-line free text is
+  a per-field judgement, and only the fields listed above have had it made.
+- **Genuinely multi-line server text keeps its line breaks**, and is indented under
+  the line that introduced it, so a continuation can never sit at column zero.
+  There are four such surfaces: the generation prompt and negative prompt
+  (`images … --meta`), the orchestrator's failure reason (`workflows get` and
+  `workflows list`), per-output exclusion reasons (`generate`), and the reviewer's
+  rejection reason / approval notes (`app status --id`). If a field you expect to
+  be multi-line arrives on one line, it was in a cell.
 - **Known limits.** A long value is not shortened, so your terminal can still
   soft-wrap it to column zero, and one hostile value widens a column for every
-  row. `U+2028` / `U+2029` are passed through (no terminal is known to break
-  lines on them).
+  row. Only `workflows list` wraps its reason text to a fixed budget; the other
+  multi-line surfaces do not. `U+2028` / `U+2029` are passed through (no terminal
+  is known to break lines on them).
 
 🔴 **None of this applies to `--json`.** That output is emitted raw, because JSON
 already escapes control characters and rewriting the bytes would corrupt what a
--- a/README.md
+++ b/README.md
@@ -118,6 +118,7 @@ contract, and **packages/submits** it for review.
 
 - [Upgrading](#upgrading)
 - [Global flags](#global-flags) — colour, `--version`, the update nag
+  - [What a table cell can contain](#what-a-table-cell-can-contain)
 - [Configuration](#configuration)
 - [Exit codes](#exit-codes)
 - [Troubleshooting](#troubleshooting) — **look the error message up here**
@@ -3747,6 +3748,40 @@ what counts, not the value.
 without passing through the presentation layer at all, so `--json` is always
 safe to pipe into `jq` regardless of how colour is configured or whether a TTY
 is attached.
+
+### What a table cell can contain
+
+Almost every value the CLI prints in a human table — a model name, a username, a
+tag, a workflow status, a submission's block id — is **text a stranger uploaded**
+or that a server chose. The human renderers put all of it through one gate before
+it reaches your terminal, and this is what that gate promises:
+
+- **Terminal escapes are removed.** Cursor moves, line clears, OSC sequences and
+  the invisible / direction-reversing characters are stripped from every
+  server-supplied string, so a hostile value cannot overwrite a line the CLI
+  already printed or reorder what you read. (What the CLI prints from what **you**
+  typed — a prompt, a path, a flag value — is echoed byte-for-byte and is
+  deliberately *not* rewritten.)
+- **A table cell is one line, and one column.** In any tabular or `label: value`
+  output, a newline or a tab inside a server value is rendered as a **space**. A
+  newline would otherwise start a line at column zero, where it is
+  indistinguishable from a row the CLI wrote; a tab is the column separator, so it
+  would add an extra, perfectly aligned column. Both read as real output, which is
+  why the value is flattened rather than trusted.
+- **Genuinely multi-line server text keeps its line breaks**, and is indented
+  under the line that introduced it — a generation prompt (`images … --meta`) and
+  the orchestrator's failure reason (`workflows get`) are the two that matter. So
+  if a field you expect to be multi-line arrives on one line, it was in a cell.
+- **Known limits.** A long value is not shortened, so your terminal can still
+  soft-wrap it to column zero, and one hostile value widens a column for every
+  row. `U+2028` / `U+2029` are passed through (no terminal is known to break
+  lines on them).
+
+🔴 **None of this applies to `--json`.** That output is emitted raw, because JSON
+already escapes control characters and rewriting the bytes would corrupt what a
+script parses. **A script that renders server strings from `--json` onto a
+terminal has to do its own sanitising** — the guarantees above are about the
+human output only.
 
 ## Configuration
 
--- a/README.md
+++ b/README.md
@@ -3762,8 +3762,8 @@ it reaches your terminal, and this is what that gate promises:
   overwrite a line the CLI already printed or reorder what you read. (What the CLI
   prints from what **you** typed — a prompt, a path, a flag value — is echoed
   byte-for-byte and is deliberately *not* rewritten.)
-- **A table cell is one line, and one column.** Every value that reaches a cell of
-  a rendered table — `models search`, `images search`, `app status`,
+- **A table cell is one line, and one column.** Every **server-supplied** value
+  that reaches a cell of a rendered table — `models search`, `images search`, `app status`,
   `workflows list`, the pre-spend cost table, and the rest — has any newline or
   tab in it replaced by a **space**. A newline would otherwise start a line at
   column zero, where it is indistinguishable from a row the CLI wrote; a tab is
@@ -3777,14 +3777,19 @@ it reaches your terminal, and this is what that gate promises:
   table, `collections get`, `app view`) have their escapes stripped but **may still
   carry a newline**, so a hostile value there can start a line at column zero.
   Telling a genuinely single-line field from legitimately multi-line free text is
-  a per-field judgement, and only the fields listed above have had it made.
+  a per-field judgement. The list above is **illustrative, not exhaustive** — more
+  fields are flattened than it names (`images … --meta`'s cfg, steps and url among
+  them). Treat it as "these definitely are", never as "only these are": the
+  authoritative answer is the ledger in `internal/cmd/tabwriter_ledger_test.go`,
+  and an enumeration in prose goes stale the moment a renderer is added.
 - **Genuinely multi-line server text keeps its line breaks**, and is indented under
   the line that introduced it, so a continuation can never sit at column zero.
-  There are four such surfaces: the generation prompt and negative prompt
+  There are five such surfaces: the generation prompt and negative prompt
   (`images … --meta`), the orchestrator's failure reason (`workflows get` and
-  `workflows list`), per-output exclusion reasons (`generate`), and the reviewer's
-  rejection reason / approval notes (`app status --id`). If a field you expect to
-  be multi-line arrives on one line, it was in a cell.
+  `workflows list`), the same reason on `generate`'s **error** path, per-output
+  exclusion reasons (`generate`), and the reviewer's rejection reason / approval
+  notes (`app status --id`). If a field you expect to be multi-line arrives on one
+  line, it was in a cell.
 - **Known limits.** A long value is not shortened, so your terminal can still
   soft-wrap it to column zero, and one hostile value widens a column for every
   row. Only `workflows list` wraps its reason text to a fixed budget; the other
--- a/internal/cmd/app_listing.go
+++ b/internal/cmd/app_listing.go
@@ -488,8 +488,11 @@ func listingStatusPayload(slug string, ref *appapi.ListingRef, view *appapi.List
 
 func printListingStatus(w io.Writer, slug string, ref *appapi.ListingRef, view *appapi.ListingEditView) {
 	tw := tabwriter.NewWriter(w, 0, 0, 2, ' ', 0)
+	// `slug` is what the USER typed, echoed exactly — sanitising the user's own
+	// bytes is the civitai/cli#393 defect. `ref.Status` is SERVER text in a
+	// tabwriter cell, so it goes through safeTermSingle (civitai/cli#552).
 	fmt.Fprintf(tw, "App:\t%s\n", slug)
-	fmt.Fprintf(tw, "Listing status:\t%s\n", ref.Status)
+	fmt.Fprintf(tw, "Listing status:\t%s\n", safeTermSingle(ref.Status))
 	fmt.Fprintf(tw, "Icon:\t%s\n", assetLabel(view.Assets.Icon.Present(), true))
 	fmt.Fprintf(tw, "Cover:\t%s\n", assetLabel(view.Assets.Cover.Present(), true))
 	fmt.Fprintf(tw, "Screenshots:\t%d\n", len(view.Assets.Screenshots))
--- a/internal/cmd/app_listing.go
+++ b/internal/cmd/app_listing.go
@@ -498,12 +498,20 @@ func printListingStatus(w io.Writer, slug string, ref *appapi.ListingRef, view *
 	fmt.Fprintf(tw, "Screenshots:\t%d\n", len(view.Assets.Screenshots))
 	_ = tw.Flush()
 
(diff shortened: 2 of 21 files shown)


ZacxDev:
Round-1 audit ran against `ea39f7f` (content-identical to `bc8ca7c` for this PR's payload — the rebase only pulled in an unrelated handoff-doc change from `main`). Two independent rounds ran: a requirements/deletion pass and a blind correctness pass. Their findings were fixed in `30f2630`.

Claims block for the delta round, so the next audit anchors correctly rather than re-reading the whole PR:

```audit-claims round=1 audited=bc8ca7c..30f26304
1. The ledger's sink seed now walks *ast.ValueSpec as well as *ast.AssignStmt, so a renderer whose sink is `var tw = tabwriter.NewWriter(...)` is no longer invisible to GREW/BARE-SAFETERM; two calibration cases added (`var x =`, grouped `var ( … )`).
2. app_status.go's rejection reason and approval notes now go through safeTerm + indentContinuation, and its live URL and blockId through safeTermSingle — closing a path where a raw ESC reached stdout.
3. app_listing.go's screenshot id and caption now go through safeTermSingle, closing a forgery that impersonated the same function's own `App:` label.
4. generate.go:1925's bare safeTerm(r.ID) unified on safeTermSingle, matching the call 13 lines above it.
5. gateUnsanitised and gateNoServerText deleted, along with maxUnsanitisedTabwriterRenderers and the RATCHET arm; the gate set is closed at two and the default arm rejects an invented spelling. No escape hatch retained.
6. The `depth > 4` local-alias laundering limit is NOT fixed; it is now enumerated as residual 5 with its reason.
7. The stale "closing condition for #552" claim in indentcontinuation_ledger_test.go corrected in both the header and the test doc comment.
8. TestTabForgeryIsNeutralisedInTheRealImagesTable deleted as strictly subsumed, with its #569 payload folded into the ledger's images-search fixture via a new alsoWant field.
9. The README's two overclaims narrowed: the "every server-supplied string" sentence now matches what the code guarantees, "the two that matter" corrected to four and enumerated, and the ~60 remaining read-path label:value sites stated as a known limit rather than silently covered.
10. The PR body's MISLABELLED-rows-at-base figure corrected from 16 to 18, with the measurement method shown.
```

Not merged. A delta re-audit of `bc8ca7c..30f2630` follows — a fix round frequently introduces the next finding, and this one touched payload in four files.

ZacxDev:
## Round 2 (delta `bc8ca7c..30f2630`) — findings fixed in `34f0257`, and **the ladder stops here**

Round 2 refuted **no shipped behaviour claim**. Every fix (1–4, 6–8, 10) verified under mutation and against `bc8ca7c`; the four findings were all false or incomplete **sentences**.

### 🔴 The one that mattered, and a third site the audit did not report

Deleting `gateUnsanitised`/`gateNoServerText` led the const block to assert *"a renderer that reaches a cell without safeTermSingle has no truthful row available."* **False.** `gatePreSanitised` resolves its `upstream` only to a NAME — it checks the function exists and calls `safeTermSingle` *somewhere*, never that it feeds **this** renderer's cells. Measured: a renderer writing a raw server string into a cell, ledgered `{…, gatePreSanitised, "printTagList", …}`, passes the whole package.

Its own comment had called itself *"A BINDING, NOT AN EXCUSE"* while naming the hazard it then implemented one level indirected — *"it is handled somewhere else, asserted in prose."*

I swept for the **shape** rather than the two reported sites, and found a **third**: residual 1 claimed the set being "CLOSED at two" meant there was "no ungated-but-declared state to move a hole into." Same claim, same falsity, in the line a reader hits first.

All three now state the residual and say **nothing justifies it**. It is **pre-existing** — the same mutation is green before this file existed — so nothing widened it; what shipped was the sentence saying it was shut.

### README

- *"Every value that reaches a cell"* — was *"a server value"* before this branch rewrote it. **Wider on the origin axis while narrower on the `label: value` axis**, the exact shape this ladder hunts. Scoping restored.
- *"four such surfaces"* → **five** (`generate`'s error path was missing).
- *"only the fields listed above"* → marked **illustrative**, with the ledger named as authoritative, because a prose enumeration goes stale the moment a renderer is added.

Plus a dangling pointer to the test the previous commit deleted, and a new **residual 7** — the cost of closing the set: an all-CLI-owned renderer has no honest row today.

---

## 🔴 Why this is the last round

Stating it rather than leaving it implicit, because a report that **ended** a ladder on this criterion is otherwise indistinguishable from one that converged.

**The rounds will not stop on their own here.** Every remaining finding is prose, so the attribution gate is structurally inert — "fixed a defect" and "reworded a warning" are the same edit, and round 3 would audit round 2's wording exactly as round 2 audited round 1's.

The stated criterion is met: **no 🔴**; blast radius is bounded by *"the document contains a false sentence"*; and the recurring shape — a guard description asserting coverage its body does not provide — was **swept at every site**, not just the reported ones, which is what turned up the third.

### Knowingly NOT fixed, so the next reader knows these are open rather than absent

- `gatePreSanitised`'s non-relational resolution (**residual, recorded, pre-existing**) — closing it means resolving `upstream` against the renderer's own cell expressions, which the scan cannot do for the struct-field case the state exists to cover.
- `depth > 4` laundering (residual 5); the all-CLI-owned renderer (residual 7).
- Soft wrap, U+2028/9, the ~60 read-path `label: value` sites, `pinnedIndentCallSites`' set-pinning.
- **#552 stays OPEN** for the soft-wrap residual.

Gate on this commit: `make ci` 21 packages ok, `golangci-lint` 0 issues, `gofmt` clean over 478 files.

Issue #575, opened 2026-09-12
Four recorded residuals from the tabwriter ledger (#552 successor)


ZacxDev:
> **STATUS, 2026-09-14 — R2, R3 and R4 were ACCEPTED IN WRITING; this issue now rests on R6 alone.**
> The decision and its reasoning are in a comment below. The 09-13 note follows unchanged.
>
> **STATUS, 2026-09-13 — read this before the list below.** Two of these are
> CLOSED and two more residuals were found while closing them. The per-residual
> headings below were written when all four were open and are annotated in place
> rather than rewritten, so the original framing stays readable.
>
> | residual | state |
> |---|---|
> | **R1** `gatePreSanitised` resolves a name | ✅ **CLOSED** — #578 (`3457c5dc`): the state was deleted rather than made relational |
> | **R2** laundering past four hops | ✅ **CLOSED** — accepted in writing 2026-09-14 by @ZacxDev (comment below): bound measured 4/5, live corpus max 2 hops, GREW/SHRANK/MISLABELLED unaffected |
> | **R3** a renderer with no server text has no honest row | ✅ **CLOSED** — accepted 2026-09-14 (comment below). 🔴 Its "points at `gatePreSanitised`" clause is STALE: #578 deleted that state |
> | **R4** the same hazard one line outside a cell | ✅ **CLOSED** — accepted 2026-09-14 (comment below); was already partly narrowed by #578 |
> | **R5** one-letter `bareIdentArgs` keys (comment below) | ✅ **CLOSED** — #582 (`095f4acf`): re-keyed `enclosingFunction::argument` |
> | **R6** a `(function, name)` key cannot express two callers with two origins (comment below) | 🟡 **open** — surfaced by R5's audit |
>
> R5 and R6 are recorded as comments on this issue, not in this body.

Successor to #552, whose closing condition was met by #573 (`6f0a8d8`). These four were **recorded in the code as open** — each is stated in `internal/cmd/tabwriter_ledger_test.go`'s residual list or const block rather than implied — and none is covered by #552's condition. Filed together because they share one root: a structural scan over tabwriter cells cannot see any of them.

## Closing condition

Each residual below is either **fixed**, or **re-stated with a decision recorded** by a named reader. R1 is the only one that is plainly a defect; R2–R4 are trade-offs that may legitimately be accepted in writing.

[...]

PR #578, opened 2026-09-12
fix(safeterm): delete the ledger gate state that resolved a name, not a relationship

ZacxDev:
Closes the R1 half of #575: the tabwriter ledger had two gate states and one of
them resolved a function **name**, never a relationship to the renderer it was
excusing.

## The defect, reproduced before touching anything

`gatePreSanitised` meant "this renderer's cells were gated upstream, through a
struct field the scan cannot follow". Its `upstream` was checked two ways — the
named function must EXIST in the package, and it must call `safeTermSingle`
SOMEWHERE. Neither tied it to *this* renderer's cells.

Measured on `main` (`d91e857`): a renderer writing a **raw server string** into a
tabwriter cell — no `safeTerm`, no `safeTermSingle` — ledgered as
`{…, gatePreSanitised, "printTagList", …}`, where `printTagList` is an unrelated
renderer that merely happens to sanitise, passes `go vet` **and the whole
`internal/cmd` package**. Pre-existing: the same row is green before the ledger
file existed.

So the set was closed at two states, but one of them was a hole any row could be
moved into.

## The fix: delete the state, don't make it cleverer

The issue proposed resolving `upstream` against the renderer's own cell
expressions. That narrows the hole to a *field name* and still leaves one.
Deleting the state closes the walk instead.

Its single user, `printGenerateQuote`, now gates at the cell. `safeTermSingle`
over already-gated text is an idempotent no-op, so **output is unchanged** — what
changes is that the gate is somewhere the scan can see.

This file had already made exactly this move once, deleting `gateNoServerText`
and `gateUnsanitised` for being machine-indistinguishable, and recorded that
*"deleting them is strictly stronger than a ceiling nobody was under."*

## Nothing was unpinned, and that was measured

The row's stated value was that deleting `describeVersion`'s gate would fail
HERE, and its message went further: *"nothing else would have said so."* Removing
that `safeTermSingle` in fact reddens **three** tests —
`TestGenerate_SanitisesServerStrings`, `TestSafeTermCallSitesAreCoveredByANamedTest`
and the ledger. The claim was false; the pin is redundant twice over.

## What holds the set shut now

Not a sentence. The row verdict (GREW / SHRANK / MISLABELLED) is factored out as
`ledgerVerdicts`, so `TestNoLedgerStateClassifiesAnUngatedRenderer` runs the
**real** rules over a synthetic ungated renderer under every gate spelling. A
test that re-implemented those rules would only prove two copies agree.

Every escape route watched, on the real tree, with the same planted renderer:

| the author writes | verdict |
|---|---|
| `"pre-sanitised-upstream"` (the deleted state) | `unknown gate` — red |
| `gateSingle` | `MISLABELLED` — red |
| no row at all | `GREW` — red |
| *all three on `main` before this change* | **green** |

Mutation-isolated, each dying with this guard's own message: neutering the
default arm reddens **only** the new test — the existing ledger stays green,
which is what shows the new test is not redundant — and neutering the
`gateSingle` arm reddens only its own case. The positive control (gated at the
cell, honestly ledgered) must be accepted, so a verdict function that rejected
everything would fail the suite.

## Behavioural half

`printGenerateQuote` is driven with the `resolvedGraph` fields set **directly**,
bypassing `describeVersion` — the case the old row could not ask: whatever
reaches the field, does this pre-spend screen forge? Watched red: deleting the
Checkpoint cell gate fails it on `gqckpt` and `gqnote` plus two column-zero
findings.

## The second commit is a near-miss I caught on myself

The first commit classified the LoRA loop variable as `"l"` in `bareIdentArgs`.
That map is looked up as `bareIdentArgs[s.arg]` — keyed by the bare **name across
the whole package**, not per enclosing function — so a one-letter key allowlists
every other `safeTerm(l)` in `internal/cmd`. That is the defect #554 proposed for
`s` and #569 refused, in a different letter.

Measured with a `safeTerm(l)` planted over the **user-typed prompt**:

- with `"l"` in `bareIdentArgs` → **SURVIVES**, harness blind
- with it removed → KILLED, named at `generate.go:1777:30`
- after renaming to `loraLabel` → KILLED, named

## Gate

`make ci` green, `make lint` **0 issues** — with a negative control: a planted
unused variable makes lint exit 1, so the zero is a reading.

`internal/cmd` also demanded rows in two *other* ledgers for the new call sites
(`safeTermCoveredBy`, `bareIdentArgs`); both are filled and the coverage row
names a test watched going red, not one read and believed.

## Against #572

#572 touches both `safeterm_coverage_test.go` and `safeterm_userinput_test.go`.
`git merge-tree` exits **0** (branched on the exit code, not a marker grep), its
new keys `partPath` / `safeTermErr` don't collide with `loraLabel`, and both PRs
*strengthen* the guards rather than one blinding the other. Full suite run on the
**merged tree**, not on either branch: 21 packages ok, lint 0 issues.

## Not in scope

R2 (laundering past four hops), R3 (a renderer with no server text has no honest
row) and R4 (the same hazard one line outside a cell) stay open on #575. R3's
text is updated: deleting `gatePreSanitised` **narrows** it rather than widening
it — that state was never an honest option for such a renderer either (the old
text called it "a lie"), so no truthful row was lost; what is gone is the
untruthful one the GREW message used to point at.

Refs #575

### DIFF
--- a/internal/cmd/generate.go
+++ b/internal/cmd/generate.go
@@ -1825,11 +1825,21 @@ func printGenerateQuote(out, errw io.Writer, built *resolvedGraph, o generateOpt
 	if o.aspectRatio != "" {
 		fmt.Fprintf(tw, "Aspect ratio:\t%s\n", o.aspectRatio)
 	}
+	// 🔴 GATED AT THE CELL, THOUGH describeVersion ALREADY GATED THE VALUE —
+	// civitai/cli#575 R1. These are the only SERVER-derived cells on this screen,
+	// and they arrive through a resolvedGraph FIELD, which the tabwriter ledger's
+	// taint walk deliberately does not follow. That used to be recorded as a
+	// ledger row naming describeVersion as the upstream gate; the row resolved
+	// only to a NAME, so any unrelated sanitising function satisfied it and the
+	// state was a hole any ungated renderer could be moved into. Gating here costs
+	// an idempotent second pass (safeTermSingle over already-sanitised text is a
+	// no-op) and buys a gate the scan can SEE, which is what let that third
+	// classification be deleted rather than made cleverer.
 	if built.checkpoint != "" {
-		fmt.Fprintf(tw, "Checkpoint:\t%s%s\n", built.checkpoint, built.checkpointNote)
+		fmt.Fprintf(tw, "Checkpoint:\t%s%s\n", safeTermSingle(built.checkpoint), safeTermSingle(built.checkpointNote))
 	}
 	for _, l := range built.loras {
-		fmt.Fprintf(tw, "LoRA:\t%s\n", l)
+		fmt.Fprintf(tw, "LoRA:\t%s\n", safeTermSingle(l))
 	}
 	// 🔴 NOT "Generatable". The server's `ready` is a RESOURCE-AVAILABILITY flag,
 	// not a prediction that the job will produce anything: it is computed as
--- a/internal/cmd/generate.go
+++ b/internal/cmd/generate.go
@@ -1838,8 +1838,15 @@ func printGenerateQuote(out, errw io.Writer, built *resolvedGraph, o generateOpt
 	if built.checkpoint != "" {
 		fmt.Fprintf(tw, "Checkpoint:\t%s%s\n", safeTermSingle(built.checkpoint), safeTermSingle(built.checkpointNote))
 	}
-	for _, l := range built.loras {
-		fmt.Fprintf(tw, "LoRA:\t%s\n", safeTermSingle(l))
+	// 🔴 NOT `l`. The bare-identifier half of safeterm_userinput_test.go's ledger
+	// is keyed by the argument NAME across the WHOLE package, so classifying a
+	// one-letter loop variable allowlists every other safeTerm(l) in internal/cmd
+	// — the same defect civitai/cli#554 proposed for `s` and #569 refused.
+	// Measured both ways before renaming: with "l" in bareIdentArgs a planted
+	// safeTerm(l) over the USER-TYPED prompt SURVIVES; with this name it is killed
+	// and named. A distinctive name is what keeps the ledger entry about THIS value.
+	for _, loraLabel := range built.loras {
+		fmt.Fprintf(tw, "LoRA:\t%s\n", safeTermSingle(loraLabel))
 	}
 	// 🔴 NOT "Generatable". The server's `ready` is a RESOURCE-AVAILABILITY flag,
 	// not a prediction that the job will produce anything: it is computed as
--- a/internal/cmd/safeterm_coverage_test.go
+++ b/internal/cmd/safeterm_coverage_test.go
@@ -263,6 +263,14 @@ var safeTermCoveredBy = map[string]safeTermCoverage{
 	"printCostMap": {"TestTabwriterRenderersCannotBeForged",
 		"the cost-map KEYS on the quote screen — server-named factors beside Buzz amounts, written " +
 			"into the PRE-SPEND table through a tabwriter its caller owns"},
+	"printGenerateQuote": {"TestTabwriterRenderersCannotBeForged",
+		"the Checkpoint cell, its substitution note and each LoRA cell on the PRE-SPEND quote screen. " +
+			"describeVersion already gates these on the way INTO resolvedGraph; they are gated AGAIN at " +
+			"the cell because a ledger row naming an upstream function resolved only to a NAME and was " +
+			"therefore satisfiable by any unrelated sanitising renderer (civitai/cli#575 R1). The named " +
+			"test drives this renderer with the fields set DIRECTLY, bypassing describeVersion, so it " +
+			"goes red when the cell gate is deleted — watched, not read: deleting safeTermSingle from " +
+			"the Checkpoint cell fails it on gqckpt and gqnote plus two column-zero findings"},
 	"classifyGenerateError": {notCovered,
 		"the server's own error message, shown verbatim when generation is refused"},
 	"buildGenerateGraph": {notCovered,
--- a/internal/cmd/safeterm_userinput_test.go
+++ b/internal/cmd/safeterm_userinput_test.go
@@ -79,7 +79,7 @@ var bareIdentArgs = map[string]string{
 	"sha":        "SERVER: a published hash",
 	"reason":     "SERVER: an orchestrator failure reason",
 	"name":       "SERVER: a published file name",
-	"l":          "SERVER-derived: one describeVersion label out of resolvedGraph.loras, re-gated at the quote screen's cell (civitai/cli#575 R1)",
+	"loraLabel":  "SERVER-derived: one describeVersion label out of resolvedGraph.loras, re-gated at the quote screen's cell (civitai/cli#575 R1). Spelled out rather than `l`: this map is keyed by NAME across the package, so a one-letter key would allowlist every safeTerm(l) in internal/cmd",
 	"h":          "SERVER: a hash out of image metadata",
 	"baseModel":  "SERVER: a base-model label",
 	"typ":        "SERVER: the primary file's published `type`, defaulted to \"Other\" when blank (nonModelFileMarker)",
--- a/internal/cmd/safeterm_userinput_test.go
+++ b/internal/cmd/safeterm_userinput_test.go
@@ -79,6 +79,7 @@ var bareIdentArgs = map[string]string{
 	"sha":        "SERVER: a published hash",
 	"reason":     "SERVER: an orchestrator failure reason",
 	"name":       "SERVER: a published file name",
+	"l":          "SERVER-derived: one describeVersion label out of resolvedGraph.loras, re-gated at the quote screen's cell (civitai/cli#575 R1)",
 	"h":          "SERVER: a hash out of image metadata",
 	"baseModel":  "SERVER: a base-model label",
 	"typ":        "SERVER: the primary file's published `type`, defaulted to \"Other\" when blank (nonModelFileMarker)",
--- a/internal/cmd/tabwriter_forgery_test.go
+++ b/internal/cmd/tabwriter_forgery_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"io"
 	"regexp"
 	"strings"
 	"testing"
@@ -464,6 +465,32 @@ func TestTabwriterRenderersCannotBeForged(t *testing.T) {
 					tw := tabwriter.NewWriter(w, 0, 0, 2, ' ', 0)
 					printCostMap(tw, "factors", map[string]float64{forgeCell("cmkey"): 2})
 					_ = tw.Flush()
+				})
+			},
+		},
+		{
+			// 🔴 THE FIELDS ARE SET DIRECTLY, BYPASSING describeVersion, AND THAT IS
+			// THE POINT — civitai/cli#575 R1. These cells are normally gated by
+			// describeVersion on the way IN, and that used to be the whole claim: a
+			// ledger row naming describeVersion as the "upstream". The row resolved
+			// only the NAME, so it was equally satisfied by any unrelated sanitising
+			// function, and nothing checked that the value in the field had actually
+			// been through one. Driving the renderer with a hostile field is what
+			// asks the question the row could not: whatever reaches
+			// resolvedGraph.checkpoint / .checkpointNote / .loras, does this screen
+			// forge? It is a PRE-SPEND surface — the table a user reads before
+			// agreeing to be charged.
+			surface: "the pre-spend quote screen (printGenerateQuote)",
+			fields:  []string{"gqckpt", "gqnote", "gqlora"},
+			render: func() string {
+				return writerOut(func(w *bytes.Buffer) {
+					built := &resolvedGraph{
+						checkpoint:     forgeCell("gqckpt"),
+						checkpointNote: forgeCell("gqnote"),
+						loras:          []string{forgeCell("gqlora")},
+					}
+					printGenerateQuote(w, io.Discard, built, generateOpts{},
+						&genapi.WhatIfResult{Ready: true, Cost: &genapi.WorkflowCost{Base: 1, Total: 1}})
 				})
 			},
 		},
--- a/internal/cmd/tabwriter_ledger_test.go
+++ b/internal/cmd/tabwriter_ledger_test.go
@@ -44,20 +44,27 @@ import (
 // satisfied by adding a call to a named helper.
 //
 // 🔴 WHAT THE STRUCTURAL HALF CANNOT SEE, stated rather than waved at:
-//  1. Server text written into a cell with NO sanitizer at all. No AST can tell
-//     a server string from a CLI-owned label, so that judgement is the LEDGER's
-//     job: every renderer carries a gate classification and a reason naming the
-//     fields, and the set of classifications is CLOSED at two.
-//     🔴 CLOSED AT TWO IS NOT THE SAME AS "no state to move a hole into", and an
-//     earlier draft of this line claimed it was. gatePreSanitised resolves its
-//     `upstream` only to a NAME, never against this renderer's own cells, so a
-//     row naming any unrelated sanitising renderer passes — see its const-block
-//     note. The hole is one NAME wide rather than one WORD wide; it is not shut.
+//  1. WHICH cells carry server text, and which are CLI-owned labels. No AST can
+//     tell the two apart, so that judgement is the LEDGER's job: every renderer
+//     carries a reason naming the fields, field by field.
+//     🔴 WHAT IS NO LONGER A RESIDUAL: a cell reaching a tabwriter with NO
+//     sanitizer at all. This file shipped a third classification,
+//     gatePreSanitised — "gated upstream, through a struct field the scan cannot
+//     follow" — whose `upstream` resolved to a NAME and never against this
+//     renderer's own cells, so a row naming ANY unrelated sanitising function
+//     passed. That made it a hole every other row could be moved into
+//     (civitai/cli#575 R1; reproduced before this change, on a renderer writing a
+//     raw server string into a cell). It is DELETED rather than made relational:
+//     its single user (printGenerateQuote) now gates at the cell, where a second
+//     idempotent safeTermSingle costs nothing and the scan can SEE it.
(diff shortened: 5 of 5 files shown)


ZacxDev:
## 🔴 Correcting this PR's body: "nothing was unpinned" was too wide

Round 0 of the audit found a coverage reduction this PR causes and whose opposite
the body asserts. Posting rather than silently editing the body, since anyone who
already read it read the wrong version.

**The body says:** *"Nothing was unpinned, and that was measured."* That is true
of the **deleted ledger row**, and false of the change as a whole.

Deleting both `safeTermSingle` calls in `describeVersion`:

| tree | tests reddened |
|---|---|
| `d91e857` (main) | **3** — `TestGenerate_SanitisesServerStrings` (behavioural), the safeTerm coverage ledger, the tabwriter ledger |
| `4c69bb2..745bdd4` (as reviewed) | **1** — the coverage ledger only, i.e. structural |

`TestGenerate_SanitisesServerStrings` drives `--dry-run`, so it reached
`describeVersion`'s output only *through* `printGenerateQuote`. Once that renderer
gained its own cell gate — correct, and the point of the PR — it sanitises the
payload whatever `describeVersion` did, so the upstream test went green on a
broken upstream. Defence in depth ate the test for the layer below it.

### Why that mattered beyond the test

`confirmGenerate` prints `Checkpoint:` and each `LoRA:` with a plain `Fprintf`
**four lines above `Generate? [y/N]:`**. A model name carrying `\n` forges
arbitrary lines at column zero in the last screen a user reads before an
irreversible Buzz spend — including a fake `Cost:` line. Those lines were ungated
at the call site and safe only because `describeVersion` gates on the way in, and
nothing ever drove that screen with a hostile value.

So the masked pin was not a bookkeeping problem: it removed the only thing
demonstrating that a money-path screen could not be forged.

### Fixed in `5ca1e1c`

1. **`confirmGenerate`'s two label lines are gated at the call site** — the
   money-path half of the residual #575 R4 records for plain `label: value`
   lines. `TestGatedRenderersDoNotForgeOutsideTheirTable` now drives that screen
   with the `resolvedGraph` fields set **directly**, bypassing `describeVersion`.
   Watched red: deleting the Checkpoint gate fails it with two column-zero
   findings.

   That case's two controls earned their place immediately — it first ran against
   an **empty buffer**, because `confirmGenerate` refuses when stdin is not a TTY
   and it never is under `go test`. Without them the forgery assertions would have
   passed over nothing.

2. **`TestDescribeVersionIsSanitisedAtItsOwnOutput`** asserts the function's
   *return value*, which no downstream gate can mask. The same mutation reddens a
   behavioural test again.

### Two smaller corrections from the same round

- The new `safeTermCoveredBy` row claimed `checkpointNote` is gated by
  `describeVersion`. It isn't — `substitutionCheckpointNote` builds a CLI-owned
  literal around a `%d`. Corrected, and it now records *why* the field is gated
  anyway: so an author extending it to name the substituted model doesn't inherit
  a row saying it was already handled.
- My one-letter-key comment stated the rule at package width while
  `bareIdentArgs` holds **five** one-letter keys (`w`, `r`, `k`, `t`, `h` of 16),
  each a live allowlist of that name at every bare-ident `safeTerm` site.
  Pre-existing, not fixed here (renaming them touches files other work is in), but
  no longer implied to be enforced. Worth tracking on #575.

### Round 0's other output, deliberately not actioned here

Round 0's verdict was `deletion candidate — the ledger's one-valued gate column`.
With `gatePreSanitised` gone, all 21 rows spell `gateSingle`, the `default:` arm
is unreachable on the committed tree, and a field that can hold one value is not a
classification — which is the same disqualification this file used when it deleted
`gateNoServerText` and `gateUnsanitised`. It's a fair argument and it's the PR's
own thesis applied one step further. Leaving it for a maintainer call rather than
widening this diff again; recorded here so it isn't lost.

`make ci` green, `make lint` 0 issues (with a negative control — a planted unused
variable moves it to `1 issues`).

ZacxDev:
## Round 1 (nine axes, dispatched blind) — findings and fixes

Round 1 re-measured every number this PR asserts and confirmed all of them,
including that `safeTermSingle` is idempotent over 198,657 runes (so rendered
output is byte-identical), that all three escape routes are red at HEAD, and that
each mutation arm is isolated. It returned two 🟡s and two 🟢s. Both 🟡s were
defects in **my prose**, not in the shipped behaviour — the shape this arc keeps
producing.

### F1 — the code said a residual was "Recorded on #575". It was not.

`internal/cmd/generate.go` carried *"Recorded on civitai/cli#575"* about the
one-letter-key residual. Measured: #575 had **zero** comments and its body records
only R1–R4. The PR comment said "worth tracking" — a plan — while the shipped code
said "recorded".

The failure that matters: an author hitting a bare `safeTerm(s)` follows that
pointer, finds nothing about it, concludes the note is stale, and adds `"s"` to
`bareIdentArgs` — the move #554 proposed and #569 refused.

Fixed by making the sentence true: posted as **R5 on #575** with the measurement
and a closing condition, and the comment now names R5. Verified after posting
rather than assumed.

### F2 — a third server-derived label line on the approval screen was still forgeable

The same comment claimed gating `confirmGenerate` was *"the money-path half of the
residual #575 R4 records"*. It was not the whole of it. `printImageDisclosure`
prints `Image:  %s` with a plain `Fprintf` **between the LoRA lines and the real
`Cost:` line**, and its value went through `safeTerm`, which deliberately keeps
`\n`. Round 1 measured the consequence — a fake
`Cost: 1 Buzz (balance 999999).` rendered directly above the true one, four lines
above `Generate? [y/N]:`.

🔴 **This line is genuinely gated upstream, and it is the only one left.**
Everywhere else this PR moved the gate to the *render* site, because a gate at the
cell is one the structural ledger can see. This line cannot follow that rule: it
interleaves **two origins** — the path the user typed, echoed byte-for-byte
(#393), and the server's blob URL. Flattening the composed string at the render
site would rewrite the user's own path, which is exactly what #393's first cut
shipped. The halves are separable only where they are joined, so that is where the
gate is: `safeTerm` → `safeTermSingle` on the URL half only.

Which means the render-site test **cannot** pin it — setting `built.images`
directly bypasses the composition, and my first attempt did exactly that and
failed. `TestImageDisclosureLineIsGatedAtComposition` drives the real upload seam
instead. Watched red: reverting to `safeTerm` fails it on both the newline arm and
the dropped-value arm. It also asserts the **user** half survives byte-for-byte, so
a "fix" that flattened the whole line would satisfy every other assertion and
silently reintroduce #393.

### F3, F4

- The `loraLabel` ledger row named one of its two call sites. It is a
  `safeTermSingle` argument in `printGenerateQuote` (a tabwriter cell) **and**
  `confirmGenerate` (a plain label line); the row now names both.
- The hand-rolled `stdinIsTTY` stub is replaced by the package's existing
  `withStdinTTY` helper.

### Base moved: #572 merged

`#572` landed as `f0cb748` while this was in flight, and it touches two of the
same ledger files. The earlier merged-tree check was against its PR head, so it
has been **re-run against the real new base**: `origin/main` merged in as
`73ceeda`, no conflicts, 21 packages `ok`, `golangci-lint` 0 issues.

### Round 0's deletion candidate — still open, still a maintainer call

`deletion candidate — the ledger's one-valued gate column`. Round 1 declined to
file it as a finding, noting a real distinction from #573's deleted arms: this
`default:` arm **is** exercised, by the new synthetic test. Recorded, not actioned.

```audit-claims round=1 audited=5ca1e1c..73ceedad
1. F1: the "Recorded on #575" claim in generate.go is now true — the residual is posted as R5 on #575 and the comment names it.
2. F2: printImageDisclosure's server-derived URL half is gated with safeTermSingle at the composition site, with the user-typed half left byte-for-byte.
3. F2: TestImageDisclosureLineIsGatedAtComposition drives the real upload seam, because a render-site test cannot pin a value composed upstream.
4. F3: the loraLabel bareIdentArgs row now names both call sites, the quote-screen cell and the approval-screen label line.
5. F4: the hand-rolled stdinIsTTY stub is replaced by the existing withStdinTTY helper.
6. Base sync: origin/main (f0cb748, #572 merged) merged in as 73ceeda; full suite and lint re-run on that tree.
```

**Ledger:** `round 1 · payload lines changed THIS round: 18+6 in internal/cmd/generate.go (since round 1: same) · scaffolding: 65 generate_test.go, 1+1 safeterm_userinput_test.go, 1+3 tabwriter_forgery_test.go`. Payload is non-zero, so the attribution gate does not fire.

ZacxDev:
## Round 2 (delta, `5ca1e1c..73ceeda`, dispatched blind) — findings and fixes

Round 2 confirmed all six of round 1's claims and proved the #572 merge clean two
independent ways (`--remerge-diff` emits no content; `merge-tree --write-tree`
returns a tree byte-identical to `73ceeda^{tree}`). It also read every overlapping
region of the two files both PRs touch and found no semantic conflict — neither
side blinded a guard the other strengthened.

Its four findings were all in **my** guards and prose. Three are corrected below;
the fourth was an instrument defect I'd have shipped.

### 🔴 A — the new test's #393 arm was vacuous, and my commit message pointed a reader at it

`f1b666c` asserted the arm would stop a "fix" that flattened the composed line. It
would not. The fixture path was `writePNG`'s own `icon.png` — pure ASCII — so
`safeTermSingle` is the identity on it and `strings.Contains(line, src)` could not
distinguish *untouched* from *flattened*. RULES.md names this exactly: a fixture
that can only ever produce the constant's own value cannot see the mutant.

**Re-measured first-hand rather than taken from the report:** gating the *composed*
string instead of the URL half is **green across all 21 packages**. Under that
mutant, `civitai generate --image 'می‌روم.png'` renders `میروم.png` on the approval
screen — ZWNJ stripped, letters joined — while the wire carries the original. That
is #393 verbatim, on the path `TestSameBytesTypedAndReceived_AreEchoedAndStripped`
does not reach (it drives `o.prompt` only).

The fixture path now carries a U+200C, with a control asserting it still does so
the arm cannot silently stop measuring again. Watched: the mutant that was green
now dies, naming the joined path.

### 🟡 B — `buildGenerateGraph` was ledgered `notCovered` while this PR made it covered

The file's own rule is "the test that goes RED when the `safeTerm` calls are
deleted". Deleting its only gate reddens the new test with its own messages — so
the row was false, on a function this PR changed, and it left a unit of ratchet
headroom. `safeterm_coverage_test.go` calls that *"not a nit: a cap above the real
count is headroom a LATER commit can spend in silence"*.

Row now names the test, and `maxUncoveredSafeTermFuncs` goes 21 → **20**, banked in
the same commit as the harness demands.

### 🟡 C — "of 16" was staled by the merge inside this PR's own range

#572 added `partPath`, so `bareIdentArgs` holds **17** keys; the five one-letter
ones are unchanged. Re-derived from the harness itself, which prints
`17 bare-identifier origins are pinned`. This is the class the brief warns about —
a claim true when an early commit wrote it and falsified by a later commit of the
same PR, inside no round's range.

The **R5 comment on #575 carried the same stale figure** and has been corrected in
place with a dated note rather than silently.

### 🟢 D and E

- The forged-`Cost:` geometry was wrong for the `Image:` line *and* labelled
  "measured". `confirmGenerate` **always** prints its img2img caveat after the last
  `Image:` line, so a forgery there is two lines above the true `Cost:` and five
  above the prompt — never "directly above". The four-line figure was carried over
  from the `Checkpoint:`/`LoRA:` comments, where it is correct. Fixed in both
  places it appears.
- The new test passed a nil `context.Context`; now `context.Background()`.

### 🔴 `make ci` passed and `make lint` did not

Worth recording because it is exactly the split AGENTS.md describes. The raw
U+200C in the new fixture tripped **staticcheck ST1018** — the same rule that once
let four control-byte fixtures reach a push — while `make ci` was green. Escaped to
`‌`, and the mutation re-run **afterwards** to confirm the escape did not
re-blind the arm. Both green now.

```audit-claims round=2 audited=73ceeda..d2e2b12
1. A: the #393 arm's fixture path now carries a U+200C, with a control on the fixture, and the previously-green flatten-the-whole-line mutant is watched dying.
2. B: buildGenerateGraph's row names TestImageDisclosureLineIsGatedAtComposition instead of notCovered, and maxUncoveredSafeTermFuncs is lowered 21 -> 20 in the same commit.
3. C: the bareIdentArgs denominator is corrected 16 -> 17 in generate.go, re-derived from the harness, and the R5 comment on #575 is corrected in place with a dated note.
4. D: the forged-Cost geometry for the Image line is corrected to two lines above the true Cost and five above the prompt, in both places it appears.
5. E: the new test uses context.Background() rather than a nil context from a bare cobra.Command.
6. The U+200C is written as an escape so staticcheck ST1018 passes, and the arm's mutation was re-run after the escape.
```

**Ledger:** `round 2 · payload lines changed THIS round: 8+/3- in internal/cmd/generate.go · scaffolding: 34+/5- generate_test.go, 8+/3- safeterm_coverage_test.go · rc 0, silent stderr, non-empty range`. Payload non-zero, so the attribution gate does not fire.

ZacxDev:
## Round 3 (delta, `73ceeda..d2e2b12`, blind) — findings, fixes, and the ladder's stop

Round 3 confirmed all six of round 2's claims by first-hand re-measurement,
including the two that mattered: the #393 arm dies for **its own** reason as the
sole failure in the repo, and the escaped `‌` is present at runtime with a
lint positive control proving ST1018 can still see a raw one. It also caught that
`origin/main` had moved again (`f0cb748` → `deacd19`) and rebuilt the merged tree
rather than trusting the earlier run.

Four findings, all in guards and prose. All four fixed in `bc2aaa5`.

### 🔴 1 — the `describeVersion` row named a test this PR had stopped from reddening

`safeTermCoveredBy["describeVersion"]` read `TestGenerate_SanitisesServerStrings`.
That was true until this PR gave `printGenerateQuote` its own cell gate — which
sanitises the payload whatever `describeVersion` did, so the named test goes green
on a broken upstream.

I measured exactly that in round 0, wrote it into the new test's doc comment,
added the replacement test — **and left the row pointing at the masked one.**

The file's own rule is *"the test that goes RED when the `safeTerm` calls are
deleted — measured by deleting them, not by reading the test and believing it"*,
and `checkCoveringTestsResolve` only proves the named test **exists**. So the row
was a name that resolves rather than a relationship that holds: precisely the
defect this PR deletes an entire ledger state for, regenerated one file over.

| tree | deleting both gates in `describeVersion` |
|---|---|
| `origin/main` | `TestGenerate_SanitisesServerStrings` red — row **true** |
| PR head, before this fix | that test stays **green** — row **false** |
| after the repoint | `TestDescribeVersionIsSanitisedAtItsOwnOutput` red — row true again |

### 🟡 2 — the ratchet's derivation asserted 21 three times above a constant of 20

And directly above its own retraction saying *"a count in a comment is a claim;
re-measure it, do not carry it through a merge."* Amended rather than
search-and-replaced: a reader hitting a RATCHET failure reads that paragraph as
the authority on what the number should be, so each bullet now says which tree its
figure belonged to, and a new one records 21 → 20 and why.

### 🟢 3 — the new fixture control pinned the *rune* while its sentence promised the *property*

Demonstrated by the auditor and re-verified here: exempt U+200C inside `saferune`
the way U+FE0F already is, and the rune assertion passes while `safeTermSingle`
becomes the identity on that path — the #393 arm vacuous again, with nothing
saying so. It now asserts the gate actually **moves** the value, covering both the
rune-lost and the class-changed cases. Watched firing.

### 🟢 4 — the header's scale figures were present tense and stale

"152 sites in 60 functions"; the harness reports **199 in 72**. Pre-existing.
Dated rather than restated, since nothing checks them and the #398 audit they
support is a claim about the tree it ran on.

---

## 🔴 The ladder stops here, on the attribution gate

Not on a clean round — round 3 found real things. On the measured gate:

```
round 2's fixes (73ceeda..d2e2b12):
   8+/3-  internal/cmd/generate.go        <- payload FILE, but 0 executable lines (comment-only)
  34+/5-  internal/cmd/generate_test.go           scaffolding
   8+/3-  internal/cmd/safeterm_coverage_test.go  scaffolding

round 3's fixes (d2e2b12..bc2aaa5):
  12+/5-  internal/cmd/generate_test.go           scaffolding
  38+/9-  internal/cmd/safeterm_coverage_test.go  scaffolding
```

**Two consecutive rounds whose fixes changed zero payload lines ⇒ the ladder has
left the PR.** Both ranges: rc 0, silent stderr, non-empty. The payload/scaffolding
call was fixed at round 1 and has not moved: this PR exists to close #575 R1, which
lives in `tabwriter_ledger_test.go` and `generate.go` — revert either and the
deliverable does not ship. Revert `generate_test.go` or
`safeterm_coverage_test.go` and it does.

Nothing is filed as a follow-up, because all four of round 3's findings are fixed
rather than deferred.

### Still open, and still a maintainer call

Round 0's `deletion candidate — the ledger's one-valued gate column`. With
`gatePreSanitised` gone, all 21 rows spell `gateSingle` and the `default:` arm is
unreachable on the committed tree. Round 1 declined to file it, noting a real
distinction from #573's deleted arms: this arm **is** exercised, by the new
synthetic test. Recorded across three rounds, actioned by none — deliberately.

### Ladder summary

| round | shape | outcome |
|---|---|---|
| 0 | requirements & deletion | coverage reduction on the money path; the PR body asserted its opposite |
| 1 | nine axes, blind | a false "Recorded on #575"; a third forgeable label line on the approval screen |
| 2 | delta, blind | the new #393 guard was **vacuous**; an unbanked ratchet unit; a count staled by my own merge |
| 3 | delta, blind | a ledger row naming a test this PR had masked; three prose claims |

Every round found something real, and **every finding across all four rounds was
in my own work** — three of them in guards or prose that read as coverage while
providing less than they claimed. That is the pattern, and it is the reason the
ladder ran to four.

```audit-claims round=3 audited=d2e2b12..bc2aaa5
1. safeTermCoveredBy["describeVersion"] is repointed to TestDescribeVersionIsSanitisedAtItsOwnOutput, and deleting the gates is watched reddening it.
2. The ratchet derivation paragraph is amended so each figure names the tree it belonged to, and records 21 -> 20.
3. The fixture control asserts safeTermSingle MOVES the path rather than that it contains U+200C, and is watched firing.
4. The header's 152-sites/60-functions figures are dated rather than restated; the harness reports 199 in 72.
5. Base moved f0cb748 -> deacd19; merged and the full gate re-run on that tree.
```

**Ledger:** `round 3 · payload lines changed THIS round: 0 · scaffolding: 12+/5- generate_test.go, 38+/9- safeterm_coverage_test.go · rc 0, silent stderr, non-empty range`. Second consecutive zero — gate fires, ladder ends.

PR #582, opened 2026-09-12
fix(safeterm): key the bare-identifier ledger per function, not per argument name

ZacxDev:
Closes **R5** of #575 — the residual found while fixing R1 and filed there with a closing condition.

`bareIdentArgs` was looked up as `bareIdentArgs[s.arg]` — keyed by the bare argument **name across the whole of `internal/cmd`** — so one row vouched for that name at every bare-identifier `safeTerm` site in the package.

## A name is not an identity, and the tree proves it

Measured over the 35 sites: **nine argument names span more than one function, covering 27 of them.**

| name | functions |
|---|---|
| `target` | 5 |
| `w`, `workflowID` | 4 each |
| `status`, `name` | 3 each |
| `loraLabel`, `k`, `r`, `note` | 2 each |

`w` alone is **four distinct values** — a mixed-type warning, the same warning at a second site, a base-model warning, and an image-metadata **weight**. `k` is a SERVER cost-map key in `printCostMap` and a key out of the **user's own `--input` file** in `joinQuoted`; the old row admitted both in one sentence, which is the defect written down rather than fixed.

## Matrix

A bare `safeTerm(w)` planted over the **user-typed prompt** in `printGenerateQuote`:

- **green at `3457c5d`** — survives; the harness is blind
- **red here** — killed and named at `generate.go:1807:30`

17 rows become 35, each a claim about one site.

## Both guards were fixed BY their own mutations, not shipped on the first draft

- `TestBareIdentLedgerIsKeyedPerFunction` drives `classifyBareIdents` with synthetic sites. 🔴 Its first draft spelled the synthetic ledger `"alpha::w"` — in the key format under test — so reverting the keying matched nothing, every case was refused, and the mutant **died on the positive control while the case meant to detect it passed for the wrong reason**. The ledger is now built through `bareIdentKey`, so the discriminating property survives any keying: alpha's own `w` accepted, beta's refused. The mutant now dies on that case alone.
- `TestBareIdentLedgerCoversAMultiFunctionName` pins the **premise**, so if no name is ever shared again the extra rows stop looking arbitrary and someone is told to say so.

🔴 **And the first draft left the rule in two places.** Extracting `classifyBareIdents` while the live test kept its own inline copy meant the synthetic guard exercised a duplicate — the "two copies agree" trap. Caught because a mutation's anchor matched twice. The live test now **calls** the shared rule, and the `sanitizerComposers` exemption moved with it: leaving it at the call site made the two disagree about exactly one site (`safeTermErr`), which is the failure a shared helper exists to prevent.

## Gate

`make ci` green; `make lint` 0 issues. No open PR touches `internal/cmd/` — #572 merged, which is what unblocked this (R5 was deferred from #578 precisely because renaming would have collided with it).

Refs #575


ZacxDev:
## 🔴 Correcting this PR's own evidence — round 0

Round 0 could not fault the re-keying, and it attacked the right thing: **R5 was
written and implemented by the same person**, which is the "requirements from
smart people are the most dangerous" case in its purest form. It found four false
sentences of mine and one deletion. Posting rather than editing the body, since
anyone who already read it read the wrong version.

### The headline demonstration was true but badly chosen

The body says a bare `safeTerm(w)` planted in `printGenerateQuote` is *"green at
`3457c5d` — survives; the harness is blind."*

That is **literally true** — I re-measured it — but the plant sits inside an
`if false` block, and `printGenerateQuote` is the **most**-guarded function in this
package for exactly this hazard: `TestGenerateDryRun_DoesNotEchoTheUncheckedPrompt`,
`TestGoldenSpendCopy` and the tabwriter ledger all watch it. The `if false` dodged
their runtime reach. A reader would conclude nothing else catches this, which is
false there.

That is the *"verified in isolation"* trap used in the reassuring direction.

**The sound demonstration, measured both ways** — reachable, at a site with no
dedicated guard of its own:

```
w := d.Name; … safeTerm(w)   in printAppDetail (apps.go)
                              — `w` is ledgered in four OTHER functions

base 3457c5d   GREEN across the whole package
PR head        KILLED and NAMED at apps.go:372:40
```

### "35 sites" was a row count

Re-measured: **35 rows govern 41 bare-identifier call sites.** Six rows cover two
sites each (`downloadOne::target`, `presentTargetSatisfies::target`,
`printReattach::workflowID`, `printSubmitted::workflowID`,
`waitAndCollect::workflowID`, `writePart::partPath`).

So *"each a claim about one site"* is false for six of thirty-five — and that
sentence is the one justifying the maintenance cost. A row is a claim about one
**(function, name) pair**. The multi-function table in the body checks out exactly:
3+3+4+4+2+2+5+2+2 = 27, of 35 rows.

### The composer exemption covers a different function than I said

The body says leaving it at the call site *"made the two disagree about exactly one
site (`safeTermErr`)"*. Measured: it is **`safeTermSingle::s`** at
`safeterm.go:122:14`. `safeTermErr`'s only call passes `err.Error()` — a
`CallExpr`, `bareIdent=false` — so the bare-ident exemption can never reach it. The
requirement is sound; my evidence sentence pointed at the wrong function.

### Two comments would have shipped describing a closed hazard as open

- 🔴 `internal/cmd/generate.go` said the ledger is keyed by argument **name**
  package-wide, that five one-letter keys each allowlist their name at every site,
  and that the residual is *"Recorded as R5 … so the next reader finds an open
  residual"*. Every clause is false once this lands. It is **#578 round-1's own F1
  in the opposite direction** — that one claimed a residual was recorded when it
  was not; this one would claim one is open when it is shut.
- `sanitizerComposers`' doc justified itself by saying an argument-name allowlist
  *"blinds this harness in all ~67 files at once"*. Gone — a row now blinds exactly
  one site. The **conclusion** stands (a file-wide exemption would pass every future
  function added to `safeterm.go` without anyone naming it), so the paragraph keeps
  its point and loses its dead mechanism.

Both retracted in `b5b63e5`.

### Deleted: `TestBareIdentLedgerCoversAMultiFunctionName`

It failed when **no** argument name was shared — an invariant guard. Losing every
shared name is not a defect, the auditor could construct no mutant it kills alone,
and both its arms were already covered (the `::` arm by the SHRANK loop, the rest
by `TestBareIdentLedgerIsKeyedPerFunction`). Its numbers were worth printing, which
is a different thing from worth asserting, so the table survives as `t.Logf`.

### Round 0 also settled the choice R5 left open

R5 offered two closing conditions as equals: rename the five one-letter keys, or
re-key by function. **Round 0 measured that option 1 was the defective half of my
own requirement.** Only three of the five one-letter keys are shared at all
(`w`, `k`, `r`); `t` and `h` are not shared, so two of the five renames buy
nothing — while `target` (5 functions), `workflowID` (4), `status` (3), `name` (3),
`note` and `loraLabel` stay under one package-wide sentence. Option 1 closes 3 of
9 collisions and leaves **19 of the 27 collision-covered rows** open while looking
finished.

### One residual, deliberately not fixed here

**Nobody resolves a row against reality.** A row is prose written by whoever adds
the site — structurally the same shape as this issue's own R1. Re-keying narrows a
wrong row's blast radius from the package to one function, which is a real gain,
but no row is checkable by anything. That should go on #575 when R5 closes rather
than shipping silently, and I'll record it there.

### Gate

`make ci` green, `make lint` **0 issues** — I ran lint myself, which the auditor
could not: `golangci-lint` was not on PATH in its environment, so its report marks
that claim unchecked rather than counting it. Worth noting because it is the one
number in this PR that only one of us has verified.

**Ledger:** `round 0 · requirements: 6 (unattributed: 1) · deletion candidates: 2`.
Round-0 trial pair for this PR: `ran: 1 · changed the outcome: 1`.

ZacxDev:
## Round 1 (nine axes, blind) — findings and fixes

Round 1 re-derived **every** number in this PR with its own independent AST
scanner rather than the repo's: 35 rows, 41 sites, the six two-site rows, the nine
shared names, the `safeterm.go:122:14` exemption site, and both arms of the
`apps.go:372:40` mutation end-to-end. All hold. It also confirmed the re-keying is
strictly *stricter* — no site that was refused can now be accepted — and that the
composer exemption kept its semantics.

Its three findings are all mine, and the first is the one that matters.

### 🔴 F1 — three rows said SERVER about a string carrying the user's own typed bytes

A row that misclassifies origin is precisely the failure this ledger exists to
prevent, and I wrote three while re-keying it. Verified first-hand:

| row | what it actually carries |
|---|---|
| `reportBaseModel::w` | `baseModelWarning(have, want)` interpolates `want` = `o.forBase` — the user's `--for-base` |
| `printDownloadPlan::note` | `routeDir` (`layout.go:145`) interpolates `root` = the user's `--root` |
| `downloadSelected::note` | the same note at its other call site |

The second is the worse one: **the line whose entire job is to tell the user where
the file is going prints a directory that is not the directory.** All three are now
`MIXED`, the label the five `::target` rows already carry.

### 🔴 F2 — `writePart::name` is a shape this ledger's key cannot express

`writePart` has **two callers with two origins**: `download.go:926` passes the
server's `f.Name`; `generate_output.go:427` passes `filepath.Base(target)` off
`planOutputTarget(o.outName, …)` — the user's `--out-name` template.

A `(function, name)` key cannot say *"two callers, two origins"*. So the row is
`MIXED` by necessity rather than by measurement — and this PR's own line, *"a row
is a claim about one (function, name) pair"*, has a shape underneath it the key
format does not reach. **Recorded on #575** rather than papered over.

Two `see targetPath` pointers went with it: on the generate-output path the target
comes from `planOutputTarget`, not `targetPath`.

### 🔴 F3 — I shipped, in two comments I edited, the exact sentence the same commit's message retracts

`b5b63e5` corrected *"each a claim about one site"* — and left:

- *"A row now blinds **exactly one site**"* live at `safeterm_userinput_test.go:126`,
  **contradicted 198 lines below in the same file by the same commit**; and
- *"a short name allowlists **exactly one site**"* at `generate.go:1880`, which is
  true of today's tree only by accident (all ten one-letter-keyed rows happen to
  cover one site each) and is asserted by nothing.

Both now say one **(function, name) pair**. Swept for the *shape* rather than the
two reported sites — the only remaining hits are the retractions quoting
themselves, confirmed with a positive control.

### Base and collisions

`origin/main` did not move. **#583 has opened since this PR's body claimed no open
PR touches `internal/cmd/`** — that claim is stale. `merge-tree` against it exits
0 (branched on the exit code, not a marker grep) and the merged suite is green.

### Two things round 1 could not verify, reported rather than counted

- **The depth-1 (shallow) tier.** `scripts/ci-shallow.sh` needs a named branch ref,
  and writing one lands in the shared common git dir, which its brief forbade. It
  checked what that tier discriminates on — `agents_split_preserved_test.go`, whose
  input is `AGENTS.md`, untouched here — so the risk is low, but it is unverified.
- **CI's pinned `golangci-lint v2.12.2`.** Both of us ran a near-neighbour (2.13.2).

### Pre-existing, not attributed to this PR

`minSafeTermCallsScanned`'s comment says *"There are 153 calls today"*; the live
count is **199**. Byte-identical at `3457c5d`. Left alone deliberately — it is not
this PR's to fix, and silently correcting it would blur what this diff did.

```audit-claims round=1 audited=b5b63e5..05b4f6b
1. F1: reportBaseModel::w, printDownloadPlan::note and downloadSelected::note are relabelled MIXED, each naming the user-typed flag it carries.
2. F2: writePart::name is relabelled MIXED and records that a (function, name) key cannot express two callers with two origins; recorded on #575.
3. F2: the two "see targetPath" pointers now name planOutputTarget for the generate-output path.
4. F3: both "exactly one site" sentences now say one (function, name) pair, and the shape was swept rather than the two reported sites.
5. The stale "no open PR touches internal/cmd" claim is corrected; #583 test-merged, merge-tree rc 0, suite green.
```

**Ledger:** `round 1 · payload attribution: not measured (first, full audit — no previous round to attribute against; starts at round 2)`. Both files changed this round are scaffolding or comments; `generate.go`'s hunk is comment-only.

ZacxDev:
## Round 2 (delta, `b5b63e5..05b4f6b`, blind) — findings, fixes, and the ladder's stop

Round 2 reproduced all three of round 1's MEASURED claims under its own probes,
re-derived every count independently (35 rows, 41 sites, six two-site rows, 29
one-site, ten one-letter rows, even the 201-line distance between a claim and its
retraction), and built the **real integration tree** — base + #582 + #583 — where
the suite, vet, gofmt and lint are all green.

Four findings. All four are mine.

### 🔴 F2 — I relabelled a row MIXED and then described its second origin wrong, the reassuring way

The `writePart::name` row said `generate_output.go:427` passes *"the USER's
`--out-name` template"*. It does not. `renderOutName` expands the **default**
template `{workflow}-{n}{ext}` with `{workflow}` = the **server's** workflow id and
`{ext}` = a bounded URL extension — so with no `--out-name` the whole basename is
the server's. User text enters only from the literal parts of an explicit flag.
Verified first-hand at `generate_output.go:79-103`.

**That direction is the dangerous one.** A value labelled user-typed is one the gate
must *not* rewrite (#393) — so the row read as cover for the ungated siblings on
that same path. Fixing an origin misclassification and introducing another one in
the same commit is the defect this PR's own subject line names, twice over.

### 🔴 F4 — I established the two-caller fact and failed to propagate it

Two **payload** comments still asserted the single origin round 1 had disproved:

- `download.go:1007` — *"`name` is the SERVER's `files[].name`"*
- `download.go:1179` — *"threaded here from `downloadOne` via `writePart`'s `name` parameter"*

The second is the `\r`-rewritten progress surface — the one its own comment calls
the highest-value. Both now name both callers, and I swept for the *shape*: the
only remaining hits are the two corrected sites.

This is the "sweep the shape, not the reported sites" rule, failed by the round
that had just quoted it.

### 🟡 F3 — an enforcement claim wider than the harness

`generate.go` said a row being a claim about one function's value *"is now enforced
rather than requested"*. `classifyBareIdents` enforces that a site **has** a key —
never that the key resolves to one origin. `writePart::name` is the live
counter-example, written by the same commit 1800 lines away. Retracted with what is
actually enforced.

### 🔴 F1 — "Recorded on #575" was false for the third time in this arc

#578 round 1 caught it. #582 round 0 caught the inverse. I wrote it again.

Now posted as **R6** on #575 with the measurement and a closing condition — and
**verified after posting** rather than asserted: #575 has 2 comments, both matching
`writePart`. R6 also records, as pre-existing and explicitly **not** attributable to
this PR, the ungated workflow-id basename on the generate-output path
(`blobStatusError`, `downloadBlobTo:407/418/430`, `downloadOutputs:487`) — #566-class
sibling splits that deserve their own issue.

---

## 🔴 The ladder stops here, on the attribution gate

Round 2 was **not** clean, so the findings-keyed rule would continue. The measured
gate ends it instead:

```
round 1's fixes (b5b63e5..05b4f6b)   4+/1-   generate.go              comment-only
                                    10+/7-   safeterm_userinput_test  strings/comments
                                             → 0 executable lines

round 2's fixes (05b4f6b..2884dc3)  21+/5-   download.go              comment-only
                                     9+/1-   generate.go              comment-only
                                     1+/1-   safeterm_userinput_test  one string
                                             → 0 executable lines
```

**Two consecutive rounds whose fixes changed zero payload lines ⇒ the ladder has
left the PR.** rc 0, non-empty ranges, and the line filter carries a **positive
control**: the same filter counts **101** over `3457c5d..dc790f5`, a range that does
contain executable change. A bare zero from an uncontrolled filter is not a reading —
and the first attempt at this measurement produced exactly that, because `zsh` does
not word-split an unquoted variable and both commands ran with an empty range.

Nothing is deferred: all four of round 2's findings are fixed.

### Ladder summary

| round | outcome |
|---|---|
| 0 | R5's own closing condition offered two options as equals; option 1 was measurably the defective half. Plus two comments that would have described a closed hazard as open |
| 1 | **Three rows said SERVER about strings carrying the user's typed bytes.** One key format cannot express two callers with two origins |
| 2 | The *corrected* row was still wrong about origin, the reassuring way. The two-caller fact was established and not propagated |

Every round found something real, and **every finding across all three was in my own
rows or prose** — which is the honest summary of what a ledger of hand-written
classifications costs, and is itself the strongest argument for the residual R6 now
records.

```audit-claims round=2 audited=05b4f6b..2884dc3
1. F2: writePart::name's second origin is corrected to server-derived-by-default, naming renderOutName's {workflow} expansion and the default template.
2. F4: download.go:1007 and :1179 now name both callers instead of asserting files[].name only; the shape was swept.
3. F3: generate.go's "now enforced" claim is retracted to what classifyBareIdents actually enforces.
4. F1: the residual is posted as R6 on #575 with a closing condition, and the pointer was verified after posting.
```

**Ledger:** `round 2 · payload lines changed THIS round: 0 executable · scaffolding and comments only · rc 0, non-empty range, filter positive-controlled at 101`.

PR #583, opened 2026-09-13
test(saferune): ledger what every saferune.* call site is GIVEN, not just who imports it

ZacxDev:
Discharges the **second clause** of issue #542's stated condition, which the issue's own closure left open.

## The gap

#542 asked for a guard that fails when "`snippet()` **or any `saferune.*` call site outside `internal/cmd`**" gains a caller whose argument is not server-supplied, **or when the caller set changes at all, bidirectionally**.

#557 shipped `pkg/civitai/snippet_args_ledger_test.go` — the right guard for the first clause, keyed per enclosing function, bidirectional, with positive controls. It enumerates `snippet(` call sites **in `pkg/civitai`** and nothing else. The issue was closed on it. Nothing covered the words "or any `saferune.*` call site outside `internal/cmd`".

Measured on `3c84b0f`, `saferune.*` appears at **three** places in the module's non-test sources:

| package | site | enclosing |
|---|---|---|
| `internal/cmd` | `saferune.Strip` | `safeTerm` |
| `internal/genapi` | `saferune.HasVisibleContent` | `hasPrintableContent` |
| `pkg/civitai` | `saferune.Strip` | `snippet` |

`snippetArgs` sees one of them. The module-root `saferune_callers_ledger_test.go` sees all three *packages* and states in its own comment that it answers "who calls", never "with what" — so a package already in it can grow a second, unrelated `saferune.Strip(userTypedFlagValue)` without the importer set moving.

**A closed issue is not a discharged condition**, and that is the durable part here: the gap sat behind a green suite, a lint run and a closed issue for a whole arc.

## What lands

**`saferune_arg_origins_test.go`** (module root) — `TestSaferuneReferenceArgumentsAreLedgered`. Every saferune reference in non-test sources needs a row keyed `<package>:<enclosing func>:<rendered reference>`. Red on an unledgered site, on a row with no site, and on a row classified as the user's own bytes.

**All three sites are `originDelegated`, and that is the finding rather than a shortcut.** Every one is a one-line wrapper whose argument is the wrapper's own parameter, so the bytes' origin is not decidable at the `saferune` call site *at all* — it is decided one level out. A ledger that only asked "are these server bytes?" would answer "cannot tell here" three times out of three, which is exactly how a guard ends up reading as coverage while providing none. So a delegated row names the `Test` that answers the question at the wrapper's own call sites, and **that name is resolved to a real declaration before it is believed** — the move `checkQuestionsResolve` already makes next door.

- `internal/cmd:safeTerm` → `TestSafeTermIsNeverAppliedToUserTypedInput`. This site **must not** be classified server-or-user here: `internal/cmd` deliberately routes two non-server values through `safeTerm` (`--input` file content, `download`'s mixed-origin target path), both named in `saferune`'s package doc as documented exceptions.
- `pkg/civitai:snippet` → `TestSnippetArgumentsAreAllServerBytes` — #557's ledger, now reachable from the module root rather than being a fact about one package's test directory.
- `internal/genapi:hasPrintableContent` → `TestHasPrintableContentArgumentsAreServerBytes`, **which did not exist**. That is the hole.

**`internal/genapi/saferune_origins_test.go`** fills it, and pins **two hops** — `hasPrintableContent`'s only caller is `dedupeReasons`, and `dedupeReasons`' callers are where the wire fields are actually read. Pinning only the first hop would be a description ("the argument is server bytes") wider than its body ("the argument is whatever `dedupeReasons` passes").

🔴 **The second hop found a live stale claim on its first run.** `dedupeReasons`' doc comment read *"The four callers are the two step types' `failureReasons` and the two workflow-level `FailureReasons`."* There is **one** step-type `failureReasons` and there are **three** callers; no `ListedStep.failureReasons` has ever existed — `ListedWorkflow.FailureReasons` reads `steps[].errors` inline. Corrected, and the set is now a **ledger, not a count**: a count of four would have been satisfied by any four functions at all.

**Incidental, and it was the consolidation that found it:** package `cli_test` held the module walk **twice** under two names, byte-identical in body (`neterr_ledger_test.go`'s `moduleGoFiles`, `saferune_callers_ledger_test.go`'s `goSourceFiles`). Adding a third is what made them visible. One copy now.

## Structural, not spelled

The scan asks "can it pass while the hazard exists in a different shape?":

- the saferune import's **local name is resolved**, so `import sr ".../saferune"` is caught (M2);
- a **dot import** is a hard refusal — it would make every reference unqualified and the scan blind;
- a **bare, non-call reference** (`f := saferune.Strip`) is reported separately rather than ignored, because it escapes the argument question entirely (M3).

## Mutation matrix — 13 mutants, each killed by the guard's own message

| # | mutant | result |
|---|---|---|
| M1 | new `saferune.Strip` call site, plain import | GREW, names the site + key |
| M2 | same, behind an **aliased** import | GREW — the spelled-guard control |
| M3 | bare non-call reference | GREW, flagged `bare reference` |
| M4 | the genapi ledger row deleted | GREW on the existing site |
| M5 | row for a site that does not exist | SHRANK |
| M6 | a row flipped to `originUserTyped` | the #393 arm |
| M7 | the delegated `Test` renamed | `pinnedBy` does not resolve |
| M8 | new `hasPrintableContent` call site | genapi GREW |
| M9 | a fourth `dedupeReasons` caller | genapi GREW |
| M10 | a `dedupeReasons` caller removed | **CONTROL floor**, not the SHRANK arm |
| M11 | import-path typo | CONTROL: `0 refs, want >= 3` |
| M12 | enclosing func of the `pkg/civitai` site renamed | GREW **and** SHRANK together |
| M13 | `dedupeReasons` renamed, ledger held fixed | GREW **and** SHRANK together |
| M14 | a ledgered caller renamed | GREW **and** SHRANK together |

Two things worth stating rather than hiding:

- **M10 dies on the positive-control floor, not on the arm it aimed at**, because the floor is set at the measured count — so removing a caller trips the floor first. M12/M13/M14 are what prove the SHRANK arms reachable on a realistic edit; without them the stale arms would be breakable-but-unreached.
- 🔴 **M13's first run SURVIVED, and the harness was the defect.** The `sed` renamed the ledger's own keys along with the sources, so the mutation was a self-consistent rename and correctly passed. Re-run with the ledger held fixed, plus a spelled control asserting the old key is still in the file, it dies on both arms.

The one-sided registration was also watched to fail rather than asserted: adding `saferune_arg_origins_test.go` to `item38CommentedFiles` **without** the decision header reddened `TestItem38FileLedgerMatchesTheDecisionHeader` by name, and the missing `TestHasPrintableContentArgumentsAreServerBytes` reddened the root guard's `pinnedBy` resolver before that guard was written.

## Gate

`make ci` green — 21 packages `ok`. `make lint` — `0 issues`. Branch cut from `origin/main` at `3c84b0f`.

## Scope note

Issue #542 is already CLOSED. This PR uses no closing keyword anywhere, on purpose — the evidence goes on the issue by hand. `internal/genapi/saferune_origins_test.go` is deliberately **not** added to `item38CommentedFiles`: that ledger is compared element-for-element against the files decision 38's own header declares, and a `genapi` guard is not item 38's scope.

Refs #542, #575.


ZacxDev:
## Audit round 0 + round 1 (nine axes), dispatched in parallel — and a correction to this PR's own body

**Correcting the description publicly rather than silently editing it**, because it has been readable since the PR opened:

- The matrix is headed **"13 mutants"** and lists **M1–M14**. Fourteen. A PR whose thesis is *"a ledger, not a count"* shipped a count that was wrong — noted rather than quietly patched.
- The scope note says *"`internal/genapi/saferune_origins_test.go` is deliberately not added to `item38CommentedFiles`: a `genapi` guard is not item 38's scope"* — while §6 of decision 38 describes that guard at length and registers the root guard (which covers `genapi`) under item 38. Round 0 called that out as the boundary being asserted both ways. Resolved in `eac7fdc` by writing the rule down in §6: **item 38 owns the pointer; the item whose seam the target guards owns the target.**

### What the audits found, and what `eac7fdc` does about it

Round 1 demonstrated **two SURVIVED mutants against relationships these guards claim to pin**. I reproduced both before fixing — an auditor's mutation result is a claim, not evidence.

**A. The genapi walk iterated `FuncDecl`s only.** Adding

```go
var untrustedReasons = dedupeReasons([]string{"from a flag the user typed"})
var reasonsFor = func(raw []string) []string { return dedupeReasons(raw) }
```

left `go test ./...` **fully green**, with `TestDedupeReasonsCallersAreLedgered` still logging `scanned 3 dedupeReasons(...) call site(s)`. Two unledgered callers of the exact function whose caller set it exists to hold — falsifying the row's own *"every caller of `dedupeReasons` is ledgered below"*, which is what the module-root ledger's whole delegation chain rests on. The `sites` floor could not see it either, because floor and walk came from the **same restricted traversal**: a positive control that shares the step it is checking.

Fixed by attributing package-level declarations, plus a **totality control** — a second traversal of the whole file node, built differently, whose *disagreement* with the attributing walk is the failure.

🔴 **The first fix for this was itself wrong, and the mutant caught it.** Attributing every file-level call to one literal `<file-level>` key collapsed both planted callers onto one row — one note vouching for two sites, the precise defect the per-function keying exists to prevent, reintroduced one level down while fixing something else. Each `ValueSpec` is now its own owner.

**B. `pinnedBy` resolved a name, not a relationship.** Gutting `TestHasPrintableContentArgumentsAreServerBytes`' body to `_ = …` left the suite green while the root ledger kept asserting `internal/genapi`'s site was covered. `moduleTestDecls` indexed every `func Test*` **module-wide**, so a stub of the right name in any package satisfied it.

That is the state **`3457c5d` (#578, *"delete the ledger gate state that resolved a name, not a relationship"*) removed from this repo one commit before this branch was cut** — reintroduced in the file that cites it. Both audit rounds landed on it independently. The neighbour it claimed to copy, `checkQuestionsResolve`, resolves inside the ledgered `pkgPath` and is strictly stronger; that claim was false and is corrected.

🔴 **Residual, stated rather than papered over:** package scoping kills the wrong-package stub, **not** a gutted test in the right package — and no static scan can. `pinnedBy` is not evidence the delegated guard is effective.

**C. The import resolver kept only the last matching local name.** A file may bind one import twice under two names; `gofmt` and `golangci-lint` both accept it. A planted `sr.Strip(userFlagValue)` beside a `saferune.` alias was never seen. Collected into a set now — the file's own 🔴 claimed it handled "whatever that file chose", and now it does.

**D. `skippedDirs` was missing `.claude` — pre-existing, and the consolidation is what surfaced it.** `agents_xrefs_test.go`'s `skippedDirs` is `{.git, node_modules, dist, bin}`; `agents_evidence_test.go`'s `skipDirsForSourceScan`, same package and same question, has always included `.claude`. Measured: `.claude/worktrees/` holds **20 agent worktrees, 9,012 `.go` and 1,495 `.md` files**, and the walk read all of them. `TestAgentsItemCrossReferencesResolve` validated **17,658** item references before, **900** after — 95% of its corpus was *other branches' checkouts*, so its verdict was a function of what happened to be checked out at suite time. It passed only because no worktree referenced a number above AGENTS.md's highest.

Fixed here rather than filed because it is one map entry and this PR's own new comment diagnoses it word for word. 🔴 And the first draft of that comment **failed the guard**: it spelled the next free item number, and the scan reads `.go` files for `item N` without distinguishing a comment from a citation.

**E. Three claims this PR made that were wrong.** `~150 safeTerm call sites` (the guard's own log says 199 — number dropped, not corrected: nothing asserts on it); `#582 fixed the bare-identifier ledger` (#582 is **OPEN**, `mergedAt: null` — past tense would cite a fix that never landed); and *"the body is five decisions"* in **both** decision 38 and `agents_split_preserved_test.go`, which §6 makes six. That last one is §5's own subject — *two prose statements of one set, both stale at once* — happening to the file that records it.

**Deleted:** `originServer`. Dead, unmutated, and an evidence-free escape hatch: its arm asked for no `why` and no `pinnedBy`, so a future direct call could self-certify.

### Matrix after the fixes

All **11 original mutants still die unchanged**. Five new:

| # | mutant | result |
|---|---|---|
| N1 | package-level + func-literal callers | both caught, **distinct** keys, 3 → 5 |
| N2 | attributing walk narrowed | totality control fires: 4 vs 3 |
| N3 | two aliases on one import | both refs caught (was: invisible) |
| N4 | `pinnedBy` → right name, wrong package | refused, naming the package |
| N5 | `skippedDirs` without `.claude` | 17,658 refs vs 900 |

`make ci` green (21 packages ok); `make lint` 0 issues.

### Round-0 items deliberately NOT taken

Filed here rather than fixed, so they read as open rather than absent: the unreachable `IndexExpr`/`StarExpr`/`ArrayType` arms in both `renderExpr` copies (they fail closed via the `default:` arm, so deleting them is churn without a safety gain), and consolidating the **four** near-identical AST expression renderers across `internal/cmd`, `pkg/civitai` and these two files — which needs a non-test helper package and is its own change, not a line in this one.

ZacxDev:
Round 1's claims block, for the next round to read. Each line is WHAT the fixes in `eac7fdc` claim to have addressed — not why it is correct.

```audit-claims round=1 audited=bff2100..eac7fdc1
1. checkGenapiCallLedger now attributes call sites in package-level declarations and package-scope func literals, not only those inside a FuncDecl; the two planted dedupeReasons callers that previously SURVIVED are now reported.
2. Each package-level declaration is its own ledger key, named after the variable it initialises, so two file-level callers cannot share one row.
3. A totality control was added: a second traversal of the whole file node, and its disagreement with the attributing walk is a CONTROL failure.
4. isGenapiCallTo is one shared spelling of the call match, used by both walks.
5. pinnedBy now resolves against Test declarations in the site's OWN package directory, not module-wide; moduleTestDecls became moduleTestDeclsByPackage.
6. The claim that pinnedBy resolution is "the same move checkQuestionsResolve makes next door" was removed as false, and the residual — a gutted test in the right package still satisfies pinnedBy, and no static scan can see it — is now stated in source.
7. The saferune import scan collects EVERY local name the file binds to internal/saferune, not only the last matching import; a second alias is no longer invisible.
8. The originServer kind and its switch arm were deleted.
9. agents_xrefs_test.go's skippedDirs now includes ".claude"; TestAgentsItemCrossReferencesResolve's corpus is claimed to drop from 17,658 references to 900.
10. The "~150 safeTerm call sites" figure was removed rather than corrected, on the stated ground that nothing asserts on it.
11. The two civitai/cli#582 citations were changed from past tense to present, on the stated ground that #582 is OPEN with mergedAt null.
12. Both "five decisions" geometry sentences (decisions/38 and agents_split_preserved_test.go) now read six decisions, and decisions/38's "three tables" now reads four.
13. decisions/38 §6 states an ownership rule for the item-38 boundary: item 38 owns the pointer, the item whose seam the target guards owns the target.
14. The PR body's "13 mutants" over a 14-row table was corrected in a PR comment rather than by a silent body edit.
15. All 11 original mutants are claimed to still die unchanged, and five new mutants (N1–N5) are claimed to die.
16. make ci is claimed green at 21 packages and make lint at 0 issues, full-clone tier.
```

ZacxDev:
Round 2's claims block, for the next round to read. Each line is WHAT the fixes in `c351022` claim to have addressed — not why it is correct.

Note for the next round: the branch was **rebased onto `095f4ac`** (#582's merge) between round 2's audit and these fixes. `11ebd82` is the rebased twin of `eac7fdc`, the tip round 2 read — verified to differ by exactly #582's three files and nothing else — so `11ebd82..c3510225` is this round's fixes with no bring-in noise.

```audit-claims round=2 audited=11ebd82..c3510225
1. genapiOwners now pairs Names[i] with Values[i] when the counts match, so a multi-name ValueSpec gives each call site its own owner key instead of keying all of them on Names[0].
2. When Names and Values counts differ (one multi-return call initialising several variables) a single owner is used, named for the first variable.
3. The blank identifier is refused STRUCTURALLY: genapiOwner carries a `blank` flag, blank-owned sites are collected in their own bucket, and they are reported before the ledger is consulted so no row can reach them.
4. The previous blank-identifier fix, which encoded the refusal in the KEY's spelling, is claimed to have been a spelled guard that a row spelling that string defeated; it was replaced, not extended.
5. The module-root guard received the same owner split via saferuneOwners, replacing the single literal "<file-level>" it used for every non-FuncDecl declaration.
6. The module-root guard received the same structural blank refusal, via an `unowned` flag on saferuneRef checked before the ledger.
7. A per-package emptiness control was added to the pinnedBy resolver: every package a row delegates into must have resolved at least one Test declaration, because the module-wide total cannot control a per-package lookup.
8. saferuneRenderExpr now renders *ast.BasicLit by value, so two calls with different literals cannot share a key.
9. The comment claiming an unrecognised shape "cannot collide with a ledgered key" was removed as false, and the residual — an unknown shape still renders by TYPE, so two of the same unknown type in one owner still share a key — is stated in its place.
10. The agents_xrefs_test.go comment that sourced the do-not-spell-the-next-item-number rule to agents_evidence_test.go is claimed to have been wrong (zero occurrences there) and now points at parseAgentsItems instead.
11. The two civitai/cli#582 citations: the module-root one is back to past tense on the ground that #582 merged as 095f4ac, and the comment now says not to restate a sibling PR's lifecycle in source at all. The internal/genapi one was NOT changed in either round and is claimed to be byte-identical to base.
12. Round 1's claim that BOTH #582 citations were changed is claimed to have been false — only one was.
13. The "no count is quoted here on purpose" sentence had both of its counts (~150 and 199) removed.
14. The branch was rebased onto current main (095f4ac) and the gate re-run there, so the claim is about the merged tree and not a pre-merge base.
15. All 11 mutants of the original battery are claimed to still die unchanged, and three new controls (P1 blank-owned refs refused and distinct, P2 absolute-dir index re-key producing a CONTROL failure, P3 two literals in one owner keyed distinctly) are claimed to fire.
16. make ci is claimed green at 21 packages and make lint at 0 issues, full-clone tier, on the rebased tree.
```

Issue #574, opened 2026-09-12
generate's blob download path prints the server-derived workflow id raw in 6 error strings, while its own Saved line is gated


ZacxDev:
Found by the round-1 adversarial audit of cli#572 (which hardens the sibling `download` path). **Pre-existing — not introduced by #572.**

`internal/cmd/generate_output.go` has the same renderer-sanitizes-while-its-twin-does-not split that #566 item 2 was about, one command over. #572 gates `downloadBlobTo`'s **`Saved`** line (`:434`, `safeTerm(target)`) — so the author already treated `target` as untrusted — while every error string on the same function's failure paths prints it raw.

## The text is genuinely server-derived

`renderOutName`'s own doc comment states it: *"🔴 BOTH INPUTS ARE UNTRUSTED. The workflow id is SERVER-supplied…"*. `{workflow}` expands to `filepath.Base(strings.TrimSpace(workflowID))`, `outputTarget` joins that into `target`, and `downloadBlobTo` derives `name := filepath.Base(target)`. So the server's workflow id reaches the leaf of both.

`cmd/civitai/main.go:57` prints `err.Error()` unfiltered.

## Sites

| line | string | carries |
|---|---|---|
| `:408` | `refusing to overwrite the existing file %s` | `target` |
| `:419` | `download %s: %w` | `name` |
| `:433` | `install %s: %w` | `target` |
| `:388`, `:390`, `:392` | `blobStatusError`'s three arms | `name` |

`blobStatusError` (`:382`) is the **structural twin of `downloadStatusError`** — same signature, same `defer civitai.TagStatus(status, err)`, same arms. #572 gated that one **once at the top** (`name = safeTerm(name)`) with a comment recording that four spellings of one rule drift apart. This twin was not gated.

## Deliberately NOT in scope

`:413` `create output directory %s: %w` is **correctly ungated**, for the same reason #572 *removed* its own copy of that gate: `dir` is `filepath.Dir(target)`, which is `outputTarget`'s `outDir` — user-supplied via `--out-dir`, never the server's leaf, because `outputTarget` refuses any `name` where `name != filepath.Base(name)`. Sanitizing it would strip user-typed bytes, which `internal/saferune`'s rule forbids. Do not "fix" it.

## Known residual, stated rather than hidden

`saferune` deliberately retains `\n` and `\t`, so routing these through `safeTerm` does not close newline line-forgery on this path either. That class is #552's, and is not part of the condition below.

## Closing condition

[...]

PR #596, opened 2026-09-14
fix(generate): gate the blob download path, which printed the workflow id raw

ZacxDev:
Closing condition of #574, the last of the three ranked items. This is the same
class as #566/#572, one command over, on the path that spends the user's money
irreversibly.

`downloadBlobTo`'s `name` is filepath.Base(target), and `target` comes from
planOutputTarget whose `{workflowId}` is read off the wire — renderOutName's own
doc says so. Six error strings printed it raw while the SAME function's success
line already read safeTerm(target): one value, two treatments, decided by
whether the download worked.

WHAT MOVED

- blobStatusError gated ONCE at the top, not at each of its arms. It is the
  structural twin of download.go's downloadStatusError — same signature, same
  defer civitai.TagStatus, same arms — and #572 gated that one exactly this way
  because #566 exists precisely because two spellings of one rule drifted.
- The overwrite refusal, both `%s: %w` pairs, and the Saved line.
- safeTermSingle, not safeTerm: saferune keeps \n and \t by design, so safeTerm
  alone leaves the server able to forge a line (#577, measured one command over
  as a fake `(SHA256 verified)` on a transfer that had not happened).
- BOTH HALVES of each `%s: %w`. #577's first pass gated the operands and left
  safeTermErr on safeTerm, so the wrapped cause still forged the line; #590
  repaired it. This path takes the repaired version from the start.

WHAT DELIBERATELY DID NOT MOVE

`create output directory %s: %w` stays UNGATED. `dir` is filepath.Dir(target) —
the user's own --out-dir, never the server's leaf, because outputTarget refuses
any name that is not its own filepath.Base. Gating it would strip USER-TYPED
bytes, which saferune's rule forbids (#393). download.go carries the
byte-identical line ungated for the same reason: #572's round 0 REMOVED a gate
someone had added there, after that gate had already survived a 20-mutant sweep,
because breakable is not reachable.

A reader now sees one gated twin and one ungated one and will try to make them
agree. TestCreateOutputDirectoryStaysUngated is the test that says which
direction is wrong.

🔴 AND THAT GUARD WAS VACUOUS ON ITS FIRST DRAFT, FOR THIS PR'S OWN SUBJECT.
It asserted strings.Contains(err, userDir) — which the WRAPPED CAUSE satisfied,
because the message is `create output directory <dir>: mkdir <dir>: …` and `%w`
renders its half raw whatever the `%s` half does. Gating the operand left
Contains true and the mutant SURVIVED. One value, printed twice, asserted on one
half — the same shape as the defect, in the test for it. It now COUNTS: ungated
renders the path twice, gated once.

MUTATION MATRIX — each site reverted alone
  blobStatusError gate removed     FORGERY named, per arm (all four)
  overwrite refusal ungated        FORGERY in downloadBlobTo (refusing to overwrite)
  the WRAPPED CAUSE ungated        FORGERY in downloadBlobTo (download %s: %w)
  the ungated line GATED           #393 REGRESSION: "appears 1 time(s), want 2"

THREE LEDGERS DEMANDED ROWS, WHICH IS THE MACHINERY WORKING
  bareIdentArgs            origins for blobStatusError::name, downloadBlobTo::name
  safeTermCoveredBy        blobStatusError added; downloadBlobTo promoted from
                           notCovered, and the RATCHET lowered 20 -> 19, reading
                           the number off the assertion's own failure rather than
                           computing it
  safeTermErrCallers       generate_output.go added — and this one caught that
                           safeTermErr's doc scoped itself to "download.go's
                           single-line error path", which my second caller
                           falsified. The ledger and the sentence moved together.

⚠ make ci was GREEN while make lint was RED: staticcheck ST1018 on a raw U+200B
in the new fixture. That is the exact rule AGENTS.md records as invisible to
make ci, hit for real. Escaped.

⚠ One unrelated timing test (TestAppStatusDriftLookupGetsADeadlineOfItsOwn)
failed once under 5 concurrent go test processes and passes in isolation, at
base, and on a re-run of the full suite. Load, not this change — stated rather
than silently re-run.

Gate: make ci green (21 packages ok); make lint 0 issues.

Refs #566, #572, #574, #577, #590.


ZacxDev:
Round 1's claims block, for the next round. Each line is WHAT the fixes claim to have addressed — not why it is correct.

A round-0 requirements pass and a first-full nine-axis pass both ran against `2d509d3`; two commits answer them. Round 0's F1 and the nine-axis round's F1 are the same finding, reached independently.

```audit-claims round=1 audited=2d509d3..19455c0c
1. downloadOutputs' batch overwrite refusal now sanitises j.target — it was FULLY RAW (raw ESC, not merely an ungated newline), and it is the reachable branch: downloadOutputs pre-checks every job before any byte moves, so downloadBlobTo's own !force refusal, which #574's table named, is TOCTOU-only.
2. The forgery through that refusal was REPRODUCED before fixing: esc=true, 2 forged newlines, 1 tab, with \x1b[1A\x1b[2K intact.
3. TestDownloadOutputsRefusalCannotForgeALine was added and DRIVES downloadOutputs end to end — plan the target, pre-create it, call the real function. Its first draft rebuilt the message inline with safeTermSingle hardcoded and the ungate mutant SURVIVED; that is claimed fixed.
4. Two more single-line surfaces in the same function are gated: the no-URL error and the duplicate-target hint's trailing workflow id.
5. generate.go's waitAndCollect re-read hint is gated. It fires only when downloadOutputs errors — the exact path this PR hardens — and printed the workflow id raw one line below the sanitised error.
6. gbfAssertOneLine was DELETED as a byte-identical duplicate of download_newline_test.go's assertOneLine in the same package; all callers repointed.
7. Two claims in safetermerr_ledger_test.go that this PR falsified are corrected: the present-tense "every caller is on download.go's single-line error path", and a failure message telling the next contributor to widen a doc comment "which currently scopes itself to download.go" — a sentence this PR had removed. The message now states the ORDER: ledger first, then the sentence.
8. The downloadOutputs coverage row named "the multi-output progress line" in a function with no print statement at all, and omitted both refusals. Rewritten; the row is now COVERED and the ratchet banked 19 -> 18.
9. install %s and the Saved line now have their OWN subtests. Measured before: ungating install's operand alone SURVIVED the whole suite and ungating Saved entirely SURVIVED it, while the test doc and the ledger row both listed them — and Saved had been honestly ledgered notCovered BEFORE this PR.
10. Both of those mutants are claimed to now die naming their own surface.
11. The bareIdentArgs row for downloadBlobTo::name is reclassified from SERVER to MIXED-mostly-SERVER. The earlier row had the two origins swapped — the --out-name template lands in the LEAF, --out-dir in the directory — contradicting writePart::name three lines below.
12. The test doc and the coverage row are narrowed to what is actually driven, rather than the four surfaces they previously listed.
13. Claimed verified clean by the nine-axis round and NOT re-checked by the fix commits: the ungated `create output directory` across 7 attack shapes; the count-2 guard (0 both-gated, 1 operand-only); exit codes and errors.Is/As through safeTermErr; the widened safeTermErr scope; ratchet correctness; AGENTS.md item 13 not violated.
14. make ci is claimed green at 21 packages and make lint at 0 issues.
```

ZacxDev:
Round 2's claims block, for the next round. Each line is WHAT the fixes claim to have addressed — not why it is correct.

Round 2 was a DELTA re-audit of `2d509d3..19455c0c` (round 1's fixes). It returned findings, including one that falsified a round-1 claim: claim 4 said the no-URL error at `generate_output.go:486` had been gated, and it had not — that commit gated a *different function's* line (`reportExcludedOutputs`, `:240`) while the coverage row was simultaneously rewritten to advertise `:486` as covered. Re-verified by hand before acting on it.

Two findings were deliberately NOT fixed, by operator decision, and are listed as claims 12 rather than omitted.

```audit-claims round=2 audited=19455c0c..a3a8ca0
1. generate_output.go's no-URL error is now safeTermSingle; a subtest of TestDownloadOutputsErrorsCannotForgeALine drives it and dies when reverted to safeTerm; its reachability was driven end-to-end through runGenerate with "available":true / "url":null.
2. The downloadOutputs coverage row no longer claims a surface no test reaches: all three surfaces are driven, the row names each one's own operand separately, and the covering test is renamed off ...RefusalCannotForgeALine.
3. The decision to KEEP safeTermSingle(j.target) is recorded at the call site, with the mixed origin, the documented saferune exception, and the --out-dir residual stated as OPEN.
4. bareIdentArgs now records that a selector expression is structurally invisible to it, so its silence about j.target means UNASKED rather than CLASSIFIED.
5. generate.go's re-read hint gate is pinned by TestWaitAndCollectReReadHintCannotForgeALine and dies when reverted to safeTerm; the six sibling sites are deliberately untouched and are filed as civitai/cli#604.
6. reportExcludedOutputs' output-id gate is pinned against the safeTermSingle->safeTerm downgrade, not only against full ungating.
7. The writePart::name row's citations are corrected to generate_output.go:449 and download.go:935.
8. safeterm.go's safeTermErr scope note says SEVEN live call sites, counted at head, with the provenance of the stale "five" traced to civitai/cli#590.
9. The "2 forged newlines" figure is re-measured as ONE at both places that carried it, with the method and the counter's reading written down.
10. The waitAndCollect hint test ships no tab assertion, and says why: ui's lipgloss styles expand \t to spaces before the bytes reach the writer, so such a guard could never fire there. This also implies any tab assertion on a ui-styled surface in this package is inert.
11. waitAndCollect moves notCovered -> covered and maxUncoveredSafeTermFuncs is lowered 18 -> 17 in the same commit, with the row stating which one of its four surfaces is driven.
12. NOT acted on, stated rather than omitted: the six generate.go sibling sites (deferred to civitai/cli#604 by operator decision), and widening bareIdentArgs' parser to see selector expressions (ruled out of scope for this PR).
13. Gate caveat carried forward rather than dropped: make lint was run under nix-shell golangci-lint v2.13.2, while CI pins v2.12.2 — same major, not the pinned binary.
```

ZacxDev:
Round 3's claims block, for the next round. Each line is WHAT the fixes claim to have addressed — not why it is correct.

Round 3 was a DELTA re-audit of `19455c0c..a3a8ca0`. It returned **no 🔴 and no 🟡** — two 🟢 findings, both prose, and its own summary says each changes what a reader concludes. That is the definition of a finding rather than a stopping nit, so the fixes below landed and the ladder continues rather than ending on the auditor's "safe to merge" verdict.

⚠ **Both findings were re-verified against the code before being acted on**, not taken on the auditor's report.

```audit-claims round=3 audited=a3a8ca0..93db5cc
1. The waitAndCollect coverage row no longer credits printSubmitted's two lines to waitAndCollect: runGenerate reaches emitSubmitHandle (generate.go:1362, which calls printSubmitted at :1399) BEFORE waitAndCollect (:1367), so that function is never reached from this one, and printSubmitted already carries its own notCovered row.
2. The row and the test header now enumerate this function's OWN three still-open sites, counted in generate.go:1436-1585: safeTerm(wf.Status) and safeTerm(workflowID) at :1514, and safeTerm(workflowID) at :1558.
3. safeTerm(wf.Status) is now named explicitly as still open. The previous enumeration spoke only of the workflow id, so a SERVER STATUS STRING on the same line — keeping \n and \t exactly as the id does — was absent from the list of what remains open here.
4. Both places that carried the wrong sentence were corrected, not just the one the audit quoted: safeterm_coverage_test.go's row and generate_blob_forgery_test.go's test header.
5. The non-existent placeholder `{workflowId}` is swept to `{workflow}` at all SEVEN sites, not only the two a3a8ca0 added. renderOutName accepts exactly {workflow}/{n}/{ext}, and outNamePlaceholder's own comment names `{workflowid}` as the typo its refusal exists to catch, so every one of those comments described a template that would be REFUSED if a reader tried it.
6. The sweep asserts rather than reports: the script exits 2 on a per-file count mismatch and on any survivor, and printed `replaced N` for each of the three files; 0 occurrences remain.
7. NO mutation matrix is claimed, because no guard changed. This commit edits comments and one ledger row's `why` string; claiming a mutation result for a prose change would be an unearned claim.
8. Gate: internal/cmd re-run fresh with -count=1 (ok, 37 targeted tests pass, 0 FAIL); make ci green; make lint 0 issues under golangci-lint v2.13.2 while CI pins v2.12.2 — same major, not the pinned binary, carried forward rather than rounded off.
9. NOT acted on, stated rather than omitted: the no-URL error's end-to-end reachability through runGenerate is still argued in a comment rather than pinned — round 3 proved it with a throwaway driver, but no SHIPPED test drives that path (the subtest calls downloadOutputs directly). Round 3 also scoped its \t-expansion measurement to lipgloss v1.1.0 as vendored here.
```

**Ledger for this round** — payload lines changed by round 3's fixes: **6** (`internal/cmd/generate_output.go` +3/−3). Named separately because the gate reads the first number: *executable* payload changed = **0** — all six are comment text. The other three files are scaffolding by the class fixed at round 1.

ZacxDev:
Round 4's claims block, for the next round. Each line is WHAT the fixes claim to have addressed — not why it is correct.

Round 4 was a DELTA re-audit of `a3a8ca0..93db5cc`. **No 🔴. Two 🟡, and the first of them was a defect the round-3 FIX introduced** — a corrected sentence that overshot into a new false claim, which is the shape this ladder is explicitly told to hunt for. Both were re-verified against the code before being acted on.

```audit-claims round=4 audited=93db5cc..c63486c
1. The claim that printSubmitted AND printReattach are "different functions this one never calls" is corrected. It is true of printSubmitted only: waitAndCollect CALLS printReattach at generate.go:1447 (errWaitTimeout) and :1451 (context.Canceled/DeadlineExceeded).
2. The three sites printReattach exposes are now named with their lines — safeTerm(workflowID) :1628, safeTerm(status) :1630, safeTerm(workflowID) :1631 — and stated as REACHABLE from waitAndCollect, on the same timeout path civitai/cli#604 measured its two counterfeit `Re-attach:` lines on.
3. safeTerm(status) at :1630 is named. The round-3 fix again said only "the id", repeating for printReattach the exact omission it had just closed for wf.Status one line above.
4. Both bad drafts are recorded in the comment rather than replaced silently, including that the PRE-round-3 text made no unreachability claim and was, on this point, less wrong than its correction — so a third draft is not derived.
5. The safeTermCoveredBy waitAndCollect row carries the same reachable-through clause. Scoping the row to the function's own call sites stays deliberate, but "the other three are still open" was being read as the complete residual for the wait path.
6. The count claims in safeterm_userinput_test.go are DELETED, not renumbered: "35 rows govern 41 bare-identifier call sites: six rows cover two sites each (...)" with its enumeration, and "one call site for 29 of the 35 rows and two for the other six".
7. Reason for deleting rather than updating, stated in the comment: every part was exact at merge base e9c51b1 and falsified by THIS PR's own new rows, which also pushed an existing pair to three sites so the list's implicit "two is the maximum" broke; nothing asserts on any of it (t.Logf, under a comment saying LOGGED NOT ASSERTED), so it can never go red; and this sentence has now been wrong three times in the same direction.
8. The property is kept and the reader is pointed at the harness's own log line, which is computed from the tree they are standing in. Verified live: 37 rows pinned, and two pairs at three sites (waitAndCollect::workflowID at generate.go :1514/:1558/:1576; downloadBlobTo::target at generate_output.go :420/:455/:457).
9. NO mutation matrix is claimed, for the second commit running and for the same reason: no guard changed. This edits comments and one ledger row's `why` string.
10. Gate: internal/cmd fresh go test -count=1 ok; make ci green (22 package lines, 0 FAIL); make lint 0 issues under golangci-lint v2.13.2 while CI pins v2.12.2 — same major, not the pinned binary.
11. NOT acted on, stated rather than omitted: round 4's four nits (the `<workflowId>` angle-bracket metasyntax at generate_output.go:56/:462, which is correct as-is; the "60 functions" figure at safeterm_coverage_test.go:444, pre-existing and already stale at merge base; the placeholder-vs-expansion looseness; and a PR-comment wording that says "six sibling sites" where #604's table lists seven).
```

**Ledger for this round** — payload lines changed by round 4's fixes: **0** (all three files are scaffolding under the class fixed at round 1: payload = `generate_output.go` + `generate.go` + `safeterm.go`).

⚠ **Read together with round 3's 6, that is NOT two consecutive zero-payload rounds, so the attribution gate does not fire.** There is a second series in which both rounds are zero — *executable* payload, since round 3's six lines were all comment text — but taking the stop on that number would mean re-deciding the payload class mid-ladder, which is the exact way that gate gets disarmed without anyone choosing to. The stop, if it comes, is taken on the round-1 class.

ZacxDev:
Round 5's claims block — and **the ladder ENDS HERE, on the stated criterion, not on a clean round.** That distinction is the whole point of writing this down: a report that ends on the escape hatch is otherwise indistinguishable from one that converged.

```audit-claims round=5 audited=c63486c..a8cfa88
1. The "reachable-through set" enumeration is DELETED from both copies rather than extended. Round 5 found six more functions reachable from waitAndCollect rendering server text through plain safeTerm and unnamed by it: printOutputURLs (generate.go:1591 via :1562), classifyGenerateError (:2039 via :1455), and both poll reporters (generate_wait.go:313/317/322/342/349 via :1441).
2. Both copies now state the PROPERTY, name the notCovered rows plus civitai/cli#604 as the authoritative list, and say DO NOT ADD A FOURTH LIST. All three bad drafts are recorded so a fourth is not derived.
3. "Only printSubmitted is unreachable from here" is corrected — printSubmitResult is too. The error was conservative, but it is what made the enumeration read as exhaustive.
4. The THIRD copy of the falsified count, at generate.go:1881 ("six of the 35 rows govern two call sites each"), is deleted. The previous commit's title was deleting that claim and it swept two copies; `git grep "35 rows"` finds the third. Live: 37 rows, distribution 30x1 / 5x2 / 2x3.
5. Both redirect sentences are corrected: the t.Logf they point at prints the ROW count and the total over ALL safeTerm expression shapes (205), NOT the bare-identifier site count (46), which no output of this package prints; and being a t.Logf it is silent under `make ci`. They now give the -v invocation and say to measure rather than write the answer back.
6. NO mutation matrix is claimed, third commit running, same reason: no guard changed.
7. NOT fixed here, by design: none of the six surfaces is gated. They are pre-existing, ledgered notCovered, and outside #574's closing condition. civitai/cli#604 has been WIDENED to carry them, with the measured stdout repro and an amended closing condition.
8. Gate: internal/cmd fresh go test -count=1 ok; make lint 0 issues under golangci-lint v2.13.2 while CI pins v2.12.2 — same major, not the pinned binary.
```

**Ledger** — payload lines changed by round 5's fixes: **14** (`internal/cmd/generate.go` +11/−3, all comment text). Series across the ladder: round 3 = 6, round 4 = 0, round 5 = 14.

---

## Why the ladder ends here, stated rather than left implicit

**The attribution gate structurally cannot fire on this PR.** It needs two consecutive rounds whose fixes changed zero payload lines. Every round's fix edits a comment inside a payload file, so the count reads non-zero (6, 0, 14) while zero executable lines have moved since `a3a8ca0`. Switching to an "executable payload" count would make it fire immediately — and would be re-deciding the payload class mid-ladder, which is exactly how that gate gets disarmed without anyone choosing to. So it is left alone and named instead.

**The three criteria are met:**

- **No 🔴, and none in rounds 3, 4 or 5.** No finding in this range has a runtime, API or output consequence.
- **No blast radius beyond "the document contains a false sentence."** Every #596 finding since round 3 has been comment-truth. The one substantive discovery — `printOutputURLs` forging a row on **stdout** — is pre-existing, is not a #596 regression, and is now tracked in #604 rather than silently swept in here.
- **The recurring shape was swept at every site, and its generator removed.** The count claim was deleted at all three sites (found by `git grep`, not by fixing the one reported). The enumeration was not corrected a fourth time — it was deleted, with the property stated and a written instruction not to rebuild it. That is the documented cure: an enumeration like this cannot be completed by thinking harder, and each attempt reads as exhaustive to the next reader.

**Why the rounds would not stop on their own:** each round's fix is prose, and prose written to explain a correction generates the next round's finding. Rounds 3 → 4 → 5 are literally that chain — round 4 found round 3's fix false, round 5 found round 4's fix incomplete. Left alone this does not terminate, and the gate cannot see it.

**What is NOT fixed, so the next reader knows it is open rather than absent:**

1. The six surfaces above are **ungated**. → civitai/cli#604 (widened, with the closing condition amended).
2. The no-URL error's end-to-end reachability is **argued in a comment, not pinned** — round 3 proved it with a throwaway driver, but no shipped test drives that path through `runGenerate`.
3. Round 3's `\t`-expansion measurement, on the strength of which a tab assertion was deleted, is **scoped to lipgloss v1.1.0 as vendored**; nothing pins it against a version bump.
4. `safeterm_coverage_test.go`'s "60 functions hold a call today" is stale (73 at head). **Pre-existing**, already stale at the merge base, and the enforced constant beside it is a floor of 40 that still fires.
5. The bare-identifier site total and its distribution are **unpinned by construction** — `t.Logf`-ed under a comment saying so.

None of these blocks the merge; all of them are open rather than absent.

Issue #599, opened 2026-09-14
app submit: MaxSubmitBodyBytes is vendored and unguarded — add a drift guard like bump-scaffold-pins


ZacxDev:
## The gap

`#585` vendored the server's request-body ceiling into the CLI:

```go
// internal/appapi/appblocks.go
const MaxSubmitBodyBytes = 10485760
```

`SubmitVersion` now refuses locally on `len(body) > MaxSubmitBodyBytes` before uploading, which is
the right shape — the alternative (react to what the server returns) cannot work, because a
proxy-matched body is truncated at a chunk boundary rather than refused, so the response carries no
size signal at all.

But the number is **vendored and unguarded**. Nothing in this repo notices the day it changes
upstream, and the decision record says so in as many words
(`claudedocs/decisions/31-pkgzip-caps-are-not-a-server-mirror.md`).

## Why that is worth a guard rather than an accepted risk

It fails in **both** directions, and one of them is silent:

- **Upstream raises the ceiling** → the CLI refuses bundles the server would happily accept.
  Visible, annoying, and `--allow-oversize` is the escape hatch.
- **Upstream lowers the ceiling** → a bundle under the vendored number clears the local guard, is
  truncated on the wire, and comes back `400: Invalid JSON` naming nothing about size. That is
  **exactly the original bug (#423) reappearing**, and `--allow-oversize` does not help because the
  user never hits the guard. This direction has no user-visible tell.

`--allow-oversize` covers binaries already on someone's machine. It does not cover future releases
shipping a stale constant, which is what a drift guard covers.

## This repo already runs this exact pattern twice

- `.github/workflows/bump-scaffold-pins.yml`
- `.github/workflows/revendor-canonical-schema.yml`

Both re-read an upstream source on a schedule, open a PR on drift, and file an issue on a red run.
`MaxSubmitBodyBytes` is a third instance of the same shape and should reuse it rather than invent
anything.

## Provenance of the number, so the guard knows what to read

All three inputs are in `civitai/civitai`:

1. `src/proxy.ts` — `config.matcher` includes `'/api/v1/:path*'`, which is what makes the submit
   route proxy-matched and therefore subject to the limit at all.
2. `next.config.mjs` — sets no `proxyClientMaxBodySize` override.
3. The pinned Next version — ships `proxyClientMaxBodySize: 10485760` as the default.

The vendored constant is correct only while **all three** hold. A guard that reads just the third
[...]

Issue #423, opened 2026-08-15
app submit: MaxBundleSizeBytes (50 MiB) is ~12x the real server ceiling, so an oversized bundle uploads and fails as an opaque "400: Invalid JSON"


ZacxDev:
`MaxBundleSizeBytes` is **50 MiB**, but the server rejects a bundle far below that. The CLI therefore packages and uploads a bundle that can never be accepted, and surfaces the rejection as `400: Invalid JSON` — an error naming neither the size, nor the bundle, nor anything the author can act on.

An app in this state is simply **unreleasable**, with no diagnostic pointing at why.

## Repro

```
civitai app submit --yes
Packaged 68 file(s) (8201270 bytes compressed, 8866319 decompressed)
Submitting gen-matrix@0.8.3…
Error: server returned 400: Invalid JSON
```

Nothing in that output suggests size. `civitai app validate` passes, the manifest is fine, and the same command succeeds for other apps — so the natural read is a malformed manifest or a server fault.

## Bounding the real ceiling

Diagnosed against a control rather than a theory, because the error names no cause:

| bundle | files | compressed | submit |
|---|---|---|---|
| a known-good app, live today | **90** | **2.32 MB** | ✅ accepted |
| `gen-matrix@0.8.3` before the fix | 68 | **8.20 MB** | ❌ `400: Invalid JSON` |
| `gen-matrix@0.8.3` after the fix | 44 | **0.176 MB** | ✅ accepted |

**File count is ruled out** — the accepted control carries *more* files (90 > 68). The real server ceiling sits somewhere in **(2.32 MB, 8.20 MB]**. I did not narrow it further because each probe costs a real submission; the bound is what I measured, not a guess at the limit.

The cause in my case was a `docs/screenshots/` directory of review PNGs, which `app submit` packages along with everything else. That is arguably fine — but the failure mode should not be an opaque 400.

## The client-side cap is ~12x too permissive

`internal/pkgzip/pkgzip.go:45`

```go
MaxBundleSizeBytes  = 50 * 1024 * 1024  // 50 MiB total (compressed upload)
MaxDecompressedSize = 200 * 1024 * 1024 // 200 MiB decompressed total
```

Both guards exist and both pass, so the CLI's own preflight reports success on a bundle the server will always refuse. The comment on the constant says *"server max"*, which is what makes it misleading: it reads as a mirror of the server's limit, and it is not.

[...]

PR #602, opened 2026-09-14
fix(submit): refuse AT the body ceiling, and pin the contract #585 published

ZacxDev:
Follow-up to #585, carrying findings from a round-0 and a round-1 audit of it.

**Rebased onto `main` @ `b251d1c`** (was 54 behind, `CONFLICTING`; re-rebased once when `main` advanced mid-review with #672, and the whole gate was re-run on the moved base rather than assumed). Three files conflicted — `README.md`, `internal/appapi/appblocks.go`, `internal/cmd/app_submit.go` — all because `main` had restructured the same prose; see *Rebase* below for what each resolution kept. The only file this shares with any other open PR is `README.md`, with #671, in a different section.

The headline is an off-by-one that a shipped test asserted was unobservable.

---

## 1. The ceiling guard was off by one, and its "unreachable" claim was false

`internal/appapi/appblocks.go` refused on `len(body) > MaxSubmitBodyBytes`, so a body of **exactly** 10485760 uploaded in full.

`submit_ceiling_value_test.go` said in prose that this could not matter — *"the envelope is 19 bytes and base64 steps by 4, so body sizes go … 10485759, 10485763 — nothing lands on 10485760 at all. There is no input that distinguishes the two operators."* That measured **one** provenance and spoke for all four. Measured through the real `SubmitBodySize`:

| provenance | envelope | mod 4 | zipLen giving body == 10485760 |
|---|---|---|---|
| none | 19 | 3 | unreachable |
| commit only | 77 | 1 | unreachable |
| commit + `Dirty:false` | 97 | 1 | unreachable |
| **commit + `Dirty:true`** | **96** | **0** | **7,864,246** |

base64 output is always a multiple of 4, so only an envelope that is *itself* ≡ 0 (mod 4) can land on the ceiling — and that is exactly what `civitai app submit --allow-dirty` sends. A 7,864,246-byte zip therefore produced a 10,485,760-byte body the guard waved through: the precise body #585 exists to stop.

Changed to `>=`.

### The reasoning is an asymmetry, not a claim about the server

There is a real unresolved contradiction here and the code, the README and this PR all say so rather than inventing a mechanism:

- Next.js's source reads `bytesRead > bodySizeLimit`, which would **accept** a body of exactly 10485760.
- An end-to-end measurement of the real endpoint read `10485759 -> 401` (the body reached auth), **`10485760 -> 413`**, `12000000 -> 400`.

One of those is wrong. Neither can be re-measured from a unit test — it needs credentials and a real submit — so **nobody should write down which**. The decision does not depend on resolving it: refusing one byte early costs an author a flag they are already told about (`--allow-oversize`), while accepting one byte too many costs the whole ~10 MB upload and returns an error naming nothing about size. The consequences are wildly unequal, so the guard takes the cheap side.

**Pinned by** `TestSubmitBodyExactlyAtCeilingIsRefused`, which drives the real `SubmitVersion` at the one input that can tell the operators apart, with a negative control one base64 quantum below (so the assertion means `>=` and not "refuses everything") and two CONTROL checks that the fixture really sits on the boundary and that the provenance survives `sanitised()`.

## 2. `MaxSubmitBodyBytes` shipped with no godoc

A missing blank line glued its doc block to `ErrBundleTooLarge`'s. Measured before:

```
go doc appapi.MaxSubmitBodyBytes  ->  the bare const line, nothing else
go doc appapi.ErrBundleTooLarge   ->  "MaxSubmitBodyBytes is the largest…" + the whole evidence chain
```

The constant three published surfaces quote literally had no godoc; the sentinel wore the constant's evidence. Each declaration now carries its own doc (verified with `go doc`, not by reading), and the derivation block that documented nothing moved onto `SubmitBodySize`, which it was always about.

This originally added a 139-line AST test asserting each doc **begins with its own identifier** rather than merely being non-empty — non-emptiness cannot see the failure that happened, since `ErrBundleTooLarge`'s doc was long, detailed, and about something else. **That test has since been deleted and the predicate handed to staticcheck instead** — see *Algorithm pass* below. The defect itself is still fixed, and now caught repo-wide rather than in one package.

## 3. The published exit-code contract had no guard

`exitcodes_doc.go` publishes that the refusal exits `1`, and — under code 2 — that *"a script branching on `2` for that case must branch on `1`"*, a documented breaking change. `appblocks.go` called the code "now assertable". It was assertable and **not asserted**: no ledger row, no floor entry, no reference to `ErrBundleTooLarge` under `cmd/civitai/`. Behaviour was correct only by falling through `exitCode`'s `default`.

Added `TestOversizeBundleExitsGeneric` (the **real** error from the real command, not a hand-tagged fixture) + `TestOversizeSentinelIsNotAnAPIKind`, mirroring `TestDirtyWorkTreeExitsGeneric`, plus two `contractClaim` rows and their `exitCodeClaimsFloor` entries.

## 4. A comment named a guard that does not enforce what it claimed

`appblocks.go` said `pkgzip/caps_claim_test.go` is *"exactly the claim … that test exists to keep OUT of that file, and the reason this constant lives here instead."* Measured with both controls:

- **Real probe** — a fully server-attributed constant (`🔴 This IS the server's number` + the whole evidence chain) using none of the four banned phrases: **PASSES**.
- **Negative control** — appending `caps mirror the server`: **FAILS** with the test's own message.

So the guard bans four literal spellings; it does not force the placement. The placement is still correct on architectural grounds (`appapi` builds the body, so the body's limit belongs beside `SubmitBodySize`) — the constant did not move, only the stated reason. `claudedocs/decisions/31` was already accurate on this and is untouched.

## 5. Three user-facing strings contradicted shipped behaviour

All three said the CLI does not know any server ceiling, while it vendors 10485760 and refuses on it.

The one with teeth is `printSubmitSizeDiagnosis`: it runs **only when the vendored ceiling did not fire**, which is precisely the situation where that number may be too high. If the platform ever lowers `proxyClientMaxBodySize`, a body under the vendored value clears the guard, is truncated, returns `400: Invalid JSON` — #423 exactly — and the diagnosis told the author no local cap existed, pointing them away from the only number that could explain it. It now names the ceiling and the direction it can be wrong in.

`pkgzip`'s wording is narrowed to the **compressed archive**, which genuinely has no known platform ceiling — as distinct from the request body, which does. `caps_claim_test.go` re-run and green; no banned phrase introduced.

## 6. Three cheap ones

- **Spinner** — `Submitting …` goes to stdout and the refusal to stderr, so the CLI announced an upload it then declined. The spinner is now skipped for a body it can predict will be refused. It **predicts, it does not decide**: `appapi.SubmitVersion` remains the only place the refusal is made, so a disagreement could only show or withhold a spinner. `app_submit_oversize_test.go` read only stderr and was blind to this; it now asserts stdout, with a positive control that it is reading a non-empty stream.
- **Mismatched command** — the refusal's trailing colon promised "`--allow-oversize` submits anyway" and then printed `civitai app submit --package-only`, which never submits. Both commands are now offered under their own sentence.
- **Self-inconsistent regex** — `10 ?MiB|10 ?MB` were listed as suspicious then exempted via `!strings.HasPrefix(m, "10 ")`, so spaced forms could never be reported while unspaced `10MiB` could. Dropped rather than repaired: enumerated over both files, `10 MiB` appears in each and in both it is pkgzip's per-**file** cap, not this ceiling (alongside `2 MiB`, `50 MiB`, `200 MiB`, `32 MB`). A rule reporting them would be wrong; one exempting them by prefix only looks like a rule.

---

## Mutation results

Each mutant applied in a `cp -a` copy with `rm -f <copy>/.git` first, and watched to fail **for its own reason**:

| mutant | result |
|---|---|
| `>=` → `>` | **RED** — `TestSubmitBodyExactlyAtCeilingIsRefused`: *"a submit body of EXACTLY 10485760 bytes was uploaded"* |
| ceiling shifted down one base64 quantum | **RED** — same test's negative control: *"a 10485756-byte body — under the 10485760 ceiling — was refused"* |
| refusal tagged `civitai.ErrBadRequest` | **RED** — `TestOversizeBundleExitsGeneric`: *"exitCode(...) = 2, want 1"*. **SURVIVES the full suite at `main`** (rc=0, 21/21 packages) |
| code-2 ledger row deleted | **RED** — the floor: *"the ledger row … is gone"* |
| each of the 4 ledger phrases reworded out of the doc | **RED** independently, one assertion each |
| the #585 glued-doc defect re-introduced | **RED** — `golangci-lint` exit 1, `ST1022` on `ErrBundleTooLarge`. This is the mutant that justifies deleting the hand-rolled test; see *Algorithm pass* |

**Red at `origin/main`, green here:** `TestSubmitBodyExactlyAtCeilingIsRefused`; the new stdout assertion in `TestAppSubmitOversizeEndToEnd`. The godoc guard that used to be listed here was deleted in the algorithm pass; `ST1022` now covers it, and was watched going red on the same defect.

`TestOversizeBundleExitsGeneric` passes at `main` unmutated — it is an **invariant guard** on the published contract, not a regression test, and is labelled as such. The mutation above is what shows it has teeth.

### One finding this PR made against itself

A first draft of the code-2 ledger phrases used the bare `"must branch on"`. That is **walkable**: code 2's `--json` bullet already contains *"must branch on the exit code first"*, so rewording the migration notice to "Scripts may need updating." left the row **GREEN**. Caught by mutating, then respelled with the `1`/`2` the claim is actually about, and re-mutated to confirm both phrases now kill independently.

## Algorithm pass

Applied to this PR's own additions, before merge: question the requirement, then delete
before simplifying.

**Deleted `internal/appapi/submit_ceiling_godoc_test.go` (139 lines).** Its stated
assertion — *"the doc names its own declaration"* — is verbatim staticcheck **ST1022**
(and **ST1020** for methods). `.golangci.yml` disabled both, with a note saying adoption
was *"out of scope for a first gate; revisit to adopt wholesale"*. So the test was a
hand-rolled, single-package reimplementation of a check the repo had already vendored and
switched off, and its own comment said as much: *"staticcheck's ST1020-ST1022 would have
caught the empty half and .golangci.yml disables them."*

**Enabled the checks instead** — the three `- -ST1020` / `- -ST1021` / `- -ST1022` lines
are gone from `.golangci.yml`.

**Measured on the rebased tree, whole module: 1 finding.**

```
internal/scaffold/scaffold.go:37:1: ST1020: comment on exported method NeedsHarness
  should be of the form "NeedsHarness ..." (staticcheck)
```

Fixed here. It is a genuine instance of the same defect class — the comment named
`SDKTemplates`, which exists as no declaration anywhere in the repo — and it lives in a
package the deleted test could not reach, which is the argument for the swap rather than a
bonus. (Measured against `origin/main` the count was **2**; the other was
`ErrBundleTooLarge` wearing `MaxSubmitBodyBytes`'s doc, i.e. the #585 defect, already
cleared by this PR's own restructuring in §2.)

### Watched red, because the deletion is only justified if the linter really catches it

In a throwaway `cp -a` copy with `rm -f <copy>/.git` run first, the #585 defect was
re-introduced exactly as it shipped: the blank line between `MaxSubmitBodyBytes`'s doc
block and `ErrBundleTooLarge`'s removed, leaving the constant bare.

The symptom reproduced first:

```
$ go doc ./internal/appapi MaxSubmitBodyBytes
const MaxSubmitBodyBytes = 10485760          # no doc at all

$ go doc ./internal/appapi ErrBundleTooLarge
var ErrBundleTooLarge = errors.New("bundle too large to upload")
    MaxSubmitBodyBytes is the largest request body the submit endpoint can
    actually RECEIVE, in bytes.
```

Then the gate:

```
$ golangci-lint run          # exit 1
internal/appapi/appblocks.go:661:1: ST1022: comment on exported var ErrBundleTooLarge
  should be of the form "ErrBundleTooLarge ..." (staticcheck)
```

**Controls, both arms stated.** The unmutated tree lints clean — exit 0, 0 findings — with
these three checks enabled, so the check can go green. The mutant's finding is *inside
`internal/appapi`* itself, which is what proves the linter reached the package rather than
reporting a reassuring zero over it. And the deleted test was run against the same mutant
and also failed, naming the same two declarations — so for this defect the two are
redundant, and the linter is the wider of them.

**Two comments that asserted the old state were corrected**, since a stale comment is a
false claim: `MaxSubmitBodyBytes`'s doc block said staticcheck could not see this and that
`go doc` was the only check, and `.golangci.yml`'s header listed ST1020-ST1022 among the
families deliberately turned off. Both were true while the checks were disabled.

**Not shrunk:** `cmd/civitai/submit_oversize_exitcode_test.go` (188 lines) stays as-is.
It rests on a real maker — AGENTS.md item 7 plus a published breaking-change contract —
and the mutation table above shows it has teeth (`SURVIVES the full suite at main` for the
`ErrBadRequest` mutant).

## Rebase

Three conflicts, each because `main` had independently rewritten the same prose. In every
case the resolution keeps `main`'s structure and re-applies this PR's *semantic* delta on
top, rather than restoring the text `main` deleted.

- **`README.md`** — `main` slimmed the whole *How big can a bundle be?* section and every
  Troubleshooting row. Kept `main`'s wording and re-applied the `>`-to-`>=` contract to it:
  *"refuses at or above it"*, and the `the server can receive` row now reads *"reached or
  exceeded"*. This PR's two new rationale paragraphs (*Why "at or above" and not "above"*)
  carry over intact. The bracket/provenance table that appears on this branch's side was
  **not** this PR's — it is pre-existing prose `main` deliberately removed — so it stays
  removed. One of this PR's three README edits is **moot**: it reworded a sentence in the
  `What this CLI sent` row that `main` deleted outright when it rewrote that row.
  Verified structurally against `main`: headings 130 → 130, table rows 257 → 257, nothing
  lost.
- **`internal/appapi/appblocks.go`** — both sides restructured this region. `main` had
  noticed a *related* godoc defect and worked around it with a floating comment (*"THIS
  COMMENT DOCUMENTS SubmitBodySize, WHICH IS ~40 LINES BELOW"*); this PR fixes it properly
  by attaching each doc to its own declaration, which makes that workaround note false, so
  it is dropped. `main`'s two new declarations — `ErrNothingSent` and
  `submitRequestSentTrace` (#637) — were lifted out of the conflict verbatim and re-homed
  after `ErrBundleTooLarge`. Final order: `MaxSubmitBodyBytes` → `ErrBundleTooLarge` →
  `ErrNothingSent` → `submitRequestSentTrace` → `SubmitBodySize`, each with its own doc.
- **`internal/cmd/app_submit.go`** — kept `main`'s new `"up to"` / #637 paragraph and
  re-applied this PR's correction to the sentence beside it, which still summarised
  `pkgzip`'s cap comment as *"the honest move is to report rather than refuse"* — a phrase
  that stopped describing the CLI at #585.

### A semantic conflict the rebase did not report

`git` reported no conflict in `internal/cmd/app_submit_nothing_sent_test.go`, and the
result did not compile. This PR widens `doUpload` with an `allowOversize` parameter; `main`
independently added that test file, which calls `doUpload` with the old arity. Disjoint
files, both sides green on their own, red on the merge:

```
vet: internal/cmd/app_submit_nothing_sent_test.go:74:87:
  not enough arguments in call to doUpload
```

Fixed in its own commit. `false` is the production default and inert for those cases —
the fixture is 17 bytes, so the ceiling guard cannot fire either way — and the call site
now says so.

## Gate

`Makefile` target `ci` = `tidy vet test build`. All four run:

- `go vet ./...` — clean
- `go test ./... -count=1` — rc=0, **21/21 packages ok, 0 failures**
- `go build ./...` — ok
- `go mod tidy` — no change

**Both tiers, stated separately, re-run after the rebase.** The above is the full-clone tier. The depth-1 tier (`scripts/ci-shallow.sh`, which is also what CI's `build-test` job runs — `actions/checkout@v4` defaults to `fetch-depth: 1`) was run too, at `4a4c919`: `ok=21/21 failing-tests=0 failing-packages=0 timeouts=0 filtered-empty=0`.

`golangci-lint` 2.13.2 locally, **with ST1020/ST1021/ST1022 now enabled**: **0 issues**. `gofmt -l .`: clean. CI's `lint` job (2.12.2) initially caught two staticcheck findings in this PR's own new test files — `ST1008` (error not last) and `SA1019` (`parser.ParseDir` deprecated in Go 1.25). Both fixed in the second commit; the godoc guard was re-verified RED at `main` *after* the rewrite rather than assumed. All 13 checks now green.

## Not verified

- Which side of 10485760 the platform actually sits on. Deliberately unresolved above, and no code comment or doc claims a mechanism for it.
- Whether enabling ST1020-ST1022 stays at 1 finding as the repo grows. It is a gate from now on, so a new violation blocks rather than accumulating — but that is a claim about the gate, not a measurement of future code.
- #599 is untouched here — no comment, no close.
- This does **not** address #599 (a drift guard that would notice the day `proxyClientMaxBodySize` moves). It makes the staleness *legible* — `printSubmitSizeDiagnosis` now names the vendored number and the direction it can be wrong in — but nothing here detects the change.


ZacxDev:
Heads-up from the session that just merged **#601** (`b727a83`): this PR and that one both edit `internal/cmd/exitcodes_claims_test.go`, and **#601 landed first**, so this branch is now 1 commit behind `main` and its shared-file measurement is stale.

`gh pr view` reports `MERGEABLE`/`CLEAN`, but that is only the **textual** claim. The two changes are in different parts of that file — #602 appends `exitCodeClaimsFloor` entries and prepends claim rows; #601 rewrote the code-6 row's `pinnedBy` (it now names two guards instead of saying *"nothing local"*) and added a comment above it. An audit test-merged an earlier pair of heads (`cc24058` + `f02c2003`, branching on `git merge-tree --write-tree`'s **exit code**, not a marker grep) and got a clean tree at **21 ok / 0 FAIL** — but #601 then gained a second commit (`af35fbf`) that touches that file again, so **that reading no longer covers this pair**.

Cheap re-check before merging, since a clean git merge is not a clean merge:

```bash
git fetch origin
git merge-tree --write-tree origin/main <602-head>; echo "rc=$?"   # non-zero = conflict
# then materialise it and run the suite, and diff for silently-dropped rows:
diff <(git show origin/main:internal/cmd/exitcodes_claims_test.go) internal/cmd/exitcodes_claims_test.go
```

The last one matters most here: `#585`'s merge silently dropped `#591`'s two Troubleshooting rows on a resolution that conflicted nowhere (fixed by `#598`). The "lines `main` has that we lack" set should be empty except for what this PR deliberately changes.

No action needed from #601's side — flagging it so the second merge doesn't repeat that.

ZacxDev:
⚠️ **Heads-up for whoever rebases this — a semantic collision that a clean `git merge` will not show you.**

Posting rather than touching the branch. This PR is another session's and is being left alone deliberately.

**#630** (README phase-2 cuts, currently open) rewrote the "How big can a bundle be?" prose and the `app submit` command-table row. In doing so it **deleted the exact sentence this PR exists to correct**, and the 14-line paragraph this PR adds explaining why the refusal is inclusive.

The rewritten text asserts `app submit` refuses **above** the ceiling — *"a body **larger than** the server can receive"*. That is correct against `main` today (`internal/appapi/appblocks.go` is `len(body) > MaxSubmitBodyBytes`) and **becomes wrong the moment this PR lands**, which changes it to refuse **at or above**, on the strength of a measured `413` at exactly 10485760 and a `401` one byte below.

**Why the usual check misses it.** `git merge-tree` reports a conflict here (rc 1), but so does `#602 × origin/main` on its own — this branch is now 16 commits behind, so the git conflict is not #630's doing and an exit-code check clears it. The hazard is that a rebase resolver, seeing the conflicting hunk is #630's own newer rewrite, takes it wholesale and silently drops the inclusive-ceiling correction.

**Concrete consequence:** the shipped README would tell an author whose `--allow-dirty` submit produces a body of exactly 10,485,760 bytes that the CLI accepts it. It refuses. That size is reachable to the byte — 96-byte envelope, base64 output always a multiple of 4, a 7,864,246-byte zip lands on it exactly.

**When rebasing, re-apply this PR's inclusive wording on top of #630's shorter paragraph** rather than accepting either side whole. The two changes are compatible; they just cannot be resolved by picking a side.

Found by a round-0 audit on #630. Nothing here is a criticism of this PR — its measurement is the more accurate one, which is exactly why it should survive the rebase.

Issue #604, opened 2026-09-14
generate: a server-chosen workflow id forges counterfeit `Re-attach:` lines on the recovery block


ZacxDev:
`internal/cmd/generate.go` prints the server-supplied `workflowID` (and `status`) through **plain `safeTerm`** at seven sites. `internal/saferune` keeps `\n` and `\t` **by design**, so `safeTerm` alone does not make a single-line surface single-line — `safeTermSingle` is the gate that does, and it is what civitai/cli#596 applied to the blob-download path beside these.

The result: an attacker-chosen workflow id forges counterfeit **column-zero `Re-attach:` lines** on the recovery block — the block a user reads *after a wait ended without a result*, i.e. about money that has already been spent and outputs they can still fetch.

## Measured

`printReattach` with `workflowID = "wf-real\n  Re-attach:   civitai workflows get wf-ATTACKER"`, against `fix/generate-blob-forgery` @ `a3a8ca0`:

```
⚠ Stopped waiting after 5m0s. The generation was NOT cancelled — it is still running server-side and has already been charged.
  Workflow ID: wf-real
  Re-attach:   civitai workflows get wf-ATTACKER
  External ID: ext-1
  Last status: processing
  Re-attach:   civitai workflows get wf-real
  Re-attach:   civitai workflows get wf-ATTACKER
Or watch it at https://civitai.com/generate
```

**Two counterfeit `Re-attach:` lines**, because the id is printed twice (`Workflow ID:` and `Re-attach:`) and each occurrence forges one. Note the first counterfeit appears **above** the genuine line, so a reader scanning top-down reaches the attacker's command first. The whole block is `fmt.Fprintf` with no styling in front of it, so nothing re-indents the continuation.

## The sites

All in `internal/cmd/generate.go`:

| line | function | operand |
|---|---|---|
| 1514 | `waitAndCollect` — non-succeeded terminal status | `safeTerm(wf.Status)`, `safeTerm(workflowID)` |
| 1558 | `waitAndCollect` — succeeded-but-no-deliverables | `safeTerm(workflowID)` |
| 1610 | `printSubmitted` — `Generation submitted — workflow %s` | `safeTerm(workflowID)` |
| 1616 | `printSubmitted` — the `re-attach any time with …` line | `safeTerm(workflowID)` |
| 1628 | `printReattach` — `Workflow ID:` | `safeTerm(workflowID)` |
| 1630 | `printReattach` — `Last status:` | `safeTerm(status)` |
[...]

PR #608, opened 2026-09-14
fix(generate): gate the thirteen sites where server text forges terminal lines (#604)

ZacxDev:
Implements the widened scope of #604 — the issue stays open for a human to retire by hand. Every `generate` site where server-supplied text can forge terminal lines. `internal/saferune` keeps `\n` and `\t` **by design**, so `safeTerm` alone does not make a single-line surface single-line; `safeTermSingle` is the gate that does.

**13 sites, 15 expressions** (the terminal-status error and the quiet reporter's `tick` each carry two). Line numbers **re-derived on this tree** (`426288f`), not taken from the issue:

| # | file:line | function | operand | verb |
|---|---|---|---|---|
| 1 | generate.go:1514 | `waitAndCollect` terminal-status error | `wf.Status` | `%q` |
| 2 | generate.go:1514 | same line | `workflowID` | `%s` |
| 3 | generate.go:1558 | `waitAndCollect` no-deliverables error | `workflowID` | `%s` |
| 4 | generate.go:1591 | `printOutputURLs` | `*o.URL` | `%s`, **stdout** |
| 5 | generate.go:1610 | `printSubmitted` submitted line | `workflowID` | `%s` |
| 6 | generate.go:1616 | `printSubmitted` re-attach line | `workflowID` | `%s` |
| 7 | generate.go:1628 | `printReattach` `Workflow ID:` | `workflowID` | `%s` |
| 8 | generate.go:1630 | `printReattach` `Last status:` | `status` | `%s` |
| 9 | generate.go:1631 | `printReattach` `Re-attach:` | `workflowID` | `%s` |
| 10 | generate.go:**2047** | `classifyGenerateError` | `apiErr.ServerMessage` | `%s` |
| 11 | generate_wait.go:313 | `(*quietPollReporter).tick` err branch | `e.status` | `%s` |
| 12 | generate_wait.go:317 | `(*quietPollReporter).tick` no-err branch | `e.status` | `%s` |
| 13 | generate_wait.go:322 | `(*quietPollReporter).finish` | `status` | `%s` |
| 14 | generate_wait.go:342 | `(*ttyPollReporter).tick` | `e.status` | `%s` |
| 15 | generate_wait.go:349 | `(*ttyPollReporter).finish` | `status` | `%s` |

Only #10 moved from the issue's numbers (it listed `:2039`); every other line is where the issue said it was. No site has disappeared and no new sibling appeared — `git grep -n 'safeTerm(' internal/cmd/generate.go internal/cmd/generate_wait.go` now returns exactly one line, `serverReasonSuffix`'s `indentContinuation(safeTerm(reason))`, which is the deliberate MULTI-line surface and stays.

## Per-site origin

Every operand is **pure SERVER** origin, so `internal/saferune`'s rule against rewriting what the USER typed (#393) is not engaged at any of them:

- `workflowID` — `result.ID` off the submit reply (`generate.go:1343`), threaded through `emitSubmitHandle` and `waitAndCollect` unchanged. Already classified `SERVER` in `bareIdentArgs` for all four of its functions.
- `wf.Status` / `e.status` / `status` — the workflow's own status off the wire, via `statusOrUnknown` (which substitutes the CLI's literal `"unknown"` when absent).
- `*o.URL` — a presigned blob URL off the workflow payload.
- `apiErr.ServerMessage` — the server's own message; `genapi`'s own doc calls it "UNTRUSTED, server-origin text".

`printReattach`'s `External ID:` row carries the CLI-minted (or `--external-id`) value and is **deliberately left ungated**; the ledger row says so.

## 🔴 One operand is not line-forgeable today, and the test says so rather than shipping an assertion that cannot fire

`wf.Status` at :1514 renders through **`%q`**, which *escapes* `\n` and `\t` — and `genapi.IsTerminalStatus` `TrimSpace`s before matching its set, so the server can pad a known terminal status with the retained runes but cannot put words of its own in that operand and still exit the poll loop. So `assertOneLine` passes **identically** with and without the gate there; shipping it as that site's guard would have scored a mutation as killed while measuring nothing.

It is gated anyway — one rule, one place, and a gate must not depend on a format verb three tokens away staying `%q` — and pinned by a new `assertNoControlEscapes`, which asserts the rendered message carries no literal `\n`/`\t` escape. The matrix below confirms the split: M1 is killed **only** by that assertion, M2 **only** by `assertOneLine`.

## Tab assertions deliberately omitted

Re-measured on this tree rather than carried over: `ui.For(w).Dim/.Info/.Warn/.Bold/.Success` all render `"a\tb"` as `"a    b"` (lipgloss expands the tab to four spaces) while passing `"a\nb"` through unchanged. So:

- **`printSubmitted` gets no tab assertion** — both its lines are `st.Success`/`st.Dim`, where such a guard could never fire.
- **`printReattach`'s ⚠ header and trailer** are styled too, but its three gated rows are bare `Fprintf`, so the block-level tab assertion IS live (and M7/M8/M9 each fire it).
- Every other surface here is bare `fmt.Fprintf` or a plain `fmt.Errorf`, so their tab assertions are live and watched red.

## Tests

- `TestPrintOutputURLsRowCountIsNotServerChosen` — asserts the **row count on stdout** (one per kept output) and **two tab-separated fields per row, numbered by the CLI**, not the absence of a payload. On the mutant a URL carrying `\n2\t<other>` produced **4 rows for 3 outputs** and shifted every subsequent row's number.
- `TestPollReportersCannotForgeALine` — drives a **NON-TERMINAL** status through the real `pollWorkflow` for both reporters, reaching `tick`'s two gated branches (attempt 1 no-error, attempt 2 a scripted 500) and `finish` via the **timeout** path, which calls it with the last non-terminal status. The `tty` reporter is constructed directly: `newPollReporter` picks it only for a real TTY.
- `TestPrintReattachBlockGeometryIsNotServerChosen` — asserts the block is the **6 lines it writes**, that no TAB survives, and that each of its four labelled rows begins **exactly one** line. State, not the absence of a word.
- `TestPrintSubmittedCannotForgeALine`, `TestWaitAndCollectTerminalStatusErrorCannotForgeALine`, `TestWaitAndCollectNoDeliverablesErrorCannotForgeALine` — the last two drive the REAL command and use the existing `assertOneLine` helper from `download_newline_test.go`; **no second copy of it was defined**.
- `TestClassifyGenerateErrorCannotForgeALine` — all **five** matching arms, each with a control that the arm actually MATCHED rather than falling through.

## Mutation matrix — 15/15 KILLED

One mutant at a time, restored from a pristine copy between each, in a `cp -a` copy **outside the module root** with its `.git` file removed first. Each mutation asserts it matched its anchor exactly once and that the mutant is on disk. Full `internal/cmd` package each time, baseline green first.

| # | site → mutant | test that went red | the failure text watched |
|---|---|---|---|
| M1 | :1514 `wf.Status` → `safeTerm` | `TestWaitAndCollectTerminalStatusErrorCannotForgeALine` | `#604 UNGATED OPERAND in waitAndCollect (terminal-status error, server status): the rendered text carries a literal \n escape…` (and `\t`) |
| M2 | :1514 `workflowID` → `safeTerm` | same test | `#577 FORGERY in waitAndCollect (terminal-status error, workflow id): rendered text holds 1 newline(s), want 0` + the TAB arm |
| M3 | :1558 → `safeTerm` | `TestWaitAndCollectNoDeliverablesErrorCannotForgeALine` | `#577 FORGERY in waitAndCollect (succeeded-but-no-deliverables error): rendered text holds 1 newline(s), want 0` + TAB |
| M4 | :1591 → `safeTerm` | `TestPrintOutputURLsRowCountIsNotServerChosen` | `#604 FORGERY in printOutputURLs: 4 row(s) on STDOUT for 3 kept output(s) — a server-chosen URL wrote 1 extra numbered row(s)…` + `row 3 is numbered "2", want "3"` |
| M5 | :1610 → `safeTerm` | `TestPrintSubmittedCannotForgeALine` | `#604 FORGERY in printSubmitted: a hostile workflow id rendered 3 line(s) where a benign one renders 2` |
| M6 | :1616 → `safeTerm` | same test | same message |
| M7 | :1628 → `safeTerm` | `TestPrintReattachBlockGeometryIsNotServerChosen` | `#604 FORGERY in printReattach: … rendered 7 line(s) where benign ones render 6` + `a TAB survived into the block` |
| M8 | :1630 → `safeTerm` | same test | same two |
| M9 | :1631 → `safeTerm` | same test | same two |
| M10 | :2047 → `safeTerm` | `TestClassifyGenerateErrorCannotForgeALine` (**all five subtests**) | `#577 FORGERY in classifyGenerateError (insufficient buzz): rendered text holds 1 newline(s), want 0` + TAB |
| M11 | :313 → `safeTerm` | `TestPollReportersCannotForgeALine/quiet` | `#604 FORGERY in the quiet poll reporter: a hostile status printed 8 line(s) where a benign one prints 5` + `a line the reporter did not write appeared in its output` |
| M12 | :317 → `safeTerm` | same subtest | `…printed 6 line(s) where a benign one prints 5` + the unwritten-line and TAB arms |
| M13 | :322 → `safeTerm` | same subtest | `…printed 6 line(s) where a benign one prints 5` + the unwritten-line and TAB arms |
| M14 | :342 → `safeTerm` | `TestPollReportersCannotForgeALine/tty` | `#604 FORGERY in the tty poll reporter: a hostile status emitted 5 newline(s) where a benign one emits 1` + TAB |
| M15 | :349 → `safeTerm` | same subtest | `…emitted 2 newline(s) where a benign one emits 1` + TAB |

No mutant died on a build error, a `CONTROL failure`, or another guard's message.

## Ledgers

`safeTermCoveredBy`: eight functions move notCovered → covered (`printOutputURLs`, `printSubmitted`, `printReattach`, `classifyGenerateError`, and all four poll-reporter methods). `maxUncoveredSafeTermFuncs` **17 → 9**, banked in the same commit; 9 is the number the RATCHET assertion itself printed (`RATCHET HEADROOM: 9 … but maxUncoveredSafeTermFuncs is 17`), not one derived by subtracting.

`waitAndCollect`'s row is rewritten: it had enumerated these very sites as still open, and it now states the `%q`/`%s` split on :1514 explicitly so neither assertion reads as a claim about the other operand. Every new/edited `why` describes only what its test's body actually drives — including the surfaces each test does **not** touch (`printSubmitted`'s `Charged … Buzz` line, `tick`'s 429 branch, `printReattach`'s `External ID:` row, the classifier's fall-through).

`bareIdentArgs` needed no new rows: `scannedSanitizers` already scans `safeTermSingle`, so no call site left the harness's view (205 sites before and after) and no argument spelling changed.

## README

The `--no-download` bullet now states the listing's row shape and its one-row-per-output guarantee, and the "`label: value` lines are a narrower promise" bullet adds `generate`'s wait-path lines to the flattened set.

## Gates

- `make ci` — green, **21 packages** `ok`, run on the committed tree.
- `make lint` — **golangci-lint 2.13.2**, `0 issues`.

Refs #604


ZacxDev:
## Round 0 (requirements & deletion) ran at PR-create time, and questioned the requirement

Recorded here for the ladder's history. **Round 0 reports and does not move the ladder** — it is not a findings-keyed stop signal and licenses no skipped round, so the nine correctness axes still run next.

Its verdict was `requirement questioned`, and it was right twice. Both defects trace to a closing condition **I** wrote on #604 during #596's audit ladder, not to the implementation.

**F1 🔴 — the closing condition's own check instrument was structurally blind.** #604 named its check as `git grep -n 'safeTerm(' internal/cmd/generate.go internal/cmd/generate_wait.go`. That enumerates operands which **already have a gate**; it cannot see a server-text operand with no gate at all. One sat at `generate_wait.go:312` — `e.err` rendered with `%v`, on the *same `fmt.Fprintf`* as an expression this PR gates. `genapi`'s `generateError` embeds the server's own `message` into every arm and `APIError.Error()` returns it verbatim. Measured at the then-head **with the gate already applied**: three forged lines, each beginning `  status ` — byte-identical in prefix to `quietPollReporter.finish`'s own line — each carrying a tab, each claiming a status the workflow never had.

The PR's own test could not see it: its fixture message was the literal `"scripted"`, and its row-prefix assertion **permitted** `  status ` as a legitimate prefix.

**F2 🟡 — the row-count invariant I specified was wrong, in the false-positive direction.** `genapi.Deliverable` does not require a URL, while `printOutputURLs` numbers by index into `kept` and skips nil-URL entries. So a server returning `url: null` on the middle of three deliverable outputs legitimately renders **two rows numbered 1 and 3** — and the assertion I demanded (`len(rows) == len(kept)`) would have fired a false `#604 FORGERY` on a valid response.

**F3/F4 🟡🟢** — two README sentences over-claimed (one of them false only until F1 landed) and one ledger `why` repeated F2's error.

### Two things round 0 got right that are worth keeping

- **It declined to manufacture a deletion**: 4 candidates examined, 0 recommended. It argued the one plausible cut — the `%q`-protected gate at `:1514` and its new assertion helper — should **stay**, because deleting the helper ships an unpinned gate, which is the incident #596 round 2 already recorded.
- **It found my justification weaker than the facts.** #604 argued the `status` operand class in on consistency ("gating six of seven identical lines is how the seventh regenerates"). The real argument is stronger: `printReattach`'s status is `statusOrUnknown(wf)` off a poll that timed out or was interrupted — a *non-terminal* status straight off server JSON, unbounded by `IsTerminalStatus`, on a bare `fmt.Fprintf`. A future reader acting on "consistency" might have dropped it.

It also established that `classifyGenerateError` has **five** callers, not three — so this PR incidentally hardens `civitai workflows get/list/cancel`, which the PR body does not claim.

### Verified independently before and after the fixes

Not taken on the agents' reports: the ratchet equality (`9 NOT COVERED (cap 9)` against 9 real `notCovered` rows); `printOutputURLs`' ungate mutant both before and **after** the assertion was widened — it still dies, now on a fixture whose legitimate rows are numbered **1, 3, 4** with the forgery filling the gap at "2"; the `:312` gate's mutant, which fires a *sequence* assertion (`finish()` writes `  status ` once, last — one in the middle is a counterfeit wearing the prefix); and that `wf.Status` renders through `%q` while `workflowID` on the same line renders through `%s`.

⚠ **Unrelated red check, not caused by this PR:** `schema-drift` is failing. `main`'s own vendored schema differs from the live canonical (upstream added `posts:write:self`), so `main` would fail it too — a stored green with an expiry date. Filed as #609. It is not a required context, so it does not block here.

`round 0 · requirements: 7 (unattributed: 2) · deletion candidates: 4`

ZacxDev:
Round 1's claims block — the first full nine-axis pass. Each line is WHAT the fix claims to have addressed, not why it is correct.

Round 1 returned **no 🔴**, two 🟡 and two 🟢. 🔴 **Only ONE of the four is this PR's own** (F4); F1–F3 are pre-existing holes in the same command, outside #604's closing condition, and are filed as **#612** rather than swept in — a fix rolled into an audit round is a fix nobody reviewed against its own requirement.

```audit-claims round=1 audited=8872a4c..d597211
1. TestWaitAndCollectTerminalStatusErrorCannotForgeALine no longer runs assertOneLine over the WHOLE error. The message ends in serverReasonSuffix, which is deliberately multi-line via indentContinuation(safeTerm(reason)) at generate.go:144 and pinned by name in indentcontinuation_ledger_test.go.
2. The assertion is scoped to the head (the text before ". The server reported:"), which carries both gated operands. The suffix is the reason's own surface.
3. assertNoControlEscapes still runs over the WHOLE string, because no surface here may carry an escape whether it is multi-line or not.
4. The narrowing did NOT weaken coverage: reverting the workflowID gate still reddens this test by name with "#577 FORGERY in waitAndCollect (terminal-status error, workflow id): rendered text holds 1 newline(s), want 0".
5. The false-red was measured as a CONTROL PAIR, not asserted: on a fixture carrying the legitimate reason "the model could not be loaded:\nout of memory on the worker", the fixed test PASSES and the reverted test FAILS with "#577 FORGERY … forged 1 extra terminal line(s)" — the guard accusing correctly-indented output on the one surface this repo keeps multi-line on purpose.
6. The unstated precondition is now stated: the original passed only because this fixture's `"steps":[]` carries no failure reason, which nothing enforced.
7. NOT fixed here and filed as #612, stated rather than omitted: generate.go:1255 (the Buzz-balance warning prints berr with %v and NO gate, forging a Cost: line above the spend confirmation) and generate.go:2110 (classifyGenerateError's fall-through returns err unchanged on the dominant error path). Both pass RAW ANSI, a wider class than #604's \n/\t. #612 also carries a DERIVED-not-measured third item.
8. Gate: internal/cmd `go test -count=1` ok; `make lint` 0 issues under golangci-lint v2.13.2 while CI pins v2.12.2 — same major, not the pinned binary.
9. Round 1 verified clean and NOT re-checked by this fix: no over-gating (all 16 operands server-origin; externalID, --out-dir/--out-name, o.prompt and the --image source half correctly stay raw); the apiErrorWithMessage/apiErrorWithStatus refactor is byte-identical and weakens no existing caller; serverReasonSuffix stays multi-line and nothing else in the diff collapses a value that should not be; all five new tests drive the real production function rather than rebuilding the message inline; all 60 ledger `why` rows read, none wider than its body.
10. Round 1's search bound, stated so the next round can widen it: all 102 fmt.Fprint*/Errorf/Sprintf sites in generate.go (93) and generate_wait.go (9), each operand traced to its origin — NOT `git grep 'safeTerm('`, which by construction only finds operands that already have a gate. generate_wait.go is now fully gated; generate.go has exactly the three ungated server operands in #612.
```

**Ledger — round 2's range, and the first one this ladder can measure** (round 1 was a first full audit with nothing to attribute against): payload lines changed by round 1's fix = **0**. The only file touched is `internal/cmd/generate_forgery_sites_test.go`, which is scaffolding under the class fixed here: payload = `generate.go` + `generate_wait.go` + `README.md`.

⚠ One zero is not the attribution gate — that needs **two consecutive** zero-payload rounds. This is the first.

ZacxDev:
Round 2's claims block — and **the ladder ENDS HERE, on the ATTRIBUTION GATE**, with no outstanding findings.

Round 2 returned two 🟢. The first was a real defect **in the fix I wrote for round 1**, and it is worth stating plainly: my guard was walkable.

```audit-claims round=2 audited=d597211..0a97292
1. The `strings.Cut(got, ". The server reported:")` narrowing is REMOVED. It was measured walkable: `workflowID` is server-origin and lands BEFORE that marker, so an id of "wf1. The server reported: ok\nSaved out.png (2.0 MiB)\tDONE" moves the real newline into the discarded tail.
2. That walk was reproduced independently before fixing: with the workflowID gate reverted AND that id, the test PASSED — a fully forged terminal line surviving every assertion. assertNoControlEscapes could not see it either, because it matches the two-character `\n` that `%q` emits, not a raw newline.
3. The no-reason test now asserts one-line over the WHOLE string, which is simply correct when no reason exists and requires no splitting. Its control is a HasSuffix on the no-reason note — serverReasonSuffix is appended LAST, so the ending is the CLI's own whatever the server spelled earlier; a Contains or Cut on that same text would be displaceable, HasSuffix is not.
4. TestWaitAndCollectTerminalStatusErrorWithAReasonStaysMultiLine is NEW and drives the reason-carrying case that no shipped fixture drove before — round 1's F4 was proved only in a commit message, on a fixture that never landed. Round 2 measured that: the narrowing was a no-op on every shipped test.
5. That test SUBTRACTS the suffix it injected (CutSuffix of the exact string production must append, rebuilt from its own known reason) rather than searching for a marker, so nothing the server spells in another operand can displace it; a mismatch fails loudly instead of silently discarding a forgery.
6. Control pair, measured both ways: the walk was PASS before and FAILS after, in both tests, by name, printing the forged line. The legitimate multi-line reason FAILED under round 1's assertion and PASSES now. So the false-red stays gone AND the walk is closed.
7. Round 2's F2 (a stale citation) is corrected in issue #612's body, not in this tree: generate.go:2110 is past EOF — the file is 2109 lines and the fall-through `return err` is :2108.
8. 🔴 Round 2's OTHER citation correction was itself WRONG and is deliberately NOT applied: `:1188` IS `return classifyGenerateError(err)`; `:1187` is the `if err != nil {` above it. Re-derived by hand before declining it.
9. Gate: internal/cmd `go test -count=1` ok; `make lint` 0 issues under golangci-lint v2.13.2 while CI pins v2.12.2 — same major, not the pinned binary.
```

---

## Why the ladder ends here — the gate, not a judgement call

**Two consecutive rounds whose fixes changed zero PAYLOAD lines.** The class was fixed at round 1 and is binding: payload = `internal/cmd/generate.go` + `generate_wait.go` + `README.md`; every `*_test.go` is scaffolding.

- Round 1's fixes (audited by round 2): `generate_forgery_sites_test.go` only → **0 payload**.
- Round 2's fixes (audited by round 3): `generate_forgery_sites_test.go` only → **0 payload**.

Cross-checked independently rather than inferred from the per-round numbers: `git diff 8872a4c 0a97292 -- internal/cmd/generate.go internal/cmd/generate_wait.go README.md` is **empty**. The shipped behaviour has been byte-identical since `8872a4c`; the last two rounds audited scaffolding the ladder itself wrote.

That is exactly what the gate exists to catch, and unlike #596 — which had to end on the stated criterion because comment edits inside payload files kept the count non-zero — this one ends on the mechanical rule.

**No findings are left unaddressed.** Round 2's F1 and F2 are both fixed, so nothing is being filed-instead-of-fixed here.

⚠ **What that does NOT mean.** The gate says the ladder has left the PR; it is not a claim that `generate` is now wholly gated. It is not — **#612** carries three operands with no gate at all, including the Buzz-balance warning that forges a `Cost:` line directly above the spend confirmation, and `classifyGenerateError`'s fall-through on the dominant error path. Those are pre-existing, outside #604's closing condition, and deliberately not swept in here.

PR #611, opened 2026-09-14
fix(upgrade): derive the release-asset extension per GOOS, so Windows can self-update (#613)

### LINKED ISSUE #613: upgrade: cannot work on Windows — the asset name is built as .tar.gz unconditionally, but goreleaser ships windows as .zip
`civitai upgrade` is advertised for every platform the release builds, and §Install explicitly advertises Windows binaries. On Windows it can never succeed.

## Mechanism

`internal/cmd/upgrade.go:199` builds the release-asset name with a hardcoded extension:

```go
tarName := fmt.Sprintf("civitai_%s_%s_%s.tar.gz", verNoV, runtime.GOOS, runtime.GOARCH)
```

`.goreleaser.yaml:42-44` overrides the archive format for Windows:

```yaml
format_overrides:
  - goos: windows
    formats: [zip]
```

So the published Windows asset is `civitai_<v>_windows_amd64.zip`, the lookup for `civitai_<v>_windows_amd64.tar.gz` finds nothing, and `upgrade.go:203-205` returns:

```
no release asset "civitai_<v>_windows_amd64.tar.gz" for windows/amd64 — upgrade manually from https://github.com/civitai/cli/releases/latest
```

## Severity

Low blast radius — it fails loudly and names the manual fallback, so nobody is left with a corrupted binary. But it is a published capability that has never worked on one of the three advertised platforms, and the error names an asset that does not exist rather than the real reason.

## Fix sketch

Select the extension from `runtime.GOOS` (zip on windows, tar.gz elsewhere) and add the matching unzip path beside the existing tar reader. `checksums.txt` already covers both archive kinds, so the integrity check needs no change.

## Related

- `upgrade.go:128` lists `/home/linuxbrew/.linuxbrew` in `brewPathMarkers`, so a Linuxbrew-path binary is told to run `brew upgrade civitai/tap/civitai` — which installs nothing, because only a macOS **cask** is published (`.goreleaser.yaml` has `homebrew_casks:` and no `brews:`). Same family, worth fixing in the same pass.
- Found during a README accuracy audit (#611), which corrected the documentation to scope the claim. That is a patch over a real defect, not a fix.

## Closing condition

`civitai upgrade` on windows/amd64 and windows/arm64 locates and installs the published `.zip` asset, and a guard asserts that the archive extension `upgrade.go` builds for a given GOOS equals the format `.goreleaser.yaml` publishes for that GOOS — so adding a platform or changing a format cannot silently desync again.

Checked by: a merged PR carrying that guard, with `go test ./internal/cmd/ -run Upgrade -count=1` exiting 0. The end-to-end Windows leg needs a Windows machine and is the maintainer's to observe; the guard is the mechanically checkable half.

ZacxDev:
**Scope widened.** This started as a documentation-only PR: it corrected six published
README claims that source contradicts. One of those six (defect **2**) documented a real
**bug** — `civitai upgrade` cannot work on Windows. On review the decision was to **fix the
bug rather than document it**, so this PR now also carries a user-visible behaviour change,
and the caveat prose it previously added is deleted on all four surfaces it reached.

Closes #613

---

## Part 1 — the fix (`internal/cmd/upgrade.go`)

`runUpgrade` built the release-asset name with a hardcoded `.tar.gz`, while
`.goreleaser.yaml`'s `format_overrides` publishes `goos: windows` as a **`.zip`**. Neither
file was wrong on its own; the defect lived in the seam between them, and `runtime.GOOS`
being compile-time meant no test on CI could execute the branch at all. Every
`civitai upgrade` on Windows died with

```
no release asset "civitai_<v>_windows_<arch>.tar.gz" for windows/amd64 — upgrade manually from …
```

having replaced nothing — permanently, on every version, not just a bad release.

**Confirmed against the real published release, not inferred from the config.** `v0.1.105`
uploads `civitai_0.1.105_windows_amd64.zip` and **no** windows `.tar.gz`; that zip holds
`civitai.exe` (16,036,352 bytes, PE magic `MZ`) at top level beside `LICENSE`/`README.md`.
The fixed code, driven with GOOS pinned to `windows`, derives that exact asset name and
extracts that exact binary out of the real downloaded archive.

| new | what it is |
|---|---|
| `archiveFormatForGOOS(goos)` | the vendored mirror of `.goreleaser.yaml`'s archive formats — the single source of the extension |
| `releaseBinaryName(goos)` | `civitai.exe` on Windows. goreleaser appends `.exe`, so extracting by the bare name would have failed on exactly the platform the zip path exists for |
| `releaseAssetName(version, goos, goarch)` | the ONE writer of that string; an inlined `Sprintf` elsewhere reopens the bug |
| `extractBinaryFromArchive(data, format, name)` + `extractBinaryFromZip` | the new reader, dispatched on the **format argument** rather than `runtime.GOOS`, so both branches are reachable from a test on any host |
| `runtimeGOOS` | a seam beside the existing `osExecutable` / `evalSymlinks` / `applyUpdate` ones, so the Windows path runs on CI |

**The integrity gate is untouched and unweakened.** `checksums.txt` already covers both
archive kinds, so nothing about it needed to change: verification still happens **before**
extraction, extraction still happens **before** `applyUpdate`, and the refusal on a missing
`checksums.txt` stands. A corrupt archive therefore leaves the running binary alone.

## Part 2 — the prose the fix makes false

Four surfaces carried the Windows caveat. All four are gone:

1. `README.md` §Upgrading — the caveat paragraph (the archive sentence now just names what
   is downloaded per platform).
2. `README.md` — the `civitai upgrade` command-reference row's `🔴 **Not on Windows**`.
3. `internal/cmd/upgrade.go`'s `Long`. The **macOS Homebrew scoping stays** — that is
   defect 1 and is still true.
4. `README.md` Troubleshooting — the `no release asset` row. Checked first: the symptom
   index holds **69** documented symptoms and the floor guard's minimum is **15**, so
   removing one row (→ 68) does not approach it. Its entry in
   `TestREADMETroubleshootingCoversTheRefusalsAuthorsActuallyHit` went with it.

## Part 3 — the guard now asserts the opposite

`TestREADMEWindowsUpgradeCaveatMatchesTheReleaseFormats` was built to **demand the caveat
exist**. It is replaced by **`TestUpgradeArchiveFormatsMatchTheReleaseConfig`**, which pins
the RELATIONSHIP the caveat only described — the durable claim, and the one #613's closing
condition names.

For **every GOOS `.goreleaser.yaml` builds**, through the real functions rather than the
source text:

1. `archiveFormatForGOOS(goos)` equals the format the release publishes for it;
2. `releaseAssetName(…)` ends in that extension — so an extension inlined anywhere *other*
   than `archiveFormatForGOOS` is still caught;
3. `extractBinaryFromArchive` can actually **open** that format — so an override naming a
   format the CLI has no reader for fails here, not on a user's machine after a verified
   download;
4. the three surfaces that **state the format in prose** — the README's §Upgrading, the
   README's `civitai upgrade` command-reference row, and the command's own `Long` — name
   the format that GOOS really publishes. Added in round 3 (see below).

🔴 **Leg attribution, corrected in round 3 — it was wrong in both directions.** Measured,
one mutant at a time, reading the failing assertion's line number:

| mutant | legs that redden | failing assertion |
|---|---|---|
| `upgrade.go` reverted to a hardcoded `tar.gz` | **(1) alone**, on windows | `upgrade_test.go:847` |
| `.goreleaser.yaml` gains `darwin: [zip]`, `upgrade.go` untaught | **(1) and (4)**, on darwin | `:847`, `:885` ×3 |
| `releaseAssetName` drops the extension | (2), on every GOOS | `:859` ×3 |
| the zip case deleted from the extractor's switch | (3), on windows | `:871` |
| any prose surface inverted, release and code untouched | **(4) alone** | `:885` ×3 |

Legs (2) and (3) never fire on a format *disagreement*: leg (1) `continue`s for exactly the
GOOS that failed it. They are reachable, by the other mutants above. Leg (4) runs in its own
pass after the loop, so that `continue` cannot suppress it — and it fires on the `darwin`
row because a RELEASE change makes the prose wrong too, which is exactly what leg (4) is
for. (The comment in the test claimed "(1) only" for that row for one commit; `af627f4`
corrects it against the re-measured numbers above.) Positive controls: the `civitai` build must declare ≥3 goos (and name
linux/darwin/windows), ≥2 archives must parse, the override list must be **non-empty**, and
the derived formats must be **non-uniform** — a uniform release could not tell
`archiveFormatForGOOS` from `return "tar.gz"`.

## Part 4 — red-before-fix matrix (11 mutants)

Counted from the runner's own per-test result lines, never an exit code. Baseline
**31 pass / 0 fail**.

| mutant | killed by |
|---|---|
| **A** `archiveFormatForGOOS` → always `tar.gz` *(the exact #613 defect)* | parity guard, `…WindowsUsesTheZipAsset`, `…ChecksumMismatchAbortsForBothFormats/windows`, `…CorruptArchiveLeavesTheBinaryAlone/windows` |
| **B** `releaseBinaryName` ignores `.exe` | `…WindowsUsesTheZipAsset` |
| **C** zip case dispatches to the tar reader | parity guard, `TestExtractBinaryFromZip`, `…CorruptInput/zip`, `…WindowsUsesTheZipAsset` |
| **D** `default:` falls through to the tar reader | parity guard (negative control) |
| **D′** `default:` falls through to the **zip** reader | parity guard (negative control) — *survived until round 3; see below* |
| **E** `zip.NewReader` error swallowed | `…CorruptInput/zip` ×3, `…CorruptArchiveLeavesTheBinaryAlone/windows` |
| **F** `gzip.NewReader` error swallowed | `…CorruptInput/tar.gz` ×2 |
| **G** extraction error ignored (corrupt archive reaches `applyUpdate`) | `…CorruptArchiveLeavesTheBinaryAlone` **both** subtests |
| **H** checksum gate short-circuited | `TestUpgrade_ChecksumMismatchAborts`, `…ForBothFormats` **both** subtests |
| **I** `runUpgrade` inlines a hardcoded name, bypassing `releaseAssetName` | `…WindowsUsesTheZipAsset`, `…ForBothFormats/windows`, `…CorruptArchive…/windows` |
| **J** `.goreleaser.yaml` gains a `darwin: [zip]` override `upgrade.go` does not honour | parity guard |
| **K** `.goreleaser.yaml` drops `format_overrides` entirely | parity guard (control) |

Mutant **A** reproduces the original symptom verbatim:
`no release asset "civitai_0.1.11_windows_amd64.tar.gz" for windows/amd64`.

🔴 **This control was wrong TWICE, in mirror-image ways, and the code was fine both times.**
Round 1: **D** — a `default:` falling through to `extractBinaryFromTarGz` — passed a green
suite, because the control fed it a **zip**; the error came from the tar reader rejecting
the zip, not from the dispatch, so the assertion was green for an unrelated reason and would
have stayed green with the `default:` deleted. The fix made the payload a well-formed
**tar.gz** — which made the control one-directional in the other direction. Round 2
measured **D′**: with that tar.gz payload, a `default:` falling through to
`extractBinaryFromZip` **SURVIVED**, 76 pass / 0 fail, identical to baseline.

The control now runs **once per known reader**, each with a payload that reader would
ACCEPT, mislabelled `"7z"`. Both D and D′ are killed above, each naming the reader it fell
through to. The comment and the assertion message were also rewritten: both previously
claimed coverage the single-payload body did not provide.

## Part 5 — behavioural tests for the new path

- `TestExtractBinaryFromZip` — a real `archive/zip` archive, flat and nested entry, through
  both `extractBinaryFromZip` and the dispatcher; a missing member is an error, not empty
  bytes.
- `TestExtractBinaryFromArchive_CorruptInput` — **both** formats × 4 corruptions
  (half-truncated, 10 bytes, empty, header overwritten), each with a positive control that
  the **intact** archive extracts the full payload first.
- `TestUpgrade_WindowsUsesTheZipAsset` — the whole `runUpgrade` path with GOOS pinned to
  windows against a server publishing **only** the `.zip`.
- `TestUpgrade_ChecksumMismatchAbortsForBothFormats` — the refusal still fires for tar.gz
  **and** zip, with a positive control that the archive really was downloaded first (so the
  refusal is the checksum gate, not an earlier asset-lookup miss).
- `TestUpgrade_CorruptArchiveLeavesTheBinaryAlone` — a truncated archive whose
  `checksums.txt` entry **matches**, so the run gets past the integrity gate and dies in the
  extractor. Asserts `applyUpdate` is never called and the on-disk binary is byte-identical.

🔴 The fixtures state what the release publishes **literally** (`wantArchiveFormat` /
`wantBinaryName` in the test file), not by calling the production helpers. A fixture built
from `archiveFormatForGOOS` would agree with it by construction, and mutant A would have
survived every end-to-end test.

---

## The original six README defects (unchanged, re-measured)

| # | Claim | Reality |
|---|---|---|
| 1 | `### Homebrew (macOS / Linux)` | `.goreleaser.yaml` has `homebrew_casks:` and **no** `brews:` stanza. A cask is macOS-only, so a Linux user following §Install gets a hard refusal. |
| 2 | `civitai upgrade` works, stated unconditionally | **Now fixed rather than documented — see above.** |
| 3 | `NO_COLOR` and `CIVITAI_NO_COLOR` presented as interchangeable | The `CIVITAI_*` pair goes through viper `GetBool` → `strconv.ParseBool`, so **`CIVITAI_NO_COLOR=yes` parses as false and silently does nothing**. |
| 4 | "the flag wins over the environment" | Backwards for `--token`, wrong for the update-check flag (an OR), and wrong for `--color` (off beats on). Only `--tunnel-endpoint` behaved as claimed. |
| 5 | Section titled "Command reference" | Its table covers the authoring/account commands only; nine read/download commands had no row and no pointer. |
| 6 | Contents says "every command, one table" | There are two `civitai …` command tables, different schemas, 351 lines apart. |

**Red-on-revert, re-measured on this tree** (baseline 77 result lines, 0 fail; one
correction reverted at a time, restored between runs):

| # | correction reverted | verdict | owning guard that went red |
|---|---|---|---|
| 1 | Homebrew macOS-only (TOC + heading + Install intro + refusal + §Upgrading scoping + `upgrade --help`) | **RED** | `TestREADMEHomebrewSectionMatchesTheReleaseConfig` |
| 2 | *(prose gone — replaced by the parity guard + 11-mutant matrix above)* | — | `TestUpgradeArchiveFormatsMatchTheReleaseConfig` |
| 3 | colour env value set (2 flag rows + block + 2 config rows) | **RED** | `TestREADMEColorEnvValuesMatchTheBooleanParser` |
| 4 | per-setting precedence → "the flag wins over the environment" | **RED** | `TestREADMEConfigPrecedenceIsPerSetting` |
| 5 | `## Command reference` pointer | **RED** | `TestREADMECommandReferencePointsAtTheReadCommandTable` |
| 6 | Contents entry → "every command, one table" | **RED** | `TestREADMETOCCommandReferenceEntryNamesTheTableSplit` |

Each revert produced **exactly one** failure, on its own guard — no cross-kills.

## Gates

- `make ci` — green (21/21 packages).
- `./scripts/ci-shallow.sh` at the pushed commit — **21/21 ok, 0 failures, 0 timeouts**.
- `golangci-lint run` — **0 issues**, with two negative controls run and reported: a
  non-compiling defect in the new zip reader → 1 typecheck issue, and a **compiling** one
  (`name = fmt.Sprintf("%s", name)`) → 1 staticcheck S1025. The instrument can go red on a
  defect `go build` would not catch.
- `gofmt -s -l .` — clean.
- **Merged-tree check** against current `origin/main` (`b37f388`) plus the only other open
  PR touching `README.md` (#602, branch `zach/585-follow-up`). Integration branch off main,
  both merged, full suite: **21/21 packages green**, no conflict. #602's nearest README
  hunk is 29 lines from one of mine, which is why it was extracted and run rather than
  reasoned about. Read the merged result: #608's `--no-download` paragraph and wait-path
  bullet survive (#602's branch predates #608 and its two-dot diff *looks* like it reverts
  them — the three-way merge does not), #602's "at or above" wording lands, and no stale
  Windows caveat remains.
- `README.md` vs `origin/main`: **4189 → 4233 lines**, **311,422 → 314,637 bytes**
  (+44 lines, +3,215 bytes).

---

## Round 3 — the last audit round (round 2 returned no 🔴, four 🟡)

Every mutant below was reproduced BEFORE the fix (confirmed surviving) and again after
(confirmed dying, naming the assertion that killed it). None failed to compile.

| # | mutant | before | after — killed by |
|---|---|---|---|
| 🟡-2 | invert §Upgrading **and** `Long` to "a `.tar.gz` on Windows, a `.zip` everywhere else" | **SURVIVED** — 21/21 packages ok, 0 FAIL | `upgrade_test.go:885` (leg 4), 6 errors: both surfaces × 3 GOOS |
| 🟡-1 | `default:` → `return extractBinaryFromZip(data, name)` | **SURVIVED** — 28 tests, 0 FAIL, incl. the parity guard | `upgrade_test.go:908`, naming the *well-formed zip payload* leg of the control |
| 🟡-3 | delete the whole `if closeErr != nil` block | **SURVIVED** — 20 pass / 0 fail | *not a kill — the branch was genuinely dead.* Deleted; see below |
| 🟡-4 | hardcoded-`tar.gz` **and** `darwin: [zip]` override | each reddened **only the leg-(1) line** | *not a code fix — the sentence was corrected to what was measured; re-measured again after leg (4) landed, see Part 3* |

### 🟡-2 — three published claims pinned by nothing

`README.md` §Upgrading, the `civitai upgrade` command-reference row and
`upgrade.go`'s `Long` each state "a `.zip` on Windows, a `.tar.gz` everywhere else". Round 2
deleted the last guard that read prose and replaced it with one that reads only code, so
all three could be inverted with the suite green.

Leg (4) now **parses the claim out of each surface** — platform clause plus catch-all — and
compares it against the same `plan.formatFor` the code legs use. Nothing is hardcoded, which
is the point: the scenario the finding named (`.goreleaser.yaml` gains `darwin: [zip]`,
someone teaches `archiveFormatForGOOS`, suite green, prose now false) was measured and now
reddens **all three surfaces** on `darwin`. Each surface is independently reachable —
mutating only the command-reference row fails only that surface.

Controls: the parse Fatals if it cannot find both a platform clause and a fallback in a
surface, and Fatals if the platform words it found intersect no GOOS the release builds
(a prose "macOS" would land there — map it to `darwin` rather than loosening the compare).

### 🟡-3 — the comment named the wrong mechanism, and the branch it guarded was dead

The comment said a truncated/corrupt member surfaces as a CRC failure on **`Close`**.
Measured directly against Go 1.25.14 `archive/zip`, three corruption shapes including
through the `io.LimitReader` production uses: `io.ReadAll` returns **`zip: checksum error`**
and `rc.Close()` returns **nil**. The `err` check on the line above is what catches
corruption; the `closeErr` branch could not fire, and deleting it survived the suite.

Branch deleted. The comment now names `io.ReadAll` as the CRC site and says plainly that
Close's error is dropped because it has nothing left to report — a false comment there
invites deleting the check that actually works.

### 🟡-4 — leg attribution

Corrected in Part 3 above, to what was measured. The guard is unchanged and not weakened.

### 🟢-1 (taken) — silent truncation over the running binary

Both extractors read `io.ReadAll(io.LimitReader(rc, maxDownloadBytes))`, which returns
**exactly the cap with a nil error** when the member is larger — and those bytes go straight
to `applyUpdate`, i.e. over the running executable. Pre-existing on the tar path, newly
duplicated on the zip path. `civitai.exe` is 16 MB against a 64 MiB cap today, so this is
headroom, not a live break.

`readArchiveMember(r, name, limit)` reads `limit+1` and refuses the overflow by name. Both
extractors call it; the README's abort list names it. Mutation-swept, bounds straddling the
cap (limit-1 / limit / limit+1) so an off-by-one has a case that separates it:

| mutant | killed by |
|---|---|
| read `limit` instead of `limit+1` (*the original bug*) | `upgrade_test.go:1097` — "returned 1000 bytes and no error" |
| `>= limit` instead of `> limit` | `upgrade_test.go:1107` — the exactly-at-cap case |
| the size check deleted | `upgrade_test.go:1097` |

## Round 3 gates

- `make ci` — green, 21/21 packages.
- `./scripts/ci-shallow.sh` at the pushed commit `868df44` — **21/21 ok, 0 failures,
  0 timeouts**, depth-1 clone confirmed at that SHA.
- `golangci-lint run` — **0 issues**, with a negative control run and reported: a
  **compiling** defect in the new `readArchiveMember` (`name = fmt.Sprintf("%s", name)`) →
  **1 issue, staticcheck S1025**; restored → back to 0. The instrument can go red on a
  defect `go build` would not catch.
- `gofmt -s -l .` — clean.
- **README revert matrix re-run** (defects 1, 3, 4, 5, 6 — one revert at a time, full
  `go test ./...` each, README restored between runs): each reddened its owning guard —
  `…HomebrewSectionMatchesTheReleaseConfig`, `…ColorEnvValuesMatchTheBooleanParser`,
  `…ConfigPrecedenceIsPerSetting`, `…CommandReferencePointsAtTheReadCommandTable`,
  `…TOCCommandReferenceEntryNamesTheTableSplit`. Two second reds, both explained rather
  than waved off: defect 4 also fails `TestReadmeAssertsFlagWinsRulePredicate`, whose own
  CONTROL Fatals when today's README already asserts the retracted claim — exactly the
  state that revert creates; and `TestFilesystemErrorsExitGenericEndToEnd` failed once
  under the defect-6 run and then passed **5/5** when re-run (3× clean tree, 2× under the
  same revert), so it was a flake, not a cross-kill.
- **11-mutant upgrade battery re-run**, baseline 37 result lines / 0 fail: **all 11 still
  killed**, same attribution as Part 4 — A(6 fails) B(1) C(5) D(1) E(7) F(4) G(3) H(4) I(5)
  J(1) K(1). Plus D′ above.
- **Merged-tree check** against **every** still-open PR touching the files this one edits:
  **#602** (`README.md`) and **#619** (same `internal/cmd` package). `git merge-tree
  --write-tree` branched on the **exit code**: rc=0 for main+HEAD, HEAD+#602, HEAD+#619.
  Run twice, because `origin/main` moved mid-round (`b37f388` → `e464652`, #618): an
  integration branch off each main with all three PRs merged, full suite **21/21 green**
  both times, plus a main-only merge of this branch at `e464652`, also **21/21**. Read the
  merged README: my three archive-format surfaces and the 64 MiB clause survive intact, and
  #602's changes do not touch them.
- `README.md` vs `origin/main` (`e464652`): **4189 → 4234 lines**, **311,422 → 314,712
  bytes** (+45 lines, +3,290 bytes).

---

**Not verified:** a real `civitai upgrade` executed on a Windows machine. The archive name,
the archive contents and every branch are exercised against the real published v0.1.105
zip, but `selfupdate.Apply`'s Windows rename-the-running-exe path is stubbed behind the
`applyUpdate` seam in tests, as it already was for every other platform.


ZacxDev:
Round 1 (nine correctness axes) ran against `07a278f` and returned five 🟡, no 🔴. All five were fixed, and the operator then widened the PR's scope to fix the underlying Go bug (#613) rather than document it. The block below covers both commits so round 2 can diff the delta.

```audit-claims round=1 audited=07a278f..0e504a59
1. The `NO_COLOR=0` claim the PR itself published is now pinned — a new behavioural leg asserts both halves at the value `0`, where the two env contracts diverge; previously only `yes` was exercised, a value where they agree, and harmonising `envSet` with `envTrue` left all 21 packages green.
2. The TOC guard's reverse branch now gates on `len(sections) < 2` alone; schema count is reporting detail only. Its previous failure message instructed the reader to re-derive the entry, which if followed would have republished defect 6.
3. The Homebrew guard's reverse direction now un-pins every surface its forward direction pins (reason, Install intro clause, heading, `upgrade --help`), and the intro clause is required forward so the reverse ban is not vacuous.
4. The retracted-sentence check now normalises whitespace and is scoped to a non-quoted context, so a re-wrapped restoration is caught while a sentence quoting the claim in order to refute it is permitted.
5. `civitai upgrade --help`'s `Long` string was scoped to match the README, and pinned as a third surface in the two guards that own those derivations.
6. Two factual errors introduced by the previous round's compression were corrected: the Homebrew "why" now attributes the `brews:` stanza to `.goreleaser.yaml` rather than to the tap repository, and a test comment no longer cites a hand-written fire-drill fixture as evidence about goreleaser's output.
7. `internal/cmd/upgrade.go` now derives the release-asset extension per GOOS (`archiveFormatForGOOS`), the binary name per GOOS (`releaseBinaryName`, `.exe` on Windows), and dispatches extraction on an explicit format argument with a new zip reader beside the existing tar reader.
8. The Windows caveat prose added by earlier rounds was deleted from all four surfaces it had reached, because the fix makes it false. The macOS Homebrew scoping remains.
9. The caveat-presence guard was replaced by a parity guard asserting that, for every GOOS goreleaser builds, the extension `upgrade.go` derives equals the archive format published for it — driven through the real function rather than a regex over source.
10. Four behavioural tests were added covering zip extraction, tar.gz regression, corrupt archives of both kinds, and checksum mismatch for both kinds. One mutant initially survived; the control was fixed rather than the code, because the surviving mutant had been fed a zip and was dying in the tar reader rather than at the dispatch.
```

PR #618, opened 2026-09-15
docs(handoff): four stale ranks, a retracted red-gate claim, and a closing condition that a merge could satisfy

ZacxDev:
An accounting pass, prompted by *"are all the objectives fully addressed?"*. They were not — and three of the four stale rows were stale **because of this session's own work**.

## Corrections

| rank | said | actually |
|---|---|---|
| 3 | "dispatched… still 0 comments" | fixed and **MERGED** (`5da5cc6467` + `d2218f54da`) |
| 12 | "R2, R3, R4, R6 remain, NONE touched" | R2–R4 **accepted in writing**; only R6 remains |
| 19 | "devdocs#5 open since 2026-07-13" | **MERGED** `23c6d46` — and the date was wrong (08-10) |
| 24 | "🔴 OVER its ceiling, reddening a SHARED gate" | **RETRACTED — the gate is green** |

## 🔴 Rank 24: a `forcing: gate` row asserting a gate that isn't red

Measured: `handoff-index-store-claims-accuracy.md` on devrc's `origin/main` is **59,764 B, under the 65,536 ceiling**. Every oversize copy (66,930–72,627 B) is in a `devrc-*` **worktree** — other sessions' unlanded work, which the gate never reads (`this_repos_corpus()` walks the repo root).

The row was inherited from a previous handoff and carried all session without a check, while citing *"a permanently-red gate trains everyone to click through"* — it was that hazard **inverted**. A `forcing:` kind is a claim, and nothing validates it; `gate` is the kind most worth re-measuring precisely because it sounds automatic.

## 🔴 The closing condition could be satisfied by a merge

It read *"…has a fix **MERGED**"* — met the moment the platform PRs landed, while the endpoint still coerced and while Rochet2 had been told on #513 that it closes on a probe. **Two bars for one thing, and the published one was stricter.** Now requires the rank-25 probe to return `"username":"0222"` quoted on a `cf-cache-status: MISS`.

Second hole: *"externally-authored"* literally admitted `#607` (`schema-revendor-bot`). Now reads **HUMAN-authored**.

## 🔴 A near-miss that is the best lesson here

The eviction making room for this cut **27,991 B instead of ~1,800**. A slice bounded only on `"\n### "` ran past the end of its section — the block was the last one — deleting the **entire ranked list** and most of the Gotchas, while printing a plausible new byte count and exiting 0.

What caught it was an integrity check counting ranked items, which returned **0**. Nothing else would have, and the next step was a commit. Nothing was committed; `git restore` on both files undid it, at the cost of redoing the edits.

Redone bounded on **both** delimiters (`min` of `\n### ` and `\n## `), with a size assertion, a no-`## `-inside assertion, and a post-condition that the ranked section survives. All three lessons recorded in Gotchas.

## Integrity

26 ranked items · 28 `forcing:` tags · 1 closing-condition · 1 rank-25 probe · 7 `## ` sections · **0** removed `## ` headings in the diff · 65,446 B of 65,536.

## What this does NOT close

The arc's remaining step is unchanged and is nobody's here to take: **the deploy**, then rank 25's probe, then `cli#513`.


ZacxDev:
## Round 0: four findings, all four applied — and one of them closed the arc

Dispatched before the merge decision, which is the only reason F4 was actionable. **Every claim below was re-verified first-hand; none was taken on the auditor's word.**

### 🔴 F4 — the deploy had LANDED, and this PR's conclusion was stale on arrival

Rank 25 said *"MERGED, NOT DEPLOYED"*. The auditor ran **the probe that row itself prescribes** and it came back quoted. Re-verified independently — cache-busted, `MISS`, `limit` varied per run:

```
?username=2428023993&limit=3   MISS  ->  "username":"2428023993"     <- the reporter's own case
?username=2428023993&limit=11  MISS  ->  "username":"2428023993"
?username=0222&limit=9         MISS  ->  "username":"0222"
?username=0222&limit=17        MISS  ->  "username":"0222"
```

And the round-trip, which is the real test: `?username=0222` → **18 items**, `?username=222` → **0**. Before the fix the printed name was `222` and `?username=222` found nothing — the value the API handed you could not be used against the API.

**`cli#513` is CLOSED** with that evidence, and the arc's closing condition is **MET**. The lesson is in the doc: *a "waiting on someone else" row is the kind that rots silently — run its own probe before believing it.*

### 🔴 F2 — the ceiling that drove five evictions does not bind in this repo

**Nothing in `civitai/cli` enforces a handoff byte budget.** The only such gate is devrc's `test_handoff_doc_size.py`, whose corpus is `REPO_ROOT`-rooted at `/home/zach/workspace/devrc` and structurally cannot see this repo. What *does* reach here is `handoff_doc.py`'s `budget_warning()`, which gates on **relpath alone** (`claudedocs/` + `/handoff-`) and is repo-agnostic — so it prints *"will go RED on `main`, and it fails for EVERYONE"* about a gate that will never run.

That is **rank 24's error one level up**: an unmeasured claim that a shared gate is red, driving real work. Cost: five evictions in two days, 35,517 B moved to `refs/`, and the 27,991 B near-loss this PR was already recording. The standing "evict before adding anything" instruction is now corrected — keeping the doc tight stays worthwhile **as judgement, not as a gate**.

Worth a separate devrc change: make `budget_warning()` repo-aware, or register `civit/cli` in `default_repos()` so the premise becomes true. Out of scope here.

### 🔴 F1 — my own "correction" was the false claim

Rank 19 said devdocs#5 was *"open since 2026-07-13"*; I "fixed" it to 08-10. **`createdAt` IS `2026-07-13`.** 08-10 was `updatedAt` when I read it — the last push — and a comment of my own moves that field. The original row was right, and *"two months stale"* was exact. In a PR whose entire payload is accuracy corrections, replacing a true fact with a false one is worse than the staleness it replaced. Now recorded as the lesson: **`updatedAt` is not an open date.**

### F3 — two of three new Gotchas were self-duplication

They restated rank 24 and the closing-condition block **added in the same diff** — 1,269 B. Dropped. Only the heading-delimited-slice lesson was new; it stays, and it is the best content here.

### What the auditor refuted

It rejected the deletion candidate I pointed it at hardest — retiring the ranked-list mechanism. `claim-work --slug-for <doc> 25` resolves to `external-issue-513-numeric-username-25`: **rank is half a live cross-session claim slug**, so renumbering silently repoints live claims. The 16 closed rows are 2,818 B of 65,446 — 4.3%. Correct call, and not the one I expected.

It also confirmed rank 24's retraction independently (59,764 B on devrc's `origin/main`, absent from `GRANDFATHERED`, worktrees are siblings of `REPO_ROOT`) and validated its instrument both ways before reading its verdict.

Ledger: `round 0 · requirements: 10 (unattributed: 5) · deletion candidates: 4` · **round-0 trial for this PR: `ran: 1 · changed the outcome: 1`.**

Issue #627, opened 2026-09-15
pkg/civitai: snippet truncates on a BYTE index, so a multi-byte body is cut mid-rune and the SDK returns invalid UTF-8


ZacxDev:
Found while implementing #605/#624, which bound a server operand for display. `pkg/civitai`'s `snippet` is the repo's existing precedent for that policy, so it was read closely before being copied — and it should not be copied.

## The defect

`pkg/civitai/read.go:609`:

```go
func snippet(raw []byte) string {
	s := strings.TrimSpace(saferune.Strip(string(raw)))
	const max = 500
	if len(s) > max {
		return s[:max] + "…"
	}
	return s
}
```

`len(s)` and `s[:max]` are **byte** operations. A body whose 500th byte falls inside a multi-byte UTF-8 sequence is cut mid-rune, and `snippet` returns a string that is **not valid UTF-8**.

## Measured

Driving the real `snippet` (throwaway test in `pkg/civitai`, deleted after):

```
prefix=""    in bytes=2400 -> out bytes=503 runes=169 VALID=false  tail=8b e6 96 e2 80 a6
prefix="v"   in bytes=2401 -> out bytes=503 runes=169 VALID=false  tail=9e 8b e6 e2 80 a6
prefix="v2"  in bytes=2402 -> out bytes=503 runes=169 VALID=true   tail=e5 9e 8b e2 80 a6
prefix="v2x" in bytes=2403 -> out bytes=503 runes=171 VALID=false  tail=a1 e5 9e e2 80 a6
emoji (4-byte runes, 1-byte prefix): out bytes=503 VALID=false
```

Body = `<prefix> + "模型文件" x 200` (3-byte runes). `500 % 3 == 2`, so exactly one of the three alignments survives; the 4-byte-rune case fails too (`499 % 4 == 3`). The truncated tail is a **partial** rune followed by the ellipsis — a terminal renders it as U+FFFD or worse, and a consumer doing rune-wise work on the string gets `RuneError`.

⚠ Scope of the measurement: it is the real `snippet`, on synthetic bodies. It does not establish that any live endpoint returns such a body — only that the function cannot be relied on to return valid UTF-8 when one does. A server body carrying non-ASCII over 500 bytes is ordinary (any localised error message, any prompt echoed back, any CJK file name in a list).

## Where the bytes go

`snippet` is not a debug helper; its output is user-facing on several paths:

- `read.go:163` — `unexpected response from %s (status %d): %s`
[...]

Issue #624, opened 2026-09-15
generate: the Buzz-balance gate bounds the rune class but not LENGTH, so the forged Cost: line survives via soft wrap — uncapped via appblocks.go:1185


ZacxDev:
Found by the round-1 audit of #619. **Pre-existing — not introduced by #619**, and the same class as #605 (soft-wrap forgery on `download`'s progress line), on a different and more expensive surface.

## The shape

`safeTermSingle` guarantees one line **per line-break rune it can see**. A terminal's **soft wrap** has no rune. So a gate that closes the ANSI/`\n`/`\t` class leaves the *length* class wide open, and the forged line lands anyway.

The surface is `internal/cmd/generate.go:1286` — the Buzz-balance failure warning, printed **immediately above** `confirmGenerate`'s real `Cost: … Buzz` line and `Generate? [y/N]:`. That is the last screen before an irreversible spend, and #619 has just gated it against the rune class.

## Measured at `beb7373`, driving the real `runGenerate`

```
BENIGN  balance error "boom"       -> 1 logical line,   108 runes
HOSTILE balance error, 5120 chars  -> 1 logical line, 5,224 runes
        (160 x "Cost: 1 Buzz (balance 999999).  ")
        logical lines: 2   ESC: none   TAB: none
```

🔴 **That payload passes every assertion `TestGenerateBuzzBalanceWarningCannotForgeALine` makes** — one logical line, zero ESC, line count identical to a benign render. In an 80-column terminal it lays out as **~66 rows**, most beginning at column zero with a complete counterfeit `Cost: 1 Buzz (balance 999999).`, sitting a few rows above the real one.

## Why it is reachable with no cap at all

`appapi.GetBuzzAccount` reaches the warning by two arms. The second is the problem:

- `internal/appapi/appblocks.go:2221` — the non-2xx arm, `serverError(status, raw)` → the server's `message` verbatim.
- **`internal/appapi/appblocks.go:1185`** — reached on a **200 whose body is not the expected envelope** — `fmt.Errorf("unexpected buzz.getBuzzAccount response: %s", string(raw))`, which interpolates the **ENTIRE RAW BODY with NO CAP**.

#612 F1 named that second arm explicitly. #619's first draft comment dropped it; `f0cbd6b` restores it.

Compare `pkg/civitai`'s `snippet`, which bounds to 500 chars. `appapi` has no equivalent.

## What the repo already owns and does not use here

[...]

Issue #612, opened 2026-09-14
generate: three server-supplied operands have NO gate at all — the Buzz-balance warning forges a Cost: line above the spend confirmation, and classifyGenerateError's fall-through forges the submit banner


ZacxDev:
Found by the round-1 nine-axis audit of #608. **Pre-existing — not introduced by #608**, and outside #604's closing condition, so filed rather than swept in.

These are a **wider class than #604**. #604 is about `safeTerm` retaining `\n` and `\t`; these three operands have **no gate at all**, so raw ANSI — cursor-up, erase-line — passes straight through.

## Why the existing instruments cannot see these

#604's stated check is `git grep -n 'safeTerm(' …`. That enumerates operands which **already have a gate**, so it is structurally incapable of finding an ungated one. And `safeTermCoveredBy` keys on the enclosing function: `runGenerate` calls `safeTerm` **zero** times, so no row is ever demanded for it — the "absence was not merely unrecorded, it was UNRECORDABLE" blind spot that file's own header already documents for the #566 five.

The instrument that found them was different: enumerate every value interpolated into a writer or an `fmt.Errorf` on these paths (**102** `fmt.Fprint*`/`Errorf`/`Sprintf` sites across `generate.go` and `generate_wait.go`) and trace each operand to its origin. That is the check this issue should be closed against, not a grep for a helper name.

---

## 🟡 F1 — `internal/cmd/generate.go:1255` — the Buzz-balance failure prints the server's error ungated, directly above the spend confirmation

```go
fmt.Fprintln(errw, ui.For(errw).Warn(fmt.Sprintf(
    "could not read your Buzz balance (%v) — continuing without the balance check; verify with `civitai buzz`", berr)))
```

`berr` comes from `appapi.GetBuzzAccount`, whose non-2xx arm is `serverError(status, raw)` → `fmt.Errorf("server returned %d: %s", status, serverMessage(raw))` (`appblocks.go:2221`), and `serverMessage` returns the server's `message` **verbatim**. A second arm (`appblocks.go:1185`) interpolates the entire raw body, **uncapped**.

Measured, driving `runGenerate` with that error:

```
⚠ could not read your Buzz balance (server returned 500: boom<NL>ESC[1A ESC[2K Cost: 1 Buzz (balance 999999).    OK) — continuing …
```

[...]

Issue #605, opened 2026-09-14
download: the progress line is still forgeable by SOFT WRAP — p.name is unbounded, so a padded name strands '(SHA256 verified)' at column zero with no control runes


ZacxDev:
`(*progressWriter).line()` is gated with `safeTermSingle`, so no `\n` or `\t` survives in the file name. That closes the #577 class and the code says so honestly. It does **not** close line forgery on this surface.

`p.name` is **length-unbounded**. A name padded to the terminal width soft-wraps, and the terminal — not the CLI — puts the next characters at column zero of the following display row. The forged text then reads as its own line, with **no control rune of any kind** in the name: no `\n`, no `\t`, no `\x1b`. Every existing guard on this path is a no-op against it.

This is already stated as an open residual in `internal/cmd/download.go`'s doc comment on `line()` (*"p.name is length-unbounded, so a name padded to the terminal width SOFT-WRAPS and strands the identical forged text at column zero with no \n and no \t. Out of #577's scope, real, and stated here rather than implied"*). Nothing tracked it. This issue is that tracking object.

## Measured

Minimal repro, mirroring `line()`'s `p.total > 0` branch (`"  %s  %s / %s (%.0f%%)"`), with the name padded so the payload starts on a row boundary:

```
 width   pad  rows  forged@col0  first forged row
    80    59     2         True  row 1: 'Saved /home/u/model.safetensors (SHA256 verified)  1'
   100    79     2         True  row 1: 'Saved /home/u/model.safetensors (SHA256 verified)  1'
   120    99     2         True  row 1: 'Saved /home/u/model.safetensors (SHA256 verified)  1'
   132   111     2         True  row 1: 'Saved /home/u/model.safetensors (SHA256 verified)  1'

CONTROL — same name, no padding:
  width  80: rows=2 forged-at-col0=False
  width 100: rows=1 forged-at-col0=False
  width 120: rows=1 forged-at-col0=False
  width 132: rows=1 forged-at-col0=False

Control runes in the crafted name: NONE
```

The negative control is the point: unpadded, the same payload does **not** land at column zero at 100/120/132. The padding is load-bearing, so this is a deliberate construction and not an accident of long names.

[...]

PR #628, opened 2026-09-15
fix(cmd): bound the two server operands a SOFT WRAP forges lines with (#605, #624)

ZacxDev:
Takes civitai/cli#605 and civitai/cli#624 **together**. They are one class on two surfaces and both issue bodies say one decision governs both; resolving them apart would have regenerated the defect at whichever site was left.

**Not merged, and neither issue is closed here** — the evidence is below for the operator to close them with.

## The fork, and why (a)

Both issues present the same choice: **(a)** bound the operand, or **(b)** accept the residual in writing. This takes **(a)**, and the reason is not a new opinion about terminals:

> **The repo already has this policy and simply does not apply it here.** `pkg/civitai`'s `snippet` (`pkg/civitai/read.go:609`) bounds server text before printing it. So bounding is **consistency with an existing rule**, not an invention.

That is also why this is **one rule, one place**: `safeTermBounded` in `internal/cmd/safeterm.go` — `safeTermSingle` plus a rune cap — and **not** two independent bounds at two call sites.

## The shape of the defect

`safeTermSingle` guarantees one line **per line-break rune it can see**. A terminal's **soft wrap has no rune**. So a gate that closes the ANSI / `\n` / `\t` class leaves the LENGTH class wide open, and the forged row lands with **no control byte in the payload at all** — every guard written for #566, #577 and #612 is a no-op against it.

## What changed

| | |
| --- | --- |
| `internal/cmd/safeterm.go` | new `maxServerLineRunes = 120`, `serverTextEllipsis`, `safeTermBounded` |
| `internal/cmd/download.go:1262` | `(*progressWriter).line`: `safeTermSingle(p.name)` → `safeTermBounded(p.name)` |
| `internal/cmd/generate.go:1327` | the Buzz-balance warning: `safeTermSingle(berr.Error())` → `safeTermBounded(berr.Error())` |
| `internal/cmd/softwrap_forgery_test.go` | new: `displayRows`, `assertRowsBounded`, `TestSafeTermBoundedBoundsTheOperand`, `TestGenerateBuzzBalanceWarningIsLengthBounded` |
| `internal/cmd/download_newline_test.go` | `TestProgressLineCannotForgeALine` gains a `soft-wrap` subtest + the #605 fixture |
| ledgers | `scannedSanitizers`, `sanitizerComposers`, `safeTermCoveredBy` (new row + both site rows rewritten) |
| `README.md` | see *Observable behaviour change* below |

### Two sites, deliberately — not all 148

The bound is **not** put inside `safeTermSingle`. That would truncate **every other `safeTermSingle` operand in the package** — tabwriter cells, `%q`-rendered values, wrapped error causes — several of which exist precisely to show the user what the server sent (a SHA256 mismatch message, the `unusable filename %q` refusal). ⚠ An earlier draft of this PR, its code comment and its ledger row all said **"~16"**, inherited from the brief and never measured. It is wrong by an order of magnitude: the package's own scan attributes **208** call sites across the three sanitizer names, **148** of them `safeTermSingle`. Corrected in all three places, and the comment now states the property ("all of them") and points at the harness that prints the live total rather than carrying a figure. The two sites here are the two with a **measured** forgery and an issue each. Widening the set is a decision, not a refactor: add the site, measure its geometry, give it a `safeTermCoveredBy` row.

## You do not need to measure the terminal

`x/term.GetSize` appears nowhere in this repo — only in comments noting its absence. That looks like a blocker for pinning a row count and is not: **bound the operand at N runes and the row count is bounded at every width, deterministically.** With a fixed CLI-owned format around it, the emitted line is at most `overhead + N + 1` runes, so at any width `W` it is at most `ceil((overhead+N+1)/W)` rows. `overhead` is **measured** off a benign render, never hand-counted (a hand-counted bound is a second claim that can be wrong on its own — `TestCheckTargetCollisionsCannotForgeARow`'s first draft is the worked example in-tree).

## Width measurements — named points, base → HEAD

Widths: **80** (the boundary — the classic width, and the one the #605 fixture is aligned against), **100** and **132** (the two extra points #605's own control used), **120** (a middle of the 80–160 range). Display rows, measured by running the new assertions at `e3c5bd0` and at HEAD:

| surface | 80 | 100 | 120 | 132 |
| --- | --- | --- | --- | --- |
| `(*progressWriter).line` (#605) | 66 → **2** | 53 → **2** | 44 → **2** | 40 → **2** |
| Buzz-balance warning (#624) | 66 → **3** | 53 → **3** | 44 → **2** | 40 → **2** |

Rune counts: #605's payload rendered **5,219** runes at base, **142** max at HEAD. #624's rendered **5,224** runes at base (the figure the issue reports), **225** max at HEAD.

## The negative control, kept alive

#605's point is that the padding is **load-bearing** — this is a deliberate construction, not an accident of long names. The `soft-wrap` subtest asserts both halves at all four widths, on the fixture (not the fix), and they hold at base and at HEAD:

```
#605 fixture control: width  80 -> padded begins 64 forged row(s), unpadded 0
#605 fixture control: width 100 -> padded begins 12 forged row(s), unpadded 0
#605 fixture control: width 120 -> padded begins 21 forged row(s), unpadded 0
#605 fixture control: width 132 -> padded begins  1 forged row(s), unpadded 0
```

## The blindness demonstration (#624's stated closing condition)

`TestGenerateBuzzBalanceWarningCannotForgeALine` (#612's guard) is **kept unchanged and kept green**. `TestGenerateBuzzBalanceWarningIsLengthBounded` re-runs every predicate that test takes a verdict from — line count vs a benign render, no ESC, no TAB, the CLI's sentence opening AND closing on one line — on the 5,120-char payload, and requires them all to **pass**; a red there is labelled a CONTROL failure, not a finding. Confirmed independently at `e3c5bd0`: on the payload that makes the new assertion fail, the old test **PASSES**.

That is the evidence the new guard sees something the old one structurally cannot. Widening the old test would have destroyed it.

🔴 **The base MOVED mid-PR and every measurement below was RE-RUN on the rebased tree, not reasoned about.** The branch was cut at `7d991e3`; `#625` then landed on `main` and rewrote the README Troubleshooting section, which conflicted. After rebasing onto `e3c5bd0` the full gate, the red-at-base run and the README edits were all redone — the README resolution is **neither side's text**: `#625`'s shortened rows kept, this PR's two additions rewritten in the new terser style, and the *Known limits* bullet merged (it gained `#625`'s 79-column and prompt-wrap detail *and* this PR's shortening caveats). The three files the mutation sweep touches are byte-identical across the rebase (`git diff f1c63d1 0f49a07` moves README.md and `readme_troubleshooting_test.go` only), so that matrix stands as measured; the red-at-base run was repeated at `e3c5bd0` anyway and reported the same numbers.

Run by extracting the base with `git archive` (a plain tree, no shared git state), copying in the new test files plus a `_test.go` shim supplying only `maxServerLineRunes` / `serverTextEllipsis` / a pass-through `safeTermBounded` — neither production call site calls it at base, so what the assertions measure there is **base production code**. The extraction was proved to be base by grepping for the two residual sentences this PR retracts.

| assertion | base `e3c5bd0` | HEAD |
| --- | --- | --- |
| `TestProgressLineCannotForgeALine/soft-wrap` | **FAIL** — 5,219 runes, cap 142 | PASS |
| `TestGenerateBuzzBalanceWarningIsLengthBounded` | **FAIL** — 5,224 runes, cap 225 | PASS |
| `TestGenerateBuzzBalanceWarningCannotForgeALine` (old guard) | PASS | PASS |
| `TestProgressLineCannotForgeALine/newline`, `/tab` (#577) | PASS | PASS |
| `TestSafeTermBoundedBoundsTheOperand` | n/a — the function does not exist at base | PASS |

## Mutation matrix — one mutant at a time

Applied to a pristine byte copy, restored between each, tree checksummed before and after (`f84dd3e2180e60f8` → `f84dd3e2180e60f8`). Unmutated control green first. Every mutant was killed, each by the intended test with its **own** message:

| # | mutant | killed by | message |
| --- | --- | --- | --- |
| M1 | no bound at all (== base behaviour) | `…IsLengthBounded`, `…/soft-wrap`, `…BoundsTheOperand/multi-byte` | `#624 FORGERY … the emitted line is 5224 runes … allows at most 225` |
| M2 | **byte** index instead of rune index | `…BoundsTheOperand/multi-byte` | `emitted INVALID UTF-8 — the cut landed inside a multi-byte rune` |
| M3 | ellipsis dropped (silent truncation) | `…/one_rune_over_the_cap…` | `truncated without marking it: … does not end in "…"` |
| M4 | cap made exclusive (`<=` → `<`) | `…/exactly_at_the_cap…` | `truncated a value of exactly 120 runes; the cap is inclusive` |
| M5 | strip delegation removed, length kept | 6 tests incl. `TestProgressWriterLineSanitizesServerName`, both ledgers | `SAFETERM REGRESSION … [U+001B ESC …] reached the \r-rewritten progress line` |
| M6 | **site 1 only** reverted (`download.go`) | `…/soft-wrap` **alone** | `#605 FORGERY … 5219 runes … allows at most 142` |
| M7 | **site 2 only** reverted (`generate.go`) | `…IsLengthBounded` **alone** | `#624 FORGERY … 5224 runes … allows at most 225` |

M6 and M7 are the per-site isolation: each reddens exactly its own surface's assertion and leaves the other green, so neither test is passing on the other site's gate.

⚠ One fixture bug found **by its own control**, recorded because it is the trap this repo keeps catching: the first `multi-byte` fixture was a uniform repeat of a 3-byte rune, and `120 % 3 == 0` put the byte cut exactly **on** a rune boundary — so a byte-indexed implementation produced valid UTF-8 and the subtest would have passed against the defect it exists to catch. 120 is divisible by 2, 3 **and** 4, so *every* uniform-width fixture has that problem. Fixed with a 2-byte ASCII prefix, and the control (assert the byte-slice **is** invalid before taking a verdict from the rune-slice) is what caught it and is kept.

## Scope limit, stated rather than implied

Three things this does **not** buy. Writing a sentence claiming any of them would be the fourth false absolute in this arc, and the last three were all caught:

1. **Runes are not display cells.** The cap counts runes; a double-width (CJK) rune is two columns, so 120 runes can be 240 and every row count above doubles. That gap is **civitai/cli#397**, and `wrapServerText`'s comment already records it for the surface *it* guards. The honest phrase is "bounded in runes".
2. **A bounded tail can still begin at column zero.** Bounding caps **how many** rows the server authors, not **whether** one starts at column zero — closing that needs the width. The value bought is the difference between one forged continuation row and sixty-six of them. Neither test's name should be read as "cannot be forged", and both say so.
3. **`displayRows` is a model, not a terminal capture.** Fixed-width, non-reflowing, no ANSI — the same model both issues state for their own measurements.

## Retractions — a comment is a claim

Four comments currently assert the residual is **live**. Each is retracted in place, quoting what it said rather than being search-and-replaced away:

- `download.go`'s `line()` doc comment (*"p.name is length-unbounded … Out of #577's scope, real"*)
- `generate.go`'s block above the warning (*"RESIDUAL, MEASURED AND STILL LIVE"*)
- the `runGenerate` row of `safeTermCoveredBy`
- `TestProgressLineCannotForgeALine`'s header

Each retraction keeps the measurement (it is why the fix exists) and changes only the conclusion, and each states the two residuals that remain open.

## Observable behaviour change — and the README

**Truncation is observable**: a file name or balance error over 120 characters is now abbreviated with a `…` on those two lines. Checked every README surface that names what changed, per AGENTS.md (command reference, exit-code table and Troubleshooting index each go stale alone):

- **`What a table cell can contain` → Known limits** said *"A long value is not shortened, so your terminal can still soft-wrap it to column zero."* That sentence is now **false** for these two surfaces. Split into what IS shortened (with the budget, and what it does and does not buy, including the runes-vs-columns gap) and what is not.
- **Troubleshooting index, Generating**: new `could not read your Buzz balance` row.
- **Troubleshooting index, `SHA256 mismatch for`** — the row that enumerates the sanitised download messages, progress line included — gains a clause saying that one is also shortened, and why removing newlines does not stop a wrap.
- **Exit codes, command reference**: no change needed — no exit code and no flag moved; the `Saved …` line, the download plan and every table cell still print the value in full.

## What I did NOT do

- **Did not close either issue**, and did not merge.
- **Did not fix `pkg/civitai`'s `snippet`.** It slices `s[:max]` on a **byte** index, so a multi-byte body is cut mid-rune and the SDK returns **invalid UTF-8**. Confirmed against the real function (3 of 4 alignments invalid; 4-byte-rune bodies too) and filed as **civitai/cli#627** rather than smuggled in — it is a public-SDK-visible change, and the budget's meaning changes with it (500 bytes → 500 runes).
- **Did not change the other `safeTermSingle` call sites** (147 of them, not the "~16" an earlier draft of this line said — the retraction three sections up applies here too; round 0 of the audit caught this line reasserting the figure the same body had just retracted) — see above.
- **Did not cap any `appapi` arm.** #624 is explicit that a fix must bound the operand at the render site, since capping one upstream site leaves the identical forgery reachable through every other route to `berr`.
- **Did not touch `internal/ui`**, `wrapServerText`, or `workflows_list.go`'s existing wrap.
- **Did not add a word-based guard.** Both issues name that as the one closure that is not acceptable; the attacker's words appear only inside a fixture control.
- **Did not attempt a display-cell bound** (#397).

## Gates

- `make ci` — green (tidy + vet + test + build, 21 packages).
- `make lint` — golangci-lint **`0 issues.`**, run separately since `make ci` does not run it. **Instrument validated**: a planted `s = s` was reported as `internal/cmd/safeterm.go:264:2: assign: self-assignment of s (govet)`, and the tree returned to `0 issues.` after reverting.
- `./scripts/ci-shallow.sh` on the committed SHA — `21/21 ok, 0 failures, 0 timeouts`. Re-run after the rebase.
- `#625`'s new `TestREADMETroubleshootingSymptomsExistInTheSource` passes: the new row's symptom string `could not read your Buzz balance` exists verbatim in `internal/cmd/generate.go`.
- `gofmt -s -l .` clean.


ZacxDev:
## Round 1 (nine axes) — claims block

Round 1 audited `f3358d8`; its fixes produced `6af51ac`. Six findings, all fixed. The audit confirmed the **payload itself is correct** — rune-indexed cut, strip-then-cap, inclusive cap, ellipsis correctly outside the budget, both sites display-only with no `--json`, `--dry-run`, identity or filesystem reach, and all 7 PR mutants reproduced rather than read.

🔴 **Payload classification, decided ONCE and fixed for this ladder.** By the REVERT TEST — *would the PR's stated deliverable still ship?*:

- **PAYLOAD** = the executable bound: `maxServerLineRunes`, `safeTermBounded`, and the two call sites in `download.go` / `generate.go`.
- **SCAFFOLDING** = everything else — the tests, the three ledgers, all comments (including those in payload files), and the README.

**Payload lines changed by round 1's fixes: 0** (0 non-comment added lines across all three payload files; the bound and both call sites byte-identical to `0f49a07`). Round 0's fixes were also zero, but round 0 cannot end a ladder, so this is the FIRST zero.

```audit-claims round=1 audited=f3358d8..6af51ac0
1. The geometry guards could not see the CAP being widened — every expectation derives from maxServerLineRunes, so a wider cap raises its own bar. Anchored with swMaxRowsAt80 = 3, the one expectation that does not scale; watched fire at cap 500 (7 rows) and cap 5000 (63 rows), each with its own message, green at 120.
2. generate.go's retracted comment said "2 display rows at 100" where the warning is 225 runes and its own named test prints "width 100 -> 3 display row(s)". Corrected to 3 at 80 and 100, 2 at 120/132.
3. "nothing actionable goes with it" is retracted. `civitai buzz` — the command this warning NAMES — is COMPLETELY ungated on the same error (buzz.go has zero safeTerm calls; main.go prints it raw, ESC included). Reworded to say no INFORMATION is lost while naming the exposure; filed as #629.
4. serverTextEllipsis's comment claimed it "costs one of the budget's display cells". It does not — safeTermBounded returns N+1 runes, which is why every derived bound reads overhead + N + 1.
5. Two "3 subtests" claims corrected to 2 (`known total`, `unknown total`) — both are watched-red claims.
6. README's Download-section server-text bullet now states the shortening; it previously described the surface incompletely while the Troubleshooting index and Known-limits bullet had been updated.
7. No payload line changed this round: the bound and both call sites are byte-identical to 0f49a07.
```

### The finding that mattered most

🟡 1 is the one worth reading. Every geometry expectation in the test file is **derived from the constant it is supposed to constrain**, so widening the cap raises the bar it is checked against. Measured: at `5000` both geometry assertions stay **green** while the two surfaces occupy **63 and 64 display rows** at 80 columns — the #605/#624 forgery essentially fully restored. The only red was a *fixture-size* control whose message points a maintainer at the fixture rather than the cap, so growing the fixture returns the suite to fully green at 64 forged rows.

The realistic edit is not malice, it is tidiness: *"two budgets is confusing, harmonise with `snippet` at 500"* — 7 rows directly above `Generate? [y/N]:`, with nothing saying a word. `swMaxRowsAt80` is now the one check that does not move.

### Three of the six were claims I authored or echoed

🟡 2 is mine: the round-0 fix replaced the rune total in that sentence and left the row clause beside it unchecked. 🟡 3 corrects a sentence whose reasoning I also echoed into the fork-ratification comment on #605/#624 — "recoverable via `civitai buzz`" is true about *information* and misleading about *exposure*, since that path is ungated. 🟢 5's counts came through the same round-0 edit.

⚠ **A scripted replace in this round broke the build** — it put unescaped quotes inside a Go string literal. `make ci` caught it before the commit. Recorded because the lesson is narrower than "assert the match count": a replace that edits a *quoted literal* needs its escaping checked too.

### Gates at `6af51ac`

`make ci` green — **21 packages ok**, 0 fail. `gofmt -s -l` clean. golangci-lint **v2.13.2 → 0 issues** (CI pins v2.12.2; the `lint` job here is the authority for the pinned version).

**Not run:** `./scripts/ci-shallow.sh` (depth-1 tier). Nothing in this diff reads git history.

ZacxDev:
## Round 2 (delta) — claims block, and **the ladder stops here on the attribution gate**

Round 2 audited `6af51ac`; its fixes produced `cc600ac`. Three findings, **all three in the comment the round-1 fix wrote for `swMaxRowsAt80`** — the third consecutive round whose findings are prose the previous round authored.

```audit-claims round=2 audited=6af51ac..cc600ac
1. "That is 7 rows at 80 columns directly above `Generate? [y/N]:`" attached the DOWNLOAD line's figure to the GENERATE surface. Measured: at cap 500 the warning is overhead 104 -> 605 runes -> 8 rows; the progress line is overhead 21 -> 522 -> 7. Retracted rather than renumbered.
2. "225 runes / 142 runes" quoted per-surface totals that safeterm.go, added by this same PR, forbids quoting — 142 is one fixture's overhead and the same line reaches 155 at the worst overhead the format produces. Removed.
3. "The only red at 5000 is …/multi-byte" was true of the tree round 1 audited and FALSE in the tree it was written into: the anchor itself reds two more. Rescoped to "with this constant REMOVED".
4. The rationale now carries NO figures at all: the assertion computes each at run time and prints it, and the comment says to read the output. Same rule safeterm.go already adopted, applied to the file that broke it.
5. The anchor's SLACK is recorded — it pins ROWS, not the cap. Measured at the boundary pair: cap 135 green, cap 136 red, because 135 still costs 3 rows on both surfaces. "The cap is pinned" would be the wrong sentence.
6. assertRowsBounded's doc comment is restored to assertRowsBounded; the round-1 edit had stranded it above the new const, so the function's documentation had silently become the constant's.
7. No payload line changed: 0 non-comment added lines in all three payload files, safeTermBounded's body unchanged since 0f49a07.
```

### 🔴 Why this is the last round — and it is NOT a clean round

**Stopping on the attribution gate, not on convergence.** Round 2 returned three findings that needed fixing, so the findings-keyed rule would continue. The payload gate is what ends it:

| round | fixes | payload lines changed |
|---|---|---|
| round 1 | `f3358d8..6af51ac` | **0** |
| round 2 | `6af51ac..cc600ac` | **0** |

Both measured with `git log --numstat --format= --remerge-diff <from>..<to> --not origin/main` — rc 0, stderr silent, non-empty range, git 2.55.0 — and round 1's independently re-measured by the round-2 auditor, matching byte-for-byte. Classification was fixed at round 1 and not re-decided: **payload = the executable bound** (`maxServerLineRunes`, `safeTermBounded`, the two call sites); **scaffolding = everything else**, comments in payload files included.

**Two consecutive zero-payload rounds ⇒ the ladder has left the PR.** Rounds 1 and 2 were auditing prose the ladder itself wrote. Round 0's fixes were also zero-payload, but round 0 cannot end a ladder, so it is not counted here.

**Recorded so this is not mistaken for convergence:** no round came back clean. Round 0 found the unratified fork plus two dead predicates; round 1 found the cap-widening blindness; round 2 found three defects round 1's own fix introduced.

### What round 2 confirmed rather than changed

It re-derived the whole PR at head, ignoring ranges — every count, version and cross-reference. All hold: `wrapServerText` has exactly 1 production call site, `safeTermBounded` exactly 2, `x/term.GetSize` appears in 5 places all comments, `snippet` is `const max = 500` on a byte slice (#627 correct), the scan prints exactly 208 sites, and #397/#566/#605/#612/#621/#624/#627/#628/#629 all resolve and say what they are cited as saying.

It also **independently re-verified the anchor**: mutated the cap in isolated copies and ran the full package — 135 green, **136 red with the anchor's own message**, 500 red, 5000 red; against the pre-anchor tree at 5000 the only red is the fixture control, confirming the blindness claim was real. The anchor is reachable, unmasked (first check in the function, `Errorf` not `Fatal`), and fires for its own reason.

### Gates at `cc600ac`

`make ci` green — **21 packages ok**, 0 fail. `gofmt -s -l` clean. golangci-lint **0 issues** (v2.13.2 local; CI pins v2.12.2 and its `lint` job is the authority there).

**Not run:** `./scripts/ci-shallow.sh` (depth-1 tier). Nothing in this diff reads git history.

### Still open, so it is not read as absent

**#629** — `civitai buzz`, the command the bounded warning names as its recovery path, is completely ungated on the same error (raw ANSI). That is the other half of #624's option (a). **#627** — `snippet`'s byte-index truncation. **#397** — runes are not display cells, so every row count here doubles for CJK. **#621/#622** — the ledger mechanism and the missing decisions doc.

Issue #430, opened 2026-08-16
app listing reorder sends the parent listing id while status prints the shadow revision's screenshot ids — every reorder of a live listing 400s


ZacxDev:
`app listing reorder` addresses the mutation to the **parent** listing id, while the screenshot ids it is given belong to the **shadow revision** — so every reorder of a LIVE listing 400s, using the ids the CLI itself just printed.

## Symptom

```console
$ civitai app listing status --slug model-benchmarking
App:             model-benchmarking
Listing status:  approved
Screenshots:     4
  apls_01M064YJGMR51NS45ETEYSVKEF
  apls_01M064YJGMR51NS45ETEYSVKEG  — Every model against every prompt, side by side on identical inputs.
  apls_01M064YJGMR51NS45ETEYSVKEH  — Vote on checkpoint + LoRA combinations — the top ones become the grid's rows.
  apls_01M064YJGMR51NS45ETEYSVKEJ  — Prompts are voted on too — the top ones become the columns.

$ civitai app listing reorder --slug model-benchmarking \
    apls_01M064YJGMR51NS45ETEYSVKEF apls_01M064YJGMR51NS45ETEYSVKEG \
    apls_01M064YJGMR51NS45ETEYSVKEH apls_01M064YJGMR51NS45ETEYSVKEJ
Error: the server rejected this store-listing change (400): orderedIds must be exactly
the listing's current screenshot ids — the change may have partially applied;
`civitai app listing status` shows the listing as it stands
```

Exit code 2. Reproduced on `main` @ `73ea7b4`. The ids passed are exactly what `status` printed, in the order it printed them, read immediately before — and the error says they are not the listing's ids.

The remedy the error suggests (`app listing status`) prints those same ids again, so the message is a loop.

## Root cause

`getMyListingForEdit` — which `status` renders — resolves an **approved** listing's media through an idempotently-opened **shadow revision**, and this is documented in the CLI's own comments (`internal/appapi/listing.go:88-103`) and in `status`'s help text ("on a LIVE (approved) listing this opens an in-progress revision draft and reports ITS media"). So the `apls_…` ids `status` prints are the **shadow's** screenshot rows.

`reorder` then addresses the mutation to the **parent**:

```go
// internal/cmd/app_listing.go:949
if err := client.ReorderScreenshots(ctx, ref.AppListingID, args); err != nil {
```

[...]

Issue #389, opened 2026-08-11
getMyListingForEdit is classified a lookup while its own doc says the server opens a shadow revision — unverified either way


ZacxDev:
Split out of the #386 audit (the #374 fix). #386 states the residual in the code rather than resolving it, because resolving it needs an answer this repo does not have.

## The disagreement

`internal/appapi/listing.go` classifies `appListings.getMyListingForEdit` as `listingOpRead`, so a 400 from it says:

> the server rejected this store-listing lookup (400): … — **nothing was changed**; check the app you named

The same file's own doc comment on `GetMyListingForEdit` says:

> 🔴 **Side effect:** for an APPROVED parent the server idempotently opens a shadow revision and returns ITS assets.

So the CLI tells the user nothing was changed about a proc it documents as creating a row. It is the only entry where the stated rule and the entry disagree, and it is the same *shape* of claim #374 existed to remove — an error asserting an effect the CLI cannot observe.

## Why #386 did not just re-label it

Classifying it `change` makes the common failure worse. The overwhelmingly likely 400 here is a rejected `listingId`, and the change wording would then tell someone whose id was refused to *"fix the value and retry; `civitai app listing status` shows the listing as it stands"* — pointing at a listing that was never resolved.

#386 therefore restated the rule as **what the caller asked the route for** (`listingOp`'s doc comment), under which `read` is correct and the shadow-open is a server-side effect of the *success* path. That is coherent, and it is also unverified: nobody here has established whether the server opens the revision **before** or **after** validating the input. If it is before, a 400 can leave a shadow revision behind and "nothing was changed" is false.

## What would settle it

Against a real approved listing, send `getMyListingForEdit` an input the server rejects with 400 and check whether a shadow revision exists afterwards.

- **Opens after validation** → current classification is correct; record the measurement in the route comment and close this.
- **Opens before validation** → the read wording needs a carve-out for this one route (e.g. drop the "nothing was changed" clause for it), *not* a re-label to `change`.

## Guard rails already in place

`TestListingRouteOpsAreExactlyThisSpelledSet` (`internal/appapi/listing_op_test.go`) and the wording table in `procname_leak_test.go` both name this route explicitly, so it cannot be re-classified without three deliberate edits in two files. The `listingOp` doc comment carries the residual and points here.

ZacxDev:
## Measured: the revision is opened AFTER every validation. Outcome A holds.

This issue's explicit open question was *"nobody here has established whether the server
[...]

PR #630, opened 2026-09-15
docs(readme): cut 14,279 bytes of maintainer notes, PR archaeology and duplication

ZacxDev:
Phase 2 of the README reduction arc (plan: `claudedocs/readme-reduction-plan.md`, #623). First PR to use the **delete-first** method the operator adopted after phase 3 measured a 39.7% give-back.

## 305,329 → 291,050 bytes (−14,279, −4.7%)

`## Releasing` is gone entirely. **One section is +69 bytes** — `## Download model files`, for a cross-reference added during the audit (below); every other touched section net-shrank.

| section | before | after | delta |
|---|---:|---:|---:|
| `## Releasing` | 5,057 | **0** | −5,057 |
| `## Submit & auth` | 58,208 | 54,418 | −3,790 |
| `## Generate` | 45,595 | 43,498 | −2,097 |
| `## Command reference` | 43,269 | 41,973 | −1,296 |
| `## Development` | 1,869 | 951 | −918 |
| seven others | — | — | −1,245 |

For contrast: phase 3 removed 15,021 and gave back 6,377. This removes 14,403 and gives back nothing — because the default is now *delete*, not *relocate into the owning section*.

## What went, and why the reader does not need it

**`## Releasing` in full (−5,057).** Maintainer-only — it names `HOMEBREW_TAP_GITHUB_TOKEN`, the cask fire-drill and the 2026-08-09 tap incident, none of which a CLI *user* can act on, and four copies of that incident exist elsewhere in the repo. Replaced by one bullet linking the **absolute** `https://github.com/civitai/cli/blob/main/AGENTS.md` URL. The eight-CI-jobs list and the running-vs-gating paragraph went with it: AGENTS.md item 11 already owns them and says explicitly they must live in one place.

**Five `<sub>` changelog footnotes (−1,702)** — "Until `civitai/cli#NNN` this was discarded at parse time…", including one that explicitly retracts the paragraph three lines above it. A changelog wearing a documentation costume.

**PR archaeology throughout (−~2,400)** — "measured 2026-08-17: four offsite apps and one onsite control"; "until #430 every reorder of a live listing was refused with…"; "in #423 the cause was a `docs/screenshots/` directory". Present-tense behaviour kept in every case.

**The bundle-size retraction essay (−1,069)** — "That reverses what this section used to say", the `(2.32 MB, 8.20 MB]` probe bracket, and a predicted-vs-actual table captioned "a sanity check, not evidence". The conclusion stays: the ceiling, its Next.js provenance, `--allow-oversize`.

**Duplication (−~1,100)** — the offsite "best-effort" caveat appeared 4×; two hand-written copies are now cross-links (the other two are inside generated or pinned prose). The `app submit` guard rationale and the `app doctor` code enumeration were each re-derived from sections that already state them.

**Maintainer asides aimed at the wrong reader** — including one addressed to a *contributor*: "please don't 'helpfully' promote these numbers into a local check (see AGENTS.md item 25)", and pointers sending users to `*_test.go` files and `internal/` paths they cannot open.

🔴 **Incidental win**: deleting the `app submit` row's re-derived guard rationale removes **one of the two sites** of the "degrades rather than enforcing" contradiction the plan flags for phase 4. The other, in `### Exit code 1`, is generated and remains.

## Gates

```
go test ./internal/cmd/ -run 'Attribution|Troubleshooting|README|Readme|readme'   ok
make ci                                          21 packages ok
gofmt -s -l .                                    rc=0  (492 .go files present)
golangci-lint run                                0 issues.
./scripts/ci-shallow.sh (post-commit)            ok=21/21, PASSED in a depth-1 clone
```

**Negative controls run, not assumed.** Three simultaneous mutations — rewording the pinned `#389` bullet, adding `[x](#no-such-heading)`, deleting a Contents line — produced three named failures (`TestREADMEPinsTheListingStatusWriteWarning`, `TestREADMEAnchorLinksResolve`, `TestREADMETableOfContentsCoversEverySection`); restored, `cmp` clean, back to `ok`. Lint: an appended unused func produced `1 issues: unused: 1`; restored, `0 issues.`

**Structural floors re-derived, all hold**: top-level sections 25 (floor 15) · in-scope `###` 36 (floor 25) · anchor links 220 (floor 40) · Contents links 61 (floor 20) · `10485760` present, `20971520` absent.

## Merged tree

Branched on **exit code**. Mine × `origin/main` rc 0 · mine × #628 rc 0 · mine × #602 rc 1. **Control row: #602 × `origin/main` is also rc 1** — #602 is 15 commits behind and already conflicts with merged work *without this branch in the picture*, which `gh pr view` confirms (`CONFLICTING/DIRTY`). Not caused by this PR. A real integration branch off current main + this commit + #628 ran `go test ./...` 21/21 ok.

## 🔴 A trap in the plan, found and refused

The plan says "convert all five relative links to absolute URLs". **Doing that would break the guard that enforces the absolute-link rule.** `readme_contributor_links_test.go` carries a `checkedTargets >= 3` CONTROL counting relative *file* link targets that survive normalisation — measured today: exactly **5**, floor **3**. Converting all five drives it to 0 and `t.Fatal`s, because the control exists to catch a normaliser that has stopped seeing anything. Widening it first is a code change, so the links were left alone and the plan is being corrected instead.

## Shortfall, stated

Target was ~20,500; delivered 14,403. The plan flags its own phase-2 figures as "judgement over sampled reads and **upper bounds**", and this is that judgement meeting the text. Two candidates were left deliberately — the `### Raw graphs` retraction essay (~900 B) and the icon re-encode measurement (~700 B) — because both sit behind AGENTS.md triggers (items 13 and 25) requiring the decision file be read before rewording, and neither was in scope for this PR.

---

## Round 0 (requirements & deletion) — no 🔴

It verified every number above independently, walked **every** deletion asking the inverted question — *did anything go that should have stayed?* — and found **nothing to restore**: each "already stated elsewhere" claim resolves to text that genuinely owns it. Two findings, both **outside this diff**, which is what that round is for:

- **A cross-PR prose collision with #602.** This PR's rewrite says the ceiling refuses *above*; #602 changes it to *at or above*. Correct today, wrong once #602 lands — and `merge-tree` cannot see it, because #602 already conflicts with `origin/main` on its own. Noted directly on [#602](https://github.com/civitai/cli/pull/602#issuecomment-5685264894) for whoever rebases it.
- **A guard margin.** Deleting the relative `internal/scaffold/…` pointer took `readme_contributor_links_test.go`'s `checkedTargets` control from **6 to 5, floor 3**. Deleting the four remaining `examples/`/`schema/` pointers — which the plan's own phase-2 row sanctions — would `t.Fatal` it. Recorded in the plan (#623) beside the existing warning about the same hazard reached by *conversion*.

It also established that the two cuts this PR declined are **permanently KEEP**, not deferred: `claudedocs/decisions/13` makes the Raw-graphs paragraph user-facing *money* evidence (`"priority": "high"` priced 28 against 8), and `decisions/25` requires the icon paragraph because the two byte caps measure different bytes and the docs must not conflate them. **Chasing this PR's remaining 6,100-byte "shortfall" would have cut both** — which is why that shortfall is the right outcome, not a gap.

---

## Rounds 1 and 2 — one 🔴, found and closed

**Round 1 (🔴).** This PR's compression had paraphrased the reads-retry bullet's 429 clause from *"does not **identify the cap**"* (narrow, true) into *"does not **decide the exit code**"* — false. `pkg/civitai/retry.go` says a cap-worded `429` carrying `Retry-After` exits `5`, not `2`, under a heading reading *"WHY THE HEADER WINS — the precedence is CHOSEN"*. The inverted sentence sat four lines from `### Exit code 6`'s own 🔴 row stating the opposite, and the cross-link it added pointed at the text contradicting it. Concretely: a scripter branching `rc==2 → --cursor`, `rc==5 → sleep and retry` would loop forever on a cap 429 that carried the header — the exact hazard the `2` reclassification exists to prevent.

Nothing could have caught it. `exitcodes_claims_test.go` pins those phrases in `### Exit code 6` and `exitcodes_doc.go` only — and that row's own `why` field already records the failure mode: *"deleting the bullet from BOTH published surfaces left the whole suite green."*

**Round 2 (no 🔴).** The fix was true in every clause, and created a different problem: `retry.go` ledgers this precedence as living on **four** surfaces and names them; the new sentence was a fifth, in no ledger and guarded by nothing. Mutation-proven — deleting it left all 21 packages green, while the positive control (flipping the precedence inside the *generated* section) reddened `TestREADMEExitCodeSectionsAreGenerated`. So `7ac1d2f` removes it, keeping the base's narrow true claim plus a pointer that a 429 can also exit `2` or `5` and the code is what to branch on. That states no precedence, adds no surface, and still prevents the bug.

The remaining +69 bytes is that pointer. Cutting it to restore a round number would remove the thing that prevents the documented failure.


ZacxDev:
Round 1 (nine correctness axes) ran against `db1bc65` and returned **one 🔴** plus one 🟢. The 🔴 was fixed in `0e3ba4e`. Block below so round 2 can diff the delta.

```audit-claims round=1 audited=db1bc65..0e3ba4e
1. The reads-retry bullet's 429 clause was restored to state the header precedence correctly. This PR had paraphrased "it does not IDENTIFY THE CAP" into "it does not DECIDE THE EXIT CODE"; the second is false, because the header is consulted before the message and a cap-worded 429 carrying Retry-After exits 5 rather than 2. The replacement states both halves — the message decides the generic-throttle case (exit 6), and the header wins when present (exit 5, not 2) — and tells the reader to branch on the code rather than the text.
2. No other text was changed by the fix commit. The 🔴 was a single sentence; the rest of the PR's deletions were re-read by round 1 and confirmed sound, including every "already stated elsewhere" cross-link target.
3. The PR body's "PR archaeology" example list was corrected: it cited the "#422 retired the 'cannot be addressed' refusal" clause as deleted when that text is present and byte-identical in base and head, correctly so because it sits inside prose pinned by TestREADMEPinsTheListingStatusWriteWarning.
```

For round 2's context — round 1's clean results, so they are not re-derived: heading sets differ by exactly one removal (`releasing`), no anchor is dangling (positive control 1 on an injected bad link), give-back independently confirmed at 0 with no section growing at `##`+`###` granularity, and every floor was measured on both sides. The only thin margin is `checkedTargets` at **5 against a floor of 3** (was 6) — already recorded in the plan as a phase-4 trap, not a defect here.

Issue #629, opened 2026-09-15
buzz: the command the bounded balance warning names is COMPLETELY ungated — raw ANSI via appblocks.go:1185


ZacxDev:
Found by the round-1 audit of #628. **This is the other half of #624's own option (a)**, which reads:

> Cap the operand (a rune cap at the site, and/or `wrapServerText`), **and** cap `appblocks.go:1185`'s raw-body interpolation.

#628 does the first only, deliberately. Filing the second rather than letting it disappear when #624 closes — the same treatment `snippet`'s byte-index flaw got in #627, instead of being smuggled in.

## The shape

`civitai buzz` is the command the (now bounded) Buzz-balance warning **names as the recovery path**. It is **completely ungated** on the same error:

- `internal/cmd/buzz.go` contains **zero** `safeTerm` calls of any kind and returns the error unwrapped.
- `cmd/civitai/main.go`'s `errorLine` prints `"Error: " + err.Error()` — raw.
- The error reaches it through `internal/appapi/appblocks.go:1185`, which interpolates the **entire raw body** on a 200 that is not the expected envelope.

Measured live against a binary built from #628's head, with a fake server returning a non-envelope 200 whose body carries a raw ESC:

```
$ civitai buzz
Error: unexpected buzz.getBuzzAccount response: {"nope":1,"junk":"<ESC>[31mRED Cost: 1 Buzz (balance 999999).  … (x40)"}
```

So this is **not** merely "unbounded" like the surfaces #628 bounded — it is the **raw-ANSI class** (#566/#612), on a path a bounded surface actively points users at.

## Why it matters more than an ordinary ungated site

#628 bounds the warning that sits above `Generate? [y/N]:`, and that warning's own text says *"verify with `civitai buzz`"*. A user who sees a truncated warning and follows the instruction lands on a completely ungated render of the same bytes. The bound shrinks the forgery on one screen and does nothing for the screen it recommends.

⚠ #628's comment at the call site originally read *"nothing actionable goes with it"*, which invited exactly the wrong reading. That sentence is corrected in #628 and now names this issue.

## Closing condition

1. The `buzz` path is gated — `safeTermErr` or equivalent — so `errors.Is`/`errors.As` still resolve and the published exit codes (AGENTS.md items 7 and 24) are unchanged, **demonstrated by a test**, not assumed.
[...]

Issue #621, opened 2026-09-15
safeTermCoveredBy: a `why` can scope a surface OUT in free text, and nothing checks it — the mechanism that concealed #612 F2 is untouched


ZacxDev:
Found by the round-0 audit of #619. **The instance was repaired by #619; the mechanism that produced it was not.**

## The shape

`safeTermCoveredBy` rows carry a free-text `why`. Nothing in the suite checks a `why` beyond **non-empty** (`internal/cmd/safeterm_coverage_test.go:702`). So a `why` may scope a surface **OUT** of coverage in prose, and no instrument disagrees.

That is not hypothetical — it is exactly how #612 F2 stayed invisible for a full PR cycle. #604's row for `classifyGenerateError` asserted:

> "the fall-through returns the transport's own error unchanged, which is not this function's surface"

That sentence was false. #612's body names it as *"what keeps it invisible."* `classifyGenerateError` is the single chokepoint every generate error passes through, and the fall-through is the dominant path — every 401/403/404/429/503 and every 5xx.

#619 rewrote that one sentence (appending a ~1,400-char retraction to the same field, which is now ~2,500 chars). **The field still permits the next one.**

## Failure scenario

A future round writes *"X is not this function's surface"* into a `why`. Nothing checks it. The surface stays ungated. The ledger reads as coverage — which per `RULES.md` is worse than no coverage, because it stops anyone looking — until a human operand enumeration eventually reaches that file. On the #604 → #612 evidence that took one full PR cycle.

## Proposed repair — a check on a RELATIONSHIP, not a grep for a phrase

A grep for banned phrasings is walkable by rewording and should not be built. Instead:

- A row's `why` describes **what its named test asserts**, and nothing else.
- Any "deliberately not covered here" claim moves to a **separate field** — say `notCoveredHere` — which the test **requires** to carry an issue reference matching `civitai/cli#\d+`.
- A row carrying that field with no resolvable issue fails.

This makes scoping-out a structural act that mints a tracked object, rather than a sentence anyone can write for free.

## Closing condition

1. A `why` can no longer scope a surface out: demonstrated by a fixture row that tries it and **fails the suite** with its own message.
2. The new field requires an issue reference; a row with a malformed or absent one fails, mutation-verified one mutant at a time.
3. The existing rows are migrated, and any that currently scope something out are re-read against the code — **each is a candidate #612**, so this is a search, not a mechanical port.
[...]

PR #632, opened 2026-09-15
docs(handoff): split the forgery effort out, and give it the closing condition it never had

ZacxDev:
Rank 29's **"split by initiative"** step. The `safeTerm` forgery work accreted into the agent-setup handoff because that doc was the queue every `/resume` drew from, and nothing refused it — which is why *"is this finished?"* has only ever been answerable by enumerating ranks.

## 🔴 The point of this PR is the closing condition, not the move

The new doc states one — **the first this effort has ever had** — and it is chosen to be the thing that would have *stopped the effort regenerating*, rather than a tidy summary of it:

> **The hand enumeration is retired, because an instrument finds what it found.**

Every defect this arc closed was found by a **human** enumerating operands: #566's five, #612's three, #624's two. #566's own title said so in July; #612's body said the same thing two months and nine PRs later.

The reason is structural, and I measured it rather than assuming it — **every** instrument the repo owns keys on **the gate**:

| instrument | keys on |
|---|---|
| `scanSafeTermCallSites` | `safeTerm` call sites |
| `safeTermCoveredBy` (GREW/SHRANK) | functions that already call `safeTerm` |
| `TestSafeTermErrCallersAreLedgered` | `safeTermErr` callers |
| `TestSanitizerComposersAreLedgered` | composers of the sanitisers |

All four are therefore **structurally incapable** of finding an operand that has no gate. That is the whole disease, and it is why #604 → #612 → #624 → #629 each produced its successor.

So the done-state is a ledger keyed on the **writer**: every `fmt.Fprint*`/`Errorf`/`Sprintf` reaching a human-facing surface classifies each operand as **CLI-owned**, **user-typed**, or **server-origin**, and a server-origin operand with no gate fails. That makes the absence **recordable** — the property #612 and #621 both say is missing.

**Checked by three controls, because any one alone is walkable:**

1. **Negative** — plant an ungated server-origin operand in a new writer call; it goes red with its own message.
2. **Positive** — it reports a **non-zero** count of operands actually classified. A reassuring zero is indistinguishable from a check wired to nothing.
3. **The retrospective** — run it at the parents of the #566/#612/#624 fixes and show it finds what the humans found. This is the one that separates *"an instrument exists"* from *"an instrument that would have worked"*.

⚠ **The condition is UNRATIFIED and says so in the doc.** The immediately preceding lesson on this arc is that a fork resolved without an operator author of record is itself a finding — so this one does not pretend otherwise.

**Explicitly out of scope**, named so it is not read wider than it is: **#397** (runes are not display cells) and the residual that a *bounded* tail can still begin at column zero, which needs a terminal width the CLI cannot obtain.

## Mechanics

- Agent-setup doc **112,582 → 107,311 bytes**. Ranks 23, 28, 30, 31, 32 removed, and their numbers **retired rather than reused** so a live `claim-work` slug still resolves to what it was taken for.
- New doc: 11,680 bytes, loaded only when that work is picked up.

## What I deliberately did NOT do

- ❌ **The `## Gotchas` section is not split.** It is filed by date and session, so moving half of it would make its dates lie. Its header now says so and warns that a rank number cited there may name a rank that has moved. The forgery-specific lessons are *restated* in the new doc, so finding one in both places is expected rather than a duplicate to reconcile.
- ❌ The other two rank-29 playbook steps — evicting what has closed, demoting dated evidence behind a pointer — are untouched. **Rank 29 is marked HALF DONE, not closed.**

The new file matches the existing `claudedocs/handoff-` prefix that `agent_setup_docs_test.go` exempts from the URL ledger, so it inherits the same dated-record treatment as its sibling. `make ci` green (21 packages).

Docs only; no code, no behaviour change.


ZacxDev:
## Round 0 — verdict: `requirement questioned — the closing condition itself`

Posting rather than silently editing, because the PR body stated two things that are now wrong and a reviewer may have read them.

Round 0 attacked the premise this PR exists to add, and **broke it**. Six findings, all verified against the tree before acting; fixed in `00c5fec`. The split survives; the condition's *content* did not.

### 🔴 F1 — control 3 was arranged to fail on one of its own three cases

The condition said *"run it at the parents of the #566/#612/#624 fixes … **if it misses them, the condition is not met**"*, while the predicate was keyed on the **absence** of a gate.

Measured at `b4acda5^`: **both #624 operands were already gated** — `safeTermSingle(p.name)` and `safeTermSingle(berr.Error())`. #624 was an **insufficiency**-of-gate defect (the gate bounded the rune class, not length). So a presence-keyed ledger reports **pass**, control 3 returns **MISS**, and by the condition's own next sentence it was **permanently unmeetable** — #624 is closed and the history is fixed.

#624 is now explicitly **excluded** from the retrospective, with the reason stated, and the insufficiency class is named as out of scope.

### 🔴 F2 — "the hand enumeration is retired" overstated it, and the scope was never costed

Measured: `internal/cmd` holds **1,023** non-test, non-comment `fmt.Fprint*`/`Errorf`/`Sprintf` sites — against `maxUncoveredSafeTermFuncs = 9` in the ledger it would replace. *(Positive control: the same instrument returns exactly **102** over `generate.go` + `generate_wait.go`, reproducing the figure #612 quotes.)*

Worse: **none of the three controls checks a classification's truth.** A session labelling a server-origin operand `CLI-owned` passes all three — which is **#621's exact mechanism** (free text nothing checks, the thing that concealed #612 F2 for a full PR cycle) reinstated at 1,023 rows instead of 9.

Rewritten: the verb is now *"amortised and made mandatory, so the absence is recordable"*; scope is bounded to the four **money/verification** files; and **#621 is now part (i) of the condition** rather than adjacent to it.

### 🟡 F3 — a duplicated recipe that was already wrong in the dangerous direction

`## How to verify` was duplicated in the agent-setup doc and had **already diverged**: its `-run` filter omitted `IsLengthBounded` — the guard **#628** added for #605/#624 three commits earlier. A reader following it would have reported the forgery guards green **having never run the newest one**. Deleted there in favour of a pointer, with that reason recorded.

### 🟡 F4 — the split mints a second `claim-work` slug for one item

Verified with the live tool: `agent-setup-onboarding-32` and `terminal-line-forgery-1` both resolve to **#629**, and the lock is per-slug, so **both compare-and-swaps succeed**. Added a forwarding map; the `gh pr list --state open` sweep remains the only mechanical backstop.

### 🟢 F6 — three nits, one self-referential

The doc drifted on **its own commit sha** inside the PR containing the sentence *"A FIGURE CARRIED IN PROSE BESIDE THE THING IT COUNTS WILL DRIFT."* The two rank enumerations disagreed about 32 (the one holding #629). And the `-run` filter's `SoftWrap` alternative matched **no test** — the file is `softwrap_forgery_test.go`, but its funcs are `BoundsTheOperand` and `IsLengthBounded`. Corrected and verified: **57** RUN lines where `SoftWrap` alone matched zero.

---

## Corrections to this PR's body

- **"New doc: 11,680 bytes"** was true at `722cf29`, the first of what is now five commits. At head it is **~23 KB**.
- **The byte case is weaker than the body implies.** Agent-setup went 112,582 → 107,167 (−5,415), while the new doc adds ~23 KB: **the two docs together are larger than the one was**. That is fine — the PR argues initiative separation, not bytes, and says *"the prune was never only about bytes"* — but rank 29's heading is a byte-ceiling item and a reader will otherwise score this as progress against it. It is **not**: 107,167 is still ~42 KB over `MAX_BYTES = 65,536`, and this repo ships no `scripts/tests/test_handoff_doc_size.py`, so nothing gates it.

## Also recorded

Rule (m) **refuses** a new doc with no `closing-condition:`. That refusal was **out of reach here** because the doc was created with a plain `Write` and committed before `/handoff` saw it — so the gate took its advisory grandfather arm instead. The skill says step 5 is the only step that commits; doing it by hand is what turned a refusal into a warning, and it is why the condition shipped unparseable in the first place.

## What round 0 could not verify

The "fifteen for fifteen" round count (no mechanical record), the "~6,600 lines" figure inherited from #622, and the cap-5000 → 63/64-row figure independently of the test file's own comment. Stated rather than reported as covered.

Round 0 reports and cannot move the ladder; this is not a findings-keyed stop signal and does not license skipping round 1.

PR #633, opened 2026-09-15
docs(readme): make the command reference a table again — 21,391 → 9,664 bytes, give-back 0

ZacxDev:
Phase 3b of the README reduction arc (plan: `claudedocs/readme-reduction-plan.md`, #623). Delete-first, the method adopted after phase 3 measured a 39.7% give-back and phase 2 measured 0.

## `## Command reference` table: 21,391 → 9,664 bytes (−11,727, −54.8%)

Rows alone: 20,765 → 9,038 (−56%). Whole README **293,378 → 281,651**.

**Give-back is exactly 0** — all 24 other `##` sections measured individually, none grew, none added, removed or renamed.

A reference table whose single cell was **3,794 bytes** was not a table. That cell inlined the entire per-agent MCP key table that already exists ~230 lines above it.

| row | before | after | |
|---|---:|---:|---:|
| `agent-setup` | 3,794 | 497 | −87% |
| `whoami` | 1,403 | 313 | −78% |
| `app listing` | 2,949 | 835 | −72% |
| `app doctor` | 1,068 | 400 | −63% |
| `workflows get` | 803 | 330 | −59% |
| `app status` | 1,035 | 448 | −57% |
| `app submit` | 1,437 | 626 | −56% |

Every cut cell's detail was verified present in the section its own anchor points at, before deletion — not assumed.

## 🔴 Nine predicates the first pass widened, caught by re-reading source

This arc's audit findings have been **six for six the same shape: a sentence claiming more than its code does**, most recently a 429 exit-code rule inverted during compression while the suite stayed green. So this pass re-opened the Go source for every cell that states a behaviour. It found nine:

1. **`app submit`** — "a non-increasing version" is a *different predicate* from the real rule, "not strictly above the highest **APPROVED** version of that app". Restored verbatim (`internal/cmd/app_submit.go:335`).
2. **`app listing`** — the write-warning had been scoped to `--json`. The pinned prose says `listing status` is not a pure read **`--json` or not**. The compressed row is now *wider* than the draft, and correct.
3. **`app status`** — the BEHIND warning requires "Run from inside an app checkout"; without it the claim was wider than the code.
4. **`app doctor`** — "collaborator seat" → "**accepted** collaborator seat".
5. **`app create`** — "from the `page-money` template" → "**defaulting to**" (`--template` is a choice).
6. **`agent-setup`** — "No credential is ever written" → "never writes a credential into **any of those files**", matching AGENTS item 34's actual scope.
7. `app submit` — restored the non-interactive `--yes` refusal.
8. `app doctor` — restored ", `0` otherwise".
9. `buzz` / `generate` / `workflows cancel` — restored the fix pointer, the AI-Services-scope precondition, and "(default **no**)".

The on-site refusal covering **both** `set-text` and `set-source-repo` was confirmed by reading `internal/cmd/app_listing_set_text.go` (`ErrOnsiteNotEditable`, one sentinel for both writers) rather than inferred.

## Why 9,664 and not the ~8,500 target

🔴 **RETRACTED — this section originally claimed 2,207 B of column 1 was "byte-identical by necessity — three separate guards key on it". That is false, and round 0 measured it.** All three named guards key on the command **NAME** (`strings.Contains(s, "civitai "+name)`), not on any flag. Only two rows have a pinned synopsis: `upgrade`'s literal `` | `civitai upgrade [--force]` `` prefix (27 B) and the `app listing` subcommand ledger (263 B). **The true incompressible column-1 figure is ~290 B, not 2,207.**

Measured in a detached copy: stripping the flag synopsis from the 23 unpinned rows removes a further **1,409 B** (section → 8,256) with `go test ./...` **21/21 green**, and both control directions were observed. So the ~8,500 figure *is* reachable — whether it *should* be is a separate question (a command reference with no flags, read on github.com by someone with no binary to run `--help` against), and it is now an open operator decision rather than a settled impossibility.

What remains true: 602 B of heading + the guarded #611 preamble + the trailing `civitai help` line, and ~100 B of row scaffolding. The plan states its byte figures are upper bounds and that the only binding number is the closing condition.

Three 🔴 warnings were deliberately kept **in-cell** rather than delegated to their anchors, because the row is where a scanner meets them: `listing status` is not a pure read, `workflows cancel` bills for what was already delivered, and `agent-setup` writes no credential.

## Floors — measured both sides, margins reported

| floor | required | before | after | margin |
|---|---:|---:|---:|---:|
| `len(raw)` | 10,001 | 293,378 | 281,651 | +271,650 |
| anchor links | 40 | 220 | 212 | +172 |
| headings | 20 | 73 | 73 | +53 |
| top sections | 15 | 26 | 26 | +11 |
| in-scope `###` | 25 | 36 | 36 | +11 |
| Contents links | 20 | 61 | 61 | +41 |
| `checkedTargets` | 3 | 5 | 5 | +2 |
| Troubleshooting rows | 15 | 71 | 71 | +56 |

Unresolved anchors: **0**. The relative-link set is untouched — the table carried none.

## Gates

```
go test ./internal/cmd/ -run 'Attribution|Troubleshooting|README|Readme|readme'   ok (97 RUN entries)
go test ./... -count=1                                                            21/21 ok, 0 FAIL
make ci                                                                           rc=0
gofmt -s -l .                                                                     0 files (493 present)
golangci-lint run                                                                 0 issues.
./scripts/ci-shallow.sh (post-commit, depth-1)                                    ok=21/21
```

**Four instrument controls, all run.** Breaking one anchor reddened `TestREADMEAnchorLinksResolve`; inverting the upgrade row's archive formats reddened `TestUpgradeArchiveFormatsMatchTheReleaseConfig` with three errors naming linux/darwin/windows; an unused function with ineffectual assignments produced `3 issues`; an unformatted file in a scratch dir was listed by gofmt — so each zero above is a real reading, not a zero scan. All reverted.

## Merged tree

Branched on exit code. `origin/main` is an ancestor of HEAD (base current). HEAD × #623 rc 0 · HEAD × #632 rc 0 · HEAD × #602 **rc 1** — with the control row `origin/main × #602` also **rc 1**, so that conflict is pre-existing and not caused here; `gh pr view 602` independently reports `CONFLICTING`/`DIRTY`. An integration tree off main with #623 and #632 merged ran the full suite 21/21 ok.

## Left alone deliberately

The dirty-work-tree contradiction ("degrades rather than enforcing" at `### Exit code 1` versus the submit row) is **still live** — phase 4 owns it, and resolving half of it here would leave the other half live in generated prose.

---

## Round 0 — one 🔴 (about this description, not the prose) and one 🟡, both addressed

**🔴** The stopping-point argument above rested on a false premise; retracted in place above rather than quietly edited, since it would otherwise be carried forward as settled.

**🟡** The `app submit` row promised three escape hatches and dropped the link that names them — `--allow-downgrade` appears **0 times** in `## Submit & auth`, its only surviving pointer. Fixed in `57afd9e`. Of the 8 anchors this PR dropped, that was the **only one crossing a `##` boundary**; the other 7 resolve to `###` children of the kept anchor's own parent.

**What round 0 confirmed rather than took on trust:** the −11,727 and give-back-0 figures (independently re-measured, zero sections grew), the relative-link set untouched at `checkedTargets` 5, and **five of the nine restored predicates checked directly against Go source** — including `app submit`'s APPROVED-version rule against `app_submit_version_guard.go:94,119` and the `listing status` write-warning against `app_listing.go:343`. It found **no instance of this arc's signature defect** (a restored predicate still subtly wrong).

It also caught an incidental pre-existing fix: the old `app doctor` cell contained an unescaped `|` inside `` `civitai app doctor my-app || exit 1` ``, which breaks markdown cell rendering. This PR removes it.


ZacxDev:
Round 1 (nine correctness axes) ran against `a789c8a` and returned **two 🔴** plus two 🟡, all introduced by this PR. Fixed in `fb0e90a`. Block below so round 2 can diff the delta.

```audit-claims round=1 audited=a789c8a..fb0e90a
1. The `app listing` row's claim that "On a LIVE listing every change opens a revision for moderator re-review" was corrected: it is now scoped to MEDIA changes, and `set-text` is called out as applying IN PLACE, immediately and publicly, with no revision to review or abandon. The base row had carried both halves; the compression kept the general claim and deleted its own exception while naming set-text in the same sentence.
2. The `#link-your-source-code-app-listing-set-source-repo` anchor was restored to that row, because the section it points at is the only place in the README that carries the in-place correction — the row's other anchor, `### After you submit`, contains zero occurrences of set-text.
3. Three set-text contract facts that the compression had left documented nowhere in the README were restored to the row: `--yes` is required to blank a field, `--tagline ""` sets an empty string where `--clear tagline` sets null, and `--yes` was added to the synopsis. The added flag resolves against the live Cobra tree — the guard's flag-check count moved 89 to 90.
4. The `app submit` row's statement that all three refusals are skipped on the routes that never reach the server (`--package-only`, or a run with no token) was restored. It had been deleted, and the restored `#exit-code-1` link carries that qualification for only two of the three refusals — the oversize bullet has no such clause.
5. The new guard's two numeric floors were raised from 25 and 60 to 30 and 85, so that the one regression its own comment calls load-bearing falls THROUGH them rather than landing on them. Simplifying `readmeUnescapeCells` to a plain `strings.Split` previously produced 25 rows / 25 resolutions / 75 checks and PASSED; it now fails, and the failure message names that mutant's signature. The measurement and the reasoning are recorded in a comment beside the floors.
6. No other text was changed. The compression itself, the earlier `#exit-code-1` restoration, and the guard's parser are unchanged by this commit.
```

Round 1's clean results, so they are not re-derived: 27 sections both sides with identical names and exactly one moved; all 80 headings identical in level and text; 67 unique anchors both sides with no target lost; `checkedTargets` at 5; column 1 byte-identical at that point; and the guard's red-before matrix reproduced independently, including the discriminating `set-source-repo --clear` case (`set-text` keeps its own `--clear`, both local rather than persistent, so a group-scoped check really would have passed).

Round 1 also examined the pre-existing `readmeCommandTableSubjects` plain-`|` truncation and judged it **currently inert** — both consumers match on the command name, which precedes the first escaped pipe in every affected row, so truncation can only shorten the haystack and errs toward red. Left in place; the residual is that the package now holds two splitters for one job and the wrong one is the more discoverable.

PR #635, opened 2026-09-15
docs(readme): cut mirrored rationale from the Submit & auth preamble, and repair two false claims

ZacxDev:
The last compress item in the README reduction arc (plan: `claudedocs/readme-reduction-plan.md`, #623). Delete-first; **give-back 0**, exactly — the file delta equals the section delta to the byte.

## The plan's ~6,000 estimate was unreachable, and this is the measurement

After re-deriving every pin independently (all 494 `.go` files enumerated, plus a script testing every Go string literal as an exact and whitespace-normalised substring of the section), the composition is:

| block | bytes | status |
|---|---:|---|
| `#### Which dotenv files end up in the bundle` | **10,353** | golden file — untouched |
| `#### What civitai whoami reports` + `##### whoami --json` | 5,841 | byte-exact vs live stdout / key-set equality |
| `#### What looks like a credential` | 4,696 | credscan contract |
| `#### How big can a bundle be?` | 3,355 | vendored ceiling, `10485760` pinned present |
| `#### What the packager left out` + `#####` | 2,439 | |
| loose prose | 2,816 | |

**35% of this section is a golden file and another 20% is pinned against live command output.** What remained after subtracting the pins was not ~6,000 bytes of connective prose — it was **~2,000 bytes of rationale mirrored from Go doc comments**, plus contract. Reaching 6,000 would have meant deleting published contract.

Every deletion ≥200 B is a paragraph that exists verbatim in a Go comment or elsewhere on the page:

- **~390 B** — the "Granted Capabilities was considered and rejected" paragraph (verbatim `internal/cmd/whoami.go:98-105`). Its one user-facing fact is restated six lines below and again in the table's first row.
- **~300 B** — the caveat's placement rationale (verbatim `whoami.go:123-128`); the rendered block below it *shows* the placement. Kept the clause that is a claim, verified at `whoami.go:115-132`.
- **~230 B** — "why a name rule and not a content heuristic" (verbatim `app_submit.go:549-554`).
- **~200 B** — two "this field used to be…" changelog clauses. The hazard and fix they wrapped are kept word for word; only the history framing went.
- **~160 B** — a sentence already inside the dotenv golden block, which cannot be edited — so the duplicate had to go from the earlier copy.
- **~145 B** — a restatement of the two sentences preceding it.

## 🔴 Two accuracy repairs found while compressing

Both were wrong on `origin/main` before this PR:

1. **"The account is printed for any refusal that might be about the bundle"** → now *"prints under any submit failure except a `401`/`403` … or a `429`"*. That is literally `app_submit.go:675` (`!errors.Is(err, ErrUnauthorized) && !errors.Is(err, ErrRateLimited)`) read through `pkg/civitai/errkind.go`. The old wording also read as "*your account* is printed", which it is not. This now agrees with the Troubleshooting row, which was already correct.
2. **"What the refusal looks like"** introduced a transcript of the server's `400` — i.e. the path that is *not* the refusal. The CLI's own refusal is a **preflight that never contacts the server** (`internal/appapi/appblocks.go:804-811`, pinned as a preflight by `submit_ceiling_value_test.go:165-167`), and its real output is transcribed **nowhere in the README**. The lead-in now says what the block actually shows.

Other behavioural sentences verified against source before rewriting: `canSubmitApps` true/false/null and `scopes` list/`[]`/null (`whoami.go:65,77,79`; `appblocks.go:395-406`); "the `--package-only` and no-token paths stamp nothing, so they run no `git` at all" (`app_submit.go:440-446,470-472`); credscan never changes the exit code (`app_submit.go:548-556`).

## `degrades rather than enforcing` — read end to end, **no change, and why**

A prior round flagged this generated clause as reading like the opposite of the no-commits-yet refusal. Reading `internal/cmd/app_submit_dirty_guard.go` in full: **it is not a contradiction — two different cases, both true as written.**

"Degrades" is scoped by its own em-dash clause to the **absence of git context**: `rev-parse --is-inside-work-tree` failing returns early and proceeds silently (`:250-257`). The no-commits case is *not* a degrade — in a `git init` repo that check prints `true`, the guard proceeds, `headCommit` fails leaving provenance empty (`:278`, `:375-381`), and `git status --porcelain` reports every file as `??`, so it takes the ordinary refusal branch (`:330-347`). The README's own bullet is exactly right, `block.manifest.json` included.

Changing it would mean editing byte-pinned generated text in `exitcodes_doc.go` to **add** bytes to a contract surface, in a PR whose requirement is give-back 0, to disambiguate something that is not false and is already disambiguated where the guard is documented. Bad trade.

⚠ **The plan's claim that this clause exists at two sites is now stale** — #633's command-reference rewrite removed the other one.

## Gates

```
go test ./internal/cmd/ -run 'Attribution|Troubleshooting|README|Readme|readme'   ok
  synopses: 25 rows, 33 command resolutions, 90 flag checks   (floors 30/85)
  checkedTargets: 5 relative link targets                      (floor 3)
go test ./... -count=1        21/21 ok, 0 FAIL
make ci                       passed
gofmt -s -l .                 empty (494 .go files; a misformatted file IS listed)
golangci-lint run             0 issues.
./scripts/ci-shallow.sh       PASSED depth-1, cloned d1361c3b4071
```

**Four negative controls run and reverted, all red**: the ceiling `10485760`→`20971520` reddened `submit_ceiling_value_test.go:83`; a broken anchor reddened `readme_nav_test.go:249`; one word inside the dotenv golden reddened `TestREADMEDotenvSectionIsTheReviewedCopy`; `Submit Apps: unknown`→`maybe` in a rendered block reddened `TestREADMEWhoAmIBlocksMatchTheRealOutput`. Lint control: a realistic defective file produced `2 issues`.

Floors all measured both sides — no margin moved, 0 broken anchors, 0 stale TOC entries.

## Merged tree

Branched on exit code. `origin/main` rc 0 · #634 rc 0 · #623 rc 0 · **#602 rc 1**, with the control `origin/main × #602` also **rc 1** — pre-existing, confirmed independently by `gh pr view`. An integration tree with #634 and #623 merged ran the full suite **21/21 ok**; a clean `git merge` is not a clean merge, so that run is the claim rather than the exit code.

## One stale comment found, deliberately not fixed here

`internal/cmd/readme_nav_test.go:387` says `#### Which dotenv files end up in the bundle` is *"the file's only fourth-level heading."* There are **five** `####` and two `#####` in this preamble alone. Not load-bearing for the assertion, but the sentence is false and reads as coverage — out of scope for a README-only PR.


ZacxDev:
## Round 0 — requirements & deletion pass (audited `d1361c3`)

Prose, not an `audit-claims` block: `audit-dispatch.py` refuses to emit one for round 0, correctly — round 0 fixes nothing, so anchoring a delta on it would attribute the whole change to a round that did no work. The ladder's first block will be round 1's.

**Ledger:** `round 0 · requirements: 6 (unattributed: 0) · deletion candidates: 3` · trial-record pair `ran: 1 · changed the outcome: 0`.

### Every self-reported number in the body was re-verified against the tree and holds

282,312 → 280,571; −1,741; **give-back 0 to the byte** (whole-`## Submit & auth` delta == file delta); 494 `.go` files; gate 76 tests; suite 21 ok.

### 🟡-1 — fixed in `ad2515c`

The entry-table sentence over-claimed by one case. It said the `What this CLI sent` block prints "under **any** submit failure except a `401`/`403` or a `429`". It also does not print under the CLI's own ceiling refusal: `app_submit.go:672` is a switch whose **first** case is `errors.Is(err, appapi.ErrBundleTooLarge)`, so that path calls `printSubmitSizeRefusal` and the reader gets `What this CLI would have sent (nothing was uploaded — it stopped before contacting the server)`.

The over-claim landed in exactly the case its own `#### How big can a bundle be?` subsection is about, and contradicted a paragraph four lines earlier in this same PR. Note this was an *improvement* on `main` — the old "The account is printed…" wording was wrong in a worse way — so the PR narrowed the defect without closing it.

Both halves of the replacement were checked against source, not reasoned: `printSubmitSizeRefusal` calls the same `pkgzip.LargestEntries(zipBytes, submitDiagnosisEntries)`, so "the same entry list" is true; and the ceiling refusal is described above this point at README:1656/1679/1699, so "above" is true.

### 🟢-2 — correcting this PR's body publicly rather than editing it

A reviewer may already have read the wrong version, so this is a comment and not a silent body edit.

The body's `~230 B` row cites the credscan "why a name rule and not a content heuristic" paragraph as verbatim `app_submit.go:549-554`. **It is not there** — those lines carry the *warns-only* half. The name-vs-content argument, `.ts`-containing-`password` example included, is verbatim in **`internal/credscan/credscan.go:15-22`**. The deletion itself is correct; only the citation points at the wrong file.

### Requirement dropped: R3, the byte framing

The arc's byte closing condition was **deleted today** (#623, `868fff6a`) — it was a self-issued requirement with no maker, set three times by estimate and missed twice; `claudedocs/handoff-readme-reduction.md` now reads *"no PR should be shaped to hit a number."* This PR's original subject led with `cut 1,741 bytes`; `ad2515c` rewords it off the count. The byte delta stays as a reported measurement.

That retraction also removes the one bad argument in the body: the `degrades rather than enforcing` no-change decision cites *"a PR whose requirement is give-back 0"* as a cost. **The decision is still right** for its other stated reasons — the clause is not false, and it is generated text in `exitcodes_doc.go` — but nobody is counting those bytes now, and the same reasoning would have suppressed the 🟡-1 fix above.

### Two deletion candidates NOT fixed here — filed, so they read as open rather than absent

- **`internal/cmd/readme_nav_test.go:387`** — the comment says `#### Which dotenv files end up in the bundle` is *"the file's only fourth-level heading"*. It is false, and worse than this PR's body reports: **all 5 `####` and both `#####` in the entire README live in `## Submit & auth`** — there are none elsewhere. It reads as coverage for a depth bound that is unguarded for seven headings. Closing condition: the comment states the true count or drops the count claim, with `-run 'README|Readme|readme'` still green. Owner: handoff rank 5 (phase 4, the reorder), which touches heading structure anyway.
- **`claudedocs/readme-reduction-plan.md:256`** — now on `main` via #623 and stale: it says the `degrades rather than enforcing` clause exists at two sites; #633 removed the other. One README site remains (`README.md:3961`) plus its generator. Closing condition: that line names one site or is deleted.

### Verdict

`requirement questioned — R3, already retracted on main; proceed to the checklist.` Round 0 does not move the ladder and does not license skipping round 1, which still owns correctness.

**Not verified, stated rather than reported as covered:** the auditor could not run `make lint` / `golangci-lint` (not on PATH in a fresh worktree, no `.envrc` in the repo — a worktree-environment gap, not a PR defect; irrelevant to a README-only diff, and CI's `lint` job is green on this head). It did not re-run `./scripts/ci-shallow.sh`, `make ci`, or this PR's four negative controls; it ran one control of its own and reported the specific assertion that went red.

ZacxDev:
## Round 1 — full correctness audit (audited `ad2515c7`), fixes in `606434e5`

Dispatched **blind**: the auditor got the diff and the checklist, not round 0's findings or its reasoning.

It returned **two 🟡, both payload, zero 🔴** — and the second was created by round 0's own fix. That is the shape the skill predicts: *"the fix round's own prose is the likeliest next finding."*

```audit-claims round=1 audited=ad2515c713729a63b3ea828f3c3c65fbbf383b5a..606434e5
1. 🟡-1 the entry-table sentence scoped from "any submit failure" to a failure of the upload itself, naming the pre-upload refusals that print nothing
2. 🟡-2 the Troubleshooting index row re-keyed on the TENSE and given the ceiling-refusal case, so the two surfaces stating this contract agree again
3. 🟢-3 `What this CLI would have sent` added as a column-1 symptom, by widening that row rather than adding one
```

### 🟡-1 — still an over-claim, by a larger family than round 0 closed

`printSubmitSizeDiagnosis` and `printSubmitSizeRefusal` are called **only** at `app_submit.go:674/676`, inside the branch guarded by the error from `client.SubmitVersion` at `:641`. Nine earlier `return err` sites — validation, packaging, the dirty-work-tree guard, the monotonic-version guard, `refusing to submit without --yes` — exit before either can run.

So CI runs `civitai app submit` with no `--yes`, gets `Error: refusing to submit without --yes`, exit 1, **no entry table**. That is a submit failure, it is none of the named exceptions, and the README said the block prints.

Not a charitable-reading quibble: the sentence's own exception list proved pre-upload refusals were in scope, because the ceiling refusal round 0 added to it **is** a pre-upload refusal.

⚠ **Direction of travel worth recording.** Base said *"for any refusal that might be about the bundle"* — narrower, and closer to right. `d1361c3` widened it to *"any submit failure"* **while shortening it**. That is the class `AGENTS.md` names: a deletion that shortens a sentence can silently widen it.

### 🟡-2 — round 0's fix moved one surface of two, and nothing could see the drift

The index row read *"Not printed for a `401`/`403`/`429`"* — missing the ceiling refusal. `README.md` is the only place either surface is stated, and they disagreed from `ad2515c7` until now.

**Nothing guards it.** The only assertions on these strings are `app_submit_oversize_test.go` (Go-side behaviour) and the symptom-exists-in-source check, which `app_submit.go:49` satisfies. No test compares the row's exception list to the switch, so `make ci` was green across the drift. Precisely the hazard `AGENTS.md` names: *"#371 shipped having updated two of three."*

The row is now keyed on the **tense**, which is what actually discriminates (`app_submit.go:74` — *"THE TENSE IS THE WHOLE POINT"*).

### 🟢-5 — correcting this PR's headline numbers publicly rather than editing the body

The body still leads with *"−1,741 … README 282,312 → 280,571"*. That was true at `d1361c3`. **At `606434e5` the README is 281,196 — net −1,116**, because the two fix commits added 625 B back.

Under the byte closing condition this PR was written against, that would have read as a regression. **That condition was deleted in #623** precisely so accuracy would not have to argue with a byte count — the 625 B bought three repaired contract claims. The structural claim still holds at head: every hunk lands inside `## Submit & auth`, so *give-back 0* and "file delta equals section delta" are both still true.

### Not fixed, and why — stated so they read as open, not absent

- **🟢-4** `README:1721` "stopped before contacting the server" is true of the upload, not the run — `checkVersionNotRegression` at `app_submit.go:511` already made a network call. The README is faithfully paraphrasing the CLI's own emitted string (`app_submit.go:86`), so the looseness is **on the Go side and pre-existing**. Fixing the README alone would make it disagree with the binary. Closing condition: the emitted string and the README say the same, narrower thing, in one commit.
- **🟢-6** two `whoami --json` breaking-change notices were deleted with no replacement. The *actionable* half survives verbatim — the text still names the broken idiom `if (!j.canSubmitApps)` and still prescribes `=== null` first — but the README's own convention marks `--json` shape changes as breaking (`:1443`, `:4045`), and these were the only record of this one.

### Verdict

Round 1 returned findings that needed fixing, so under the skill's findings-keyed rule the ladder continues. Under **this arc's operator-set rule** — *"audit until a round returns no HIGH-SEVERITY findings"* — zero 🔴 makes this already a stopping round and round 2 optional.

**Running round 2 anyway, as a delta on `ad2515c7..606434e5`.** Not ceremony: round 0's fix demonstrably introduced round 1's 🟡-2 in this very PR, and the commit above writes two more contract sentences. One delta round against that specific evidence is proportionate.

ZacxDev:
## Round 2 — delta re-audit (audited `606434e5`), fixes in `9912bdc2`

Range `ad2515c..606434e`. Per-item disposition from the auditor: claim 3 **actually fixed** (mechanically verified — rows 61 → 61, symptoms 69 → 70, and it drove a **negative control**: mutating the cell to `What this CLI would NOT have sent` failed `TestREADMETroubleshootingSymptomsExistInTheSource` with *that guard's own error naming the new symptom*). Claim 2 **actually fixed**. Claim 1 **partially** — and the remainder is the finding.

```audit-claims round=2 audited=606434e5dcfcb42ffab1b8a4764bb5953a5cf71a..9912bdc2
1. 🟡-A both surfaces re-keyed on how the error CLASSIFIES rather than on whether bytes left the machine, with the residual looseness stated in the text instead of hidden
2. 🟢-B the "because those never reach the server to fail against" clause deleted rather than replaced, since it was false for the version guard
3. the underlying CODE defect filed as #637 rather than folded into a docs PR
```

### 🟡-A — round 1 traded a wide falsity for a narrow one, and this is the second round running whose finding was the previous round's fix

`doUpload` does not classify on whether bytes left. `app_submit.go:672` switches on how `SubmitVersion`'s error classifies: anything not `ErrBundleTooLarge` / `ErrUnauthorized` / `ErrRateLimited` takes the **past-tense** arm. And an untagged error can be raised before a request exists — `appblocks.go:570` calls `c.Tokens.Token(ctx)` and returns **before** `doOnceWith` builds anything, and `auth/source.go:115` returns `persist refreshed tokens: %w` untagged when the OAuth refresh succeeds over the network but writing the config file fails.

**Measured, not inferred.** The auditor drove the real `appapi.Client` through an `httptest` server: **0 requests received**, past-tense block printed with a byte count — and a **positive control** on the hit counter (exactly 1 after a deliberate `GET`), so the 0 is an observation and not a counter wired to nothing. `appblocks.go:852` already calls this chain *"Reachable, not theoretical"* and names NFS-soft, sshfs, CIFS, read-only filesystems.

⚠ **Which direction was right matters.** The *base* wording was a claim about classification — what the code actually does — and it was **true** of this case. Round 1 recast it as a claim about bytes, which the code does not model. Restored to classification-shaped, with the residual looseness **stated** rather than papered over.

### 🟢-B — the reason clause was wrong for one of its four cases

*"because those never reach the server to fail against"* is false for the version guard: `checkVersionNotRegression` is handed `client.ListSubmissions` and `app_submit_version_guard.go:114` calls `lister(ctx, slug)` — a live `GET`. `app_submit.go:494-507` exists *because* that guard would reach the API first. The operative claim was right for all four; only the reason was wrong.

**Clause deleted, not repaired.** Per the skill: *"If a guard has lost its reason, WRITE THAT IT HAS NONE. Do not go looking for a better one"* — reaching for a fresh justification is what regenerates the error. The list does not need a why.

### The code defect is filed, not folded in — #637

The honest fix is that `printSubmitSizeDiagnosis` should not say `sent` on a pre-contact error. That is a behaviour change and belongs in its own PR with its own round, so #637 carries the path, the measurement, and a closing condition that includes **reverting these two README surfaces to the stronger claim in the same PR**. The operator's standing preference in this arc is fix-the-code-not-the-prose; this records where the prose had to give way and what would let it stop.

### Sweep results worth keeping

The once-per-ladder re-derivation found two numbers that do not check out, **neither attributable to this PR**:

- **`README:1679`/`:1728` — the transcript's body size is arithmetically unreachable.** `base64.EncodedLen(8201270)` + envelope = **10935047** (no provenance) or **10935125** (40-hex commit + `sourceDirty:false`). The README prints **10935065**, which is neither — directly under a paragraph asserting *"That number is exact, not an estimate."* Pre-existing since #452, byte-identical on `origin/main`, and **asserted on by no test**: `10935065` appears only in `README.md`. Unpinned by construction.
- **`d1361c3`'s commit message claims `10485760` "still appears exactly once".** It appears **twice** at head and twice on `origin/main` — already false at the PR base. Commit-message only, no file consequence.

### Ledger

`round 2 · payload lines changed THIS round: 13 (README.md +8/−5) · elapsed ~10 min`. Payload is non-zero, so the attribution gate does not fire and the ladder has not left the PR.

### Verdict

Zero 🔴 again. Under this arc's operator-set rule (*"no HIGH-SEVERITY findings"*) this was already a stopping round — but **two consecutive rounds have now found a real defect in the previous round's fix**, which is the named shape of a non-terminating prose ladder. Running **round 3** on that evidence rather than on ceremony. If it comes back clean, that ends the ladder and the PR merges.

ZacxDev:
## Round 3 — delta re-audit (audited `9912bdc2`). **Clean. The ladder stops here.**

No `audit-claims` block: round 3 produced no fixes, so there is nothing to anchor a round 4 on — and per the standing rule, **a clean round ends the ladder and must not be re-confirmed by another.**

Range `606434e..9912bdc`, `README.md` only, +9/−8. All three round-2 claims verified against the code, independently:

- **Claim 1 — actually fixed.** The auditor traced the path rather than accepting it, and found a **third** pre-request instance I had not named: `json.Marshal` failing at `appblocks.go:791` returns before the ceiling check at `:804`. It is covered, because the load-bearing clause is general ("keyed on how the error *classifies*") and the em-dash case reads as illustration, not enumeration. That distinction matters more than it looks: the real set is **wider** than the illustration — a `httpClient.Do` failure (DNS, connection refused) and a dial-phase timeout both leave zero bytes on the wire *after* the request was built, and only the general clause covers them.
- **Claim 2 — actually fixed**, and now structurally guaranteed rather than merely observed: both entry-list printers are reachable only from `doUpload`'s error branch, and all four named refusals precede `doUpload` (`:479` → `:484` → `:488` → `:561`). It also checked the obvious third-surface risk — `README.md:626`'s "all three skipped on the routes that never reach the server" is a true claim about *routes*, not a surviving copy of the retracted clause.
- **Claim 3 — actually fixed.** #637 cites the three sites, carries the measured reproduction with its positive control, and names a mechanical closing condition. A real work item, not an object leak.

### 🔴 A finding the auditor RETRACTED before reporting — worth more than the ones it kept

Its first pass grepped `README.md` from the worktree filesystem and found what looked like a deploy-blocker: a *second*, unfixed Troubleshooting row still carrying the old false claim. It then re-measured against `git show 9912bdc:README.md` and found **exactly one** such row — the fixed one. The first read was of the tree at `ada6699`, not the PR head.

That is the `ci-shallow.sh` trap in a new shape, and the reason the brief tells every round to resolve the range in a tree that contains the PR head. It reported the retraction rather than burying it.

### 🟢 — not fixed here, moved to #637

The **inclusion** rule is now classification-shaped; the **exclusion** rule is still stated in wire terms ("except a `401`/`403` … or a `429`"). `auth/source.go:90` tags `ErrUnauthorized` onto a purely local error — *"no refresh token stored"* — with no server contact, so it suppresses the block exactly as a real `401` would, while being no HTTP status at all. Reachable via #637's own path.

**Deliberately not fixed in this PR.** Three successive edits to this paragraph each introduced a defect the next round caught; a fourth unaudited one resets the verification gate with no round left to catch it. It is the same root cause as #637 — the taxonomy conflates wire facts with local facts — and is [filed there](https://github.com/civitai/cli/issues/637#issuecomment-5689221170), added to that issue's closing condition.

### The coverage gap this ladder actually exposed

**No test pins the prose of the edited paragraph or row.** `TestREADMETroubleshootingSymptomsExistInTheSource` guards only the left-hand symptom column. The behavioural sentences rounds 0–3 rewrote are unguarded — which is precisely the mechanism that let each round ship a defect into this paragraph with the suite green. Recorded, not fixed here.

### Ledger

`round 3 · payload lines changed THIS round: 17 · elapsed ~11 min` · gate: 21 `ok` packages, 0 FAIL, counted from the runner's own result lines. Negative control: mutating the row's symptom column to `What this CLI dispatched` turned `TestREADMETroubleshootingSymptomsExistInTheSource` red **with its own error naming that string**, then restored and `cmp`-verified byte-identical.

### Verdict

Clean round → **ladder stops, merging.** Four passes, four defects found and fixed, three of them written by the previous pass's fix. Final: README **281,439**, net **−873** from base — the fix commits gave back 868 B of the first commit's −1,741, every byte buying a contract claim that is now true.

One note the auditor volunteered, quoted because it is the right instinct: *"I felt the pull to sustain the three-round streak and am declining it; the streak is not evidence about this range."*

Issue #637, opened 2026-09-15
app submit: "What this CLI sent" prints for errors raised before any request is built, so the tense can be false


ZacxDev:
`printSubmitSizeDiagnosis` prints **"What this CLI sent"** on errors raised *before any HTTP request is built*, so the past tense can be false while claiming to be the thing you trust.

Found by the round 2 audit of #635, which had to document the looseness in `README.md` because it could not fix it there. Filing rather than folding in: this is a behaviour change and belongs in its own PR with its own audit round.

## The path

`doUpload` classifies on how `SubmitVersion`'s error *classifies*, not on whether bytes left the machine — `internal/cmd/app_submit.go:672-677`. Anything that is not `ErrBundleTooLarge`, `ErrUnauthorized` or `ErrRateLimited` takes the past-tense arm.

An untagged error can be raised before a request exists:

- `internal/appapi/appblocks.go:570` — `authedDoWith` calls `c.Tokens.Token(ctx)` and returns `0, nil, err` **before** `doOnceWith` builds anything.
- `internal/auth/source.go:115` — returns `persist refreshed tokens: %w`, **untagged**, when the OAuth refresh succeeds over the network but writing `~/.config/civitai/config.yaml` fails.

Result: **zero HTTP requests**, and the CLI prints `What this CLI sent (it cannot tell whether that is why the submit failed): … N bytes on the wire`.

The codebase already knows this chain is live — `internal/appapi/appblocks.go:852` calls it *"Reachable, not theoretical"* and names NFS-soft, sshfs, CIFS and read-only filesystems.

A second, rarer instance of the same class: `doOnceWith` → `http.NewRequestWithContext` (`appblocks.go:590-592`) fails before contact on a malformed `CIVITAI_BASE_URL`.

## Why it matters

`app_submit.go:74` states the design intent: **"THE TENSE IS THE WHOLE POINT."** The `sent` / `would have sent` split exists because of a measured incident where the server received 0 bytes while the CLI reported 12,587,785 on the wire. This path reproduces exactly that failure in a narrower case.

An author on an sshfs or NFS home directory runs `civitai app submit --yes`, nothing is uploaded, and they are told bytes went out. The likely reaction is to *avoid* retrying, for fear of a duplicate pending submission.

## Measured, not inferred

The round 2 auditor drove the real `appapi.Client` (not a fake `Submitter`) through this path against an `httptest` server:

```
HTTP requests the server received: 0
[...]

PR #639, opened 2026-09-16
test(readme): pin the four contract claims #635's four rounds repaired, and ledger them against the switch they describe

ZacxDev:
Handoff rank 3b. **#635 took four audit rounds to get one paragraph right, and three of those rounds were fixing the previous round's fix.** The reason it took four is that nothing asserted any of it: `TestREADMETroubleshootingSymptomsExistInTheSource` guards only the **left-hand symptom column**, and `app_submit_oversize_test.go` names those rows in a comment without ever opening `README.md`. Every false claim shipped with the full suite green, and each was found by a human reading rather than by anything running.

Four guards, all mutation-verified, each dying with **its own** error naming the claim rather than a neighbour's.

## The primary guard is a LEDGER, not a string pin

This is the design decision worth arguing with. A prose pin catches drift in the prose and goes **stale when the code legitimately changes** — it would happily keep asserting a sentence about a switch that no longer has those cases.

`TestREADMESubmitEntryBlockSentinelLedger` reads the sentinel set out of `doUpload`'s switch in `app_submit.go` and fails when that set **grows or shrinks** — which is exactly the moment the prose needs rewriting. Membership, not a count: add-one/delete-one cannot swap a case past it, which is the defeat `symptomAttributionsFloor` was converted away from in #605.

## The prose pins are whole normalised sentences, not keywords

A guard on words is walkable by rewording, and **each of these four claims was replaced by a differently wrong one at least once during the ladder** — a keyword pin waves that through. Comparisons go through the **existing** `collapseWS` in `generate_dryrun_label_test.go` rather than a second copy, so a reflow passes and only a reword fails.

## And a behavioural case

A structural guard type-checks past a wrong reality. The ledger proves the README's exception list matches the switch; it cannot prove a pre-upload refusal never *reaches* that switch. That is the claim the arc actually got wrong.

## Mutation matrix

Run in a `cp -a` copy with `.git` removed **first** — a worktree's `.git` is a FILE, so a commit inside a naive copy lands on the real branch.

| # | mutant | dies at |
|---|---|---|
| M1 | `"any error the upload call reports"` → `"any submit failure"` — **round 1's own defect** | `ClaimsArePinned`, naming R1 |
| M2 | classification claim → `"prints only when the upload itself failed"` — **round 2's own defect** | `ClaimsArePinned` naming R3, **and** `SurfacesAgree` |
| M3 | delete the named pre-upload refusals | `ClaimsArePinned` naming R1b |
| M4 | delete "account, not a guarantee" | `ClaimsArePinned` naming R3b |
| M5 | revert the Troubleshooting row to `"Not printed for a 401/403/429"` — **the real round-1 drift** | `SurfacesAgree`, both assertions |
| M6 | add an `ErrNotFound` case to the switch | `Ledger`, "no row covers it" |
| M7 | remove the `ErrRateLimited` case | `Ledger`, "still carries ErrRateLimited" |
| M8 | print the block at the `--yes` refusal | `PreUploadRefusal`, naming the header |

Both **CONTROL** paths proven reachable rather than assumed: breaking the switch locator and breaking the paragraph anchor each **fire the control** instead of passing silently.

## One trap hit and fixed in passing

The same one the plan documents as Correction 2. The ledger test was originally `TestSubmitEntryBlockSentinelLedger` and **did not match** the handoff's gate filter `-run 'Attribution|Troubleshooting|README|Readme|readme'`. It ran only under a full `./...`, so the gate would have reported green over an unrun guard. Both it and the behavioural test now carry the `README` prefix.

Gate count **76 → 80**, measured with the same grep pattern as the 76 baseline rather than a different one — I miscounted once by dropping the subtest pattern and re-ran rather than quoting it.

## Gates

```
go test ./internal/cmd/ -run 'Attribution|Troubleshooting|README|Readme|readme'   80 PASS
go test ./... -count=1        21/21 ok, 0 FAIL
go vet ./...                  clean
gofmt -s -l .                 empty, over 495 .go files found
golangci-lint run             0 issues
```

Closes handoff rank 3b. Its closing condition was: *reverting any one of the four repairs turns a test red with that test's own error naming the claim, and the PR records which mutant it watched for each.* The matrix above is that record.


ZacxDev:
## Round 0 — requirements & deletion (audited `dc01963`), fixes in `3e53f7b`

Prose, not an `audit-claims` block — `audit-dispatch.py` refuses one for round 0, correctly, since round 0 fixes nothing and anchoring a delta on it would attribute the whole change to a round that did no work.

**Ledger:** `round 0 · requirements: 4 (unattributed: 2) · deletion candidates: 2` · trial-record pair `ran: 1 · changed the outcome: 1`.

Dispatched with the requirement deliberately put in question, because this PR **adds guards** and earlier in this arc a README byte-ceiling test was designed and discarded under `the-algorithm` step 5. The auditor was asked to interrogate rank 3b's provenance, which is identical in shape to the deleted byte threshold's.

Its discriminator, which I accept: **the byte threshold had no incident** — it was a proxy a session invented. Rank 3b has a measured one. The incident is the maker; the session only wrote it down.

### 🔴 F1 — an assertion of mine could not fail, and my own matrix scored it live. Fixed in `3e53f7b`.

`troubleshootingEntryBlockRow` returned the **whole table line**, whose left-hand symptom column literally reads `` `What this CLI would have sent` ``. So `strings.Contains(row, "would have sent")` was satisfied by the symptom column regardless of the cause cell — and that column is frozen independently, since `TestREADMETroubleshootingSymptomsExistInTheSource` requires it verbatim in non-test source and `app_submit.go:86` emits it. The string could never leave the line.

The auditor built the isolated mutant I should have: a cause cell reading *"The block never appears before an upload"* — **verbatim what my own error message calls false** — and it **PASSED**.

My M5 scored it killed because it deleted both words at once, so the sibling `classifies` assertion went red first. `RULES.md` names this exactly: *a mutation test still passes when a DIFFERENT guard's error kills your test.*

Now asserts on **column two only**, with a column-count control and an ≥80-byte control. Re-run isolated, the surviving mutant **dies at `:313` with this guard's own error**.

Making it reachable immediately found a second thing: the cell writes the tense as `` `What this CLI **would have** sent` ``, so the literal substring does not occur — the bold markers sit inside the phrase. The first reachable run failed on that. The extractor now strips `**`.

### D1, D2 and the failure-message pointer — all applied

- **D1:** ~30 lines duplicated verbatim from `app_submit_r2_test.go:98-126`, down to the literal token string. Extracted as `nonTTYSubmitRefusal`, which additionally fails the test when the refusal path was not the one taken — controlled by passing `--yes` and watching it fatal.
- **D2:** dropped the paragraph half of the `would have sent` loop; the ledger's `ErrBundleTooLarge` row already pins that sentence, and in `README.md` they are one sentence.
- **The `classifies` failure message now carries a #637 pointer.** Without it the message argues *for* the weakened wording — and #637's closing condition is to **revert** that weakening once the code stops printing the past tense for a pre-contact error. A maintainer landing there mid-fix would have been told not to make the fix.

### 🟡 F3 — correcting two overstatements in this PR's body publicly rather than editing it

A reviewer may already have read them.

1. **The matrix's M5 row says `SurfacesAgree`, "both assertions".** It killed **one**. That is the same non-isolation that hid F1.
2. **The opening attributes all four of #635's rounds to the absence of these pins.** At most **two** of the four would have been caught at introduction. The honest claim is narrower: these catch **regression** of repaired claims, plus future code/prose drift.

The auditor also established the **inverse of my own ordering**: `SurfacesAgree` — not the ledger — is the guard with real at-introduction value; `ClaimsArePinned` is friction rather than detection, and it is the one rank 3b explicitly asked for. Worth naming rather than resolving.

### Things the auditor checked that I had not

- **`app_submit_size_test.go:159`/`:244` already pin the 401/403 suppression behaviourally on `main`.** It built a mutant of its own (drop both `!`s, inverting the polarity) — **all four of my new guards survive it**, and those two pre-existing tests kill it. That further deflates my "nothing asserted any of it" framing: the **behaviour** was pinned; the **prose** was not.
- **Gate count.** It could not reproduce 76 → 80 with either obvious pattern (top-level gives 38 → 42, all-depth gives 98 → 102). The **+4 delta is confirmed** and all four run under the handoff's filter. My numbers come from `'^--- PASS\|^    --- PASS'` — top-level plus one level of subtest indent. Stating the pattern rather than the bare number, since that is the second time a count in this arc has been unreproducible without it.
- **A pre-existing flake, not mine:** `TestAppStatusDriftLookupGetsADeadlineOfItsOwn` (`app_status_drift_test.go:1558`) failed once under `./...` and passed on re-run, in code this diff does not touch. Named so nobody re-derives it as a regression.

### Gates at `3e53f7b`

```
go test ./internal/cmd/ -run 'Attribution|Troubleshooting|README|Readme|readme'   80 PASS
go test ./... -count=1        21/21 ok, 0 FAIL
go vet ./...                  clean
gofmt -s -l .                 empty, over 495 .go files found
golangci-lint run             0 issues
```

### Verdict

`proceed to the checklist.` Round 0 reports and does not move the ladder, and it does not license skipping round 1 — which now has a real delta to audit, including a guard that was vacuous until this commit.

ZacxDev:
## Round 1 — full correctness audit (audited `3e53f7b`), fixes in `3998c2f`

Dispatched **blind**: the auditor got the diff and the checklist, not round 0's findings or its reasoning.

**Three 🟡, zero 🔴 — and all three were my prose claiming more than my code does**, which is the defect class this very file exists to catch. Second round running with that shape.

```audit-claims round=1 audited=3e53f7b84291875e2fddf771e7ebd9ece9850ca0..3998c2f4
1. F1 the sentinel ledger now pins BOTH surfaces per case (paragraphSays + rowSays), so the Troubleshooting row's exception list is asserted rather than keyword-checked
2. F2 the sentinel extractor widened from a two-package allowlist to any qualified-or-bare errors.Is expression, and the ledger keys on the qualified name
3. F3 the duplicated non-TTY submit setup actually consolidated — app_submit_r2_test.go now calls nonTTYSubmitRefusal instead of carrying its own copy
4. F4/F5 two false comments rewritten from what the code does: doUploadSwitch's bounds, and the <400 control's stated cause
5. SurfacesAgree's row-side `would have sent` keyword check deleted as strictly weaker than the ledger's rowSays for the same cell
```

### F3 — the cleanest instance, and the most embarrassing

The comment read **"🔴 EXTRACTED, NOT COPIED"** and cited `AGENTS.md` *"One rule, one place."* **The PR touched one file.** Extracting is not something a one-file diff can do. `app_submit_r2_test.go` still carried its own verbatim copy down to the literal token string — so before this PR there was **one** site and after it **two**, while the comment claimed the reverse and named the rule it was breaking.

Now actually consolidated, and r2's test gains the `--yes` control it never had.

### F1 — the ledger covered one of the two surfaces it exists to keep in sync

`entryBlockSentinel` carried only what the **paragraph** must say. The row's exception list — **the half that actually drifted in #635 round 1**, the incident this guard cites — was reachable only by two keyword checks. Round 1 rewrote the cell to *"including a `401`/`403`/`429`"*, the exact inverse of `doUpload`'s second case, and the package stayed green. The published index would have told a reader the block **does** print for a credential refusal.

And the file argues at length that keyword pins are walkable by rewording — then pinned the row with two keywords. The struct now carries `paragraphSays` **and** `rowSays`, both whole sentences, asserted per live sentinel.

### F2 — the "grows" direction was spelled, not structural

The regex was `(?:appapi|civitai)\.(\w+)`. A case on any other package was invisible, while the doc said flatly that a new sentinel is red. `case errors.Is(err, os.ErrDeadlineExceeded)` — `os` is already imported by `app_submit.go` — left nothing red, and the `len(found) >= 2` control could not see it because the three known sentinels still matched. Widened to `errors\.Is\(err,\s*([\w.]+)\)`, ledger keyed on the qualified name.

### The three round-1 survivors now die

| mutant | before | after |
|---|---|---|
| **MH** invert the row's exception list | survived | Ledger `:238`, names **both** sentinels |
| **MG** drop `429` from the row only | survived | Ledger `:238`, names **both** sentinels |
| **MD** add an `os.ErrDeadlineExceeded` case | survived | Ledger `:207`, "no row covers it" |

Original matrix re-run and still holding: M1/M3/M4 → `ClaimsArePinned`, M7 → `Ledger`.

### The meta-point, since it is now three rounds of one shape

I wrote each doc comment **before** the implementation it describes, and every finding across round 0 and round 1 is the gap that opened between them. `RULES.md` already says this — *"Write the claim AFTER the code, from what the function does"* — and I did the opposite **while building a guard against precisely that**. The comments in `3998c2f` were written from the code.

### What the auditor confirmed rather than found

Worth recording, because it is what makes the guard worth keeping: **all 8 published mutants behaved exactly as claimed**, all six controls fire and none is assumed, the `--yes` control behaves as the commit message said, and gate counts reproduce at **76 → 80** with the stated pattern (38 → 42 top-level). **M7 is caught by nothing else in the tree** — the ledger's earned value, measured.

One nit it caught that I am not fixing: the file says *"Nine earlier `return err` sites"*; there are 8 literal `return err` plus one `return fmt.Errorf`. Correct as "early error returns", off by one as written. Changes nothing a maintainer does.

### Gates at `3998c2f`

```
go test ./internal/cmd/ -run 'Attribution|Troubleshooting|README|Readme|readme'   80 PASS
go test ./... -count=1        21/21 ok, 0 FAIL
go vet ./...                  clean
gofmt -s -l .                 empty, over 495 .go files found
golangci-lint run             0 issues
```

PR #640, opened 2026-09-16
docs(handoff): rank 3b closes — #639's guards land, and both ladders reduce to one named defect pattern

ZacxDev:
Handoff update for the readme-reduction arc, after #639 landed.

## What changed

- **Rank 3b closed** — #639 merged (`079dc14`): four mutation-verified guards over the paragraph #635 repaired and over `doUpload`'s switch. The ledger catches something nothing else in the tree does (measured: removing the `ErrRateLimited` case is caught *only* by that file).
- **#637 filed** — a **code** defect, not a docs one: `printSubmitSizeDiagnosis` prints `What this CLI sent` for errors raised before any HTTP request is built. Measured against a real `httptest` server — 0 requests received, block printed, with a positive control on the hit counter.
- **Phase 5 re-scoped by measurement** — a quality item, not a size one; realistic yield ≤5,465 B.
- **The worktree cost is now measured, not asserted** — `git worktree prune` was run and removed nothing; it blocked `--delete-branch` on all five merges and produced a false 🔴 in #635's round 3.

## The pattern both ladders produced

**Six of eight audit findings across #635 and #639 were "a sentence claiming more than its code does" — and half were written while building the guard against that exact class.** The mechanism is nameable: the doc comment was written *first* and the implementation after. `RULES.md` already says *"Write the claim AFTER the code"*; it was read in-session and the opposite done three times in one file.

Also recorded: two mutation-testing errors of mine, both "green for the wrong reason" — a non-isolated mutant that scored a **vacuous** assertion as killed, and a positive-control floor sized to today's set that is blind to growth by construction.

## Honest notes carried in the doc

- **#635 landed at −873, not the −1,741 its body advertises** — four audit rounds added 868 B back, each byte buying a contract claim that is now true.
- **The `app_submit_r2_test.go` consolidation in #639 was never audited.** It landed in the round-1 fix commit; operator merged knowing this. Named rather than hand-waved.
- ⚠ **7 of 8 ranked items declare `forcing: none`** — accepted by the gate and stated plainly: almost the whole remaining queue is self-generated and not eligible to be worked. Only #602 carries an external forcing function.

## Why this is a branch and not a commit on `main`

`AGENTS.md`: *"Feature branches only — never work on `main`."* The handoff tool will commit wherever the checkout sits, so this was authored in a worktree on its own branch instead.

Docs-only, no test surface — no audit round run on this PR. Stated, not skipped silently.


ZacxDev:
Closing as **superseded**, with the one surviving finding salvaged rather than dropped.

**Why it cannot be rebased.** This branch is based on `079dc14`, so its `README.md` half *reverts* #641 (`db68d1a`) — it restores the pre-#641 agent-setup section, deletes the five `###` subheadings and their `## Contents` entries, and reinstates ~4,375 B of prose that duplicated `civitai agent-setup --help`. Its handoff half describes the tree at `079dc14` and states rank 1 as "−4,936 B", a mid-ladder figure that #641 landed at **−4,064**. A rebase would either drag both backwards or be a full rewrite of every hunk. #642 (`4572d6f`) has since carried the correct state onto `main`.

**The finding that was worth keeping**, verbatim from this branch:

> 🔴 **The `app_submit_r2_test.go` consolidation in #639 was never audited.** It landed in the round-1 fix commit, and no round read it. Operator merged knowing this.

That audit has now been run: **#645**. The consolidation was *not* clean. `nonTTYSubmitRefusal` returned `serverHit` and both callers asserted `if hit { … }` — branches that **cannot fire**, because reaching the endpoint means the run got past the refusal, so the driver's own control already fataled. Measured with the gate removed from `confirmSubmit`: the recorder **was** hit, neither branch ran, and both tests died on *"the caller is measuring a different path than it claims"* — where the pre-#639 copy of the same test died on *"submit endpoint was hit — the gate did NOT prevent the submission"*. A message naming the test in place of one naming production. #645 moves the assertion into the driver, restores the wording, and proves the two controls it now runs ahead of are still individually reachable.

So this branch's estimate — *"the exposure is small and named, not hand-waved"* — was right about the size and wrong about it being clean.

Nothing else here is unrepresented: the #639 ladder's lessons are on `main` in `claudedocs/handoff-readme-reduction.md`, and #644 carries the rank-1 closure.

ZacxDev:
Superseded — see the comment above. The surviving finding is discharged by #645.

PR #641, opened 2026-09-16
docs(readme): give the agent-setup section an internal map, and cut the rationale that already lives in decisions 34-36

ZacxDev:
docs(readme): give the agent-setup section an internal map, and cut the rationale that already lives in decisions 34-36

Rank 1 of the readme-reduction arc, measured rather than estimated.

MEASURED, not projected:
  section  16,971 -> 14,265 B  (-2,706, -15.9%)   265 -> 236 lines
  README  281,439 -> 279,038 B (-2,401, -0.85%)   (+305 B of Contents entries)
  subheadings 0 -> 5, each with a Contents line

The structural change is the point, not the bytes. 265 unbroken lines with no
internal map became five subsections -- The two MCP servers / Where each agent's
config lives / No credential is ever written / Verify a setup (--check) / Exit
codes and --dry-run -- now reachable from `## Contents`.

WHAT THE TRIAL MEASURED, and it is the finding worth keeping:

1. NOTHING PINS THIS SECTION'S BODY. Maximal mutation -- delete all 264 body
   lines, keep the heading -- leaves `go test ./...` fully green (21 packages,
   0 failures). The guard set was then confirmed reachable by a positive
   control: mutating the heading turns readme_nav_test.go red in both
   directions. So the constraint on this section was never the suite; only the
   heading and its Contents entry are pinned.

2. THE SECTION IS 84.8% PROSE (14,385 B), not tables. Tables (1,503 B) and code
   fences (1,007 B) are byte-identical after the cut; all 2,867 B of reduction
   came out of prose, and 11,518 B of prose remains. A deeper cut is
   structurally available and was declined: what remains is published contract
   (exit codes, the three --json shapes, the no-credential rule, the per-agent
   spellings), and with the byte threshold deleted there is no reason to trade
   contract for bytes.

Deleted rationale is not lost -- it already sits in claudedocs/decisions/34, /35
and /36 (103,879 B), which the section now links by absolute URL per the
link-out policy. All three targets verified present on origin/main.

Every behavioural sentence was checked against source, per the arc's own lesson:
  - agent_setup.go checkCountsTowardVerdict excludes exactly TWO names, so the
    shipped "Three rows ... never fail its verdict" was loose -- an absent
    Authorization header is not a check row, it is a condition inside the
    ok:true mcp-* rows. Now stated as the code computes it, with the absent
    header noted separately (decision 34; TestCheckDoesNotFailOnAnAbsentHeader).
  - agent_setup_mcp.go's own comment confirms "not a read/write split: the
    anonymous initialize handshake itself is refused".
  - The Local development branch keeps all THREE cases it named (npm project /
    template with no package.json / nothing scaffolded yet), so the shortened
    version does not imply every project has a package.json.
  - The per-file blocked-row enumeration is stated once under Exit codes instead
    of twice; no case was dropped.

Gates: go test ./... green (21 pkg); the wide README filter runs 80 tests, all
pass (baseline 76); make ci green; golangci-lint 0 issues.

NOT DONE: no adversarial audit round was dispatched -- this session is not
permitted to spawn subagents. The repo's practice is a ladder, and this PR has
not had one.


ZacxDev:
## `pins-vs-published` is red, and it is NOT this PR

This PR changes `README.md` and nothing else. `pins-vs-published` contacts npm.

**Discriminating control, run before the plausible theory** — the same test, on a
clean `main` working tree (`git status` empty, HEAD `079dc14` = `origin/main`):

```
--- FAIL: TestScaffoldPinsSatisfyPublished
  template: templates/page-money/package.json.tmpl
  package:  @civitai/app-sdk
  pinned:   ^0.40.0   (pre-1.0 caret locks the minor)
  published:0.41.0
```

So it fails identically with zero of my changes. `@civitai/app-sdk` published
0.41.0 after `main`'s last green run (079dc14, 15:43Z today, `pins-vs-published:
success`) — the guard is correct and the pin is stale.

**`pins-vs-published` is a required context**, so this PR cannot merge until the
pin is bumped. The other four required gates pass here (`build-test`,
`scaffold-currency`, `ready-ack-runtime`, `template-page-vite`), as do `lint` and
`schema-drift`.

The fix is `bump-scaffold-pins` and it is **operator-owned** — I have not run it
and have not touched the pin. No bump PR is currently open.

ZacxDev:
## Round 0 — requirements & deletion (prose; round 0 gets no `audit-claims` block by design)

Read tip `8c38a5a`. **Ledger: `round 0 · requirements: 8 (unattributed: 1) · deletion candidates: 5`.**
**Verdict: `deletion candidate`** — explicitly *not* rework: every number the first commit claimed was re-verified and correct, and both mutants were reproduced independently (including that deleting the whole 264-line body leaves the suite green).

### Acted on in `2673c55`

**D1 — ~4,375 B of the section restated `civitai agent-setup --help`.** The plan's own order (`claudedocs/readme-reduction-plan.md:29-31`) is *"DELETE what the binary already says … → LINK OUT → relocate"*, and the first commit applied **reword** instead. My justification for keeping it — the link-out policy's `.goreleaser` argument — **does not reach this case**: the release archive ships the *binary* beside `README.md`, so anything `--help` already says is reachable offline by every tarball/Homebrew/npm reader. That argument protects `claudedocs/` link-outs, not `--help` duplication.

**D2 — the 590 B maintainers blockquote the first commit *added*.** `AGENTS.md` items 34–36 already route maintainers to those three files, and the blockquote's own label says its audience has a checkout.

### Declined, with reason

Round 0 recommended deleting the exit-code and `--json`-shapes spans too (its D1 tail and all of D5). **Kept**, because `AGENTS.md` assigns them to the README in writing: *"the README states the contract — exit codes, `--json` shapes, the command reference."* Also kept as README-only: the symlink-followed/broken-symlink-refused rule, and "Every row reports PRESENCE, never acceptance". The no-credential headline promise and the mint-your-key action were **compressed, not dropped** — a safety property and a user action.

D3 (Windsurf/Devin paragraph) and D4 (two-token-stores stated three times) were reported as 🟢 and left alone.

### Cumulative measurement

| | section | lines | README |
|---|---:|---:|---:|
| `origin/main` | 16,971 B | 265 | 281,439 B |
| first commit `8c38a5a` | 14,265 B | 236 | 279,038 B |
| **now `2673c55`** | **12,035 B** | **202** | **276,808 B** |
| | **−4,936 (−29.1%)** | **−63** | **−4,631** |

### Still open — filed, not fixed

**R6:** this PR rewrites behavioural prose and adds **no guard pinning it**. That is the same gap that cost #635 four rounds and that #639 closed for that PR (handoff rank 3b). Round 0 flagged it; it is not a round-0 finding and has not been addressed.

Also noted by round 0, unactioned: the five new anchors were not added to `TestREADMEWidenedTOCEntriesSlugAsExpected`, whose doc comment establishes that precedent for newly-written headings.

The nine correctness axes have **not** run yet.

ZacxDev:
## Round 1 — first full audit (nine axes, dispatched blind)

Read tip `2673c55`. Findings: **3 🟡 behaviour + 1 🟡 guard + 6 🟢**, zero 🔴. Verdict was *merge after fixing the three 🟡 prose defects*. All three were re-measured by hand against the built binary before any fix, and all three reproduced.

Fixes are in `bc33d094`.

```audit-claims round=1 audited=2673c55..bc33d094
1. F1 — the `--help` pointer promised "JSONC/symlink handling"; `--help` contains 0 occurrences of "symlink" (JSONC: 3). The pointer now promises the JSONC re-encoding caveat only.
2. F2 — the section said an absent Authorization header is "reported" by `--check`; a real run emits six rows and none is about a header. It now says the header gets no row of its own, that nothing fails because of one, and that the config file is where to look.
3. F3 — "Paths in `--json` are always absolute" was widened by this PR's own relocation and is false for a `manual` row (`"path": ""`). The manual-row exception is now stated.
4. 🟢5 — the three-writes-are-independent sentence was restored; without it the surviving "refused by name" sentence described the pre-fix behaviour.
5. 🟢6 — "`--agent other` prints the exact line" corrected to a template plus known spellings. The first draft of this fix claimed Zed prints "the exact line"; Zed prints a placeholder too, so the distinction was deleted rather than re-sourced.
6. 🟢8 — the stdout-cannot-be-written qualifier was restored to "a payload for every failure except a usage error".
7. 🟢7 — the `@civitai/*` pin paragraph moved out from under `### No credential is ever written`.
8. Finding 4 (guard gap) — added `internal/cmd/readme_agent_setup_claims_test.go`: four guards pinning RELATIONSHIPS rather than strings, mutation-tested with five mutants (2 README-side, 3 code-side), each killed by its own guard's own error message.
```

### Not fixed, deliberately

- 🟢9 — `8c33a5a`'s commit message says the deleted rationale is "linked by absolute URL"; `2673c55` then removed that blockquote, so the message is stale. A pushed commit message cannot be amended without a force-push; recording it here instead. The rationale itself is preserved in `claudedocs/decisions/34-36`.
- 🟢10 — the exported-OAuth-token-expires hazard left the README and is not in `--help`. In scope for the PR's own delete-what-the-binary-says policy, so left alone rather than restored on one reader's behalf.
- Round 0's note that the five anchors were not added to `TestREADMEWidenedTOCEntriesSlugAsExpected`. `TestREADMEAnchorLinksResolve` already covers all five in both directions, mutation-confirmed by round 1.

### Round 1's independent confirmation of the gap this closes

Round 1 reproduced the mutation itself: delete all 200-odd body lines of the section while keeping the five `###` headings, and `go test ./...` stays green (21 packages). The headings were guarded; not one sentence under them was. That is the same gap that cost #635 four rounds and that #639 closed for `## Submit & auth`.

ZacxDev:
## Round 2 — delta re-audit (`2673c55..bc33d094`)

All seven of round 1's repairs confirmed **actually fixed** (claims 5 and 8 partially). Round 2 then found **7 more: 2 🟡 payload + 2 🟡 scaffolding + 3 🟢 scaffolding**, zero 🔴. **Four of the seven are defects in the guards round 1 wrote** — the pattern this repo has now measured on #611, #635 and here.

Fixes in `2581fc5`. Every finding was re-measured by hand before fixing.

```audit-claims round=2 audited=bc33d094..2581fc5
1. 🟡1 — the Zed/`other` header sentence was still false after round 1 rewrote it; round 1 measured Zed only in the token-present state. Measured both: zed/no-token prints 0 header lines, zed/token 1, other/no-token 1. The README no longer promises the line for Zed and states what holds in every branch.
2. 🟡2 — "Read those two rows yourself", introduced by round 1, is wrong for `claude` where `claude-md` counts. The claude case is now named.
3. 🟡3 — the verdict ledger held four of six check names (`mcp-*` are bare literals, not constants); it now derives them from `civitaiMCPServers` AND reads the exemption SENTENCE rather than the section.
4. 🟡4 — the `--help` pointer's reverse direction was pinned for one word; it now pins the whole normalised sentence, so widening it is red by construction.
5. 🟢5 — the absolute-path assertion exempted the empty string; now `!filepath.IsAbs` alone.
6. 🟢6 — the prose fixture is case-insensitive and its failure message no longer gives a false diagnosis on a legitimate reword.
7. 🟢7 — the header-row guard asserted two words; it now asserts the emitted row SET in both directions.
```

### The finding inside the finding

🟡3's **first** fix was also wrong, and only re-running the mutant caught it. Adding `mcp-site`/`mcp-orch` to the ledger made the guard ask *"is this name mentioned in the section?"* — and `mcp-orch` is mentioned three paragraphs away, in the JSON sample and the unreadable-config paragraph. **The mutant survived a second time.** Presence of a word is not a statement about it. The guard now extracts the exemption sentence and compares the names inside *it* against the set derived from `checkCountsTowardVerdict`, failing in both directions.

### Mutation re-run — every mutant that survived round 2, now killed by its own guard's own error

| mutant | round 2 | now |
|---|---|---|
| R2-M3b code exempts `mcp-orch` | SURVIVED ×2 | KILLED `:145` |
| R2-M8 pointer widened with a topic `Long` lacks | SURVIVED | KILLED `:335` |
| R2-M6 every row path emptied | SURVIVED | KILLED `:272` |
| R2-M7 row named `mcp-auth` | SURVIVED | KILLED `:221` |
| M-new README over-claims an exemption (reverse) | — | KILLED `:153` |
| M1 README drops the claude condition (regression) | — | KILLED `:165` |

🔴 R2-M6's first run reported SURVIVED **because the substitution never applied** — caught by an explicit "mutation APPLIED" grep before the result was read. Round 2's auditor independently hit the same trap and reported it. A mutant that never ran scores SURVIVED and is indistinguishable from a vacuous guard; the applied-check is what makes the number mean anything.

### Corrections to my own earlier commit message

Both immaterial to the conclusion, recorded because round 2 checked them: `--help` has **2** case-insensitive `JSONC` occurrences, not 3; and my "80 tests" is the **wide** filter the handoff prescribes (`Attribution|Troubleshooting|README|Readme|readme`), where round 2 used the narrow one and saw 74/46 — different patterns, not a discrepancy.

### Not filed

The example payload's `"detail": "v0.1.103"` against tag `v0.1.105` (illustrative, untouched by this PR). Round 2's once-per-ladder count sweep re-derived the PR's byte and line figures at head and they hold.

Ledger: `round 2 · payload lines changed THIS round: 39 (since round 1: 379)`.

ZacxDev:
## Round 3 — delta re-audit (`bc33d094..2581fc5`)

Both prose repairs confirmed **true in every branch** (round 3 built the binary and ran six: `zed`/`other`/`claude`/`vscode` × token/no-token). Five of seven guard repairs real and mutation-confirmed. **3 findings: 2 🟡 scaffolding + 1 🟢**, zero 🔴 — and **two of them are regressions round 2 introduced into the guards while fixing them.**

Fixes in `7ffc94b`.

```audit-claims round=3 audited=2581fc5..7ffc94b
1. 🟡F1 — round 2 weakened the merge-rules fixture from "MERGED into, preserving every other server" to "merged into"+ToLower on the stated grounds that `Long` does not spell it that way; the tree contradicts that (grep at bc33d094 -> 1, committed fixture was green). Restored case-sensitive and specific.
2. 🟡F2 — the same ToLower over-applied to the vendor header spellings, where case is load-bearing. Restored case-sensitive. The improved failure message round 2 added is kept.
3. 🟡F3 — the claude clause shipped unpinned while the derived set sat unused two lines away. Now CONSTRUCTED from exemptForClaude + the conditional name and compared WHOLE.
4. 🟢F4 — the sentence parser gave a false diagnosis when its `.**` terminator moved (blaming the README for a neighbouring paragraph); it now asserts the terminator was found, bounds the slice, and says "fix the parse, not the README".
5. 🟢F5 — a wrong line number in round 2's message (R2-M8 dies at :366/:367, not :335). Recorded here; a pushed message cannot be amended without a force-push.
```

### The finding worth keeping from this round

**Round 2 weakened a working guard and its stated reason was false.** I conflated an earlier *draft* fixture (`"MERGED INTO"`, genuinely wrong, already fixed) with the *committed* one, and paid for a non-existent problem with real coverage. Two mutants survived the whole package under the weakened form:

| mutant | weakened form | restored form |
|---|---|---|
| M-G `Long` keeps "merged into", drops the merge rules | SURVIVED | KILLED `:410` |
| M-K `Long` mis-cases to `${ENV:civitai_token}` | SURVIVED | KILLED `:410` |

M-K is the one with user consequence: env var names are case-sensitive and `${env:` is the documented form, so `--help` could print a spelling that silently does not authenticate.

### Three attempts at F3, two walked past by a mutant

The spelled-guard class, three times in one PR:

| attempt | result |
|---|---|
| (a) assert the name appears in the clause | SURVIVED — it appears in the shared sentence too |
| (b) assert `` `x` alone `` / `` is `x` `` | SURVIVED — a reworded falsehood carries neither |
| (c) **construct the expected clause from the derived set, compare whole** | kills M-J, M-J2 (reworded falsehood), M-L (code-side) |

### Round 3's correction to round 2's evidence

Round 2's "zed/no-token prints 0 Bearer/Authorization" was measured on `printAgentSetupAuthNote` in isolation; the **full run** does print a ⚠ note telling the user to add the header. The conclusion is unchanged — the README should not promise the header *line* — and the shipped wording is true under the wider measurement too. Recording it because the narrower measurement is what my round-2 message cited.

### 🔴 Attribution gate — first zero-payload round

`git log --numstat --format= --remerge-diff 2581fc5..7ffc94b --not origin/main` → rc 0, stderr silent, 1 commit:

```
93  9  internal/cmd/readme_agent_setup_claims_test.go
```

**`round 3 · payload lines changed THIS round: 0 (since round 1: 389)`** — README.md is absent from the range. Payload for this PR is `README.md`; the claims test is scaffolding a round wrote.

That is **one** zero-payload round. The gate fires on **two consecutive**, so round 4 runs — but if round 4's fixes are also scaffolding-only, the ladder has left the PR and stops there regardless of what it finds.

ZacxDev:
## Round 4 — delta re-audit (`2581fc5..7ffc94b`) — **LADDER STOPPED**

All four round-3 repairs confirmed **actually fixed**, each re-verified by re-running the mutation rather than reading the claim — including two that round 4 independently measured as SURVIVED under the previous tree and killed under this one. Verdict: **safe to merge**. Findings: **2 🟢 scaffolding**, zero 🔴, zero payload.

Fixes in `c864827`.

```audit-claims round=4 audited=7ffc94b..c864827
1. 🟢F-A — the claude-clause slice ran to the next blank line, so an ordinary paragraph split made a TRUE statement fail with "does not state what the code computes". The parse is DELETED rather than hardened; the assertion is a constructed whole string, so position is irrelevant.
2. 🟢F-B — the comment credited mutant M-L to the new construction; M-L actually lands on the agent-dependence premise above, which t.Fatalf's first. The comment now states which direction each covers and says not to delete that premise.
```

Re-measured after F-A's fix: legitimate paragraph split → **GREEN** (was a false red); M-J (clause restated falsely) → **RED**; M-J2 (strongest reword, carrying both backticked names *and* "alone") → **RED**. The kills are retained and the false red is gone.

### 🔴 Why this ladder ends here — the attribution gate, not a clean round

| round | findings | payload lines its fix changed |
|---|---|---:|
| 0 | requirements & deletion — 1 major deletion candidate | (reports only) |
| 1 | 3 🟡 behaviour + 1 🟡 guard + 6 🟢 | 88 |
| 2 | 2 🟡 payload + 2 🟡 scaffolding + 3 🟢 | 39 |
| 3 | 2 🟡 scaffolding + 1 🟢 | **0** |
| 4 | 2 🟢 scaffolding | **0** |

`round 4 · payload lines changed THIS round: 0 (since round 1: 389)` — `README.md` is absent from `2581fc5..c864827`. **Two consecutive zero-payload rounds means the ladder has left the PR**: rounds 3 and 4 audited guards the ladder itself wrote, not the deliverable. Per the gate, stop — regardless of what a round 5 might find.

The published contract has been unchanged since `2581fc5`. Both remaining findings were scaffolding-only and are fixed here rather than filed.

### What the ladder cost and bought

**Bought:** the deliverable nearly doubled (round 0: −2,706 → **−4,936**, −29.1%), three false contract claims removed, two more caught before shipping, and four guards where there were none — on a section where deleting all 200 body lines previously left the suite green.

**Cost, recorded because it is the useful part:**
- Round 2 **weakened a working guard on a premise the tree contradicts**, and two mutants survived the whole package until round 3 restored it. One let `--help` advertise `${ENV:civitai_token}`, a spelling that does not resolve.
- Pinning one sentence took **three attempts**; a mutant walked past the first two. Only constructing the expected text from the derived set and comparing it whole held.
- A **mutant that never ran** scored SURVIVED twice in this ladder — once for an auditor, once for me. Every mutation now asserts `mutation APPLIED = True` before its result is read.
- A **docstring claiming coverage its body lacked** was found in four separate rounds of this one PR.

### Merged-tree gate

`origin/main` is merged into this branch (`efb8aac`), bringing in #643. Full suite on the **merged tree**: 21 packages ok, 0 failures; `TestScaffoldPinsSatisfyPublished` **PASS**. The branch's earlier greens said nothing about the tree its merge creates — this does.

**Not claimed:** golangci-lint at CI's pinned **v2.12.2**. This host has 2.13.2 (clean); read the PR's own `lint` run, which is not a required context here.

PR #645, opened 2026-09-16
test(submit): #639's consolidation left two dead `if hit` branches and a message that blames the test

ZacxDev:
Salvages the one surviving finding from #640 — *"the `app_submit_r2_test.go` consolidation in #639 was never audited"* — by running that audit. #640 itself is superseded (its diff reverts #641's agent-setup work, because it is based on `079dc14`) and is being closed.

The exposure #640 described as *"small and named"* is small. It is not nothing.

## What #639 left

`nonTTYSubmitRefusal` returned `(stdout, stderr, serverHit)`; both callers asserted `if hit { … }`. **Those two branches cannot fire.** Reaching the endpoint means the run got past the refusal, so the driver's own control — `err == nil`, or an error that is not the `--yes` refusal — has already fataled. The branch naming the regression is dominated by the branch naming the harness.

## Measured, not inferred

Mutation: `confirmSubmit`'s non-TTY arm returns `nil` — the exact regression `TestAppSubmit_NonTTYRefusesWithoutYes_NoNetworkCall` exists to catch. Applied and re-read from disk (`mutation APPLIED = True`) before either result was read.

| tree | what the failure says |
|---|---|
| `079dc14^` (pre-#639) | `submit endpoint was hit — the gate did NOT prevent the submission` |
| `main` (post-#639) | `CONTROL failure: refused, but not by the --yes gate — the caller is measuring a different path than it claims` |

The recorder **was** hit in both runs. Post-#639, neither `if hit` executed. So the consolidation traded a message naming **production** for one naming the **test** — a maintainer following it edits the test while the defect is in the gate. Both tests still go red, which is why this is a repair rather than a regression report; red-with-the-wrong-diagnosis is the failure mode `RULES.md` names as *"a guard's failure MESSAGE is part of the guard."*

## The fix

The driver owns the claim, because both callers depend on the same fact and neither can assert it. The hit check runs **first** — it is the most specific thing knowable about a failed run — and keeps the pre-#639 wording. `serverHit` leaves the signature, so there is no dead branch to re-add. The footgun test keeps its **name** (greppable as itself, not only as a side-condition of a README guard) and its comment says where its assertion now lives.

Also moved `TestREADMEPreUploadRefusalPrintsNoEntryBlock`'s doc comment back onto that test: #639 appended the driver's comment to it, so ~18 lines documenting the test were godoc for the helper and the test had none.

## Mutation matrix

Each applied and re-read from disk, each isolated to the narrowest expression that can be wrong.

| mutant | result |
|---|---|
| MU1 `confirmSubmit` non-TTY arm → `return nil` | **both** tests, at the **caller's** line, `submit endpoint was hit — the gate did NOT prevent the submission` (`app_submit_r2_test.go:109`, `readme_submit_entry_block_test.go:416`) |
| MU2 reword the refusal error only | third control, `refused, but not by the --yes gate` — isolated; the hit check did not swallow it |
| MU3 submit returns `nil` after a refusal, no network | second control, `the bare non-TTY submit did not refuse` — isolated |
| MU4 print the would-have-sent block at the `--yes` refusal | the README guard's **own** error at `:422`; the r2 test correctly **passes** — it asserts nothing about the block |

MU2 and MU3 exist because the fix adds a check that runs before two existing ones: they prove the fix did not make those two unreachable in turn, which is the same defect one level up.

## Verified

Gate **81 PASS** (`-run 'Attribution|Troubleshooting|README|Readme|readme'`, counted with the same `'^--- PASS\|^    --- PASS'` pattern as the 76/80 baselines) · `go test ./...` **21/21 ok, 0 FAIL** · `go vet` clean · `gofmt -s` clean · `golangci-lint` **0 issues** (2.13.2 locally; CI pins 2.12.2, and `lint` is not a required context).

**No production code changes.** Test scaffolding only.

Closes #640.


ZacxDev:
## Round 0 — two corrections to the body above, posted rather than silently edited

A reviewer may have read the original wording, so both stay visible.

**1. The body attributed a rule to `RULES.md` that is not in it.** It read *"red-with-the-wrong-diagnosis is the failure mode `RULES.md` names as 'a guard's failure MESSAGE is part of the guard.'"* That sentence is **not in `RULES.md`** (140 lines; grepped with `command grep`, not the gitignore-blind wrapper). Its only home is `claudedocs/handoff-readme-reduction.md:255,329` — **a prior round of this same ladder, one day old**. Per the audit brief that is the *highest*-scrutiny class of requirement, not the safest, and quoting it as a global non-negotiable let it skip the questioning it should have had.

**2. The "Measured, not inferred" table reports a subset of the run it measured, and the premise is weaker than stated.** MU1 (`confirmSubmit`'s non-TTY arm → `return nil`) reddens **four** tests in `internal/cmd`, not two:

| test | line | message |
|---|---|---|
| `TestConfirmSubmit_NonTTYRefusesWithoutYes` | `app_submit_r2_test.go:55` | `non-TTY without --yes must refuse` |
| `TestAppSubmit_NonTTYRefusesWithoutYes_NoNetworkCall` | `:103` | `CONTROL failure: refused, but not by the --yes gate …` |
| `TestAppSubmit_NonTTYRefuseHappensBeforePackaging` | `:127`, `:130` | `refusal must happen BEFORE packaging …` |
| `TestREADMEPreUploadRefusalPrintsNoEntryBlock` | `:458` | `CONTROL failure: refused, but not by the --yes gate …` |

The **first** failure in that run names production plainly, fifty lines above the caller in the same file. So *"a maintainer following the message edits the test while the defect is in the gate"* overstates it — they would have to ignore two clearer signals in the same output.

**What survives, and it is still the defect:** the one test whose whole subject is the network call died with a sentence about the harness, in the guard meant to name the gate, while the endpoint had in fact been hit. Both `if hit` branches were dead. The repair is unchanged and re-verified; only its justification was too wide. The comment now states the narrow version.

## Also fixed in `0e96791`

- **The reason for keeping `TestAppSubmit_NonTTYRefusesWithoutYes_NoNetworkCall` was wrong, and the real one is stronger.** I wrote "greppability". In fact `app_submit.go`'s two ordering comments (at `checkWorkTreeClean` and at `checkVersionNotRegression`) each cite this test **by name** as what pins a bare non-TTY submit to zero API reads. The comment says that now.
- **`stdout` had zero consumers** while the driver's comment promised "the run's streams for a caller". Cut to `stderr`. Same class as correction 2, committed inside the fix for that class.
- Deleted the duplicated #639 history and the "EXTRACTED, NOT COPIED" banner (it described a verbatim duplicate that has not existed since #639).

## Filed, not fixed here

`TestItem38CommentsCiteTestsThatExist` would catch a dangling test citation, but it ledgers only item 38's file set and `internal/cmd/app_submit.go` is not in it — so deleting a test that production comments cite is silent outside that set. Widening it is its own change: that test's own comment records why (23 sites repo-wide, 19 of them correct prose, needs a suppression convention first). **Closing condition:** a merged PR that globs the citation check repo-wide with that convention, checked by that PR's diff.

**Verified after the fixes:** MU1 → 4 red, the two touched here at `app_submit_r2_test.go:113` and `readme_submit_entry_block_test.go:416` with `submit endpoint was hit — the gate did NOT prevent the submission` · gate **81 PASS** · `go test ./...` 21/21 ok · `golangci-lint` 0 issues.

```audit-claims
round=0
audited=000e069cbec06d1ae9c49e4c85fdba08e3a0a6ae
fixed=0e96791
claims:
- F1 attribution corrected: the quoted rule is this arc's handoff doc, not RULES.md — fixed in the body (above) and not restated in code
- F2 the driver comment now states the measured scope: MU1 reddens 4 tests, two of which name production; the defect is the ONE test whose subject is the network call
- F3 the r2 test's comment now cites app_submit.go's two ordering comments as the reason its name is load-bearing; the ledger gap is filed, not widened
- F4 nonTTYSubmitRefusal returns stderr only — stdout had zero consumers
- D3/D4 duplicated #639 history and the EXTRACTED banner deleted
```

ZacxDev:
## Round 1 — the corrected count was also wrong, and the reason is the trap the correction was about

One 🟡 and three 🟢, all scaffolding, none in the repair. Fixed in `b614724`.

**F1 is the third iteration of the same defect in the same sentence.** Round 0 corrected *"the message misdirects a maintainer"* to *"it reddens FOUR tests, two of which name production"*. That four came from a **`-run`-filtered** sweep — the exact harness defect `RULES.md` names — committed inside the paragraph that was fixing an overclaim about measurement.

Re-measured **whole-package**, and the answer is better than "it is five": **the count depends on how wide the mutant is, and both numbers are right for their own mutant.**

| MU1 variant | red |
|---|---|
| **isolated** — `confirmSubmit`'s non-TTY arm returns `nil`, message literal kept in the file | **4** |
| **wide** — the whole `if !stdinIsTTY()` block deleted | **5** |

The fifth is `TestREADMETroubleshootingSymptomsExistInTheSource`: *"README Troubleshooting documents the symptom … but no non-test source file contains that string."* That is a fact about a **string**, not about the gate — the README indexes the refusal message as a symptom, so deleting the literal reddens it whether or not the gate still works. `RULES.md` says to mutate the narrowest expression that can be wrong, so the isolated mutant is the one that measures this gate. What made both drafts wrong is quoting a bare count as if it were mutant-independent. The comment now names the mutant beside the number, says the run was unfiltered, and says which of the five is about string presence.

**F2 — "50 lines above the caller" was staled by this PR's own earlier commit.** It was 49 on `main`; the round-0 commit added 13 comment lines above the caller, making it 63, and following the instruction lands the reader in a different test. A textbook instance of the delta-round blind spot: true when written, falsified by a later commit of the same PR, inside no round's range. The distance is gone — naming the file is enough.

**F3 — the driver's doc said it "asserts the whole contract of that refusal".** It asserts two thirds: the refusal happened, and the endpoint was not reached. The packaging half (no `Packaged` line, no `.zip` on disk) is `TestAppSubmit_NonTTYRefuseHappensBeforePackaging`, which keeps its own setup deliberately — no recorder server, a real base URL — so a third caller written against that sentence would skip re-asserting it. Now says what it covers and names what it does not.

**F4 is recorded, not fixed.** `TestAppSubmit_NonTTYRefusesWithoutYes_NoNetworkCall` has no assertion of its own, so deleting the driver's `if hit` would leave it green under a name claiming the opposite. It is strictly better than `main`, where **both** `if hit` branches were dead, and the alternative is a third ledger for one relationship.

**Verified:** whole-package `go test ./internal/cmd/` **ok** · `golangci-lint` 0 issues · `gofmt` clean.

```audit-claims
round=1
audited=0e967917e5848e67b7d1372786f59cd8b3fd8497
fixed=b614724
claims:
- F1 the MU1 count is now stated per-variant (isolated 4 / wide 5), measured whole-package rather than through a -run filter, and names the fifth as a string-presence guard rather than a gate guard
- F2 the "50 lines above" distance is deleted — it was staled by this PR's own round-0 commit
- F3 the driver's doc now says it asserts the refusal and the no-contact half, and names the packaging half as a different test that does not use it
- F4 not fixed: the bodyless test has no assertion tying it to the driver's `if hit`. Recorded as a residual — strictly better than main, where both branches were dead
```

ZacxDev:
Re-posting round 1's claims block in the format the dispatcher parses — the earlier one carried `round=` and `audited=` as body lines, so the header read empty and a round-2 delta brief was refused. No content change.

```audit-claims round=1 audited=0e96791..b614724b
1. F1 — the MU1 count is now stated per-variant (isolated 4 / wide 5), measured whole-package rather than through a -run filter, and names the fifth red test as a string-presence guard rather than a gate guard
2. F2 — the "50 lines above the caller" distance is deleted; it was staled by this PR's own round-0 commit, which added 13 comment lines above the caller
3. F3 — the driver's doc now says it asserts the refusal and the no-contact half, and names the packaging half as a different test that deliberately does not use the driver
4. F4 — NOT fixed, recorded as a residual: the bodyless test has no assertion tying it to the driver's `if hit`, so deleting that check leaves it green under a name claiming the opposite
```

ZacxDev:
## Round 2 — and the ladder stops here, on the attribution gate

Three 🟡 and two 🟢, every one in the same doc-comment paragraph, none a defect in the repair. Fixed in `4b2162b`.

**The retraction is the one that matters.** Round 1 said round 0's count of four came from a `-run`-filtered sweep that *"could not see the fifth by construction"*. Re-measured: **whole-package under the ISOLATED mutant also returns four.** So round 0's count was the right answer for its own mutant, the filter was a latent hazard rather than the cause — and the repo's own gate filter at the head of the file matches the symptom guard by name, so "by construction" was false for the filter this repo actually uses.

Round 1 had a real finding (a count quoted without its mutant) and then supplied a *mechanism* for it that measurement does not support. `RULES.md`'s rule about reaching for a fresh justification under pressure, in one paragraph. What was wrong every time is now stated once: quoting a number as if it were mutant-independent.

Two further overclaims, both introduced by round 1's fix:

- *"Under either, three tests name production plainly"* — **false under ISOLATED**, where the symptom guard does not fire and only two do. A maintainer applying the mutant this very comment recommends would count two and conclude the paragraph is stale, three lines below the sentence warning that counts are mutant-dependent. Now split per mutant.
- Deleting the line distance **orphaned "the same file"**, which then read as the file the comment lives in. `TestConfirmSubmit_NonTTYRefusesWithoutYes` is in `app_submit_r2_test.go`. Round 1's commit message said "naming the file is enough" and then did not name it.

Two attribution repairs:

- **`AGENTS.md: "One rule, one place."` is not in AGENTS.md**, or anywhere in this repo — it is the operator's private, unshipped rules file. Round 0's F1 was exactly this defect, and **round 0's own fix commit minted this instance four lines above round 1's hunk.** Neither round saw it.
- *"which deliberately keeps its own setup"* asserted intent with no source. `git` dates `TestAppSubmit_NonTTYRefuseHappensBeforePackaging` to #168, years before this driver existed, so "never consolidated" fits as well. The comment now says that — and records the cost the word "deliberately" was hiding: **under a gate-removing mutant that test issues a LIVE request to `https://civitai.com`** and dies on `unauthorized (401)`. Calling it deliberate would tell the next maintainer not to make the fix.

**Filed, not fixed:** eight Go files under `internal/` cite `RULES.md:` and the repo ships no `RULES.md`, so those pointers are unopenable for any contributor. Two are in this file. Fixing two of eight is the patch-the-second-copy shape; it wants one consolidating change. **Closing condition:** a merged PR that either ships the referenced rules or rewrites all eight citations to name something in-repo, checked by a grep in that PR's diff.

## Why this is the last round

```
round 2 · payload lines changed THIS round: 0 executable (37 comment lines) · since round 1: 0 executable
```

Round 1's fix changed 49 lines, **all comments** (measured by the round-2 audit). Round 2's changed 37, **all comments** — mechanically checked: zero changed lines fail to match `^[+-]\s*//`. That is two consecutive rounds whose fixes changed **zero executable payload lines**, so the attribution gate fires: the ladder is auditing text the ladder itself wrote, not the PR. It stops here rather than on a clean round — and round 2 was explicitly *not* clean, which is the point of gating on what a round *changes* rather than on what it *finds*.

**Verified:** `go test ./internal/cmd/` ok · `golangci-lint` 0 issues · `gofmt` clean.

```audit-claims round=2 audited=b614724b..4b2162b
1. retracted the "a filtered sweep could not see the fifth by construction" diagnosis — whole-package under the ISOLATED mutant also returns four, so the filter was not the cause
2. "three tests name production plainly" is now split per mutant: two under ISOLATED, three under WIDE
3. the orphaned "the same file" now names app_submit_r2_test.go
4. the `AGENTS.md: "One rule, one place."` attribution is removed — that phrase is in no file this repo ships
5. "deliberately keeps its own setup" no longer claims intent, and records the measured cost: that test hits the live civitai.com under a gate-removing mutant
6. NOT fixed, filed: eight internal/ files cite a RULES.md the repo does not ship
```

PR #646, opened 2026-09-16
fix(submit): stop claiming "What this CLI sent" when nothing was sent, and take the README's hedge back off

### LINKED ISSUE #637: app submit: "What this CLI sent" prints for errors raised before any request is built, so the tense can be false

ZacxDev:
Closes #637.

`printSubmitSizeDiagnosis` opens **"What this CLI sent … N bytes on the wire"**. It printed that for failures where the server received **zero** bytes — because `doUpload`'s switch branched on how the error *classified* and on nothing else, while two reachable failures classify as ordinary with nothing having left the machine.

`app_submit.go:74` states the intent: **THE TENSE IS THE WHOLE POINT.** The `sent` / `would have sent` split exists because of a measured incident where the server received 0 bytes and the CLI reported 12,587,785 on the wire. This path reproduced it, and the author's likely reaction is to **avoid retrying** for fear of a duplicate pending submission.

## The two paths

| path | why it lied |
|---|---|
| an unwritable config after an OAuth refresh | `internal/auth/source.go` returns `persist refreshed tokens: %w` **untagged**, and `authedDoWith` returns on `Tokens.Token()` before `doOnceWith` builds a request. `appblocks.go`'s own comment calls that chain *"Reachable, not theoretical"* and names NFS-soft, sshfs, CIFS and read-only filesystems |
| a dial that never connects | the request **is** built and **is** handed to the transport; the connection is never established |

The second is why the fix is not "tag the errors raised before a request exists". That gate passes it straight through and prints the same false byte count one step later.

## The signal is a wire fact, not another error kind

`appapi.SubmitVersion` attaches an httptrace hook and tags **`ErrNothingSent`** on any failure where `WroteRequest` did not report the request written. `doUpload` gets a third arm, above the `401/403/429` one, that prints nothing.

The kinds in `pkg/civitai` answer *what kind of failure is this*. They cannot answer *did anything leave*, and two of them are not wire facts at all — `internal/auth` tags `ErrUnauthorized` onto purely local errors. That is **#637's mirror case**, and the same arm fixes it: a locally-refused credential now lands on the wire fact instead of borrowing the vocabulary of a status code that never existed. Same output, correctly attributed — which is what makes the README sentence below sayable.

Four sites tag: the token/transport failure, the request-build failure, the JSON marshal, and — separately, because `timedOutSubmitError` formats its cause with `%v` and drops a tag applied before it — a timeout that preceded the write. A timeout **after** the write keeps the diagnosis, the case it is most useful for.

**No exit code moves.** `civitai.Tag` preserves `Error()` byte-for-byte and keeps the original chain reachable, and `ErrNothingSent` matches no arm of `cmd/civitai`'s `exitCode` switch. `TestNothingSentPreservesClassification` pins that a locally-refused credential still exits `3` (AGENTS.md item 7).

## The README goes back to the stronger claim

This is the other half of the issue's closing condition. #635 spent four audit rounds on one paragraph and could only make it stop promising more than the code delivered — *"keyed on how the error `classifies`, not on whether bytes left the machine"* and *"read `sent` as the CLI's account and not a guarantee"*. Both surfaces now say the past tense is exact and name the three reachable causes that print nothing.

The guard that required the word `classifies` said in its **own failure message** that #637 was the one change it must not block. It is rewritten here, in the same commit as the code, which is what that message asked for.

## Tests, watched red first

`internal/cmd/app_submit_nothing_sent_test.go` drives the **real `appapi.Client`**, not the fake `Submitter` — a fake returns whatever error the test hands it, so every case would pass vacuously against one while the real chain lied.

| case | at `origin/main` (`be4c34a`) |
|---|---|
| token-persist failure | **RED** |
| dial failure | **RED** |
| locally-refused credential | green — invariant guard, labelled as one (output unchanged, attribution changed) |
| positive control | green |

The positive control is not optional: three of the four assert a string is **absent**, and all three pass against a `doUpload` that prints nothing ever. It asserts the recorder saw **exactly 1** request *and* that the block printed.

## Mutation matrix

Seven mutants, each applied and **re-read from disk** (`mutation APPLIED = True`) before its result was read, each killed by a named guard.

| mutant | killed by |
|---|---|
| M1 drop `doUpload`'s `ErrNothingSent` arm | `SentinelLedger` + both cmd cases |
| M2 never tag (drop `!wrote` on the ordinary arm) | 5 tests across both packages |
| M3 always tag (ignore the wire fact) | `DoesNotTagWhenTheFailureCameAfterTheWrite` |
| M4 drop the timeout arm's own tagging | `TagsNothingSentWhenATimeoutPrecededTheWrite` |
| M5 revert the README paragraph to the hedge | `ClaimsArePinned` + `SurfacesAgree` + `Ledger` |
| M6 revert the Troubleshooting row | `SurfacesAgree` + `Ledger` |
| M7 count a FAILED write as sent | `SubmitWroteTraceIgnoresAFailedWrite` |

**M3 and M7 both SURVIVED the first run, and each survivor was a real gap.**

- **M3** showed the `!wrote.Load()` condition had no falsifying case: every other failure in the battery is a transport error where nothing was written anyway, and a non-200 is not an error at that point at all. The **over-cap response-body read** is the one reachable failure arriving *after* the request went out — the case where suppressing the block would **delete** the #423 diagnosis rather than narrow it.
- **M7**'s first fix was a test driving a real dropped connection, and it was **flaky by construction**. Probed directly: against a server that closes every connection the instant it accepts it, `WroteRequest` fired on all 4 runs — `info.Err` **set** on two and **nil** on the other two, where the write had completed into the socket and the failure surfaced on the read. The same physical event resolves either way. The hook is now a named function, `submitWroteTrace`, and the test calls it both ways.

## Considered and declined

A new AGENTS.md *intentional decisions* item. The one thing a future reader might undo — folding `ErrNothingSent` into `pkg/civitai`'s error kinds — is answered in the sentinel's own doc comment, which is the file they would be editing. An item costs a trigger line in every session.

## Verified

Gate **81 PASS** (`-run 'Attribution|Troubleshooting|README|Readme|readme'`, counted with `'^--- PASS\|^    --- PASS'`) · `go test ./...` **21/21 ok, 0 FAIL** · `go vet` clean · `gofmt -s` clean · `golangci-lint` **0 issues** (2.13.2 locally; CI pins 2.12.2). Swept every `.md` and `.go` in the tree for other copies of the retracted claim: none — `README.md` is the only place either surface appears, exactly as its guard says.


ZacxDev:
## Round 0 — the strengthened sentence was false for the case named three lines below it

Two 🟡 corrected in the payload, three more repaired, three 🟢, plus one finding the re-run mutation battery produced on its own. Fixed in `c1279a9`.

### Correction to the body above, posted rather than silently edited

**"Four sites tag" was a count of declarations, not instances.** A mutant proved one of them dead: dropping the tag at the request-build site **survived**, because that error travels back through `authedDoWith` and the `!wrote` arm tags it anyway. The site is deleted. Three remain, and the marshal one is **unreachable for these field types** — `submitEnvelopeLen` says exactly that of the identical `json.Marshal` 130 lines above, which is stronger than the "not a failure anyone has seen" I wrote. It is tagged so the set of tagged sites is the set of nothing-was-sent sites, and it is **not** counted as covered.

### F1 🟡 — the universal had its counterexample on the same page

The sentence read *"A failure that sent nothing prints nothing"*. The **ceiling refusal sends nothing and prints a full block** — the would-have-sent one — and that exception is named three lines later in the same paragraph and one sentence later in the same cell. Both surfaces now say **"never prints the past tense"**, which is what the code computes; the two guard strings moved in the same commit.

This paragraph's own recorded lesson, from #635: *a fix that makes a sentence more specific can make it false, and it reads as an improvement because specificity looks like rigour.* The requirement to strengthen came from a prior round of this same ladder — the class that gets adopted verbatim because pushing back reads as ignoring evidence.

### F2 🟡 — I condemned overlapping assertions and then wrote two

The file says *"the weaker of two overlapping assertions is the one that teaches a maintainer the wrong bar"*, and then: `TestREADMEEntryBlockSurfacesAgree`'s `precondition` was a substring of the ledger's `paragraphSays` **and** its `rowSays`, so it could only fail after the ledger already had — the PR's own matrix shows no mutant ever killed by `SurfacesAgree` alone. And R3's `want` was **byte-identical** to the new sentinel row's `paragraphSays`, tested against the same extracted text.

`SurfacesAgree` is **deleted** and R3 is gone from `readmeEntryBlockClaims`. R4 — "the two surfaces must not drift apart" — has been the ledger's job since #639 gave it `rowSays`: it asserts, per live sentinel, that the paragraph carries its sentence and the cause cell carries its own. The drift #635 shipped is M6 and still red.

### F3 🟡 — the per-attempt reset contradicted the sentence the PR exists to make true

README says `sent` is a fact about bytes that left this machine. If attempt 1 writes a full 12 MB body, the server answers `401`, and the retry fails to dial, **12 MB left this machine** — and the reset would have reported "nothing was sent". The flag is sticky now.

⚠ **Recorded as unpinned:** restoring the reset scores **SURVIVED** against the whole battery, and the scenario separating the two needs the server to die between the 401 and the retry — a race, and the flaky-by-construction shape this PR already retracted once.

### F4 🟡 — and the half of it that was attempted and reverted

Suppressing the byte block on a timeout that wrote nothing left the sentence underneath still reading *"the upload may not have completed … check whether it landed before resubmitting"* — precisely the hesitation #637 is about. It now says nothing was uploaded when that is what happened.

🔴 **Skipping the futile recovery poll was tried and reverted.** It breaks `TestSubmitVersionRecoversFromADeadlineExceededTokenError`, which pins a ctx deadline arriving through the **token seam** as still entering recovery and is the positive control for the whole *"a filesystem error is not a timeout"* family beside it. Deleting the work would have deleted that control. The wasted polls stay, recorded.

### F5 🟡 — the comment argued the case that does not need httptrace

It defended the mechanism against *"a request was built"* and never named the cheaper alternative a simplifier will find: `authedDoWith` already returns a status, **0** for every failure with no response — covering the token lookup, the build failure *and* the dial. What `status == 0` gets wrong is a timeout or drop **after** the body was written but before a response header: status 0, bytes gone. That is the #423 flagship path, where suppressing **deletes** the diagnosis rather than narrowing it. That is what forces httptrace, and it is what the comment says now.

### F6 🟡 · F8 🟢 · F9 🟢

- The request-build tag site was named in the issue, **reachable** (`internal/config` binds `CIVITAI_BASE_URL` with no `url.Parse` anywhere between there and the call) and pinned by nothing. Now covered by a control-char base URL — and that test immediately earned its keep by exposing the redundant tag above.
- `TestNothingSentPreservesClassification` is labelled an **invariant guard**: no mutation of this PR's code can redden it. Kept because exit 3 for a locally-refused credential is published contract and this is the package attaching the tag.
- Declining an AGENTS.md item was right for a harder reason than I gave: **AGENTS.md is 30,493 bytes against a 30,500 ceiling**, and the evidence ledger is bidirectional. The real hole was elsewhere — the reader most likely to undo this is someone adding a kind to `pkg/civitai/errkind.go`, and that file never mentioned wire facts. It has a back-pointer now.

### Mutation matrix, re-derived after every change

| mutant | result |
|---|---|
| M1 drop `doUpload`'s `ErrNothingSent` arm | Ledger + both cmd cases |
| M2 never tag on the ordinary arm | 6 tests across both packages |
| M3 always tag | `DoesNotTagWhenTheFailureCameAfterTheWrite` |
| M4 drop the timeout arm's tag + wording | `TagsNothingSentWhenATimeoutPrecededTheWrite` |
| M5 revert the README paragraph | `ClaimsArePinned` + Ledger |
| M6 revert the Troubleshooting row | Ledger |
| M7 count a FAILED write as sent | `SubmitWroteTraceIgnoresAFailedWrite` |
| M8 restore the per-attempt reset | **SURVIVED** — recorded above as unpinnable |
| M9 drop the `!wrote` tag on the ordinary arm | 6 tests |

**Verified:** `go test ./...` 21/21 ok, 0 FAIL · `go vet` clean · `gofmt -s` clean · `golangci-lint` 0 issues.

`round 0 · requirements: 10 (unattributed: 3) · deletion candidates: 6`

```audit-claims round=0 audited=6971fd5..c1279a9
1. F1 both README surfaces say "never prints the past tense" — the old universal was false for the ceiling refusal, which sends nothing and prints a block
2. F2 TestREADMEEntryBlockSurfacesAgree deleted and R3 removed from readmeEntryBlockClaims — both were subsumed by the sentinel ledger's per-surface whole-sentence checks
3. F3 the wrote flag is sticky across the 401 retry; the per-attempt reset is gone and its unpinnability is recorded in the comment
4. F4 a timeout that wrote nothing now says so instead of "may not have completed"; skipping the futile poll was tried and reverted because it deletes a sibling family's positive control
5. F5 the sentinel comment now names the status==0 alternative and the case that defeats it, instead of arguing the dial case
6. F6 the request-build site is covered by a control-char base URL test; the redundant tag AT that site was found by a surviving mutant and deleted
7. F8 TestNothingSentPreservesClassification is labelled an invariant guard
8. F9 pkg/civitai/errkind.go carries a back-pointer for the reader most likely to undo this
```

ZacxDev:
## Round 1 — this PR called 200 KB "nothing"

Two 🟡 and one 🟢. **F1 is a defect this PR introduced, in exactly the class this PR exists to remove** — the second time in this ladder the fix has reproduced the defect it was fixing. Fixed in `9718319`.

### F1 — the predicate was "the write COMPLETED", and that is the wrong question

`wrote` was set only when httptrace reported `WroteRequest` with **no error**. The reasoning: a failed write delivers an unknown number of bytes, possibly none, so counting it would put an unsupported number under "What this CLI sent". Measured, that reasoning **inverts**. A submit whose timeout fires **mid-body** takes exactly that branch — reproduced independently before touching anything:

```
run 1: server received 204594 body bytes
run 2: server received 233266 body bytes
run 3: server received 204594 body bytes
err  = "submit timed out before any of the bundle was sent … nothing was uploaded
        and no submission was created, so retrying is safe"
errors.Is(err, ErrNothingSent) = true   → the entry list was suppressed too
```

The CLI asserted **zero** over a fifth of a megabyte, and deleted the entry-list diagnosis for the large-bundle-on-a-slow-link author — who is the same author the sentinel's own comment cites as the reason this mechanism is not just `status == 0`.

The property is now the one the CLI can stand behind: **did this request get onto the connection.** `submitWroteTrace` becomes `submitRequestSentTrace` and stores true whenever `WroteRequest` fires, error or not. The three causes README names — no usable credential, an unwritable config, a connection that never opened — all fail *before* `WroteRequest` can fire, which is what keeps that sentence true.

**Residual, stated rather than implied:** a write that fails immediately fires the hook having delivered little or nothing, so the block's byte count is an **upper bound**, not a receipt. That is why the prose now says the request went out rather than that N bytes landed. Staying silent instead is what produced the false claim above.

Both README surfaces moved, since *"so `sent` is a fact about bytes that left this machine"* is the half F1 makes false:

> A failure that never reached the connection never prints the past tense … so `sent` means a request that really went out, not one the CLI only built.

🔴 **Two false versions of that one sentence have now shipped for one commit each, in OPPOSITE directions** — *"prints nothing"* (false for the ceiling refusal, which sends nothing and prints a block) and *"bytes that left this machine"* (false for a write cut short). Both counterexamples are recorded on the ledger row so the next edit has them in front of it.

### F2 — "this cannot be pinned without a race" was false

And that kind of claim is the one that stops anyone looking. The separating seam is the **dialer**, not the server: a transport whose `DialContext` succeeds once and then fails gives attempt 1 a full written body and a `401`, and attempt 2 a failure with nothing written — deterministically, and httptrace still fires because it is still an `*http.Transport`. `TestSubmitVersionKeepsTheSentFactAcrossA401Retry` pins the stickiness now, with controls on the hit count and the dial count.

### F3 — a comment describing a tag site the previous commit deleted

It contradicted `appblocks.go`'s own note four lines from that call. Fixed to say what the test covers: the `!sent` arm, reached by a failure that never gets near the connection.

### Mutation matrix — nine mutants, zero survivors

Re-derived from scratch after the predicate change. First time this PR's matrix has had none.

| mutant | killed by |
|---|---|
| M1 drop `doUpload`'s `ErrNothingSent` arm | Ledger + both cmd cases |
| M2 never tag on the ordinary arm | 6 tests across both packages |
| M3 always tag | `FailureCameAfterTheWrite` + `A401Retry` |
| M4 drop the timeout arm's tag + wording | `ATimeoutPrecededTheWrite` |
| M5 revert the README paragraph | `ClaimsArePinned` + Ledger |
| M6 revert the Troubleshooting row | Ledger |
| M7 ignore a failed write (**the design F1 retracts**) | `DoesNotTagWhenTheWriteWasCutShort` |
| M8 restore the per-attempt reset | `KeepsTheSentFactAcrossA401Retry` |
| M9 drop the `!sent` tag on the ordinary arm | 6 tests |

**M7 and M8 were the two survivors of the previous round.** Both are now dead, and each died to a test that exists because an auditor refused to accept a comment saying the case was unreachable or unpinnable.

**Verified:** `go test ./...` 21/21 ok, 0 FAIL · `go vet` clean · `gofmt -s` clean · `golangci-lint` 0 issues · gate **80 PASS** — 80 rather than 81 because `c1279a9` deleted `TestREADMEEntryBlockSurfacesAgree`; re-measured here rather than inferred, as round 1 asked.

```audit-claims round=1 audited=c1279a9..9718319
1. F1 the tag now keys on the request reaching the connection, not on the write completing — a mid-body timeout that delivered ~200 KB is no longer called "nothing sent", and the entry list is no longer suppressed for it
2. F1 both README surfaces say "never reached the connection" and "a request that really went out", replacing the "bytes that left this machine" clause the cut-short case falsifies
3. F2 the sticky flag is pinned by TestSubmitVersionKeepsTheSentFactAcrossA401Retry, via a DialContext that fails on the second dial; the "unpinnable without a race" claim is removed
4. F3 the control-char base-URL test's comment no longer describes a deleted tag site
5. the mutation matrix is re-derived: nine mutants, zero survivors, including the two that survived the previous round
```

ZacxDev:
## Round 2 — the count moved too

One 🟡 and four 🟢. The 🟡 is **the other half of round 1's own fix**, and it is the third time this ladder has reproduced the defect it was fixing — this time by repairing one half of a sentence and leaving the other. Fixed in `f79ddf9`, which also carries the merge of `main` (#645).

### 🟡 the block said "N bytes on the wire" and N is what the CLI *built*

Round 1 widened the tag so a cut-short transfer stops being called "nothing sent" — restoring the entry list for the author it had been deleted from. It did not touch the one line stating **how much**. Reproduced here:

| case | server received | printed |
|---|---:|---:|
| mid-body timeout | 233,266 | **8,388,627** |
| dropped at accept (≈nothing arrives) | ~0 | **8,388,627**, on 4 of 5 runs |
| …with round 1's predicate restored | ~0 | printed on **0** of 5 |

That is #423's own defect — *"the server received 0 bytes while the CLI reported 12,587,785 on the wire"* — reached through a door round 1 opened. And my justification comment pointed at prose that does not exist: it said *"the prose says the request went out rather than that N bytes landed"* while the **printed** prose said N bytes. The block's other hedge is about causation, not quantity.

The line now reads **`up to N bytes on the wire`**. `printSubmitSizeRefusal` keeps the exact form — nothing is sent there, so there is no transfer to fall short of. The doc comment's "here is what left this machine" goes with it, and README's sample output moves in the same commit.

🔴 **The qualifier was unpinned until a mutant said so.** `app_submit_size_test.go` asserts the line *contains* `"<n> bytes on the wire"`, satisfied with or without it — so deleting the word survived the whole battery. Now pinned as a constructed whole line (**M10** → `TestSubmitDiagnosisPrintsWhenBytesReallyLeft`).

### 🟢 the four others

- **Two test-file headers still stated the retracted property** — *"the tag is present exactly when no request bytes reached the network"* — the sentence the sentinel's own doc comment retracts, one of them contradicting line 206 of its own file. A maintainer deriving the invariant from a header would reinstate the narrowing that caused the 🟡 above. Round 1's F3, one level up, in the file it had just edited.
- **The cut-short test had no control proving the write was cut short.** It asserted the *server* received something — equally true of a run where the whole body wrote cleanly into socket buffers, since the handler stops reading either way. On such a host the test passes while the mutant it is credited with killing survives: **the assertion holding either way does not make the kill hold either way.** A counting `net.Conn` now measures what the *client* wrote, and the control was watched to fire — draining the body gives `the client wrote 8389316 of 8388627 body bytes`.
- **That test cost 6.93 s** because the recovery poll hit the same stalling handler three times. Answering the submissions route immediately: **2.54 s** (package 8.57 s → 4.02 s).
- **"ordered as well as safe" was stronger than net/http guarantees on the protocol production uses.** It holds on HTTP/1 (`mapRoundTripError` waits on `writeLoopDone`); h2's `roundTrip` returns on `ctx.Done` without waiting for the request goroutine, and production is h2. Probed 3/3 each way without reproducing it, so it is recorded as an **unproven ordering**, not a defect — and the losing direction is conservative: a Ctrl-C mid-upload would suppress a block it should have printed, never make a false claim.

### Instrument notes, because both cost real work

- **M8 scored `APPLIED=False` rather than `SURVIVED`** — the mutant text *contains* its own anchor, so my "old text absent" check could never hold. Re-scored with an insertion check, it dies to `KeepsTheSentFactAcrossA401Retry`. A mutation harness is an instrument like any other.
- **`git checkout -- <file>` restores from the INDEX**, so reverting a mutant discarded two unstaged fixes in that same file. Stage before a battery, or copy aside.

**Mutation matrix: ten mutants, zero survivors.**

**Verified:** `go test ./...` 21/21 ok, 0 FAIL · `go vet` clean · `gofmt -s` clean · `golangci-lint` 0 issues · gate **80 PASS**.

**Merged-tree note:** this branch now carries `main` (`7819352`, #645), which touches the same `internal/cmd` test file. The merged tree is green and the overlapping region was read by hand rather than trusted to a clean merge exit code.

```audit-claims round=2 audited=9718319..f79ddf9
1. 🟡 the diagnosis line says "up to N bytes on the wire"; the doc comment and README's sample output moved with it; printSubmitSizeRefusal deliberately keeps the exact form
2. 🟡 the qualifier is pinned by a constructed whole-line assertion (M10), which nothing covered before
3. 🟢 both test-file headers now state "never reached the connection" instead of the retracted "no bytes reached the network"
4. 🟢 the cut-short test counts CLIENT-side bytes via a counting net.Conn, and the control was watched to fire
5. 🟢 the cut-short test answers the recovery-poll route immediately: 6.93s -> 2.54s
6. 🟢 the memory-ordering comment now claims h1 only, records h2 as unproven, and names the conservative losing direction
7. the branch carries a merge of origin/main (#645); merged tree green, overlapping region read by hand
```

PR #648, opened 2026-09-17
docs(readme): phase 4 — make the document's order the order `## Contents` claims, and pin it

ZacxDev:
Phase 4 of the README reduction (rank 7 in `claudedocs/handoff-readme-reduction.md`). Byte-neutral: the file goes 277,826 → 277,820, and the six bytes are the six `#` characters the promotion below removes.

## The defect

`## Contents` described an outline nobody had written.

- it listed `Command reference` last in "Get started" while the document put it after two "Author an App" sections;
- it listed `The blockId`, `Templates`, `The host handshake`, `Local dev loop`, `Preview in the real host` and `Examples` as flat top-level entries while all six were `###` subsections nested inside `## Command reference`;
- it grouped `Browse the public API`, `Download model files` and `Scripting with --json` under "Use the API" while the document had them sandwiched between the authoring sections and `## Validate fidelity`.

Every navigation guard in `readme_nav_test.go` was green throughout, and correctly so: all of them assert **presence** — a heading has a TOC line, a TOC line has a heading. A table of contents can satisfy both directions and still be a map of a different document.

## What changed in `README.md`

1. the six App-authoring subsections are promoted from `###` to `##`. Anchors are computed from heading **text**, not level (`readmeAnchorSlug` takes the text and nothing else, and its own doc comment says so), so every existing link still resolves;
2. the `##` blocks are permuted into the order `## Contents` lists them.

**Proven to be a permutation, not a rewrite:** with the six promoted headings normalised, `sort README.md` is byte-identical before and after. No sentence was edited.

## Why the document moved to the TOC, and not the TOC to the document

Both directions make the two agree and both are byte-neutral, so the choice needs a reason.

**It is structural, not scroll distance.** `## Contents` is not a flat list — it carries four editorial groups (**Get started** / **Author an App** / **Use the API** / **Reference**). On `main` the three read-path sections sat *inside* the authoring run, and `## Generate` sat between `## App metrics` (176,357) and `## Upgrading` (225,598). So repairing the TOC to match the document would have had to **split "Use the API" into fragments interleaved with "Author an App"** — destroying a real reader affordance to preserve an accident. Moving the document is the only direction that keeps the grouping, and topical coherence is the benefit: a reader working the authoring track linearly stops hitting three API sections mid-stream.

🔴 **An earlier version of this body argued it on byte distance instead, and that argument refutes itself.** Round 0 caught it. It is kept here rather than quietly dropped, because the figures are real and the *inference* is what failed:

| | before (`4cb7c17`) | after |
|---|---:|---:|
| `## Scripting with --json` → `## Exit codes` | 160,391 B | 22,339 B |
| top of file → `## Validate fidelity` | 87,546 B | 59,014 B |
| top of file → `## Download model files` | 64,782 B | 159,634 B |

The third row is a regression, and the old argument dismissed it on the grounds that the reader still has a Contents link — **which, if true, voids the first row too.** One affordance cannot be decisive in one row and irrelevant in another. ⚠️ Row 2's "before" also **read 99,003 in the original body, which is the offset of `## Submit & auth`**; corrected to 87,546 above, so the real improvement is 28,532 B and not the 39,989 implied. [Full correction in a comment below.](https://github.com/civitai/cli/pull/648#issuecomment-5709555402)

The durable reason now lives in `readme_outline_order_test.go`'s doc comment (`62aca22`), with the byte model recorded there as refuted, so it is not derived again.

## The guard (`readme_outline_order_test.go`, new)

`TestREADMEContentsListsSectionsInDocumentOrder` — the sequence of headings the TOC must list is the document's own sequence, for all 66 of them (`##`, plus every `###` whose parent is not in `readmeTOCExemptSections`).

**Watched to fail.** At `4cb7c17` the two sequences hold the *same* 66 slugs and diverge at index 9 — document `set-up-your-coding-agent-agent-setup`, contents `command-reference`. Green at this head with no other change. Both halves observed, not predicted.

Mutation battery, each mutant asserting `APPLIED = True` (new text present *and* old absent, re-read from disk) before its result was read:

| mutant | verdict |
|---|---|
| swap two adjacent `##` blocks (`Upgrading` ⇄ `Global flags`) | KILLED — "diverge at position 58" |
| swap two Contents lines (`Configuration` ⇄ `Exit codes`) | KILLED — "diverge at position 61" |
| reorder two `###` inside one section (`Install`: npm ⇄ Homebrew) | KILLED — "diverge at position 2" |
| delete one Contents line | KILLED via the **presence** branch, which names the two guards that report it better |

It asserts nothing about whether the order is *good*. Contents is the authority and this only holds the document to it, so moving a section is still a judgement — it is now a judgement that has to be made in both places at once.

## The consolidation the reorder exposed

`readmeCommandTableSubjects` and `readmeCommandSynopses` each open-coded "the `## Command reference` table ends at its first `### `". That was a proxy that held only while those six subsections were nested there.

With them promoted the proxy did not fail — it silently **widened**: body 10,304 B → 12,193 B, reading 1,889 bytes of `## Set up your coding agent`, with the whole suite green because the extra window happened to carry no `| civitai` row.

**Control, because "costs nothing today" is a claim about today.** Planting one row — `` | `civitai app validate --definitely-not-a-flag` | … | `` — in those 1,889 bytes is invisible under the new bound (correct: it is not in the table) and reddens `TestREADMECommandSynopsesNameRealFlags` under the old one. The old bound was adjudicating a flag claim in a section it does not own.

Both now call one `readmeCommandReferenceTable`, which uses `readmeSectionByAnchor` — no raw scan at all, and fence-aware — and carries its own positive control. `readmeContentsLinks` likewise now derives its set from a single ordered extractor, so the two views of the Contents block cannot disagree about what is in it.

`readme_nav_test.go`'s comment claiming order was "not a hole this guard can close" is corrected in the same commit.

## Round 0 — run by hand at PR-create time, then dispatched properly (findings in the comment below)

- **Should this exist?** The requirement came from `claudedocs/readme-reduction-plan.md` — the same provenance as the byte threshold that was deleted for exactly that reason. It survives, but **not on the reasoning it was first given** — round 0 refuted that (above). What survives is the TOC/document divergence itself, which is a fact about the file rather than a projection, plus `AGENTS.md`'s own 🔴 *"README.md IS THE PUBLISHED USER CONTRACT"*; and the deliverable is byte-neutral and reversible.
- **The plan's "single most consequential misplacement" is already discharged.** It asked for the `listing status`-is-not-a-pure-read warning to be cross-linked from the scripting section; `README.md:3548-3551` already carries it (`🔴 And app listing status --json is a read with a server-side SIDE EFFECT … it must not be polled`). No work needed. It is still unpinned by any test — filed, not fixed here.
- **Positional prose swept.** Every `](#anchor)` followed by "above"/"below" (3), plus every by-name cross-section reference carrying a positional word (`Listing media requirements`, `Local dev loop`, `Download model files`, `What looks like a credential`, `Raw graphs`), checked to still point the right way. Clean on both trees.

## Verification

- `make ci` green; `go test ./... -count=1` 21/21.
- `make ci-shallow` — **21/21 in a depth-1 clone**, 0 failures, 0 timeouts.
- wide README gate: `go test ./internal/cmd/ -run 'Attribution|Troubleshooting|README|Readme|readme' -count=1` → `ok`, **81** tests (80 at `c51fab9`; +1 is this PR's new guard).
- `golangci-lint run` → 0 issues. ⚠️ at **2.13.2** locally; CI pins **2.12.2**, so read this PR's own `lint` run for that version. `lint` is not a required context.


ZacxDev:
## Correction to this PR's body — one number was wrong, and the argument around it was worse

Round 0 ran on this PR and found both. Posting rather than silently editing the body, because a reviewer may already have read the wrong version.

### 1. The reader-cost table compared two different sections

The row `top of file → ## Validate fidelity | 99,003 B` is **wrong**. Measured on `4cb7c17`:

```
  64782  ## Download model files
  77191  ## Scripting with `--json`
  87546  ## Validate fidelity      <- the real figure
  99003  ## Submit & auth          <- what the row actually quoted
```

`## Validate fidelity` is at **87,546**, so the row is `87,546 → 59,014` — an improvement of **28,532 B**, not the 39,989 implied. **The claim's direction survives; its magnitude was inflated 40%.** I took the number from the adjacent row of my own heading map. That is the same defect class as this arc's last three merged commits, which is not a mitigating remark.

The other two rows re-derived exact.

### 2. The bigger finding: the justification refuted itself

Both directions — move the document to the TOC, or the TOC to the document — make the two agree, and both are byte-neutral. So the choice needs a reason, and the one this body gave does not hold: it dismissed the one reader the reorder made **worse** (`## Download model files`, 64,782 → 159,634) on the grounds that they still have a Contents link. **If a Contents link settles it for that reader, it settles it for the CI reader too** — and the 160,391 → 22,339 headline the whole change was argued on is then worth nothing. One affordance cannot be decisive in one row and irrelevant in another.

**The reason that does hold is structural**, and it is now in `readme_outline_order_test.go`'s doc comment (`62aca22`) so it does not live only in a PR body:

`## Contents` is not a flat list — it carries four editorial groups (**Get started** / **Author an App** / **Use the API** / **Reference**). On `main` the three read-path sections sat *inside* the authoring run, and `## Generate` sat between `## App metrics` (176,357) and `## Upgrading` (225,598). So repairing the TOC to match the document would have had to **split "Use the API" into fragments interleaved with "Author an App"** — destroying a real reader affordance to preserve an accident. Moving the document is the only direction that keeps the grouping, and topical coherence is the benefit, not scroll distance.

The byte figures stay in the comment labelled as **refuted**, so nobody derives them again.

### What round 0 did *not* find

No deletion candidates (2 examined, 0 recommended). It specifically retracted its own first theory that the new order guard subsumes the two existing presence guards — they also validate `readmeTOCExemptSections` hygiene and split *stale* from *off-limits*, neither reachable from the order guard. It re-derived every other number in the body exactly, including the watched-to-fail A/B and the planted-row control in both directions, and confirmed `build-test` is a required context so the new guard both runs and gates.

Verdict: `requirement questioned — R1's direction, and the byte-offset model that chose it`. Not a request to close or rework.

⚠️ `claudedocs/handoff-readme-reduction.md` on `main` carries the same wrong figure at line 532 (it merged in #649 before this was caught). Being corrected separately.

ZacxDev:
## Round 1 — the nine correctness axes. Four findings, all scaffolding; `README.md` untouched.

Dispatched blind to the author's reasoning. It independently re-derived the permutation proof (line-multiset, and separately the *inside-fence* and *outside-fence* multisets — so no fenced block was split by a block move), the watched-to-fail matrix, the extractor-bound measurements and the planted-row control in both directions, and re-ran the mutation battery with its own mutants. All confirmed. It also corrected my scope reporting: the diff is **4 files against the merge base `4cb7c17`**, and the handoff file that appears in `git diff 5383ba4 62aca22` landed on `main` *after* this head rather than being part of it.

Fixed in `e0024d7`:

**🟡 1 — a false census, which this PR made falser.** `readme_nav_test.go`'s anti-vacuity comment on `readmeTOCMinSubsections` read *"the real in-scope count is 33, so exempting any of the substantial sections (`Generate` 12, `Command reference` 6, `Install` 5, `Scripting` 5) drops it under the floor."* Three things wrong: the count is **41 before this PR and 35 after**, never 33; **`Command reference` has no `###` children at all since the reorder**, so exempting it would retire nothing; and at 35 against a floor of 25, only `Generate` crosses it — exempting `Install` or `Scripting` leaves the gate green. The floor is a backstop against the worst case, not the proof of coverage it claimed. Restated without a census, which is the lesson the 🔴 block one screen below in the same file already records.

**🟡 2 — the population-mismatch branch could send a maintainer to two green tests.** `…CoversEverySection` skips `contents` and so did `readmeOutlineOrder`, but `…ListsNothingElse` does **not** — it allows a `[Contents](#contents)` line. So that line made the populations differ by one, and my mismatch message named two guards that both pass and name nothing. Now skipped on both sides. Re-run with the self-link inserted: **all four guards PASS**, where before only this one went red pointing at them. It failed *safe* (a confusing red, never a false green), but the wasted round was real.

**🟢 3 — every reader-path figure labelled `bytes` is a CHARACTER count.** `awk`'s `length($0)` in a UTF-8 locale. 160,391 chars is 161,628 bytes; 22,339 → 22,532; 64,782 → 65,276; 159,634 → 160,868. The 58% is unaffected. Relabelled — and the instance is kept beside the lesson, because the same trap (Python's `len(str)`) was hit and written up **earlier in this same session**, and a lesson that does not survive one session needs its instance recorded next to it.

**🟢 4 — wrong antecedent.** *"byte-identical to the `\n## ` scan it replaces"* — what it replaces is the `\n### ` scan (12,193 B); the `\n## ` scan is the alternative repair **not taken** (10,304 B, identical). Also now recorded: `readme_install_accuracy_test.go` reads the same section directly and deliberately does not route through the new helper, because it slices the preamble and the section's end is not an input there.

**Not fixed here because it is already discharged:** the auditor's third finding — the wrong `99,003` figure on `main` — was corrected in **#650 (`effa189`)**, which merged at 05:57Z while this audit was running. Its worktree was detached at `62aca22` and test-merged `5383ba4`, so it read the pre-#650 state; the finding was correct for the tree it could see. Verified on current `main`: the row is gone. ⚠️ Its *unit* defect (finding 3 above) does still apply to that table on `main` and is being corrected separately.

**Verified after the fixes:** `gofmt -s -l .` clean · `go vet ./...` rc 0 · `go test ./... -count=1` 21/21 · wide gate `ok` at 81 · mutation battery re-run with a fifth mutant added, **5/5 KILLED**, each asserting `APPLIED = True` from disk first, positive control PASS.

```audit-claims round=1 audited=62aca22..e0024d7
1. readme_nav_test.go readmeTOCMinSubsections comment: removed the "in-scope count is 33" census and the claim that exempting any substantial section crosses the floor; stated the floor as a worst-case backstop only.
2. readme_outline_order_test.go: `contents` is now filtered from the TOC side as well as the document side, so a [Contents](#contents) line no longer routes to the population-mismatch branch and its two-green-tests message.
3. readme_outline_order_test.go: reader-path figures relabelled from `bytes` to `characters`, with the byte equivalents and the awk cause recorded.
4. readme_nav_test.go: corrected "byte-identical to the `\n## ` scan it replaces" to name the `\n### ` scan as the antecedent and the `\n## ` scan as the untaken alternative; recorded readme_install_accuracy_test.go as a third reader of the section that deliberately does not use this bound.
```

PR #652, opened 2026-09-17
fix(appapi): one 401 message, and withdraw phase 5 — the binary already said it

ZacxDev:
Rank 8, the last named phase of the README-reduction arc. **It is withdrawn**, and the one row that survived the measurement is shipped.

## Phase 5's premise is false for 20 of the 21 rows in its own target band

The phase said: *"for each row where the README still says more than the binary does, move the difference into the Go error string and delete the row."* Measured row by row against the live Troubleshooting section:

| | rows | why it is not phase-5 work |
|---|---:|---|
| the binary already carries the remedy in full | **11** | the cell restates it |
| the cell holds what a per-site string structurally cannot | **3** | two exit codes off one message (`rate limited (429)`); *"the two rows below are the 403s that are not about your grant"*; an exit code the binary never prints as text |
| not an error message at all | **6** | a `whoami` row label, a `workflows list` legend, an `errors.New` sentinel, a note, two output labels — and one string **your app prints at runtime**, which the CLI cannot reach |
| genuine | **1** | `not logged in (401)` — shipped below |

Three examples, because the pattern is uniform rather than marginal:

- **`HEAD is on no remote`** — README: *"Push the branch."* Binary already: *"Push the branch so the submitted version can be traced back to a commit."*
- **`nothing index.html loads reaches it`** — README: *"Copying `civitai-host.js` in is only half the fix — it has to be referenced too."* `readyAckRemedy` already: *"Copying the file in is not enough on its own — a browser never fetches a file nothing references."*
- **`no lockfile is committed`** — README: *"Generate it with the package manager."* Binary already: *"Run `%s` and commit the %s it writes."*

**The transferable finding:** the error strings were already carrying their remedies, because `AGENTS.md` has said *"make errors actionable — name the next command to run"* the whole time. The plan's scope note is corrected in this PR too: round 0 on #625 measured that the **cells restate the binary**, and phase 5 read that as *the binary lacking what the cells hold* — the opposite claim, which does not follow.

### The instrument mattered, and the first one was wrong

A one-line regex over non-test Go called **12 rows "no remedy in the binary"**. Every one I hand-checked was a **false negative** — these messages are `+`-joined across many lines and a single-line read truncates them. That is the same overcounting the handoff already recorded for the crude 25-of-49 pass. The classification above comes from reading whole statements.

### The stale ceiling, corrected

The ≤5,465 B figure was measured when `README.md` was 282,312 B, three PRs ago. Live: the section is **20,834 B / 61 rows**, the non-floor 150–250 band is **18 rows / 3,480 cause-bytes**, and the incident floor now protects **10** rows, not 7.

## The one genuine row turned into a consolidation

`not logged in (401)` existed at **five** sites in `internal/appapi`. Four spelled the remedy `` run `civitai login` (or set CIVITAI_TOKEN) ``; the fifth — `submissionsError` — said `` run `civitai login` ``, **dropping the environment-variable route for the command group most likely to run in CI**. Nobody chose that; it is what an open-coded predicate does at N sites, wrong at N−1 in the same direction. Consolidating is what made it audible.

All five now call `unauthorizedError`, which also names the **personal-API-key route the README carried and no user-facing surface did** (an enumerated grep found it only in code comments). The wording is taken verbatim from the one surface that already had it right — `analytics.go`'s 403 arm — so the CLI names one way to do this, not two.

### Guards (all new)

| | |
|---|---|
| `TestUnauthorizedCallersAreLedgered` | **bidirectional** — fails when a file gains a 401 arm without the helper **and** when a ledgered one loses its arm; also fails on a re-introduced literal. Floor on the directory walk. |
| `TestUnauthorizedErrorNamesBothRemedies` | the whole constructed string, **plus four literal clauses** so it is not the constant compared against itself |
| `TestUnauthorizedErrorIsStatusTagged` | drives the **real arm** and asserts `errors.Is(err, civitai.ErrUnauthorized)`, with a 404 control so the tag check can discriminate |

**Mutation battery — 5/5 KILLED**, each asserting `APPLIED = True` from disk before its result was read: restore an open-coded literal; add an unledgered 401 arm; delete a ledgered arm; drop the key clause from the remedy; drop the `defer TagStatus`. Positive control PASS.

🔴 **An earlier draft of the third test asserted only the message text while its docstring claimed it guarded the tag** — the *"docstring names a relationship, body inspects one side"* class this repo's rules name, caught here before it shipped. The tag mutant is the proof: it **survives** the draft and dies against the committed version. The tell is kept in the comment rather than quietly corrected.

## README

The 401 cell stops repeating the remedy and points at the message instead (186 → 184 B). Its **symptom column is unchanged**, so the source pin still holds.

## Verification

`make ci` green · `go test ./... -count=1` 21/21 · **`make ci-shallow` 21/21 in a depth-1 clone** · wide README gate `ok` at 81 · `gofmt -s -l .` clean · `go vet` rc 0 · local `golangci-lint` **0 issues** ⚠️ at 2.13.2, not CI's pinned 2.12.2 — read this PR's own `lint` run.

## What this closes

With this merged, the arc's third clause — *"each remaining named phase has landed or been **withdrawn** by a merged PR"* — is satisfied, and all three clauses are green.


ZacxDev:
## Round 1 — the nine axes. Two 🔴, two 🟡, two 🟢. All verified against the tree before acting.

The headline: **there are EIGHT 401 arms, not seven**, and the guard round 0 widened was *still* narrower than the code.

**The count went 5 → 7 → 8, and every correction came from a wider instrument rather than a wider read.** grep the literal → 5. grep `case http.StatusUnauthorized:` → 7. Parse the AST → 8.

### 🔴 F1 — the eighth arm

`withdrawError` (`appblocks.go:2307`) answered 401 with `case http.StatusUnauthorized, http.StatusForbidden:` — a shape no line-regex sees — returning *"not authorized (check your API key / Apps invite)"*, which names no runnable command. Live, via `civitai app withdraw`, and **seventy lines from the code under change**.

Split: the 401 returns the helper, the 403 gets its own message. They cannot share one — a 401 is *a credential the server did not accept*, a 403 is *a credential it accepted and an account without the grant*. Logging in again fixes exactly one, so one message was always wrong for one of them.

### 🔴 F2 — the guard was wrong the same way twice

File-granular (round 0 caught it), then arm-granular but blind to a case **list** and to a **branching** clause body — both shapes already present in this package, both proven by surviving mutants.

**Replaced, not hardened a third time:** a `go/parser` walk (`unauthorized_arms_test.go`), which cannot have a pattern blind spot because it reads the grammar the compiler reads. The regex ledger is **deleted** rather than kept alongside — a superseded guard that still passes reads as coverage while accepting states the real guard rejects. Its floor is now the **actual** arm count (8), not comfortably below it: the old `total < 5` against a real 7 let a mutant de-convert two arms in silence, demonstrated.

### 🟡 F3 / F4 — claims the tree contradicted

- *"the personal-key route no user-facing surface carried"* — **false**. `internal/genapi/errors.go` names both routes on its own 401; ~10 surfaces name the key route in several spellings. `internal/cmd/login.go`'s `spendCredentialRoutes` already states the governing convention: other packages cannot import it, so wording is corrected in place and **kept in step by hand**. This is another copy under that convention, not a unification. Corrected.
- *"taken verbatim from the one surface that already had it right"* — neither verbatim nor the one surface. Corrected.
- The helper's doc named one caller passing `""`; `listingError` does too, and discards its `msg` with **no comment at all**. Both now named, with the asymmetry stated rather than tidied away.

### 🟢 G1 / G2

- *"the refresh failed too"* was asserted even when a refresh **succeeds** and the retry still 401s (server-side revocation). Now *"already failed twice"* — true of the retry pair without claiming which half failed.
- The doc's *"only 2 of the band's rows are floor-protected"* sat under the composite explanation while meaning the **incident** floor specifically. Both numbers disambiguated.

### Two more pins, and they are why the merge stayed invisible

`withdraw_test.go` and `app_withdraw_test.go` asserted `"not authorized"` for **401 and 403 alike** — one expectation covering two different problems. Now distinct, so a future re-merge reddens.

### Verification

Mutation battery re-run against the AST guard: **7/7 KILLED**, each asserting `APPLIED = True` from disk first, positive control PASS. It includes both shapes that **survived** the regex (case list; branching body) and a re-merge of `withdrawError`'s arms.

`make ci` green · `go test ./... -count=1` 21/21 · `make ci-shallow` **21/21 depth-1** · wide gate `ok` at 81 · gofmt/vet clean · `golangci-lint` 0 issues (2.13.2 local).

⚠️ `pins-vs-published` is red on this PR and **is not this PR's fault** — reproduced on a clean `origin/main` worktree at `321fcd8` with zero local changes: `@civitai/app-sdk` pinned `^0.41.0` / published 0.42.0, `@civitai/blocks-react` pinned `^0.50.0` / published 0.51.0. Today's scheduled bump ran green at 12:35Z; both published after. It blocks every open PR until a pin bump lands.

```audit-claims round=1 audited=9963e08..ad426fec
1. withdrawError's 401 is split from its 403 into its own case and returns unauthorizedError; the 403 gets its own message naming `civitai whoami`.
2. The regex-based ledger (unauthorizedLedger + statusUnauthorizedArm) is DELETED and replaced by a go/parser walk in unauthorized_arms_test.go that matches case lists and branching clause bodies.
3. The arm-count floor is raised from 5 to the actual arm count, 8, so the control fires on the first unrecognised arm rather than the third.
4. unauthorized.go no longer claims the key route was carried by no user-facing surface, nor that its wording is verbatim from one surface; it names login.go's spendCredentialRoutes keep-in-step convention instead.
5. unauthorized.go's doc names BOTH callers that pass "" (devTunnelError, listingError) and states that only the first justifies the drop.
6. The remedy says "already failed twice" instead of "the refresh failed too", which no longer asserts a failure when an OAuth refresh succeeded and the retry still 401d.
7. withdraw_test.go and app_withdraw_test.go no longer assert the same string for 401 and 403.
8. claudedocs/readme-reduction-plan.md says eight arms and four spellings, narrows the no-surface-carried-it claim to this package, and disambiguates the incident floor from the composite.
```

ZacxDev:
## Round 2 — delta re-audit. Six findings, three in payload. All verified against the tree before acting.

**The centrepiece of round 1 was refuted in one pass.** The `go/parser` walk was *also* pattern-blind — it visited `*ast.CaseClause` with an `*ast.SelectorExpr` and nothing else, so it never saw `if status == http.StatusUnauthorized`, which is at **`appblocks.go:580`, in the same file**. Three surfaces told the reader it "cannot have a pattern blind spot at all".

| draft | keyed on | what it missed |
|---|---|---|
| 1 | `Contains(src, "unauthorizedError(")` | four arms in an already-converted file |
| 2 | regex on `case http.StatusUnauthorized:` | `case A, B:`; branching clause bodies |
| 3 | AST, `CaseClause` + `SelectorExpr` | `if`; `switch{case expr}`; bare `case 401:` |
| **4** | **does a 401-conditioned region CONSTRUCT a message?** | stated in the comment, not denied |

Widening a fourth time would have made `appblocks.go:580` — the transparent refresh-and-retry, which builds no message — a false positive needing an exemption, and an exemption list is a to-do list that grows one line per omission. So draft 4 stops asking what syntax encloses the 401. The enclosing form stops being an input; the retry passes for the *right* reason; and indirect routing through a wrapper stops firing, which draft 3 called a violation while telling the reader to "give it its own case" — advice that did not apply.

### 🔴 The new 403 asserted a diagnosis nothing sources

It said *"withdrawing needs Apps-author access (invite-only beta); check which account you are on."* `WithdrawPath`'s own vendored note records **200/404/409 and no 403 at all**, and puts the not-yours case on **404** — so the single cause that message named is the one the route does not use. Meanwhile `listingError`'s `isInsufficientScopeMsg` arm proves this token family *does* 403 for a stale scope, whose remedy is re-running `civitai login` — the opposite of switching accounts. Now hedged and naming both, as the neighbouring under-determined 403s already do, with the sourcing recorded at the site.

### 🟡 The remedy's falsehood had moved, not gone

Round 1 replaced *"the refresh failed too"* (false when the refresh **succeeded** and the retry 401'd) with *"already failed twice"* — false when the refresh **errored** so no retry ever ran. The message must not name **which** half failed; it now says the CLI has already tried to refresh and retry, true of every OAuth path.

### 🟡 Two claims about guards that were not true

- `unauthorized.go` named `TestUnauthorizedCallersAreLedgered` — **a test the same commit deleted** — and described a per-file ledger that no longer exists.
- *"reports file:line for every arm it finds"* was false: it reports only arms it **rejects**, i.e. never in the green state.

### 🟢 Two more

Indirect routing was a false positive with misdiagnosing advice (fixed by the re-key). And the guard's docstring claimed both blind shapes *"already exist in this package"* — only the case-list did; the branching one was proven by a mutant. Corrected, in a file whose whole subject is instruments claiming more than they measured.

**Also removed a dead branch draft 4 introduced** — `if s.Routed && len(s.Builds) == 0` after an identical earlier `continue`. That is the #645 shape, caught in my own code.

### Verification

Battery re-run against draft 4: **10/10 correct outcomes**, each asserting `APPLIED = True` from disk first, positive control PASS. Six shapes KILLED including all three that survived draft 3; the indirect-routing check correctly **SURVIVES**.

⚠️ In the mutant that disables the `if` matcher the **floor fires first**, so the form control's own assertion never runs. Said so in its comment rather than crediting it with the kill — its distinct value is the case the floor cannot see, a form lost while a site is added.

`make ci` green · `go test ./... -count=1` 21/21 · `make ci-shallow` **21/21 depth-1** · wide gate `ok` at 81 · gofmt/vet clean · `golangci-lint` 0 issues (2.13.2 local).

⚠️ `pins-vs-published` remains red and **environmental** — reproduced on a clean `origin/main` at `321fcd8` with zero local changes (`@civitai/app-sdk` 0.42.0, `@civitai/blocks-react` 0.51.0, both published after today's 12:35Z bump ran green). A bump run has been dispatched; merging that PR is the operator's by standing decision.

```audit-claims round=2 audited=ad426fe..b92789bd
1. The 401 guard is re-keyed from enclosing syntax to message CONSTRUCTION: it flags any fmt.Errorf/errors.New inside a region conditioned on StatusUnauthorized, covering case, case-list, if, switch{case expr} and bare 401 alike.
2. The refresh-and-retry at appblocks.go:580 now passes because it builds no message, not via an exemption; indirect routing through a wrapper no longer fires.
3. The guard's comment states what it does NOT cover (a message built in a helper not associated with a 401 condition; one assembled from a variable) instead of claiming no blind spot.
4. withdrawError's 403 is hedged and names both causes plus the 404-is-not-yours fact, with WithdrawPath's vendored contract cited at the site.
5. unauthorizedRemedy no longer names which half of refresh-and-retry failed.
6. unauthorized.go no longer names the deleted TestUnauthorizedCallersAreLedgered, and no longer claims a per-file ledger exists.
7. The "reports file:line for every arm" claim is corrected to say it reports only rejected arms.
8. The docstring no longer claims the branching shape already existed in this package; it says the mutant proved it.
9. A dead branch introduced by this round's own draft is removed, and the floor is raised 8 -> 9 for the newly-visible `if` site.
10. claudedocs/readme-reduction-plan.md records that draft 3 was pattern-blind too, and that the fix was to re-key rather than widen a fourth time.
```

Issue #227, opened 2026-08-06
docs: app listing is absent from the README, and the download example in --help errors out


ZacxDev:
Two first-impression docs defects found by a blind dogfood run. Both are small edits with outsized value.

## 1. `civitai app listing` has zero mentions in the README

```
$ grep -c "app listing" README.md
0
```

Zero in an 80KB document — not in the command-reference table, not in the quickstart, not in the scaffolded project README. It **gates publication**: `A store listing must have an ICON and a COVER before it can go live`. The dogfooder found it only by idly reading `civitai app --help`.

Nearly saved by the tool itself — `app submit` surfaces it, unprompted, at exactly the right moment:

> Store listing: your listing needs an icon AND a cover before it can publish.
> You can add them NOW, while the app is in review — they carry forward on approval:
>   civitai app listing set-icon <file>

That placement is excellent. But someone planning a release from the docs budgets zero time for artwork.

## 2. The headline `download` example fails

Printed in `civitai --help` **twice** (description block and Examples block) and in the README quickstart:

```
$ civitai download 128713 --layout comfyui --root ~/ComfyUI
Error: 128713 is ambiguous — it's both model "Airi Akizuki … (鬼父)" and
version 128713 (of model "DreamShaper").
rc=2
```

The error itself is good — it names both interpretations and gives both remedies. But the first command a curious new user copy-pastes errors out, and the message lands them on an unrelated NSFW-titled model. `--version 128713` works fine.

Pick an unambiguous id for the example.

PR #656, opened 2026-09-17
docs(readme): link the read path out where the site actually covers it — and measure where it does not

ZacxDev:
Rank 1 of `claudedocs/handoff-readme-slimming.md` (L1 + L2), claimed as `readme-slimming-1`.

## Result: **−5,480 B** (277,572 → 272,092, −1.97%)

The handoff projected a **−37 KB floor**. It does not hold. Establishing *why* is the durable half of this PR, because the same false premise would have driven L3 and L4.

## The −37 KB floor assumed two things. Both were measured false.

Probed against the live `developer.civitai.com/site/guide/cli` and a binary built from this commit (`v0.1.105-37-gb6e894b`):

**1. `## Scripting with --json` (10,426 B) would delete wholesale.** Only **2,780 B** is site-covered. The remaining 7,646 B — the read-body repair ([#525](https://github.com/civitai/cli/issues/525)), the numeric-username shape ([#513](https://github.com/civitai/cli/issues/513)), `### Generation --json`, and the 🔴 `app listing status --json` side-effect warning — is on no site page. The guide explicitly does not cover `civitai app …` or `generate`.

**2. `## Download model files` and `## Browse the public API` are "duplicated".** The site's Download coverage is **one sentence per topic**. It has **no type→folder table**, and nothing on: the ambiguous model-id/version-id safety stop, `--force`, `--yes`, pickle/executable safety, the ControlNet preprocessor note, SHA256-verifies-integrity-not-authenticity, filename sanitisation, retry/backoff, or exit codes.

## 🔴 And the site is wrong in two places — ⚠ two further rows were RETRACTED by round 0

| # | site says | measured | |
|---|---|---|---|
| 1 | `# Homebrew (macOS / Linux)` → `brew install civitai/tap/civitai` | `.goreleaser.yaml:85` has `homebrew_casks:` and **no** `brews:`; a cask is macOS-only. `TestREADMEHomebrewSectionMatchesTheReleaseConfig` exists *specifically* to stop this README making that claim | ✅ |
| 2 | `model-versions get 999999999 --json` **exits 1** | binary exits **4**; this README says 4 and is correct. A **scripting contract**, so the site's version is the damaging direction | ✅ **new** |
| 3 | download flag list | omits `--force` and `--yes` (the ambiguous-id **safety** stop) | ✅ |
| ~~4~~ | ~~documents `--json` on `download`~~ | **RETRACTED.** Every `--json` mention is scoped to READ subcommands; the page twice excludes download from that set. The binary half is true, but the site never claims otherwise | ❌ |
| ~~5~~ | ~~omits the "Prebuilt binary" install method~~ | **RETRACTED.** The page carries *"Prebuilt binaries for linux/macOS/windows × amd64/arm64 are on the GitHub Releases page."* | ❌ |

🔴 **How the two false rows got in, because it is the transferable half:** they were taken from an **LLM summary of the page** rather than the page. Asked to list the download flags, the summarizer folded the read-commands' `--json` into the Download section and dropped a sentence — both invisible in its output, both reading exactly like a measurement. Re-derived by curling the page, stripping tags and grepping the sentences. **Diff against the raw page, never a summary of it.**

**No conclusion moves.** Keeping `## Install` rests on row 1; declining to cut `## Download model files` rests on thinness, which round 0 re-derived independently (no type→folder table, nothing on the same-named-file refusal, pickle safety, the ControlNet note, saferune sanitisation, the `429`/`Retry-After` rule, `--force` idempotency, or exit codes).

**`## Install` is kept local** as a result (operator's call): linking it out would retire a bidirectional goreleaser↔README guard *and* hand Linux readers an instruction that cannot work.

⚠ **Nothing is filed against `civitai/civitai-developer-docs`** — outward-facing, and the operator's call. Rows 1–3 are the ones that would be filed.

## What changed

| Δ bytes | section |
|---:|---|
| −2,074 | `## Set up your coding agent` — L1's two subsections, whose content `agent-setup --help` carries in full (verified string by string; the facts `--help` lacks — per-agent paths, Windsurf Cascade-vs-Devin, symlinks — are linked to decisions 34/35, both confirmed present on `origin/main`) |
| −2,370 | `## Scripting with --json` — the four recipe subsections the site covers (Gotchas near-verbatim, Worked example byte-identical heading). Preamble and `### Generation --json` **kept** |
| −353 | `## Contents` |
| −227 | `## Quickstart: browse & download` → short form (login, one search, one download, then the guide) |
| −429 | `## Browse the public API` — base-model tutorial only. The **command table stays**: it is the anchor target for 6 inbound links and is pinned by `TestREADMECommandReferencePointsAtTheReadCommandTable` |
| −27 | `## Download model files` — a signpost, not a cut. (Was **+98** until round 0: the "what the guide does not carry" clause was an unguarded negative claim about a third-party page, and is gone.) |

## New guard: `internal/cmd/readme_external_links_test.go`

The handoff's standing decision — *a link-liveness guard ships WITH the links, not after the first 404*. Sections were deleted in favour of a URL, so a 404 on that URL is content loss, not a blemish.

- Network-gated on `CIVITAI_CHECK_README_LINKS=1`, following `internal/scaffold/pins_guard_test.go`; `go test ./...` stays offline.
- **404/410 FAIL. Unreachable / 5xx / 401 / 403 / 429 SKIP** — a permanently-red gate is worse than no gate. `www.npmjs.com` answers **403 to every user-agent including a browser one** (measured with curl), so failing on it would redden this forever.
- Floor **20**, deliberately slack (live count 38). It was 37 — the exact count — until round 0 pointed out that a floor on the count is a ratchet: any honest link removal went red telling you to lower the constant. What defends the guard is the `mustFind` positive control. Plus a positive control asserting the CLI-guide URL is among what was extracted, so a regex matching 37 GitHub URLs and zero `developer.civitai.com` ones cannot pass.
- **Its first live run went RED on three findings** — that is the negative control. Two were deliberate placeholders (`example.com/a.jpg`, `github.com/me/my-app`) and one the npmjs 403; all three are now classified with their reason in the source.

`download_example_id_test.go`: README `minHits` **8 → 5**, the actual count, justified in the same commit. It is a positive control on the *regex*, never a floor on how many examples the README owes; every surviving example is still checked against `unambiguousExampleIDs`, so #227 coverage is unchanged.

`readme-reduction-plan.md`: retires the *"user contract content never moves out"* clause for content the site genuinely covers (operator, 2026-09-17) — and records the measurements above, so the next session does not re-derive a −37 KB estimate from a matching heading list.

## Verification

- `go test ./... -count=1` — **21 packages, 0 failures** (`-count=1` is load-bearing: the input under test is a non-Go file)
- wide README gate — `ok`, **82** tests
- `make ci` green; `make ci-shallow` — **21/21** in a depth-1 clone, run against the commit, not a dirty tree
- `golangci-lint run` — **0 issues** (locally 2.13.2; CI pins 2.12.2, so read this PR's own `lint` run — it is not a required context)
- live link check — **38 extracted, 35 fetched, 3 skipped, 0 dead**

## Scope left undone, deliberately

`## Download model files` and the behavioural half of `## Browse the public API` were **not** cut. The evidence says the site does not carry them, and the arc's standing rule is not to trade published contract for bytes. The remaining bytes are in `## Submit & auth` (52,774) and `## Generate` (43,194) — 35% of the file — which are L3/L4, not this PR.


ZacxDev:
## Round 0 — requirements & deletion — recorded as prose

`round 0 · requirements: 12 (unattributed: 3) · deletion candidates: 6`

Round 0 reports and does not move the ladder, so this is a record, not a stop signal. It ran at PR-create against `7cff466`. **All of its findings are fixed in `678bd0b`.**

### 🔴 It refuted two of the four "measured" site-drift rows this PR published

Both retracted, and the PR body is corrected rather than quietly edited — a reviewer may have read the wrong version:

- ❌ *"the site documents `--json` on `download`"* — every `--json` mention on the page is scoped to **read** subcommands, and the page twice excludes download from that set. The binary half is true; the site never made the claim.
- ❌ *"the site omits the Prebuilt binary install method"* — it carries *"Prebuilt binaries for linux/macOS/windows × amd64/arm64 are on the GitHub Releases page."*

**How they got in:** both were taken from an **LLM summary of the page** rather than the page. Asked to list the download flags, the summarizer folded the read-commands' `--json` into the Download section and dropped a sentence — invisible in its output, and reading exactly like a measurement. Re-derived by curling the page, stripping tags and grepping sentences.

That probe also found a **real** item the summary had hidden, and a worse one than either retraction: the site states `civitai model-versions get 999999999 --json` **exits 1**; the binary exits **4**, and this README says 4. It is a scripting contract.

**No conclusion moved.** Keeping `## Install` rests on the Homebrew row; declining to cut `## Download model files` rests on thinness, which round 0 re-derived independently.

### Other findings, all fixed

| finding | fix |
|---|---|
| 🟡 `notAPage`'s comment claimed *"EVERY ENTRY HERE WAS PUT IN BY A MEASURED RED"* — false of 2 of 3 | rewritten per-entry; `orchestration.civitai.com/mcp` **deleted** (measured 401, already SKIPped, so inert); `mcp.civitai.com/mcp` kept with its real reason (measured **405**, which would fall through to the `>= 400` arm) |
| 🟡 3 of 5 `placeholderHosts` match nothing in the README | `www.example.com`, `example.net`, `example.org` deleted |
| 🟡 the `+98` signpost is maintainer prose in the shipped contract, and the three *"which that guide does not carry"* clauses are **unguarded negative claims about a third-party page** | all three clauses dropped. This is the finding I think was most worth having: the PR guarded the positive claim (the URL resolves) and not the negative one, which is the half justifying 12,515 B. `## Download model files` goes **+98 → −27** |
| requirement questioned: `readmeMinExternalURLs = 37` is a ratchet, not an invariant | **37 → 20.** Set on the exact count, any honest link removal went red telling you to lower the constant. The #652 lesson it cited was about a floor too *slack*; applying it here overshot. `mustFind` is what actually defends the guard |
| 🟢 byte figures off by one; one verification figure did not reproduce | **corrected, and it is the arc's own defect class**: `−5,294` was measured on a tree the same PR then edited again. Re-derived with two instruments at this tree: **−5,480 B**, 277,572 → 272,092 |

### 🟡 Left OPEN deliberately — needs your call

**`TestREADMEExternalURLsResolve` is executed by no CI job and no cron.** `CIVITAI_CHECK_README_LINKS` appears in exactly one file — the test defining it. The precedent it copies does *not* have this gap: `pins_guard_test.go`'s variable is set by a dedicated `pins-vs-published` job.

Stated precisely, because the PR should not over-claim: the **extraction** half (`TestREADMEExternalURLsAreExtractable`) runs under `go test ./...` in `build-test` and does gate. The **liveness** half is operator-run only.

Fixing it means adding a scheduled workflow, and `AGENTS.md` puts `.github/workflows/*` under **"Ask first"** — so it is not done here. Either wire a cron job on the `release-homebrew.yml` model, or delete the network half; what it must not do is read as coverage it does not provide.

### Could not verify

Round 0 could not run `golangci-lint` (absent in its worktree; `make lint` correctly refused to degrade). Run locally on this branch: **0 issues** at 2.13.2. CI pins 2.12.2 and `lint` reports without gating — read this PR's own `lint` run, which is green.

ZacxDev:
## Round 1 — the nine correctness axes

Dispatched **blind** (the diff and the checklist, not round 0's conclusions — a framed audit verifies the frame). Read `678bd0b`; fixes produced `4c69017`.

**Verdict: safe to merge, no 🔴.** Under the operator's stop rule for this arc — *audit until a round returns no high-severity findings* — and its scoped depth — *round 0 at PR-create + one correctness round, not an open ladder* — **this ends the ladder.** No round 2. All four findings are fixed below rather than deferred.

### It validated the new guard rather than trusting it

Both controls, run independently:

- **negative** — appended a bogus `/site/guide/this-page-does-not-exist-xyz` to a copied README → `FAIL … DEAD LINK … answers 404`.
- **positive (`mustFind`)** — rewrote the four guide URLs to `/site/guide/other`, holding the count at 38 so the floor could not be what fired → `FAIL: CONTROL failure`.

It also probed two bogus paths on `developer.civitai.com` and confirmed both return a real **404**, not an SPA 200 — so a dead guide link cannot land in the SKIP bucket. That is the check I should have run and did not.

Every figure in the PR body was re-derived and held, including both retractions, the exit-`1`-vs-`4` item, and the content-survival trace for each deleted passage.

### Findings — all four fixed in `4c69017`

| # | finding | fix |
|---|---|---|
| 🟡 2 | `README.md:430` told readers the per-agent **key names** are "derived and evidenced in" decisions 34/35. Measured: `mcpServers` **0** occurrences in both, `context_servers` **0** in both | the section now names the keys inline and points at `--help`, which does carry them. A Zed/VS Code user needs that key precisely because no header is written for Zed — they'd have followed the signpost into nothing |
| 🟡 3 | `minHits` 8→5 made it an **exact-count ratchet** — and the base had **9**, so the old constant carried slack | **8 → 3.** This is the identical mistake this PR had *just* removed from `readmeMinExternalURLs` (37→20), reintroduced two files away |
| 🟢 4 | `README.md:241` still said "the behaviour that guide does not cover" — a **fourth** instance the previous commit's sweep missed | dropped. Swept by meaning this time, not by string |
| 🟢 5 | `README.md:2601` claimed the guide works through `--base-model` "with examples" for both `models search` **and** `images search`; the guide's images-search flag list omits the flag entirely | now says which half it covers |

Findings 2 and 4 are the same class the arc keeps producing — *a sentence claiming more than its evidence* — and finding 3 is worse than that: an inconsistency **inside a single PR**, where I fixed a ratchet in one file and wrote one into another.

### Left open, needs a decision (unchanged from round 0)

**`TestREADMEExternalURLsResolve` is run by no CI job and no cron.** Round 1 independently confirmed it: `CIVITAI_CHECK_README_LINKS` appears nowhere outside the file defining it, while the precedent it names (`pins_guard_test.go`) has a dedicated job **and** a positive control asserting its `-run` pattern matched something.

Precisely: the **extraction** half gates via `build-test`; the **liveness** half is operator-run only. `.github/workflows/*` is "Ask first" per `AGENTS.md`, so this is the maintainer's call — wire a cron job on the `release-homebrew.yml` model, or delete the network half.

### Could not verify

Round 1 could not reproduce "wide README gate — 82 tests": no such script or CI job exists, and its own `-run 'README|Readme'` gave 75. That is the filter difference, not a discrepancy — the gate is the literal command in the handoff, `-run 'Attribution|Troubleshooting|README|Readme|readme'`, which is wider. Flagging it because "82" is unattributed unless the command is quoted with it.

### Byte figure — it moved again, and that is the lesson

**−5,111 B** (277,572 → 272,461) at `4c69017`. Four totals were reported across this PR — −5,294, −5,293, −5,480, −5,111 — each true when taken and wrong when quoted, because the PR kept editing. **Quote from the merge commit.**

```audit-claims round=1 audited=678bd0b..4c690171
1. README:430 no longer claims decisions 34/35 evidence the per-agent key names; the keys are named inline and `--help` is cited instead.
2. download_example_id_test.go README minHits set to 3, removing the exact-count ratchet introduced at 5.
3. README:241's fourth "the guide does not cover" negative claim about a third-party page removed.
4. README:2601 narrowed to say the guide's `--base-model` examples cover `models search` only, not `images search`.
5. readme-reduction-plan.md's yield figure corrected to -5,111 B and annotated with all four superseded totals.
```

Issue #665, opened 2026-09-18
agent-setup: the documented --prefix install remedy leaves the CLI unreachable from every later shell


ZacxDev:
## What happens

On any machine where npm's global prefix is not writable, `developer.civitai.com/agent-setup/prompt.md` §2 "If the install fails" correctly tells the agent to install into a user-owned prefix and prepend it to PATH:

```bash
npm install -g --prefix="$HOME/.npm-global" @civitai/cli
PATH="$HOME/.npm-global/bin:$PATH"
```

The agent then completes the setup and reports back. But `civitai` is reachable only from the shell that installed it:

```
$ command -v civitai            # a fresh non-login shell
$ zsh -lic 'civitai --version'  # command not found
$ ls ~/.npm-global/bin          # civitai   <- it is right there
```

## Why it matters more than "the user has to edit their profile"

Step 3 writes an `AGENTS.md` into the project, and that file tells **every future agent session** to run `civitai …`. Every future session gets a new shell. So the durable artifact the setup exists to produce names a binary the setup left unreachable.

## Measured

19 blind dogfood trials, 4 models × 4 container environments (civitai/cli#663; harness at `scripts/dogfood/`, full record at `claudedocs/refs/agent-setup-dogfood-matrix-2026-09-18.md`).

- **8 of 8** trials on a non-writable prefix ended this way — all four models (claude-sonnet-5, gpt-5.6-terra, gemini-3.8-flash, grok-4.6). Not model-dependent.
- **8 of 8** relayed the PATH line to the user, as step 5 requires. The prompt's instruction is obeyed.
- **6 of 8** additionally warned it would not persist and named the profile file. The agents are not the weak link.
- **0 of 8** connected it to the `AGENTS.md` they had just written.

The prompt itself notes this environment is "normal on NixOS, in some container images, and wherever Node came from a system package manager" — so it is a common path, not an edge case.

## Why step 4 cannot catch it

Step 4's verification is `civitai --version`, run in the **same shell** as the install — the one shell in which it works.

## Candidate fixes — 🔴 the ranking is abandoned; this is a measurement

Three successive drafts of this section ranked these three remedies, and all three
rankings were wrong. Replacing the ranking with what was measured.

- **(a) Verify in a NEW shell** (`bash -lc 'civitai --version'`) — does not fix the
  install; stops the failure being silent.
- **(b) Install into a prefix whose `bin` is already on PATH.**
[...]

PR #671, opened 2026-09-19
fix(agent-setup): tell every AGENTS.md what to do when `civitai` is not on PATH

ZacxDev:
Addresses cli#665. **Third revision** — a second round-0 audit found the previous one had reintroduced the class the first round raised, so the login-shell probe is **deleted** (251 lines, plus its tests and the per-machine render axis). What ships is one unconditional sentence.

Net diff vs `main`: **3 files, 116 insertions, 0 deletions.** Was 631.

## The defect, unchanged

On a machine where npm's global prefix is not writable, the documented remedy puts the CLI on PATH for the **installing shell only**. The managed block then names `civitai …` in every row, and no later session can run any of them. Measured **8 of 8** trials (#663) — and **0 of 8** agents connected it to the `AGENTS.md` they had just written.

🔴 **That last number is the actionable one, and it is an absence in this file — not a missing detection.** The agents were not the weak link: 8/8 relayed the PATH line, 6/8 warned it would not persist. What nobody had was the *connection*, in the file the next session actually reads.

So it goes there — unconditionally, naming the symptom (`command not found`), the cause, and the action (`command -v civitai` in a shell where it works, then add that directory to the profile).

## Why the probe is gone

Two drafts **detected** the condition by running a login shell with `PATH` stripped, and wrote the binary's absolute path when it looked unreachable. Each was measured wrong, and each wrong *in a way the previous fix introduced*:

| draft | fix | what it broke |
|---|---|---|
| 1 | strip `PATH` | left the profile's idempotence sentinel set, so the profile built **no PATH at all** — a correctly-installed CLI read as unreachable on the maintainer's own platform |
| 2 | strip sentinels too, add `WaitDelay` for an unrelated hang | `exec.ErrWaitDelay` returns with the shell **exited 0 and the resolved path already on stdout** — scored "not reachable", answer discarded |

And **neither fix had a guard that could catch its removal.** Measured on the full package:

```
delete cmd.WaitDelay            -> ok  (suite green)
revert the sentinel strip       -> ok  (suite green)
```

Both guards re-implemented their subject instead of calling it — the exact sin the round before them was raised to kill — and one asserted the opposite in its own comment.

🔴 **The shape, not any one bug: a per-machine signal whose only consumer was a write into a committed file.** Every inversion channel — an unlisted distro sentinel, a backgrounding profile, a shell that would not run — put a developer's home directory and a bold-red false directive into a repo other people pull. The sentinel list was a denylist, incomplete by construction; `__ETC_PROFILE_NIX_SOURCED` gates PATH on every multi-user Nix host and was missing from it.

**So: this block may depend on the PROJECT, never on the MACHINE.** A per-project sentence is wrong for nobody; a per-machine claim is wrong for everyone who did not generate it, and nothing detects the staleness — the `agents-md` row is satisfied by block *presence*, not by comparing against a freshly rendered block.

## Verification

The new guard renders **every** project shape and asserts both directions: the note is present, and no absolute path or machine-specific directive ever is. **Watched red** — removing the note from the template fails it with its own message.

A repo guard also caught my first wording (`npm prefix -g`): a static project has no `package.json`, and `TestAStaticProjectIsNeverToldToRunNPM` refuses to let the block name a package manager. Reworded to name none — a good catch by an existing guard.

`make ci-shallow` — depth-1 tier, committed state — **21/21**.

`README.md`, `agent_setup.go`, `agent_setup_files.go`, `agent_setup_test.go` and `scaffold_prompt_test.go` are all byte-identical to `main` again.

## Not closing the issue

⚠ Deliberately **not** `Closes #665`. The issue's own body says candidate (c) *"satisfies neither arm"* of its closing condition, and that remains true — this ships no install fix. The cause-side remedy is candidate (b), in the hosted prompt (ranked item 35).


ZacxDev:
Round 0 ran before the merge decision and returned **needs rework**. All three 🔴 were reproduced independently before acting on them, and all three are now fixed — the NixOS sentinel false-negative (with an A/B control on the same host), the unbounded `Output()` wait, and a test that claimed to exercise the real probe while calling a copy of it.

The `reachable` check row is **removed**, not defended: its stated justification ("a false negative merely warns") was refuted by finding 1, since the row and the AGENTS.md block key on the same boolean. `README.md` and the ledger test are now byte-identical to `main`.

`Closes #665` has been dropped from the body — the issue's closing condition needs a hosted-prompt decision this PR does not make.

ZacxDev:
Second round 0 returned **needs rework** again, and it was right: the previous revision's own fix had reintroduced the class the first round raised. Verified independently before acting —

- `exec.ErrWaitDelay` returns with the shell **exited 0 and the resolved path on stdout**, which the probe scored `Reachable=false` and discarded. The fix for the hang created a third false-negative channel.
- Both headline fixes were **unguarded**: deleting `WaitDelay` → suite green; reverting the sentinel strip → suite green. Both guards re-implemented their subject rather than calling it.

So the probe is **deleted**, not patched a third time — along with the per-machine render axis that made every inversion write into a committed file. What remains is one unconditional per-project sentence, guarded in both directions and watched red.

Net diff is now 3 files / 116 insertions, down from 631.

PR #683, opened 2026-09-21
feat(dogfood): a credential path that does not leak, and a brief that exercises generation + posting

ZacxDev:
Closes the blocker #682 records: **the only credential path into the dogfood harness was a leak.** Adds a second brief that exercises generation and posting, and makes the `dogfood-3` run bounds mechanical where they can be.

🔴 **No trial was run. Nothing was spent.** No OpenRouter call, no `civitai generate`, no `app submit`, no Buzz. Every measurement below is offline, against stubs and hand-written fixtures. Container `dogfood-ab-curve-01` was not touched.

---

## Part 1 — the credential path

`--agent-env VAR=VALUE` was the only way to get a variable into a trial container, and `runner.py` wrote its value verbatim into the `start` record of `runs/<trial>/transcript.jsonl`. **`--credential-file`** replaces it. The flag takes a **path**; the file is `docker cp`'d in and installed by a fixed script run as root.

| surface | what keeps the secret off it |
|---|---|
| **process argv** (`ps`) | the value is never an argument. `docker run -e CIVITAI_TOKEN=…` would expose it to every process on the host; `TestDogfoodCredentialIsInjectedByPathNotByValue` fails if anyone reintroduces it |
| **transcript / command log / stdout** | a redactor keyed on the credential's own strings filters every value at the write, so even a trial in which the model `cat`s the config records `[REDACTED:<sha8>]` |
| **container shell history** | `docker exec … bash -lc` is non-interactive, so bash writes no history file — and nothing in the install path holds the value anyway |

Recorded instead: `credentialed: true` and `credential_sha256`, a 12-hex prefix of the file's digest. Not reversible; not a substring (pinned in both directions).

⚠ **What this does NOT do.** The credential is a real file in the container and the model runs shell as a user who can read it — it must, or the CLI cannot authenticate. The guarantee is about what leaves the container and lands in the artifacts you keep.

### The leak test, with its positive control

`TestDogfoodCredentialNeverLeaks` plants a high-entropy canary in a config file, runs a full offline trial in which the model runs `cat ~/.config/civitai/config.yaml` and `echo "token=$CIVITAI_TOKEN"`, then greps four surfaces.

```
positive control  countHits(un-redacted credential file) = 2   ← the same grep, same pattern
transcript.jsonl                                            0
commands.log                                                0
process argv (every subprocess argv the runner spawned)     0
runner stdout                                               0
```

The control runs **first** and `t.Fatal`s if it is below 2 — a grep that finds nothing because its pattern is wrong is indistinguishable from one that finds nothing because the secret is absent. A second control requires `[REDACTED:` to appear in the transcript, so a zero cannot come from a fixture whose output never carried the credential at all.

**Mutation control** (the red-at-base arm cannot prove this — at `origin/main` the flag does not exist, so there is nothing to leak). Deleting the two lines of `Redactor.text` that do the replacing:

```
M1  TestDogfoodCredentialNeverLeaks  killed:
    the credential leaked into transcript.jsonl: 4 occurrence(s) of the planted secret
```

---

## Part 2 — `genpost`, the generation + posting brief

`briefs/genpost.brief.txt` (one line):

> Build a Civitai App Block that generates an image from a prompt and then posts it to Civitai: a text input with data-testid="prompt", a button labelled Generate, a button labelled Post that starts disabled and only becomes enabled once a generation has succeeded, and an element with data-testid="status" holding exactly ready before anything is clicked and exactly generating while a generation is in flight.

🔴 **The assertion cannot see a generation or a post happen, and never will here.** The oracle's host emulation is the SDK's `InlineTransport` — a v1 stub whose `sendRequest` rejects and which delivers no host pushes. An assertion that waited for an image or a post id would time out against a perfect app. It grades the block's own state machine on the near side of the request: the `ready` resting state, the Post gate being **closed** before anything has been generated, and the Generate click driving the machine into `generating`. Read a green cell as *"the flow is wired and gated correctly"*, never as *"the money path works"*. `briefs/genpost.md` says this at the top.

**Why the Post gate and not generation.** `page-money` already ships a prompt field and a Generate button wired to a real `submitWorkflow`. A generation-only brief is satisfied by an untouched scaffold. Posting is the half no template ships — no `posts:write:self` in any scaffold manifest, no Post control, no gate.

### 🔴 The scaffold control — measured, with observed output

`civitai app init`, no edits, `npm install && npm run build` for the two page templates. Chromium 153.0.8010.47 / node 24.20.0 / CLI `v0.1.106-8-g39c2785`, 2026-09-21. **All three FAIL:**

| template | `pass` | reason | body at failure |
|---|---|---|---|
| `static` (the default) | `false` | `timed out after 8000ms waiting for [data-testid="prompt"] to appear` | `<main id="app"><h1>Sc Static</h1><p>This is a Civitai App (static template)…</p><button id="ping">Click me</button></main>` |
| `page-vite` (built) | `false` | same | `<div id="root"><main class="app">…<button>count is 0</button><button>reset</button></main></div>` — React had **mounted**, so a real absence, not a load failure |
| `page-money` (built) | `false` | same | the **full generation form** rendered: `Sc Money / Text to image / Prompt* / Describe what you want to generate…` — the host handshake worked and the block was alive; it has `pm-`-prefixed test ids, no `data-testid="prompt"`, no `data-testid="status"`, no Post control |

⚠ **The scaffold control is not sufficient on its own here**, and the PR would be weaker if it stopped there: all three fail at *step 1*, the test id — which says nothing about posting. So `neg-nopost` is the control that matters: **a complete, working generate-only app carrying the exact test ids the brief names, and it fails** (`no clickable element labelled "post" — nothing posts`, `buttonLabels: ["Generate"]`). It is a row in `TestOracleGradesTheGenpostBrief`, not just a doc table.

### Oracle wiring

`oracle.sh` already dispatches on `briefs/<name>.assert.mjs`, so `genpost` is graded by extension, not a fork. One field added: **`scopes=`**, read from the trial's `block.manifest.json` and **reported, deciding nothing** — same standing as `validate_gate=`. A manifest is a declaration, not a behaviour. Pinned in both directions (`none` declared + app works → `yes`; both scopes declared + app broken → `no`).

The CDP/browser plumbing the two assertions shared is extracted to `briefs/_cdp.mjs`. `celsius.assert.mjs` keeps its constants and its comments and behaves identically — the existing `TestOracle*` suite drives it through a real browser and is green.

---

## Part 3 — the run caps: what is MECHANICAL and what is PROSE

The operator approved a credentialed run under the `dogfood-3` bounds. #682 notes those bounds were prose and that *"nothing enforces it"*. Four of the five are now enforced in code.

### Mechanical — the command is judged BEFORE the `docker exec` and a capped one never runs

| cap | flag | how |
|---|---|---|
| **app-name prefix** | `--app-prefix` | a mismatched `app init` / `create` / `submit` / `listing` is refused. With no slug on the command line it reads **every** `block.manifest.json` under `/work` and requires all of them to carry the prefix — the container starts empty, so every manifest there was created by the trial |
| **generation cap** | `--max-generations N` | the N+1th `civitai generate` is refused |
| **submission cap** | `--max-submissions N` | the N+1th `civitai app submit` is refused |
| **`app withdraw`** | refused by default; `--allow-withdraw` to permit | it permanently destroys a listing's captioned screenshots (`claudedocs/handoff-dogfood-3.md`) and names a publication-request id no prefix check can resolve to an app |
| **command log** | always on | `runs/<trial>/commands.log`, one line per command with its verdict and the counters, refused commands included |

`TestDogfoodCapsRefuseBeforeTheCommandRuns` keys on the **absence of a `tool` record**, not on the refusal string — a "cap" that ran the command and then complained would pass a string-keyed test. `TestDogfoodCapsAllowUpToTheLimitThenRefuse` proves it is a **count**, not a switch: `[run, run, refused]` at `N=2`.

### 🔴 PROSE — state plainly, do not dress up

- **The classifier reads COMMAND TEXT.** A model that writes a command into a file and runs the file, or builds it at runtime from pieces the classifier does not see, is **not stopped**. There is a fail-closed narrowing — a segment naming the CLI *and* a spending or publishing verb with no invocation the harness can parse (`eval "civitai app submit"`, `c=civitai`) is refused — and it does not close the gap.
- **The threat model is a COOPERATIVE agent**, the one `handoff-dogfood-3.md` settled on. These stop the ordinary accident: a loop that regenerates, a submit aimed at the wrong directory, a withdraw typed against a real listing. They are not an adversarial bound. That needs a second uid, a container with no credential in it, or a platform-capped token — not a better classifier.
- **"Do not touch the operator's existing apps" is still partly prose.** The prefix check makes `app init/create/submit/listing` against a foreign slug mechanical, and `withdraw` is refused outright. But a read-only command against a real listing (`app view sensei`, `app status`) is *allowed* by design, and any mutating surface the CLI grows without a matching entry in `APP_MUTATING` is ungated until someone adds it. That list is a hardcoded tuple in `runner.py`; nothing tells you when the CLI outgrows it.
- **The "one agent, one model" and "blindness" bounds remain convention**, unchanged by this PR.
- **The caps arm themselves when a credential is present**, so an operator does not have to remember three flags for the bound to exist.

---

## The default is unchanged — proved, not asserted

`TestDogfoodUncredentialedRunIsUnchanged` pins the exact key set of the `start` and `end` records on an uncredentialed run and fails if a `caps` or `refused` record appears. With no credential and no cap flag, nothing is judged at all.

⚠ **That test is an INVARIANT GUARD, not regression coverage — it PASSES on `origin/main`**, which it has to, because what it pins is that nothing changed. It is labelled as such in the file and is excluded from the red-at-base count below. It was watched to go red the only way it can: mutating `armed` to `True` (M5).

One deliberate asymmetry: `credentialed`/`credential_sha256` appear **only** on a credentialed run, so *absence* is the signal for "no credential". That is weaker than the positive assertion `brief` gets, and it is the price of byte-identity. The key set is pinned on both sides.

The command log *is* written on every run, credentialed or not. It is evidence, not enforcement, and it is the one behaviour change a default run sees.

---

## Tests

**Full suite, `go test ./...`, browser resolved:**

| | PASS | FAIL | SKIP |
|---|---|---|---|
| `origin/main` (`39c2785`) | 5800 | 0 | 7 |
| this branch | **5835** | 0 | 7 |

The 7 skips are pre-existing and network/env-gated (`TestREADMEExternalURLsResolve`, `TestScaffoldPinsSatisfyPublished`, …), identical on both sides.

### 🔴 Red at base

New tests copied onto a clean `origin/main` worktree (implementation at base), with `testdata/fake_trial.py` also held at HEAD for the caps arm so the red is about the missing cap and not about the fixture:

| test | failure at `origin/main` |
|---|---|
| `TestDogfoodCredentialNeverLeaks` | `reading …/runs/faketrial/commands.log: no such file or directory` |
| `TestDogfoodCredentialMarkerIsNotDerivedFromTheValue` | `a credentialed run did not record 'credentialed: true' — a reader cannot tell which trials reached an account` |
| `TestDogfoodCredentialIsInjectedByPathNotByValue` | `no 'docker cp' of the credential file — how did it get in?` |
| `TestDogfoodCapsRefuseBeforeTheCommandRuns` | `"civitai generate \"a cat\"": verdicts [run], want [refused]` |
| `TestDogfoodCapsAllowUpToTheLimitThenRefuse` | `with --max-generations 2 the verdicts were [run run run], want [run run refused]` |
| `TestDogfoodCommandLogRecordsEveryCommandAndItsVerdict` | `reading …/commands.log: no such file or directory` |
| `TestDogfoodCommandLogExistsWithoutACredential` | `reading …/commands.log: no such file or directory` |
| `TestDogfoodRefusesAnUnusableCredentialFile` | `the refusal does not name the problem ("No such file")` |
| `TestGenpostBriefAndAssertionAgree` | `open scripts/dogfood/briefs/genpost.brief.txt: no such file or directory` |
| `TestEveryBriefHasASiblingGuard` | `found 1 brief(s) under scripts/dogfood/briefs — this ledger cannot have checked anything` |
| `TestOracleGradesTheGenpostBrief` | `exit 2, want 0` (no assertion for brief `genpost`) |
| `TestOracleCarriesTheGenpostStatusSequence` | `exit 2, want 0` |
| `TestOracleReportsScopesWithoutDeciding` | `exit 2, want 0` |

13 red at base, 13 green at HEAD. `TestDogfoodUncredentialedRunIsUnchanged` passes at both and is the invariant guard noted above.

⚠ Two of those reds are honestly weak — they name a missing file, not a missing behaviour. `TestDogfoodCredentialNeverLeaks` in particular *cannot* go red at base on the leak itself, because at base there is no credential to leak. That is what the mutation battery is for.

### 🔴 Mutation battery — 11 mutants, 11 killed

Each applied in place, one at a time, with the killing test named and the tree verified byte-identical after restore.

| # | mutation | test | killed by |
|---|---|---|---|
| M1 | `Redactor.text` stops replacing | `…CredentialNeverLeaks` | `the credential leaked into transcript.jsonl: 4 occurrence(s)` |
| M2 | `credential_sha256` = 12 chars of the token | `…MarkerIsNotDerivedFromTheValue` | `credential_sha256 "civitai-dogf" is a SUBSTRING of the credential file` |
| M3 | `Caps.judge` always allows | `…CapsRefuseBeforeTheCommandRuns` | `verdicts [run], want [refused]` |
| M4 | the refusal branch is never taken | `…CapsAllowUpToTheLimitThenRefuse` | `verdicts were [run run run], want [run run refused]` |
| M5 | caps armed unconditionally | `…UncredentialedRunIsUnchanged` | `an uncredentialed run emitted a "caps" record` |
| M6 | `_prefix_ok` always allows | `…/a_foreign_slug` | `"civitai app listing … --slug sensei": verdicts [run], want [refused]` |
| M7 | the command log write is dropped | `…CommandLogRecordsEveryCommand…` | `commands.log has 1 line(s), want 2` |
| M8 | `scopes=` hardcoded | `TestOracleReportsScopesWithoutDeciding` | `scopes=ai:write:budgeted,posts:write:self, want none` |
| M9 | the Post-disabled check removed | `TestOracleGradesTheGenpostBrief` | `RENDER=yes, want no` |
| M10 | the status predicate forced true | `TestOracleGradesTheGenpostBrief` | `RENDER=yes, want no` |
| M11 | the resting-state check removed | `TestOracleGradesTheGenpostBrief` | `RENDER=yes, want no` |

🔴 **M9, M10 and M11 SURVIVED the first run, and that found a real coverage gap.** Every failing fixture in the original table threw *before* those three predicates were evaluated, so each guard was **unreachable** — breakable in principle, never executed in practice. Three fixtures were added to reach them: a Post control that is present but **enabled** at rest, a Generate button that does nothing, and an app whose status machine moves but says `submitting` instead of `generating`. The three near-miss rows are worth having on their own merits; the mutants are what surfaced that they were missing.

### Both skip/fail arms measured

Following the repo's precedent (skip locally, FAIL under `$CI`), for both gates:

```
browser gate,  no CIVITAI_CHROME, no CI  -> --- SKIP: TestOracleGradesTheGenpostBrief
browser gate,  no CIVITAI_CHROME, CI=1   -> --- FAIL: TestOracleGradesTheGenpostBrief
                                            "Under CI that is a defect: build-test would report `ok` having run nothing."
python3 gate,  PATH without python3, no CI -> --- SKIP: TestDogfoodCredentialNeverLeaks
python3 gate,  PATH without python3, CI=1  -> --- FAIL: TestDogfoodCredentialNeverLeaks
```

`go vet ./...` clean. `shellcheck -S warning` clean on `oracle.sh` / `grade.sh` / `driver.sh`.

---

## What I could NOT verify

- **No end-to-end credentialed trial was run**, by instruction. The injection path is exercised only against the stubbed Docker in `testdata/fake_trial.py`, so **`install_credential`'s real behaviour against a real daemon is unproven**: `docker cp` into a running container, `getent passwd` resolving the trial user's home in each of the four images, and the `chown`/`chmod` landing correctly for the non-root (`dev`) images. The argv SHAPE is pinned; the effect is not. **Run one uncredentialed-container smoke first** — `docker exec … test -s ~/.config/civitai/config.yaml` — before spending a cell on it.
- **`civitai` itself was never invoked against the account.** The `APP_MUTATING` / `APP_DESTRUCTIVE` lists were derived by reading `internal/cmd/app.go` and `app_listing.go`, not by exercising the commands.
- **Whether a model can actually satisfy `genpost`** is unmeasured — that is the trial, and the trial has not been run. The assertion is proved to discriminate between hand-written fixtures and to reject all three scaffolds; nothing here says an agent can pass it.
- **The classifier's coverage against a determined model** is untested, and by construction untestable from here. What is tested is the set of shapes in the table.
- **`ruff` count went 4 → 21** on the two Python files. `ruff` is not a gate in this repo (absent from `.github/workflows/` and the `Makefile`); the new findings are stylistic (`UP031` %-format, `C408`, `SIM102`). Left alone rather than churned.


ZacxDev:
## Addendum — the second commit, and a pre-existing red on `main`

### `c6c10f2` — a bug I introduced while hardening, and the `set -e` trap under it

Reviewing the installer script I found it could install the credential **in the wrong place, silently**: it runs as root (necessarily — `docker cp` stages the file as root, and a non-root trial user could neither `chown` it nor remove it from a sticky `/tmp`), so `$HOME` inside it is **root's** home, not the trial user's. On an image without `getent`, a `dev`-user trial would have got the credential at `/root/.config/civitai` and then graded as an ordinary *"not authenticated"* failure — the capability confound, arriving through the setup step.

🔴 **The first fix was itself wrong**, and shellcheck did not catch it. Under `set -e`:

```sh
[ -n "$h" ] || { [ -d "/home/$u" ] && h="/home/$u"; }   # exits the script when /home/<u> is absent
```

`A || B` with a failing `B` fails the whole compound, so `set -e` kills the script **at the first fallback** — the second fallback is unreachable and the diagnostic three lines below never prints. Rewritten as plain `if`s.

Tested by running the script itself on the host against a stubbed `getent`, with the staging path substituted so nothing touches a real `~/.config/civitai`:

- resolvable user → installed at that home, content matches, **mode 0600**, staging copy **removed**
- unresolvable user → non-zero exit carrying `cannot resolve a home directory for container user '…'`

That closes the biggest of the "could not verify" items in the body — the *logic* of the installer is now measured on a host. What is still unproven is `docker cp` and `getent` against a real daemon in the four trial images.

### 🔴 `main` is intermittently red on a browser-launch flake, and this PR roughly doubles the exposure

Not introduced here, and worth knowing before merge. The push of #680 to `main` (run `35557830290`, `2026-09-21T03:31Z`) failed:

```
--- FAIL: TestOracleGateNeverDecidesTheVerdict/red_gate_over_a_working_block (17.09s)
    dogfood_oracle_test.go:381: exit 2, want 0
    {"assertion":"celsius","pass":false,"reason":"harness error: browser never printed a
     DevTools endpoint:\n[…] Failed to connect to the bus: Could not parse server address …"}
```

Chromium did not print its DevTools endpoint inside `WAIT_MS` on the runner, so the oracle correctly exited 2 (*nothing was measured*) and the test correctly refused to treat that as a verdict. Everything behaved as designed — the failure is the **runner's browser**, not the harness. It is intermittent: the three PR runs around it were green, and **`build-test` on this PR is green** (`c6c10f2`, all 13 checks pass).

⚠ **This PR adds browser-driven oracle tests, so it roughly doubles the number of Chromium launches per CI run** (5 browser-gated test functions → 8, and ~9 launches → ~19 counting subtests). It does not make any individual launch more likely to fail, but it raises the probability that *some* launch in a run does — i.e. this PR makes an existing intermittent red more frequent.

**I have deliberately NOT fixed that here**, because it means changing shared plumbing (`briefs/_cdp.mjs`'s `launch()`) for a reason outside this PR's scope, and a launch retry is the kind of change that can quietly become a way to not notice a broken browser. The obvious fix is a bounded retry in `launch()` — `findChrome()` already throws before any attempt, and a browser that exits immediately rejects in milliseconds, so a genuinely absent or broken browser would still fail fast rather than costing N × 15 s. **Your call whether that belongs in this PR, a follow-up, or nowhere.** A permanently-flaky gate is worse than no gate; this one is not permanent yet, and I would rather flag it than expand scope without being asked.

